PR #13, opened 2020-07-22
LF-97

aliyamohamed13:
dynamic link for button added for other applicable emails - witheld consent email and confirmation email

### DIFF
--- a/packages/api/src/templates/withheld_consent_email.html
+++ b/packages/api/src/templates/withheld_consent_email.html
@@ -53,8 +53,8 @@ To request any of these actions, send an email to <a href="mailto:data@litefarm.
 <p>
 Otherwise, if you change your mind, you can log-in (on any device) at any time and accept our privacy policy using the button below:
 </p>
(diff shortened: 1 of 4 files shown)


packages/api/src/templates/withheld_consent_email.html
```diff
@@ -53,8 +53,8 @@
 <p>
 Otherwise, if you change your mind, you can log-in (on any device) at any time and accept our privacy policy using the button below:
 </p>
-<a style="color:white" href="https://www.litefarm.org"  target="_blank" class="button">Log In</a>
-<p>If you do so, you’ll once again be able to access the LiteFarm system.
+    <a style="color:white" href=""  target="_blank" class="button" id="email-button">Log In</a>
```
dustxd, line 56: I think we talked about this in private message the other day...Are we just leaving the href blank and appending the link somewhere?

aliyamohamed13: yes! the href gets modified with sendEmailTemplate.js

PR #26, opened 2020-08-19
Avoiding address assignation on null addresses and showing message

kikemarquez8:


kikemarquez8:
Also solved LF-165 Must bypass "Choose a farm" screen when no user-farm associations

PR #33, opened 2020-08-25
weird dropdown was showing up when select was setting size as 1 only

aliyamohamed13:
https://lite-farm.atlassian.net/browse/LF-89


kikemarquez8:
image
Much better! Thanks!!

PR #48, opened 2020-09-09
Lf 232 field crop permissions

JimmyRowland:


packages/api/src/controllers/fieldCropController.js
```diff
@@ -27,6 +28,11 @@ class FieldCropController extends baseController {
     return async (req, res) => {
       const trx = await transaction.start(Model.knex());
       try {
+        // TODO Move validations that involve database queries
+        const field = await fieldModel.query().select('area').where('field_id', '=', req.body.field_id);
+        if(field && field[0] && field[0].area < req.body.area_used){
+          return res.status(400).send('Area needed is greater than the field\'s area');
+        }
```
kikemarquez8, line 35: can we move this to a middleware?

packages/api/src/controllers/fieldCropController.js
```diff
@@ -27,6 +28,11 @@ class FieldCropController extends baseController {
     return async (req, res) => {
       const trx = await transaction.start(Model.knex());
       try {
+        // TODO Move validations that involve database queries
+        const field = await fieldModel.query().select('area').where('field_id', '=', req.body.field_id);
```
kikemarquez8, line 32: Use `first()` so you spare the need to look for field[0]

packages/api/src/controllers/fieldCropController.js
```diff
@@ -81,6 +87,25 @@ class FieldCropController extends baseController {
...
+      try {
+        const field_crop_id = req.params.field_crop_id;
+        const rows = await baseController.getIndividual(fieldCropModel, field_crop_id);
+        if (!rows.length) {
+          res.status(404).send('Field crop not found');
+        } else {
+          res.status(200).send(rows);
+        }
+      } catch (error) {
+        //handle more exceptions
```
kikemarquez8, line 101: this is something we need to do generally. No need to point it out here

kikemarquez8 (changes requested):

PR #49, opened 2020-09-09
Lf 247 crop field

JimmyRowland:


kikemarquez8:
You have multiple PRs that hold the same commits, can you check ? The LF232 and LF194 seem to have similarities with this one

JimmyRowland:
They are all from the same local branch.

PR #50, opened 2020-09-10
Lf 157 disable save field button

kaavyalakshmanan:
- Save field button disabled until user enters field name
- Save field button changes color based on active/disabled state
- Input field changes color based on active/disabled state
- Colors are consistent with figma

### DIFF
--- a/packages/webapp/src/containers/Field/NewField/index.js
+++ b/packages/webapp/src/containers/Field/NewField/index.js
@@ -68,7 +68,7 @@ class NewField extends Component {
       this.setState({center: {lat: farm.grid_points.lat, lng: farm.grid_points.lng}});
     }
 
-    
+
   }
 
   getMapOptions = (maps) => {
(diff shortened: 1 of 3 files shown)


packages/webapp/src/containers/Field/NewField/index.js
```diff
@@ -373,46 +401,23 @@ class NewField extends Component {
...
-                      boxShadow: "0px 2px 8px rgba(102, 115, 138, 0.3)",
-                    }} outline
+                { 
+                this.state.isSavePlanDisabled ?
+                <Button style={this.getButtonStyles()} outlinonClick={() => {
+                  this.handleModeChange(CREATE_FIELD);
+                  this.setState({ step: this.state.step + 1 });
+                }}>Save Field</Button>
+                :
+                <Button style={this.getActiveButtonStyles()}
```
kikemarquez8: Doing this in a function will trigger this on every state change, I think it might be better just to handle button styles as a constant, also having two functions for that seems to be a little overkill. just append the active styles to the active button like this
``` style={{...buttonStyles, ..additionalActiveStyles }}```

kikemarquez8 (changes requested):

packages/webapp/src/containers/Field/NewField/index.js
kaavyalakshmanan: @kikemarquez8 is `..additionalActiveStyles` another constant with he active properties?

kikemarquez8 (approved):

PR #71, opened 2020-09-14
Fix error message style

JimmyRowland:


packages/webapp/src/containers/Log/FertilizingLog/saga.js
```diff
@@ -51,6 +51,7 @@ export function* addFertilizerToDB(payload) {
 
   try {
     const result = yield call(axios.post, fertUrl + '/farm/' + farm_id, fert, header);
+    console.log(result);
```
kikemarquez8, line 54: Remove console log

kikemarquez8 (changes requested):

PR #72, opened 2020-09-15
Lf 155 bypass consent screen

JimmyRowland:


JimmyRowland:
I thought farm data is only fetched once. It turned out farm data is fetched in some components but not the others. Data will be refetched even if cache is already in place.

kikemarquez8 (approved):

PR #98, opened 2020-09-23
Lf 205 fix varietal name bug

kaavyalakshmanan:


packages/webapp/src/components/Forms/NewCropModal/index.js
```diff
@@ -98,6 +98,7 @@ class NewCropModal extends React.Component {
     const currentState = this.state;
     let validated = true;
     let errors = '';
+    let varietyError = '';
```
kikemarquez8, line 101: I don't think there is a need to have varietyError, can you use the same structure but use the errors variable? I know the story says the toaster should read "Variety Name is required", but since every other field is reading "not filled out" I would rather be consistent.

kikemarquez8 (changes requested):

kikemarquez8 (approved):

PR #102, opened 2020-09-24
Lf 215 preselected fertilizer  LF-300 props.crops undefined check

JimmyRowland:


packages/webapp/src/containers/Log/FertilizingLog/index.js
```diff
@@ -228,7 +234,7 @@ class FertilizingLog extends Component{
                 <div className={styles.inputNunit}>
                   <Control.input
                   type="number" step="any" model=".fertLog.quantity_kg" default={0}
-                  validators={{required: (val) => val, positive: (val)=> val>=0?1: 0 }}
+                  validators={{required: (val) => val===0 || val, positive: (val)=> val>=0?1: 0 }}
```
kikemarquez8, line 237: can you format this line to 
```
val === 0 || val, positive: (val) => val >= 0 ? 1: 0
```

kikemarquez8 (approved):

PR #105, opened 2020-09-25
Lf 295 seeded data tests

JimmyRowland:


kikemarquez8:
image
There are 2 tests failing here, can you review them?

JimmyRowland:
After each afterEach block, database is still getting cleaned up by the previous beforeEach block. 
Walk around:
image

kikemarquez8 (approved):

PR #130, opened 2020-10-04
Lf 276 farm data permissions

kaavyalakshmanan:
- Created migration file to revoke permissions for workers
- Added middleware to routes
- Minor changes to controller
- Wrote tests


packages/api/src/controllers/userFarmDataController.js
```diff
@@ -18,37 +18,44 @@ const Knex = require('knex');
...
         const farm_id = req.params.farm_id;
-        const data = await knex('farmDataSchedule').where({ farm_id, is_processed: false }).returning('*');
-        res.status(200).send(data);
+        const result = await userFarmModel.query().where('farm_id', farm_id);
+        if (!result.length) {
+          res.sendStatus(404)
+        }
+        else {
+          res.status(200).send(result);
+        }
```
kikemarquez8, line 51: I don't think this is right, this is supposed to be fetching farmDataSchedule data. not userFarm

packages/api/tests/userFarmData.test.js
```diff
@@ -0,0 +1,220 @@
+/*
```
kikemarquez8, line 1: please format this file, 2 spaces instead of tabs for indentation

kikemarquez8 (changes requested):

packages/api/src/controllers/userFarmDataController.js
kaavyalakshmanan: @kikemarquez8 done

packages/api/tests/userFarmData.test.js
kaavyalakshmanan: @kikemarquez8 done

kikemarquez8 (approved):

PR #142, opened 2020-10-07
LF-275 Implement status change restrictions

beetai:
Status are set such that the following status changes are allowed:

- "Active" <-> "Inactive"
- "Invited" -> "Active"
- "Invited" -> "Inactive"


beetai:
Adding to PR description: Now includes refactor of the userFarm test suite and fix an issue in the userFarm controller where transaction commits occur after response codes are sent.

kikemarquez8 (approved):

PR #156, opened 2020-10-09
fix shift duration

JimmyRowland:


packages/webapp/src/containers/Shift/MyShift/index.js
```diff
@@ -31,9 +31,11 @@ class MyShift extends Component {
 
   componentDidMount(){
     const shift = this.props.selectedShift;
+
     // const fields = this.props.fields;
     // const crops = this.props.crops;
     let tasks = shift.tasks;
+    console.log(tasks);
```
kikemarquez8, line 38: can you remove the console log

PR #241, opened 2020-11-02
Organic certifier saga and api

JimmyRowland:
Replace [disable] = useState
with watch('fieldName', default)


packages/api/src/server.js
```diff
@@ -126,6 +127,7 @@ app.use(bodyParser.json())
   .use('/farmdata', userFarmDataRoute)
   .use('/user_farm', userFarmRoute)
   .use('/roles', rolesRoutes)
+  .use('', organicCertifierSurveyRoutes)
```
kikemarquez8, line 130: I don't like the empty route here, if we need to have the get by farm lets add it to the farm routes.

kikemarquez8 (changes requested):

kikemarquez8 (approved):

PR #293, opened 2020-12-01
LF-468 Create user endpoint with password

beetai:
- user creation endpoint now receives password
- migrate: create password hash column in user table
- send email upon creation
- token is generated and sent back


packages/api/src/controllers/loginController.js
```diff
@@ -0,0 +1,58 @@
...
+const SECONDS_IN_A_WEEK = 60 * 60 * 24 * 7;
+
+class loginController extends baseController {
+  static authenticateUser() {
+    return async (req, res) => {
+      // uses email to identify which user is attempting to log in, can also use user_id for this
+      const { email, password } = req.body;
+      try {
+        const data = await userModel.query()
+          .select('user_id', 'first_name', 'last_name', 'email', 'password_hash')
```
beetai, line 31: I was thinking we can alter the payload since it currently depends on what is selected in this line. Should we include all info pertaining to the user (minus the password_hash) in the payload?

kikemarquez8 (approved):

kikemarquez8 (approved):

PR #294, opened 2020-12-01
LF-49 Add important flag to email link font colour

beetai:


beetai:
tested this, I'm 99.99% sure that this should suffice as a fix.

kikemarquez8 (approved):

PR #300, opened 2020-12-05
stage

JimmyRowland:


beetai (changes requested):
encountered this message when testing a commit

> husky > pre-commit (node v12.16.2)
sh: lint-staged: command not found
husky > pre-commit hook failed (add --no-verify to bypass)

beetai (approved):

PR #304, opened 2020-12-08
Lf 466 525

kaavyalakshmanan:


packages/webapp/src/components/Signup/EnterPasswordPage/index.js
```diff
@@ -44,7 +47,9 @@ export default function PureEnterPasswordPage({ title = 'Welcome back', onLogin
         </>
       }
     >
-      <Title style={{ marginBottom: '32px' }}>{title}</Title>
+      <Title style={{ marginBottom: '32px' }}>
+        {title + ',' + ' ' + userName.userName.userName}
```
beetai, line 51: Having `userName.userName.userName` does not look good. This leads me to find that the `manualSignUpReducer` is being set incorrectly (with nested objects) when an email is being input. Fix this and we won't need to use `userName.userName.userName`.

packages/webapp/src/containers/CustomSignUp/index.js
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
+import { useForm } from 'react-hook-form';
+import { useDispatch } from 'react-redux';
+import PureCustomSignUp from '../../components/CustomSignUp';
+import { customSignUp } from './saga';
+
+function CustomSignUp() {
+  const { register, handleSubmit, errors, watch } = useForm({ mode: 'onBlur' });
```
beetai, line 8: Can you tell me what `onBlur` is for?

packages/webapp/src/components/CreateUserAccount/index.js
```diff
@@ -0,0 +1,102 @@
...
+
+  const goBack = () => {
+    history.push({ pathname: '/' });
+  };
+
+  const onSubmit = (data) => {
+    if (isValid) {
+      dispatch(customCreateUser(data));
+    }
+    // onLogin();
```
beetai, line 47: remove comment

packages/webapp/src/components/CustomSignUp/index.js
```diff
@@ -0,0 +1,61 @@
...
+import LineBreak from '../../assets/images/signUp/lineBreak.svg';
+import Button from '../Form/Button';
+import Input from '../Form/Input';
+import React from 'react';
+import Footer from '../Footer';
+import GoogleLoginButton from '../../containers/GoogleLoginButton';
+
+export default function PureCustomSignUp({ inputs = [{}], ssoSignUp, onSubmit, disabled }) {
+  var color = '';
+  disabled ? (color = '#D4DAE3') : (color = '#FCE38D');
```
beetai, line 12: Instead of setting the background colour of the button explicitly, it is perhaps better to use the `disabled` props that the `Button` component already has.

packages/webapp/src/components/CustomSignUp/index.js
```diff
@@ -0,0 +1,61 @@
...
+                style={{ width: 312 }}
+                className={styles.manualInputContainer}
+                {...inputs[0]}
+              />
+            </div>
+          </div>
+        </div>
+      </div>
+      <Footer
+        style={{ position: 'sticky', bottom: '0' }}
```
beetai, line 38: Might be better to put styling in `styles.scss` and use `className`

packages/webapp/src/containers/CustomSignUp/index.js
```diff
@@ -0,0 +1,44 @@
...
+  const disabled = !email || !required || !validEmailRegex.test(email);
+
+  const dispatch = useDispatch();
+
+  const onSubmit = (data) => {
+    const { email } = data;
+    dispatch(customSignUp(email));
+  };
+
+  const ssoSignUp = () => {};
```
beetai, line 24: If we don't need this function, we can likely remove it from the props of `PureCustomSignUp`

packages/webapp/src/components/CustomSignUp/index.js
beetai: Might also have the benefit of making the button not clickable before a proper email is inputted

packages/webapp/src/containers/CustomSignUp/saga.js
```diff
@@ -0,0 +1,55 @@
...
+      });
+    } else if (!result.data.exists && !result.data.sso) {
+      yield put(saveUserEmailSuccess({ email }));
+      history.push({
+        pathname: '/create-user-account',
+      });
+    }
+  } catch (e) {
+    console.log('error is');
+    console.log(e);
```
beetai, line 49: Do you think using a `toastr` here would be a good idea?

packages/webapp/src/containers/CustomSignUp/saga.js
```diff
@@ -0,0 +1,55 @@
...
+
+const axios = require('axios');
+const loginUrl = (email) => `${url}/user/${email}`;
+
+export const customSignUp = createAction(`customSignUpSaga`);
+
+export function* customSignUpSaga({ payload: email }) {
+  try {
+    const result = yield call(axios.get, loginUrl(email));
+    if (result.data.user != null) {
```
beetai, line 30: I'm not quite sure what this check is for. What is the logic of having `saveUserEmailSuccess` be called two times here?

packages/webapp/src/containers/GoogleLoginButton/index.js
```diff
@@ -15,10 +15,12 @@ function GoogleLoginButton({ disabled }) {
   };
   return (
     <GoogleLogin
+      buttonText="Login"
       onSuccess={onSuccess}
       onFailure={onFailure}
       disabled={disabled}
-      style={{ width: '100%', height: '48px', fontWeight: 500 }}
+      // style={{ width: '312px', height: '48px', fontWeight: 5000 }}
+      // style={{ theme: 'dark'}}
```
beetai, line 23: remove commented code

packages/webapp/src/components/CreateUserAccount/saga.js
```diff
@@ -0,0 +1,61 @@
...
+        token,
+        user: { user_id },
+      } = result.data;
+      localStorage.setItem('id_token', token);
+
+      yield put(loginSuccess({ user_id }));
+      history.push('/farm_selection');
+    }
+    const token = result.data;
+    if (result.length) {
```
beetai, line 49: Why are we checking for the `result`'s length? Does the endpoint return an object or an array?

packages/webapp/src/components/CreateUserAccount/saga.js
```diff
@@ -0,0 +1,61 @@
...
+      console.log(result.length);
+      history.push('/welcome');
+    }
+  } catch (e) {
+    console.log('error is');
+    console.log(e);
```
beetai, line 55: remove console.log in favour of `toastr`?

packages/webapp/src/containers/CustomSignUp/signUpSlice.js
```diff
@@ -0,0 +1,36 @@
...
+  name: 'manualSignUpReducer',
+  initialState,
+  reducers: {
+    saveUserEmailSuccess: (state, { payload: email }) => {
+      state.userEmail = email;
+    },
+    saveUserNameSuccess: (state, { payload: name }) => {
+      state.userName = name;
+      console.log('inside splice');
+      console.log(state.userName);
```
beetai, line 19: remove `console.log`

beetai (changes requested):
Here are a handful of things that should be checked, some more urgent than others.

@kikemarquez8 should we use i18n for this PR?

packages/api/src/controllers/loginController.js
```diff
@@ -67,6 +67,46 @@ class loginController extends baseController {
...
+          else {
+              res.status(200).send({
+                  user: data,
+                  exists: true,
+                  sso: false
+              });
+          }
+      }
+    } catch (error) {
+        console.log("error was thrown")
```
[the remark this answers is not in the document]
kaavyalakshmanan: Done.

packages/api/tests/login.test.js
```diff
@@ -0,0 +1,121 @@
...
+const server = require('../src/server');
+const knex = require('../src/util/knex');
+const { tableCleanup } = require('./testEnvironment');
+jest.mock('jsdom');
+jest.mock('../src/middleware/acl/checkJwt');
+jest.mock('../src/jobs/station_sync/mapping');
+const mocks = require('./mock.factories');
+
+describe('Sign Up Tests', () => {
+	let middleware;
```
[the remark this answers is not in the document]
kaavyalakshmanan: Done.

packages/api/tests/login.test.js
```diff
@@ -0,0 +1,121 @@
...
+				expect(res.status).toBe(200);
+				expect(res.body.exists).toBe(true);
+				expect(res.body.sso).toBe(false);
+				expect(res.body.user[0].first_name).toBe(name);
+				done();
+			});
+		});
+
+
+		// xtest('Get status of user from email address should return 200 and SSO true', async (done) => {
```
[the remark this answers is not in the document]
kaavyalakshmanan: This test was commented out because for some reason this test was hitting an old Auth0 table -- because Auth0 was using google_user_id as ID. This was resulting in errors in the build. I think that table needs to be removed first for this test to be used. Same with the ssoUser mocks in mock.factories. I had originally mocked it so that the user ID was generating a random digit as you say.

packages/api/tests/login.test.js
```diff
@@ -0,0 +1,121 @@
...
+		// 	getRequest({ email: user.email}, (err, res) => {
+		// 		expect(res.status).toBe(200);
+		// 		expect(res.body.exists).toBe(true);
+		// 		expect(res.body.sso).toBe(true);
+		// 		expect(res.body.user[0].first_name).toBe(name);
+		// 		done();
+		// 	});
+		// });
+
+		test('Get status of user from email address should return 200 and sso true because user does not exist', async (done) => {
```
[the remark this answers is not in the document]
kaavyalakshmanan: Yes, thank you! Done.

packages/webapp/src/containers/GoogleLoginButton/index.js
kaavyalakshmanan: Done.

packages/webapp/src/components/CreateUserAccount/saga.js
kaavyalakshmanan: Done.

packages/webapp/src/containers/CustomSignUp/signUpSlice.js
kaavyalakshmanan: Done.

packages/webapp/src/components/Signup/EnterPasswordPage/index.js
```diff
@@ -7,6 +7,8 @@ import PropTypes from 'prop-types';
 import { useForm } from 'react-hook-form';
 import { validatePasswordWithErrors } from '../utils';
 import { PasswordError } from '../../Form/Errors';
+import { useSelector } from 'react-redux';
+import { manualSignUpSelector } from '../../../containers/CustomSignUp/signUpSlice';
 
 export default function PureEnterPasswordPage({ title = 'Welcome back', onLogin }) {
```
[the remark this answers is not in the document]
kaavyalakshmanan: Jimmy will be working on that.

packages/webapp/src/components/CreateUserAccount/index.js
```diff
@@ -0,0 +1,102 @@
...
+        disabled
+        defaultValue={email.userEmail.email}
+      />
+      <Input style={{ marginBottom: '28px' }} label={'Full name'} name={NAME} inputRef={refInput} />
+      <Input
+        style={{ marginBottom: '28px' }}
+        label={'Password'}
+        type={PASSWORD}
+        name={PASSWORD}
+        icon={<Underlined>Forgot password?</Underlined>}
```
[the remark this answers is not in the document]
kaavyalakshmanan: Done.

packages/webapp/src/components/CreateUserAccount/index.js
```diff
@@ -0,0 +1,102 @@
...
+import { useForm } from 'react-hook-form';
+import { validatePasswordWithErrors } from '../Signup/utils';
+import { PasswordError } from '../Form/Errors';
+import history from '../../history';
+import { customCreateUser } from './saga';
+import { useDispatch, useSelector } from 'react-redux';
+import { manualSignUpSelector } from '../../containers/CustomSignUp/signUpSlice';
+import styles from './styles.scss';
+
+export default function PureCreateUserAccount({ title = 'Create new user account' }) {
```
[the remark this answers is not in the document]
kaavyalakshmanan: So in this file, I am using Jimmy's EnterPasswordPage pure component. I was running into some issues with keeping this as just a pure component and having a container pass in the logic. I discussed this with Kike this morning, and I think we are going to keep this as it for now and then probably make that change to this and EnterPasswordPage for the next sprint.

packages/webapp/src/components/CreateUserAccount/index.js
```diff
@@ -0,0 +1,102 @@
...
+        label={'Password'}
+        type={PASSWORD}
+        name={PASSWORD}
+        icon={<Underlined>Forgot password?</Underlined>}
+        inputRef={inputRegister}
+      />
+      {showErrors && (
+        <div>
+          <Text>Hint</Text>
+          <PasswordError
```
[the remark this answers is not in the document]
kaavyalakshmanan: Nice catch! Yes, I agree, this needs to be flushed out more more current Auth0 users. However I think for the sake of this PR, we'll probably leave the password requirements as is to demonstrate that functionality, and flush this out more in the next sprint.

packages/webapp/src/components/CreateUserAccount/saga.js
```diff
@@ -0,0 +1,61 @@
...
+const userUrl = () => `${url}`;
+
+export const customCreateUser = createAction(`customCreateUserSaga`);
+
+export function* customCreateUserSaga({ payload: data }) {
+  try {
+    const name = data.name;
+    const full_name = name.split(' ');
+    const first_name = full_name[0];
+    const last_name = full_name[1];
```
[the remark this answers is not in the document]
kaavyalakshmanan: That's a good point. I'll ask Kevin this, but for the sake of time, I think we can leave this as is, and once I hear from him I'll make another PR just for that. (Unless I hear back before this PR is merged.)

packages/webapp/src/components/Signup/EnterPasswordPage/index.js
```diff
@@ -21,6 +23,7 @@ export default function PureEnterPasswordPage({ title = 'Welcome back', onLogin
   } = validatePasswordWithErrors(password);
   const inputRegister = register({ validate: () => isValid });
   const [showErrors, setShowErrors] = useState(false);
+  const userName = useSelector(manualSignUpSelector);
 
   const onSubmit = (data) => {
```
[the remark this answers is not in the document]
kaavyalakshmanan: Yep, Jimmy will be working on that as soon as this PR is merged!

packages/webapp/src/Routes.js
```diff
@@ -135,6 +138,7 @@ const Routes = () => {
         <Switch>
           <Route path="/" exact component={Home} />
           <Route path="/home" exact component={Home} />
+          <Route path="/custom-sign-up" exact component={CustomSignUp} />
```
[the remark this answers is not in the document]
kaavyalakshmanan: I removed the one on line 237 (which shouldn't be there), but the route is added to line 141 because it is part of the onboarding flow. If removed, nothing will be rendered when a user tries to create a new account.

packages/api/src/controllers/loginController.js
```diff
@@ -67,6 +67,46 @@ class loginController extends baseController {
     }
 
   }
+
+  static getUserNameByUserEmail() {
```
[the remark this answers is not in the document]
kaavyalakshmanan: Done.

packages/webapp/src/containers/CustomSignUp/index.js
kaavyalakshmanan: So that the in-line error message appears when a user clicks away from the input field.

packages/webapp/src/components/CreateUserAccount/index.js
kaavyalakshmanan: Done.

packages/webapp/src/containers/CustomSignUp/index.js
kaavyalakshmanan: Done.

packages/webapp/src/containers/CustomSignUp/saga.js
kaavyalakshmanan: I feel like since it wasn't explicitly stated in the ticket, might be better to ask Kevin about the general rule for when to use toastr? I did add Toastr to your other comment though so we can show him what it looks like and ask about it.

packages/webapp/src/Routes.js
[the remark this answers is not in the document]
beetai: Why should a user create a new account when logged in?

packages/webapp/src/containers/CustomSignUp/saga.js
beetai: That's a good idea, let's see what Kevin has to say about it.

packages/webapp/src/components/CreateUserAccount/saga.js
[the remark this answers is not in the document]
beetai: For sure, if he gets back to you before we can merge, then we can resolve this fully.

packages/webapp/src/containers/CustomSignUp/saga.js
kikemarquez8: Also, always try to use `!==`  instead of `!=` Or, Strict isntead of Abstract equality [the why](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness). Unless there's a specific reason to avoid it.

packages/webapp/src/Routes.js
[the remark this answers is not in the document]
kaavyalakshmanan: A user would not even be able to arrive at this route is they are not logged in -- that's why this route is placed under if `(!isAuthenticated)`

packages/webapp/src/components/CreateUserAccount/saga.js
kaavyalakshmanan: Oops, looks like that is just unused code that doesn't do anything. Removing that.

packages/webapp/src/containers/CustomSignUp/saga.js
kaavyalakshmanan: I was saving it twice because it hits two different cases -- if a user is a new user, `result.data.user` will be null. I need to save their email to retrieve it in create user account. If they are an existing user, I also want to save it so that after a user inputs their password in the password page, the email can be passed to the `authenticateUser` -- but I just realized this endpoint does not even require the email. So can remove the latter.

packages/webapp/src/components/CustomSignUp/index.js
kaavyalakshmanan: Done

packages/webapp/src/components/CustomSignUp/index.js
kaavyalakshmanan: Done.

packages/webapp/src/components/Signup/EnterPasswordPage/index.js
kaavyalakshmanan: Done.

packages/api/src/controllers/loginController.js
```diff
@@ -67,6 +67,46 @@ class loginController extends baseController {
...
+    return async (req, res) => {
+      const { email } = req.params;
+      try {
+      const data = await userModel.query()
+        .select('*').from('users').where('users.email', email)
```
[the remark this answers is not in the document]
kaavyalakshmanan: Done.

packages/webapp/src/components/CustomSignUp/index.js
```diff
@@ -35,11 +34,12 @@ export default function PureCustomSignUp({ inputs = [{}], ssoSignUp, onSubmit, d
...
       </div>
       <Footer
-        style={{ position: 'sticky', bottom: '0' }}
+        className={styles.footer}
         children={
           <>
             <div className={styles.bottomButton}>
               <Button
+                disabled={disabled}
                 style={{
```
beetai, line 43: Should also pull this into `styles.scss` and use `className`

beetai (changes requested):
One more change and I believe I'll be ready to approve

beetai (approved):
🔥 👌

PR #306, opened 2020-12-08
Lf 525 connect login flow

JimmyRowland:


packages/webapp/src/containers/CustomSignUp/index.js
```diff
@@ -2,25 +2,40 @@ import React from 'react';
...
+    dispatch(customLoginWithPassword({email, password}));
+  }
+
+  const enterPasswordOnGoBack = () => {
+    history.push({ pathname: '/', state: {email: user.email, emailPage: true} })
+  }
+
+  const hasUser = user && user.user_id;
+  if(hasUser){
+    return <PureEnterPasswordPage onLogin={onLogin} title={`Welcome back ${user.first_name}`} onGoBack={enterPasswordOnGoBack}/>
```
beetai, line 36: This feels inconsistent to me: if we are going to change what component we are showing instead of using `history` to switch the page, why doesn't this also apply to the `PureCreateUserAccount` page? I think we should either apply this to both of the pages or neither.

beetai (changes requested):

packages/webapp/src/containers/CustomSignUp/index.js
kaavyalakshmanan: Yes, I agree with this.

packages/webapp/src/components/Signup/EnterPasswordPage/index.js
```diff
@@ -7,8 +7,6 @@ import PropTypes from 'prop-types';
 import { useForm } from 'react-hook-form';
 import { validatePasswordWithErrors } from '../utils';
 import { PasswordError } from '../../Form/Errors';
-import { useSelector } from 'react-redux';
-import { manualSignUpSelector } from '../../../containers/CustomSignUp/signUpSlice';
```
kaavyalakshmanan, line 10: Super minor but make sure the title says "Welcome back, <Name>!" (with comma and exclamation mark) as per ticket requirements.

packages/webapp/src/components/Signup/EnterPasswordPage/index.js
```diff
@@ -7,10 +7,8 @@ import PropTypes from 'prop-types';
...
 
-export default function PureEnterPasswordPage({ title = 'Welcome back', onLogin }) {
+export default function PureEnterPasswordPage({ title = 'Welcome back', onLogin, onGoBack }) {
```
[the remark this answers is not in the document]
kikemarquez8: After talking with Jimmy the use of useForm here doesn't make it a "smart" component, its basically segwaying the data to the container imposed logic.

packages/webapp/src/containers/CustomSignUp/index.js
JimmyRowland: I will fix this later after I have separated redux from pureCreateUserAccount

JimmyRowland: We don't want users to go password page without entering an email and we don't want to introduce more checks in Router.js if those checks can be resolved in a local component

packages/webapp/src/components/CustomSignUp/index.js
```diff
@@ -35,27 +40,24 @@ export default function PureCustomSignUp({ inputs = [{}], onSubmit, disabled })
...
-              />
-            </div>
-          </>
-        }
-      ></Footer>
+      >
+        <div className={styles.bottomButton}>
+          <Button
+            disabled={disabled}
+            style={{
```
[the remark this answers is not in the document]
JimmyRowland: We will have to overwrite our own components using inline style by passing style object into classes/style properties. We can't overwrite a component using style sheet as css sheet loading order is indetermainate.

packages/api/src/controllers/loginController.js
```diff
@@ -51,13 +50,19 @@ class loginController extends baseController {
...
-        if (!user) {
+        const user = await userModel.query().where({ email }).first();
+        const isUserNew = !user;
+        if (isUserNew) {
           const newUser = { user_id, email, first_name, last_name };
           await userModel.query().insert(newUser);
         }
-        const id_token = await createAccessToken({ user_id, email, first_name, last_name });
-        return res.status(201).send({ id_token, user: { user_id } });
+        const isPasswordNeeded = user && user.user_id && user.user_id !== user_id;
```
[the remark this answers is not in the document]
JimmyRowland: Addressing 9. Attempting to log into SSO with an gmail account that is already associated with an existing user (that did not use SSO) will cause the login endpoint to fail, noting to fix this in the future

PR #313, opened 2020-12-09
UI for notification teaser modal

kaavyalakshmanan:
- UI for notification teaser + new translation object for text


JimmyRowland:
We probably need to remove props.history from the floater and replace all inputs with svg component.

PR #319, opened 2020-12-10
Signup storybook

JimmyRowland:
- Remove google login button from pureSignInComponent
- remove useSelector/useReducer from pure components
- pureSignInComponent/pureCreateAccount stories
- handle the error that user enter an email address that exist in the database as SSO user
- handle the error that user try to login to an non-sso account using google login button
- Incorrect password error
- set up backend prettier


packages/api/src/controllers/loginController.js
```diff
@@ -50,34 +47,45 @@ class loginController extends baseController {
...
       const { email } = req.params;
       try {
-        const data = await userModel.query()
-          .select('*').from('users').where('users.email', email).first()
+        const data = await userModel
+          .query()
+          .select('*')
+          .from('users')
+          .where('users.email', email)
+          .first();
```
kikemarquez8, line 88: I think the previous formatting made more sense

packages/api/src/controllers/userController.js
```diff
@@ -176,10 +166,13 @@ class userController extends baseController {
       try {
         const id = req.params.user_id;
 
-        const data = await userModel.query().findById(id).select('first_name', 'last_name', 'profile_picture', 'email', 'phone_number', 'user_id');
+        const data = await userModel
+          .query()
+          .findById(id)
+          .select('first_name', 'last_name', 'profile_picture', 'email', 'phone_number', 'user_id');
```
kikemarquez8, line 172: Same applies for this none

beetai (approved):

PR #329, opened 2020-12-15
test

JimmyRowland:


packages/api/src/controllers/passwordResetController.js
```diff
@@ -16,22 +16,22 @@
...
       // get from db user_id and first_name from email (user table)
-  
+
       // get entry in db (password table) from user_id
-  
+
       // generate token
       // payload: user_id, reset_token_version, email, first_name
-  
+      const resetPasswordToken = await createResetPasswordToken(req.body);
```
beetai, line 31: I'll be taking care of this. Just so we can avoid some conflicts, perhaps revert the changes in this file?

beetai (approved):

PR #337, opened 2020-12-16
Setting user language preference on app start

kikemarquez8:
Also, in the profile tab, you should now be able to check a language selector, that will save the language selection you make to your user table. 

TODO: get language preference from user and compare it to browser one (make user override browser).


packages/webapp/src/index.js
```diff
@@ -73,6 +74,11 @@ if (process.env.NODE_ENV === 'development' && module.hot) {
     store.replaceReducer(newRootReducer);
   });
 }
+if(!localStorage.getItem('litefarm_lang')) {
```
JimmyRowland, line 77: Is it better to have a language selector in userFarmSlice

JimmyRowland (approved):

PR #344, opened 2020-12-17
Fix query dependency on fieldCrop

kaavyalakshmanan:
- Fix query dependency on fieldCrop

### DIFF
--- a/packages/api/src/controllers/insightController.js
+++ b/packages/api/src/controllers/insightController.js
@@ -132,7 +132,7 @@ class insightController extends baseController {
           FROM "field" f
           LEFT JOIN "fieldCrop" fc
           ON fc.field_id = f.field_id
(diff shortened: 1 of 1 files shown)


packages/api/src/controllers/insightController.js
```diff
@@ -132,7 +132,7 @@ class insightController extends baseController {
           FROM "field" f
           LEFT JOIN "fieldCrop" fc
           ON fc.field_id = f.field_id
-          WHERE f.farm_id = ? and fc.deleted = false and fc.end_date >= NOW()
+          WHERE f.farm_id = ?
```
kikemarquez8, line 135: Not sure if this is what we want to do, the thing is, if we take this out, the Biodiversity will include fieldCrops that this user no longer has on his farm. So we don't want to include those.

kikemarquez8 (changes requested):

PR #375, opened 2021-01-07
Set crop translation key to common name

JimmyRowland:


kikemarquez8 (changes requested):
This won't actually work as a long term solution, the problem is, if a user were to create a new crop, this crop won't have its crop translation key set. We can either handle it on the FE and default it on the crop_common_name if no translation key exist, or default it to crop_common_name on crop creation on the backend

kikemarquez8 (approved):

PR #387, opened 2021-01-12
Feature/lf 581 Farm Map indication

kikemarquez8:


kaavyalakshmanan (changes requested):
- I would double check the error messages with Kevin - on the ticket it just says "Invalid entry please try again" but looks like we have 3 different error messages for 3 different cases. 
- I was not able to test the loading svg - as soon as I type a valid address I see a 1 sec delay where pin is rendered on an empty map and then I see the map and pin.
- If I type an incorrect address I see error message - but after typing a valid address, I still see error message unless I click 'x' and retype. It might be better to clear the error message as soon as user begins retyping address.
- I don't see buttons to zoom in and out of map - but can zoom using trackpad. I would double check if buttons are required. (Although I see buttons when I click full screen on the map so if that is the desired functionality then it's fine!)

kaavyalakshmanan (approved):

PR #388, opened 2021-01-12
Lf 693/detect browser non chrome users

kaavyalakshmanan:


kikemarquez8 (changes requested):
Careful, this is deleting a lot of entries from the message.json in all languages

kikemarquez8 (approved):

PR #472, opened 2021-01-28
Worker consent modified, step one showing worker label only

kikemarquez8:


packages/webapp/src/components/Shift/StepOne/index.js
```diff
@@ -98,15 +98,20 @@ function PureStepOne({
...
-        label={t('SHIFT.EDIT_SHIFT.CHOOSE_WORKERS')}
-        options={workerOptions}
-        onChange={setWorker}
-        value={worker}
-        style={{ marginBottom: '24px' }}
-        defaultValue={defaultWorker}
-      />
-      <Semibold>What tasks did you do today?</Semibold>
+      { Number(farm.role_id) === 3 && (
+        <Semibold>{`${t('SHIFT.EDIT_SHIFT.WORKER')}: ${defaultWorker.label}`}</Semibold>
```
beetai, line 102: Perhaps it might be better to for this to display full name instead of first name for clarity? Even if the worker may only submit shifts for themselves, multiple workers may share the same first name.

beetai (changes requested):

packages/webapp/src/components/Shift/StepOne/index.js
kikemarquez8: done

beetai (approved):

PR #476, opened 2021-01-28
Preventing continue click if currently loading

kikemarquez8:


beetai (commented):
Perhaps this change can be applied to other "Proceed" buttons on the onboarding flow?

beetai (approved):

PR #477, opened 2021-01-29
Continue button disabled until both fields filled

kaavyalakshmanan:


beetai (commented):
Since this change makes it so that inline field errors are no longer in use, we could still apply them by altering the form to be `onTouched` similar to what was done in https://github.com/LiteFarmOrg/LiteFarm/pull/465/files

beetai (approved):

packages/webapp/src/components/AddFarm/index.js
```diff
@@ -10,14 +10,21 @@ const style = {
...
   const { t } = useTranslation();
   return (
     <Form
       onSubmit={onSubmit}
       buttonGroup={
-        <Button type={'submit'} disabled={loading} fullLength>
+        <Button type={'submit'} disabled={disabled} fullLength>
           {loading ? t('common:LOADING') : t('common:CONTINUE')}
         </Button>
       }
```
kikemarquez8, line 30: The button needs to be disabled on loading as well that was part of 21-136

kikemarquez8 (approved):

kikemarquez8 (approved):

PR #488, opened 2021-01-29
Adding translation tags to users table on People, adding Svg without text

kikemarquez8:


packages/webapp/src/lang/en/translation.json
```diff
@@ -451,6 +457,12 @@
     "FARM_OWNER": "Farm Owner",
     "TITLE": "What is your role on the farm?"
   },
+  "ROLES": {
+    "EXTENSION_OFFICER": "Extension Officer",
+    "MANAGER": "Manager",
+    "OWNER": "Owner",
+    "WORKER": "Worker"
+  },
```
beetai, line 465: wonder if we can use `role.json` instead?

beetai (approved):

PR #495, opened 2021-01-30
[F21-184] add loading indicator to submit button

beetai:


packages/webapp/src/containers/Help/index.js
```diff
@@ -6,9 +6,11 @@ import history from '../../history';
 import { userFarmSelector } from '../userFarmSlice';
 export default function HelpRequest() {
   const dispatch = useDispatch();
+  const [loading, setLoading] = useState(false);
 
   const handleSubmit = (file, data) => {
-    dispatch(supportFileUpload({ file, form: data }));
+    setLoading(true);
+    dispatch(supportFileUpload({ file, form: data, setLoading }));
```
kikemarquez8, line 13: Passing a function in an action is not a good idea. Actions should be immutable objects. add a loading indicator in the reducer, and get it on the component

kikemarquez8 (changes requested):

kikemarquez8 (approved):

PR #522, opened 2021-01-31
Feature/harvest use add log

kaavyalakshmanan:
- UI changes discussed last week
- Persist frontend between steps
- Works for imperial and metric
- Disable edit for harvest log


packages/webapp/src/containers/Log/HarvestAllocation/index.js
```diff
@@ -40,18 +46,29 @@ class HarvestAllocation extends Component {
...
     let sum = Object.keys(val).reduce((sum, key) => sum + Number(val[key]), 0);
+    console.log(sum);
 
-    if (sum !== Number(this.props.formData.quantity_kg)) {
-      toastr.error('Total does not equal the amount to allocate');
-    } else {
+    if (
+      sum >= Number(this.props.formData.quantity_kg) - 0.1 ||
+      sum < Number(this.props.formData.quantity_kg) + 0.1
+    ) {
```
beetai, line 54: I think during allocation we should make the values add up on the dot rather than +- 0.01.
Just to take an example, the system would allow you to distribute your 0.01 harvest to multiple distributions of 0. A discrepancy of 0.01 is okay after a conversion but probably shouldn't be allowed when manually allocating the harvest amount.

packages/webapp/src/containers/Log/LogDetail/index.js
```diff
@@ -90,7 +90,7 @@ class LogDetail extends Component {
 
     for (let d of diseases) {
       if (d.disease_id === d_id) {
-        return d.farm_id ? this.props.t(`disease:name.${d.disease_name_translation_key}`) : d.disease_common_name ;
+        return this.props.t(`disease:name.${d.disease_name_translation_key}`);
```
beetai, line 93: A merge conflict exists here. Be sure to accept the change that is on develop.

packages/webapp/src/containers/Log/HarvestAllocation/index.js
beetai: Also make this an and instead of an or

beetai (changes requested):

beetai (approved):

PR #616, opened 2021-02-22
Fix various bugs

kaavyalakshmanan:


packages/webapp/src/containers/Log/HarvestUseType/index.js
```diff
@@ -26,40 +26,57 @@ function HarvestUseType() {
...
-        }
+      tempProps.selectedUseTypes.map((item) => {
+        defaultData.selectedUseTypes.map((item1) => {
+          let name = isEditStepTwo.isEditStepTwo
+            ? item1.harvestUseType.harvest_use_type_name
+            : item1.harvest_use_type_name;
+          if (item.harvest_use_type_name === name) {
+            item.quantity_kg = item1.quantity_kg;
+          }
+        });
```
kikemarquez8, line 41: There is a better way to do this, try something like
```
  const defaultQuantities = defaultData.selectedUseTypes.reduce( (obj, item) => {
      let name  = isEditStepTwo.isEditStepTwo ? 
            item.harvestUseType.harvest_use_type_name : 
            item.harvest_use_type_name;
      return {...obj , [name]: item.quantity_kg } 
   });
   const newData = tempProps.selectedUseTypes.map((item) => ({
      ...item,
      quantity_kg: defaultQuantities[item.harvest_use_type_name] ? 
      defaultQuantities[item.harvest_use_type_name] : item.quantity_kg
    })
```
that way it stays N

packages/webapp/src/containers/Log/HarvestUseType/index.js
```diff
@@ -26,40 +26,57 @@ function HarvestUseType() {
...
       });
     }
     defaultData.selectedUseTypes = tempProps.selectedUseTypes;
```
kikemarquez8, line 44: Can you change this? since defaultData is retrieved from state, we should not be mutating it it should always be changed to a new value. instead in the next line that dispatches. do 
```
dispatch(harvestLogData( {...defaultData, selectedUseTypes: (newData or tempProps.selectedUseTypes)} ))
```

packages/webapp/src/containers/Log/HarvestUseType/index.js
```diff
@@ -26,40 +26,57 @@ function HarvestUseType() {
...
   const onNext = (data) => {
-    const tempProps = JSON.parse(JSON.stringify(data));
-    if (defaultData.selectedUseTypes.length > 0) {
+    if (isEditStepTwo.isEditStepTwo) {
+      data.selectedUseTypes.map((item) => {
+        defaultData.selectedUseTypes.map((item1) => {
+          if (item.harvest_use_type_name === item1.harvestUseType.harvest_use_type_name) {
+            item.quantity_kg = item1.quantity_kg;
+          }
+        });
```
kikemarquez8, line 56: Try to do the same as above. In fact if you could wrap the reduce function and the map function into separate functions that would be great.

packages/webapp/src/containers/Log/HarvestUseType/index.js
```diff
@@ -26,40 +26,57 @@ function HarvestUseType() {
...
+      const tempProps = JSON.parse(JSON.stringify(data));
       if (defaultData.selectedUseTypes.length > 0) {
         tempProps.selectedUseTypes.map((item) => {
           defaultData.selectedUseTypes.map((item1) => {
-            if (item.harvest_use_type_name === item1.harvest_use_type_name) {
+            let name = item1.harvest_use_type_name
+              ? item1.harvest_use_type_name
+              : item1.harvestUseType.harvest_use_type_name;
+            if (item.harvest_use_type_name === name) {
               item.quantity_kg = item1.quantity_kg;
```
kikemarquez8, line 70: this is pretty much the same , if you can do the same as above.

packages/webapp/src/containers/Log/HarvestUseType/index.js
```diff
@@ -26,40 +26,57 @@ function HarvestUseType() {
...
             }
           });
         });
       }
+      defaultData.selectedUseTypes = tempProps.selectedUseTypes;
```
kikemarquez8, line 75: same here

kikemarquez8 (changes requested):

kikemarquez8 (approved):

PR #643, opened 2021-03-03
Refactor subject translation

JimmyRowland:
and
export fixes
and map page height fix


kikemarquez8 (approved):

packages/api/src/controllers/loginController.js
```diff
@@ -65,7 +65,8 @@ const loginController = {
             reason_for_failure: 'password_mismatch',
           });
           return res.sendStatus(401);
-        };
+        }
+        ;
```
beetai, line 69: Why was this change made

PR #654, opened 2021-03-05
Logic to render correct my-farm icon based on language

kaavyalakshmanan:


packages/webapp/src/components/Navigation/NavBar/index.js
```diff
@@ -288,7 +291,13 @@ export default function PureNavBar({
                 className={classes.iconButton}
                 onClick={farmButtonOnClick}
               >
-                <MyFarmIcon />
+                {selectedLanguage === 'en' ? (
```
JimmyRowland, line 294: default should be english

JimmyRowland (changes requested):

JimmyRowland (approved):

PR #668, opened 2021-03-11
Lf 790/filter map

JimmyRowland:


packages/webapp/src/stories/Pages/MapFilter/MapFilter.stories.js
```diff
@@ -1,9 +1,9 @@
 import React from 'react';
 import decorators from '../config/decorators';
-import MapFilter from '../../../components/MapFilter/';
+import MapFilter from '../../../components/MapDrawer/';
```
beetai, line 3: rename var to MapDrawer for less confusion

packages/webapp/src/containers/Map/constants.js
```diff
@@ -20,18 +19,27 @@ export const isArea = (type) => {
...
+  greenhouse: 'greenhouse',
+  ground_water: 'ground_water',
+  natural_area: 'natural_area',
+  buffer_zone: 'buffer_zone',
+  creek: 'creek',
+  fence: 'fence',
+  gate: 'gate',
+  water_valve: 'water_valve',
+  farm_bound: 'farm_bound',
+};
```
beetai, line 45: Should we use camelCase instead?

beetai (changes requested):
A couple of other things to address:
- the condition under which the map header appears should also be tweaked
- icons that appear next to each location type in the drawer have disappeared

packages/webapp/public/locales/en/translation.json
```diff
@@ -105,6 +105,7 @@
       "LINES": "Lines",
       "NA": "Natural area",
       "POINTS": "Points",
+      "Residence": "Residence",
```
beetai, line 108: all caps the key. same applies to the other language files

packages/webapp/src/stories/Pages/Map/SpotLight/Drawer.stories.js
```diff
@@ -0,0 +1,36 @@
...
+    <div style={{ flexGrow: 1 }} />
+    <PureMapFooter {...args} />
+  </div>
+);
+
+export const SpotLight = Template.bind({});
+SpotLight.args = {
+  isAdmin: true,
+  showSpotlight: true,
+};
```
beetai, line 33: For this story, we run into `resetSpotlight is not a function` once spotlight is closed.

packages/webapp/src/containers/Map/index.js
```diff
@@ -217,7 +183,9 @@ export default function Map() {
       dispatch(setMapFilterSetting(payload));
     }
   };
-  const handleAddMenuClick = (asset) => {};
+  const handleAddMenuClick = (asset) => {
+    startDrawing(asset);
```
beetai, line 187: Perhaps we can rename `asset` to something like `locationType`? might be more clear

beetai: same with `handleFilterMenuClick` above

beetai (changes requested):

beetai:
Another change that needs to be made: the Add Location drawer has the title "Filter the map". Please change this

beetai (approved):

PR #686, opened 2021-03-16
implement basic adjust and fix boundary fill

beetai:


kikemarquez8 (changes requested):
Drawing is now weird, no feedback on click while drawing and missaligned buttons on the drawing screen

beetai:
update:
- farm boundary should no longer have a colour fill
- addition of adjust button: adjust button will show a modal that demonstrates how to adjust drawings

kikemarquez8 (approved):

PR #711, opened 2021-03-26
[LF-1039/1041/1042] spotlight

beetai:


kikemarquez8:
I could not see the drop_point and adjust_line spotlights, in drop_point after clicking adjust I get an empty modal with the button "Got it"

kikemarquez8 (approved):

PR #752, opened 2021-04-13
Lf 973/fix insight controller

JimmyRowland:


packages/api/src/controllers/insightController.js
```diff
@@ -141,20 +145,24 @@ const insightController = {
...
+          JOIN "area" on figure.figure_id = area.figure_id
           LEFT JOIN "fieldCrop" fc
-          ON fc.field_id = f.field_id
-          WHERE f.farm_id = ?
-          AND f.deleted = false
-          GROUP BY f.grid_points`, [farmID]);
+          ON fc.location_id = f.location_id
+          WHERE location.farm_id = ?
+          AND location.deleted = false
+          GROUP BY area.grid_points`, [farmID]);
```
kikemarquez8, line 157: Sorry Jimmy, could we make this query only query farm_site_boundaries for now 
I think that would require a JOIN before the location join.

kikemarquez8 (approved):

PR #831, opened 2021-04-28
default is searchable

JimmyRowland:
F21-361
delete log


kikemarquez8 (approved):

packages/webapp/src/components/Form/ReactSelect/index.js
```diff
@@ -135,6 +135,7 @@ const ReactSelect = ({
             </Underlined>
           ),
         }}
+        isSearchable={options?.length > 8}
```
kikemarquez8, line 138: Why is 8 an important number here?

PR #837, opened 2021-04-28
Replace north icon

kaavyalakshmanan:


beetai (commented):
Is the border shadow supposed to be there for the new icon?

JimmyRowland (approved):

PR #839, opened 2021-04-28
Remove % for ph input

kaavyalakshmanan:


beetai (commented):
Is there a way to make the ph input box the same width as the notes input box?

JimmyRowland (approved):

PR #881, opened 2021-04-30
[F21-286] introduce farm map

beetai:


packages/webapp/src/components/Modals/NotifyUpdatedFarmModal/index.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import ModalComponent from '../ModalComponent/v2'
+import { useTranslation } from 'react-i18next';
+
+export default function UnableToRetireModal({ dismissModal }) {
```
kaavyalakshmanan, line 5: Change name of function to be NotifyUpdatedFarmModal

kaavyalakshmanan (commented):
2 things:

1. Not able to dismiss the new modal by clicking elsewhere, it only gets dismissed when I click the 'farm map'. 
2. Not able to dismiss 'my farm' spotlight or click anywhere except 'farm map'

JimmyRowland (approved):

PR #918, opened 2021-05-05
fix missing translation of error message

Tbrid:
Changed hardcoded error message to be using translations.


beetai (changes requested):
Let's use template literals instead of concatenation. (eg `` `${this.props.t('translation:FIELDS.EDIT_FIELD.CROP.CROP_COMMON_NAME')}, ` ``)

PR #938, opened 2021-05-12
Views for Crop detail and management

kikemarquez8:


beetai (changes requested):
I noticed that for the stories, the router tab does not highlight the corresponding active tab. Should also add eol whitespace

beetai (approved):

PR #998, opened 2021-05-21
fix s3 import

JimmyRowland:


JimmyRowland:
@aws-sdk/client-s3 is significantly larger than v2

kikemarquez8 (approved):

PR #1021, opened 2021-06-03
LF-1339/choose planting location

Tbrid:


packages/webapp/src/components/Crop/detail.js
```diff
@@ -27,6 +27,7 @@ function PureCropDetail({
     handleSubmit,
     watch,
     control,
+    setValue,
```
beetai, line 30: this is also unused, can probably just remove this too

packages/webapp/src/components/PlantingLocation/index.js
```diff
@@ -0,0 +1,89 @@
...
+PurePlantingLocation.prototype = {
+  selectedLocation: PropTypes.object,
+  onContinue: PropTypes.func,
+  onGoBack: PropTypes.func,
+  onCancel: PropTypes.func,
+  setSelectedLocation: PropTypes.func,
+  persistedFormData: PropTypes.func,
+  useHookFormPersist: PropTypes.func,
+  persistedPath: PropTypes.array,
+};
```
beetai, line 89: EOF whitespace pls

packages/webapp/src/components/PlantingLocation/index.js
```diff
@@ -0,0 +1,89 @@
...
+import PropTypes from 'prop-types';
+import PageTitle from '../PageTitle/v2';
+import Button from '../Form/Button';
+import ProgressBar from '../ProgressBar';
+import LocationPicker from '../LocationPicker';
+import { useTranslation } from 'react-i18next';
+import { useForm } from 'react-hook-form';
+import Form from '../Form';
+
+export default function PurePlantingLocation({
```
beetai, line 12: A couple comments on this file:
- recall we are no using this page as a form. with that being the case, we should use `<Layout/>` instead of `<Form/>`
- since there is no form to register, we can remove the use of react hook form in this component.
- make indentation consistent

packages/webapp/src/stories/CropManagement/Transplant.stories.js
```diff
@@ -0,0 +1,16 @@
...
+    component: Transplant,
+    decorators: decorators,
+};
+
+const Template = (args) => <Transplant {...args} />;
+
+export const TransplantForm = Template.bind({});
+Transplant.args = {
+    disabled: true,
+};
```
beetai, line 16: EOF whitespace pls

beetai (dismissed):

packages/webapp/src/components/PlantingLocation/index.js
Tbrid: Added

packages/webapp/src/stories/CropManagement/Transplant.stories.js
Tbrid: Added.

kikemarquez8 (approved):

PR #1052, opened 2021-06-08
Added "Can this be grown as a cover crop?" radio y/n to "Add a new cr…

audreyylu:
…op" modal for planting method crop management


packages/webapp/src/components/AddNewCrop/index.js
```diff
@@ -115,6 +117,11 @@ export default function PureAddNewCrop({
         )}
       />
 
+      <div style={{ marginBottom: '20px', fontSize: '16px' }}>{t('CROP_CATALOGUE.COVER_CROP')}</div>
+      <div style={{ marginBottom: '20px' }}>
+        <RadioGroup hookFormControl={control} name="cover_crop_radio"></RadioGroup>
```
kikemarquez8, line 122: I think this should be "can_be_cover_crop" which is the property that's defined in the "cropModel"  instead of cover_crop_radio

kikemarquez8 (changes requested):

kikemarquez8 (approved):

PR #1083, opened 2021-06-16
Lf 1276

audreyylu:


packages/api/src/middleware/validation/activeManagementPlanCheck.js
```diff
@@ -0,0 +1,29 @@
...
+
+async function activeManagementPlanCheck(req, res, next) {
+  const currentDate = new Date();
+  const {body} = req;   // body should be for cropVarietyModel??
+  if (body.crop_variety_id) {
+    const doesCropVarietalHaveAManagementPlan = await managementPlanModel.query().where({crop_variety_id: body.crop_variety_id})// check if there is a management plan that has the crop varietal
+    if (doesCropVarietalHaveAManagementPlan !== null) { // crop varietal is in a management plan
+      if (body.seed_date === null) { // seed date is null
+        return next();
+      }
```
kikemarquez8, line 12: This is not valid, but there is no need to check it here. Objection will check it for you

packages/api/src/middleware/validation/activeManagementPlanCheck.js
```diff
@@ -0,0 +1,29 @@
...
+      if ((body.seed_date.getTime() < currentDate.getTime()) &&
+        (body.termination_date.getTime() < currentDate) &&
+        (body.harvest_date.getTime() < currentDate)) { // active crop management plan
+        return res.status(400).send({message: 'Cannot delete! Active or future crop management plans exist for this varietal'});
```
kikemarquez8, line 16: There can't be a termination_date AND a harvest_date. termination dates exist only on cover crops and harvest_date on all other. So there should be 2 separate checks. Also it should be checking if the termination or harvest dates are in the _future_ not in the past. Something like this
```
(!!body.harvest_date && body.harvest_date.getTime()  > currentDate.getTime() ) ||  ( !!body.termination_date && body.termination_date.getTime() > currentDate.getTime() )
```

kikemarquez8 (changes requested):

kikemarquez8 (approved):

PR #1122, opened 2021-06-23
LF-1428/Patch - Fix document tile null type

Tbrid:


packages/webapp/src/containers/Documents/DocumentTile/index.js
```diff
@@ -14,23 +15,32 @@ export default function PureDocumentTile({
...
+        {type && (
+          <>
+            <div
+              className={styles.type}
+              style={{
+                marginTop: '4px',
+                marginBottom: date ? '4px' : '8px',
+              }}
+            >
+              {t(`DOCUMENTS.TYPE.${type}`)}
```
kikemarquez8, line 40: If the type is null, woudln't this do the same as it did before?

kikemarquez8 (approved):

PR #1218, opened 2021-07-15
Lf 1464 pin button and logic

audreyylu:
subtask LF-1700: drop a pin and remove pin components


packages/webapp/src/containers/Crop/AddManagementPlan/PlantingLocation/index.js
```diff
@@ -48,24 +48,36 @@ export default function PlantingLocation({ history, match }) {
...
-      }
-    } else {
+    if (isWildCrop && !isTransplant) {
       dispatch(setPlantingLocationIdManagementPlanFormData(selectedLocationId));
-      history.push(`/crop/${variety_id}/add_management_plan/planting_method`);
+      history.push(`crop/${variety_id}/add_management_plan/name`);
+    } else if (isWildCrop && isTransplant) {
+      dispatch(setPlantingLocationIdManagementPlanFormData(selectedLocationId));
+      history.push(`/crop/${variety_id}/add_management_plan/choose_transplant_location`);
+    } else {
```
kikemarquez8, line 79: I think this else would go to planting_location, not inground_transplant_method. also you are missing the setPlantingLocation here.

kikemarquez8 (approved):

PR #1242, opened 2021-07-21
1449 changes to planting date component

kikemarquez8:


packages/webapp/src/components/Crop/PlantedAlready/constants.js
```diff
@@ -0,0 +1,4 @@
+export const seedingType = {
+  "SEEDLING" : "SEEDLING_OR_PLANTING_STOCK",
+  "SEED" :"SEED"
+}
```
beetai, line 4: eof whitespace

packages/webapp/src/components/MonthCalendar/index.js
```diff
@@ -0,0 +1,111 @@
...
+            }
+            {[ ...Array(nextMonthDays) ].map(() => (
+              <div style={{ flex: '1 0 13%', marginTop: '10px', alignSelf: 'center' }}/>
+            ))}
+          </div>
+
+        </div>
+      </div>
+    </>
+  );// 08-30-2021
```
beetai, line 100: comment

beetai (changes requested):
There are two situations that do not fully line up with the mock ups and need to be checked:
- In Ground | Not Wild | Harvest | Transplant
- In Ground | Not Wild | Cover Crop| Transplant
In these cases, the plant date (green) is shown in the calendar. If this is removed, it should be all good.

beetai (approved):

PR #1268, opened 2021-08-03
 Fixing middleware depending on logs. Adding POST endpoint for tasks

kikemarquez8:


packages/api/src/middleware/acl/hasFarmAccess.js
```diff
@@ -191,35 +192,36 @@ async function fromActivity(req) {
...
 }
 
 async function fromSale(sale_id) {
-  return await knex('sale').where({ sale_id }).first();
+  return knex('sale').where({ sale_id }).first();
 }
 
 async function fromOrganicCertifierSurvey(survey_id) {
-  return await knex('organicCertifierSurvey').where({ survey_id }).first();
+  return knex('organicCertifierSurvey').where({ survey_id }).first();
```
beetai, line 224: If you are removing `await` in these functions, would you also remove the function's `async` declaration? or does it stay because the function call after the `return` is a promise?

packages/api/src/models/taskModel.js
```diff
@@ -31,7 +31,7 @@ class TaskModel extends BaseModel {
   static get jsonSchema() {
     return {
       type: 'object',
-      required: ['task_id', 'due_date', 'type'],
+      required: ['due_date', 'type'],
```
beetai, line 34: why is task_id no longer required?

packages/api/src/controllers/taskController.js
```diff
@@ -39,6 +39,56 @@ const taskController = {
...
+  createTask(typeOfTask) {
+    const nonModifiable = getNonModifiable(typeOfTask);
+    return async (req, res, next) => {
+      try {
+        // OC: the "noInsert" rule will not fail if a relationship is present in the graph.
+        // it will just ignore the insert on it. This is just a 2nd layer of protection
+        // after the validation middleware.
+        const result = await TaskModel.transaction(async trx =>
+          await TaskModel.query(trx).context({ user_id: req.user.user_id })
+            .upsertGraph(req.body, { noUpdate: true, noDelete: true, noInsert: nonModifiable, relate: ['locations', 'managementPlans'] })
```
beetai, line 52: can probably break this over multiple lines

beetai (approved):

packages/api/src/middleware/acl/hasFarmAccess.js
kikemarquez8: you are right, now they can just be functions

packages/api/src/models/taskModel.js
kikemarquez8: it cant be required for posts (because it doesn't exist) it follows the same pattern as all other models

PR #1269, opened 2021-08-03
Lf 1570

audreyylu:


kikemarquez8 (approved):

kikemarquez8:
I don't love the name for the "TaskNote" component, but we can change that later

PR #1385, opened 2021-08-24
move moment

JimmyRowland:


packages/webapp/src/components/TaskCard/index.js
```diff
@@ -81,10 +79,7 @@ const PureTaskCard = ({
   if (!locations.length) console.error('Task should be associated with at least one location');
   const locationText = locations.length > 1 ? t('TASK.CARD.MULTIPLE_LOCATIONS') : locations[0].name;
   const cropText = crops.length > 1 ? t('TASK.CARD.MULTIPLE_CROPS') : t(`crop:${crops[0]}`);
-  const dateText = moment(new Date(dueDate))
-    .locale(getLanguageFromLocalStorage())
-    .utc()
-    .format('MMM D, YYYY');
+  const dateText = getTaskCardDate(dueDate, 'MMM D, YYYY');
```
beetai, line 82: Looks to me like this function has one parameter

beetai (changes requested):

PR #1399, opened 2021-08-29
LF-1820/Patch - Fix extra harvest_tasks persisting for other types

Tbrid:


JimmyRowland:
It's better to handle this before post.

Tbrid:
But then it will leave this empty array in the persistedFormData

PR #1405, opened 2021-08-30
LF-1847/Harvest Tasks posting

Tbrid:


packages/webapp/public/locales/en/translation.json
```diff
@@ -1043,7 +1043,7 @@
     "SEED_OR_SEEDLING": "How will you plant this crop?",
     "SEEDING_DATE": "Seeding date",
     "SEEDLING": "Seedling or planting stock",
-    "SEEDLING_AGE": "What age is the seedling or planting stock?",
+    "SEEDLING_AGE": "What age is the the seedling or planting stock?",
```
beetai: remove the second "the"

packages/webapp/public/locales/es/common.json
```diff
@@ -14,7 +14,7 @@
   "EDIT": "Editar",
   "EDIT_DATE": "Editar fecha",
   "EG": "Ejemplo",
-  "EXPORT": "Exportar",
+  "EXPORT": "MISSING",
```
beetai: should these translations be removed? same with other translation files

beetai: it looks like Spanish translations are being removed and Portuguese are being reverted. is this intentional?

packages/webapp/src/containers/AddTask/TaskAssignment/index.js
```diff
@@ -61,8 +65,30 @@ export default function TaskManagement({ history, match }) {
...
+    let harvestTasks = [];
+    for (let harvest_task of taskData.harvest_tasks) {
+      let id = harvest_task.id.split('.');
+      let location = id[0];
+      let managementPlan = id[1];
+      let h = {};
+      h.location_id = location;
+      h.management_plan_id = Number(managementPlan);
+      h.harvest_everything = harvest_task.harvest_everything;
+      h.quantity = harvest_task.quantity === undefined ? 0 : harvest_task.quantity;
```
beetai, line 78: could probably stick with just checking `harvest_task.quantity` rather than using `=== undefined`. same applies to the line below here

beetai (changes requested):

beetai (approved):
you could choose to address a comment I made in the last review but apart from that looks good

packages/webapp/src/containers/AddTask/TaskAssignment/index.js
Tbrid: Not sure what you mean. Like just `harvest_task.quantity ? .... : ... `  ?

beetai: Yea. Would that work?

Tbrid: No, that would check it's null only.

beetai: alright then, can just resolve this

PR #1468, opened 2021-09-21
ST-4/show-plant-task-spotlight-on-tile-click

JimmyRowland:


packages/webapp/src/containers/Task/TaskTypeSelection/index.js
```diff
@@ -17,7 +17,7 @@ function TaskTypeSelection({ history, match }) {
   const continuePath = '/add_task/task_date';
   const customTaskPath = '/add_task/manage_custom_tasks';
   const persistedPaths = [continuePath, customTaskPath];
-  const { planting_task } = useSelector(showedSpotlightSelector);
+  const { planting_task: shouldNotShowPlantTaskSpotLight } = useSelector(showedSpotlightSelector);
```
beetai, line 20: as we discussed before, this it no longer a spotlight. i imagine you should probably make changes in the backend to remove the column.

beetai (changes requested):

beetai (approved):

PR #1476, opened 2021-09-21
fix transplant task container note margin

JimmyRowland:


packages/webapp/src/components/Task/TaskReadOnly/index.js
```diff
@@ -206,12 +187,8 @@ export default function PureTaskReadOnly({
...
             </div>
           )}
-          {(!task.happiness) && (
-            <Checkbox
-              label={t('TASK.PREFER_NOT_TO_SAY')}
-              disabled={true}
-              defaultChecked={true}
-            />
+          {!task.happiness && (
+            <Checkbox label={t('TASK.PREFER_NOT_TO_SAY')} disabled={true} defaultChecked={true} />
```
beetai, line 191: for boolean props, we can just put the prop name for true.
e.g.: `disabled={true}` -> `disabled`

beetai (approved):
pretty much ready to approve, just one small thing

PR #1489, opened 2021-10-05
patch-management-start-date-on-plant-task-completion

JimmyRowland:


packages/api/src/controllers/taskController.js
```diff
@@ -237,22 +237,41 @@ const taskController = {
...
           return res.status(403).send('Not authorized to complete other people\'s task');
         }
-        const result = await TaskModel.transaction(async trx =>
-          await TaskModel.query(trx).context({ user_id: req.user.user_id })
-            .upsertGraph({ task_id: parseInt(task_id), ...data }, {
-              noUpdate: nonModifiable,
-              noDelete: true,
-              noInsert: true,
-            }),
+        const result = await TaskModel.transaction(async trx => {
```
beetai, line 240: would the code be cleaner if you initialized a trx, followed by a `.commit` or `.rollback` depending on the result? i reckon that would be less nesting if that was the case.

beetai (changes requested):
the code within the trx should be properly indented. apart from that, just got a comment to ask about trx

beetai (approved):

PR #1607, opened 2021-10-29
ST-352/Add required error on supplier of product

Tbrid:


packages/webapp/src/components/Task/AddProduct/index.js
```diff
@@ -105,6 +105,7 @@ const AddProduct = ({
         style={{ marginBottom: '40px' }}
         disabled={disabled}
         hasLeaf={true}
+        errors={watch(SUPPLIER).match(/^\s*$/) && t('common:REQUIRED')}
```
beetai, line 108: You should use react hook form's validation functionality for regex. If I recall correctly, you can designate conditions for the input's validation in its register call.

beetai (changes requested):

packages/webapp/src/components/Task/AddProduct/index.js
beetai: if you're not checking for regex, but rather whether an input is there or not, you should use `getInputErrors()`, (passing in the proper parameters), and ensure that the input is required when registered.

beetai (approved):

PR #1650, opened 2021-11-02
St 428/cache map max zoom

JimmyRowland:


packages/webapp/src/components/Task/TaskReadOnly/index.js
```diff
@@ -130,12 +132,15 @@ export default function PureTaskReadOnly({
         />
       )}
       <LocationPicker
-        onSelectLocation={() => {}}
+        onSelectLocation={() => {
+        }}
```
beetai, line 136: does this function need to be here?

packages/webapp/src/containers/Map/useMapAssetRenderer.js
```diff
@@ -124,7 +126,8 @@ const useMapAssetRenderer = ({ isClickable }) => {
 
     markerCluster.addMarkers(markers, true);
     maps.event.addListener(markerCluster, 'click', (cluster) => {
-      if (map.getZoom() >= 20 && cluster.markers_.length > 1) {
+      console.log(maxZoomRef?.current);
```
beetai, line 129: console log

beetai (approved):

PR #1673, opened 2021-11-23
get client ip

JimmyRowland:


smattingly (approved):
So, I take it that the bug fix is to use the first item (the client) from the list, instead of the last, which is probably a proxy.

smattingly:
@JimmyRowland Is the usual practice for approving reviewer to merge it, even when it is not a story?

PR #1676, opened 2021-11-24
LF-2066

smattingly:
- [x] create migration
- [x] add POST routes
- [x] figure out access control
- [x] front end work


smattingly:
@JimmyRowland Here is WIP on a ticket that is listed as a blocker for one of yours.

smattingly:
Sorry @JimmyRowland but I changed the URL routing when I realized how to better match other routing.

smattingly:
Need feedback on this please. I waited to implement similar changes for garden and greenhouse, in case you guys show me a much better way to do this. On the other hand, this is temporary in that a fuller UI will later drive the organic history requests. (See LF-2066)

JimmyRowland (approved):

PR #1679, opened 2021-11-26
Lf 2045/record a

JimmyRowland:


packages/api/src/jobs/certification/record_a_generation.js
```diff
@@ -0,0 +1,145 @@
...
+        fill: 'D9D9D9',
+      });
+
+
+      workbook.sheet(0).cell('A1').value(t('RECORD_A.HEADER'));
+      workbook.sheet(0).cell('A2').value(`${t('RECORD_A.OPERATION_NAME')}:`);
+      workbook.sheet(0).cell('B2').value(farm_name);
+
+      workbook.sheet(0).cell('K2').value(`${t('RECORD_A.YEAR')}:`);
+      workbook.sheet(0).cell('L2').value(new Date().getFullYear());
```
smattingly, line 78: I think this is supposed to contain the date range.

packages/api/src/controllers/organicCertifierSurveyController.js
```diff
@@ -259,6 +265,75 @@ const organicCertifierSurveyController = {
...
+
+    const booleanTrueToX = bool => bool ? 'x' : '';
+    return locations.map(location => {
+      const locationOrganicStatus = location[location.figure.type].organic_status;
+      return ({
+        name: location.name,
+        crops: Array.from(locationIdCropMap[location.location_id] || []),
+        area: location.figure?.area?.total_area || location.figure?.line?.total_area || 0,
+        isNew: booleanTrueToX(new Date(location.created_at).getTime() > fromDateTime),
+        isTransitional: booleanTrueToX(locationOrganicStatus === 'Transitional'),
```
smattingly, line 329: LF-2045 describes other logic for determining transitional/organic/non-organic status.

packages/api/src/controllers/organicCertifierSurveyController.js
```diff
@@ -259,6 +265,75 @@ const organicCertifierSurveyController = {
...
+        ]`);
+
+    const booleanTrueToX = bool => bool ? 'x' : '';
+    return locations.map(location => {
+      const locationOrganicStatus = location[location.figure.type].organic_status;
+      return ({
+        name: location.name,
+        crops: Array.from(locationIdCropMap[location.location_id] || []),
+        area: location.figure?.area?.total_area || location.figure?.line?.total_area || 0,
+        isNew: booleanTrueToX(new Date(location.created_at).getTime() > fromDateTime),
```
smattingly, line 328: It's currently possible to generate a spreadsheet where `isNew` (column F) is marked with X, as well as one of G-K. But the text of the form seems to say F-K should be mutually exclusive.

smattingly (changes requested):
I think some of the logic is different that what LF-2045 specifies. Specific comments mark those places.

smattingly (approved):
I've checked this as carefully as I can. Some of the knex logic is very hard to follow, and I don't fully understand the domain logic (for example, how transplanting affects things). I really think that there should be tests for at least some of this logic. 
I don't feel that I know enough to be confident that this is ready for production, so I'm going to put that decision in Jimmy's hands by Approving the PR but letting him choose whether to merge it as is.

smattingly:
When this is merged, watch out for the path to the xls which is using the farm name. I think that the change to use a uuid instead has already been merged to `develop`.

PR #1684, opened 2021-11-30
[Lf-2059] Label Record I with user's preferred unit

smattingly:


JimmyRowland:
Will recreate this pr once record a is merged.
need to pass in measurement instead of units and all mass is stored as kg in db

PR #1685, opened 2021-11-30
add organic history on location edit and creation

JimmyRowland:


packages/api/db/migration/20211122153642_organic_history.js
```diff
@@ -7,17 +7,17 @@ exports.up = async function (knex) {
     t.uuid('location_id')
       .references('location_id')
       .inTable('location').notNullable();
-    t.enu('to_state', ['Non-Organic', 'Transitional', 'Organic']).defaultTo('Non-Organic');
-    t.dateTime('effective_date').notNullable().defaultTo(knex.fn.now());
-    t.boolean('deleted').defaultTo(false);
+    t.enu('organic_status', ['Non-Organic', 'Transitional', 'Organic']).defaultTo('Non-Organic');
+    t.date('effective_date').notNullable();
```
JimmyRowland, line 11: At some point, we decided to use date instead of DateTime for date input by user

packages/api/db/migration/20211122153642_organic_history.js
```diff
@@ -7,17 +7,17 @@ exports.up = async function (knex) {
...
     t.string('created_by_user_id').references('user_id').inTable('users');
     t.string('updated_by_user_id').references('user_id').inTable('users');
     t.dateTime('created_at').notNullable();
     t.dateTime('updated_at').notNullable();
   });
 
-  const query = `insert into organic_history (location_id, to_state, effective_date, created_at, updated_at)
-    select location_id, organic_status, '2000-01-01', NOW(), NOW() from`;
+  const query = `insert into organic_history (location_id, organic_status, effective_date, created_at, updated_at, created_by_user_id, updated_by_user_id)
+    select location_id, organic_status, '2000-01-01', NOW(), NOW(), '1', '1' from`;
```
JimmyRowland, line 20: user_id: '1' is used for migration

packages/api/src/controllers/locationController.js
```diff
@@ -66,7 +66,7 @@ const LocationController = {
         const result = await LocationModel.transaction(async trx => {
           return await LocationModel.query(trx).context({ user_id: req.user.user_id }).upsertGraph(
             { ...req.body, location_id: req.params.location_id },
-            { noInsert: true, noDelete: true, noUpdate: nonModifiable });
+            { noInsert: [...figures, ...assets, 'figure'], noDelete: true, noUpdate: nonModifiable });
```
JimmyRowland, line 69: a new organic_history entry will be created when req.body include organic_history
req.body: {...location, field: {...field, organic_history: {organic_status, effective_date}}}
We can also remove organic_history from req.body and add it in post/put endpoints, but effective date might be off by one day because of timezone difference

packages/api/src/controllers/organicHistoryController.js
```diff
@@ -13,118 +13,20 @@
...
 
 module.exports = {
-  async addEntry(req, res) {
-    if (!locationModel.isCropEnabled(req.body.location_id)) {
-      return res.status(400).send('Location must be crop enabled.');
-    }
 
-    const trx = await transaction.start(Model.knex());
-    try {
-      const result = await baseController.postWithResponse(model, req.body, req, { trx });
```
JimmyRowland, line 29: baseController should be deprecated and const trx = await transaction.start(Model.knex()); syntax causes error when await trx.commit(); or await trx.rollback(); are missing

JimmyRowland: Model.transaction() is better
https://vincit.github.io/objection.js/guide/transactions.html#creating-a-transaction

packages/api/src/middleware/validation/organicHistoryLocationCheck.js
```diff
@@ -0,0 +1,16 @@
+const locationModel = require('../../models/locationModel');
+const organicHistoryLocations = [
+  'field', 'garden', 'greenhouse',
+];
+
+async function organicHistoryLocationCheck(req, res, next) {
+  const location = await locationModel.query().join('figure', 'figure.location_id', 'location.location_id')
+    .where('location.location_id', req.body.location_id)
+    .whereIn('type', organicHistoryLocations);
+  if (!location.length) return res.status(400).send('Location must be crop enabled.');
```
JimmyRowland, line 10: validation should be in middleware

packages/api/src/models/fieldModel.js
```diff
@@ -44,6 +44,14 @@ class Field extends Model {
   static get relationMappings() {
     // Import models here to prevent require loops.
     return {
+      organic_history: {
+        modelClass: require('./organicHistoryModel'),
+        relation: Model.HasManyRelation,
+        join: {
+          from: 'field.location_id',
+          to: 'organic_history.location_id',
```
JimmyRowland, line 52: need relationMappings so that upsertGraph can insert new organic_history while adding/updating location

packages/api/src/models/locationModel.js
```diff
@@ -190,15 +190,6 @@ class Location extends baseModel {
...
   }
-
-  static async isCropEnabled(locationId) {
-    const location = await this.query().whereNotDeleted().findById(locationId)
-      .withGraphJoined(`[
-           field, garden, greenhouse
-        ]`);
-
-    return location?.field || location?.garden || location?.greenhouse;
-  }
```
JimmyRowland, line 201: I don't think we should complicate objectionModels

packages/api/src/models/organicHistoryModel.js
```diff
@@ -31,35 +30,17 @@ class OrganicHistory extends baseModel {
...
       },
       additionalProperties: false,
     };
   }
-
-  static get relationMappings() {
-    // Import models here to prevent require loops.
-    return {
-      location: {
-        relation: Model.BelongsToOneRelation,
```
JimmyRowland, line 50: We only need relational mapping when we want to insert/update location using organicHistoryModel.upsertGraph. Then upsertGraph will be able to insert/update all tables that linked to locationModel. Should add relationMappings with caution. 
https://vincit.github.io/objection.js/api/query-builder/mutate-methods.html#upsertgraph

packages/api/src/routes/locationRoute.js
```diff
@@ -78,8 +79,9 @@ router.post('/garden', hasFarmAccess({ body: 'farm_id' }), checkScope(['add:gard
 router.post('/farm_site_boundary', hasFarmAccess({ body: 'farm_id' }), checkScope(['add:farm_site_boundary']),
   modelMapping['farm_site_boundary'], createLocation('farm_site_boundary'));
 
-router.post('/:location_id/organic_history', hasFarmAccess({ params: 'location_id' }), checkScope(['add:organic_history']),
-  organicHistoryController.addEntry);
+
+router.post('/organic_history', hasFarmAccess({ body: 'location_id' }), checkScope(['add:organic_history']),
```
JimmyRowland, line 83: foreign key should be passed through body for post req

packages/webapp/src/containers/fieldSlice.js
```diff
@@ -11,7 +12,10 @@ export const getLocationObjectFromField = (data) => {
       ...pick(data, figureProperties),
       area: pick(data, areaProperties),
     },
-    field: pick(data, fieldProperties),
+    field: {
+      ...pick(data, fieldProperties),
+      organic_history: { effective_date: getDateInputFormat(), organic_status: data.organic_status },
```
JimmyRowland, line 17: add organic_history in post/put req

packages/api/src/controllers/organicHistoryController.js
```diff
@@ -13,118 +13,20 @@
...
-      await trx.commit();
-      res.status(201).send(result);
-    } catch (error) {
-      await trx.rollback();
-      res.status(400).json({
-        error,
-      });
-    }
+  addOrganicHistory() {
+    return async (req, res, next) => {
```
smattingly, line 22: Why does LF frequently complicate things by returning a function when it's not necessary? Is there some advantage to this?

packages/api/src/middleware/validation/organicHistoryLocationCheck.js
smattingly: I'm ok with this for now, but apart from LF convention I don't think that validation must necessarily go in middleware. This change complicates the codebase without gaining anything. I don't think there's anyplace else this would be re-used, and if there were it would likely be more useful as "crop enabled".

packages/api/src/models/locationModel.js
smattingly: Why not? MVC puts data and application logic primarily in models. I'd like to see future LF codebase eliminate database manipulation from controllers.

packages/api/src/routes/locationRoute.js
smattingly: Is there a source that LF uses to define REST approaches? I was following this one: http://microformats.org/wiki/rest/urls

smattingly (commented):
Wow, you did a lot on this-- thanks! I knew there had to be a better way, but you could have just pointed me toward the database graph operations. Anyway, all the stuff that makes it work is great. I'm not submitting an "Approve" review only because I want to start a conversation about some of the other things I've commented on. There are some LF coding practices that I'd like to change, and your modifications touch on some of them.

PR #1687, opened 2021-11-30
LF-2046/pdf-style-patch

JimmyRowland:


smattingly (approved):

smattingly:
@JimmyRowland I did not move LF-2046 to QA because I was unsure if this was the complete work for it. So feel free to move it if appropriate.

PR #1688, opened 2021-12-01
LF-2059 rebased

JimmyRowland:


beetai (approved):

packages/api/src/jobs/certification/record_i_generation.js
```diff
@@ -82,7 +84,11 @@ module.exports = (data, exportId, from_date, to_date, farm_name, isInputs) => {
       workbook.sheet(0).cell('A9').value(rowNine).style({ wrapText: false });
       workbook.sheet(0).cell('A10').value(t('RECORD_I.TABLE_COLUMN.PRODUCT_NAME'));
       workbook.sheet(0).cell('B10').value(t('RECORD_I.TABLE_COLUMN.SUPPLIER'));
-      workbook.sheet(0).cell('C10').value(t('RECORD_I.TABLE_COLUMN.QUANTITY'));
+      workbook.sheet(0).cell('C10').value(t('RECORD_I.TABLE_COLUMN.QUANTITY', {
+        unit: measurement === 'metric' ?
+          'kg' :
```
smattingly, line 89: I think this removes a translation that was in the original PR? I know that the en/es/pt translations of 'kg' and 'lb' are all the same, but per David the translation could be valuable if we later have non-latin alphabets (see Jira).

smattingly (approved):

PR #1694, opened 2021-12-01
add 404 on get organic history by location id

JimmyRowland:


packages/api/src/middleware/validation/organicHistoryLocationCheck.js
```diff
@@ -4,14 +4,6 @@ const organicHistoryLocations = [
   'field', 'garden', 'greenhouse',
 ];
 
-async function organicHistoryLocationCheckOnGet(req, res, next) {
-  const location = await locationModel.query().join('figure', 'figure.location_id', 'location.location_id')
-    .where('location.location_id', req.params.location_id)
-    .whereIn('type', organicHistoryLocations);
-  if (!location.length) return res.status(400).send('Location must be crop enabled.');
```
JimmyRowland, line 11: We can return 404 if there is no organic history. I think either way is fine.

smattingly (approved):

PR #1713, opened 2021-12-08
fix is planting method known prefix

JimmyRowland:


packages/webapp/package.json
```diff
@@ -138,7 +138,7 @@
     "i18next-parser": "^3.6.0",
     "jest-environment-jsdom-fourteen": "1.0.1",
     "jest-resolve": "24.9.0",
-    "jest-watch-typeahead": "^0.6.5",
+    "jest-watch-typeahead": "^1.0.0",
```
smattingly, line 141: The tests won't run with that version.
screenshot-6bf5d6db

smattingly (changes requested):

packages/webapp/package.json
JimmyRowland: Couldn't find a fix. Rollback to 0.6.5

smattingly (approved):

PR #1714, opened 2021-12-08
fix add farm location validation

JimmyRowland:


packages/webapp/src/containers/AddFarm/index.js
```diff
@@ -253,6 +245,15 @@ const AddFarm = () => {
     navigator.geolocation.getCurrentPosition(handleGetGeoSuccess, handleGetGeoError, getGeoOptions);
   };
 
+  //TODO: update hookform, move all states in hookform, and rewrite the component
+  useEffect(() => {
+    if (farmAddress) {
+      setTimeout(() => {
```
smattingly, line 251: `setTimeout` may improve the UI behavior, but I think that there is probably a deeper and better fix by reworking some of the container. I'm working on that.

smattingly (commented):
I've already made some good progress by reworking some more of the container. I'll ask you to review that when I've taken it as far as I can.

PR #1716, opened 2021-12-14
[ST-392] Consider the expenses which occurred on the final date of the report ("To")

hernandcb:
The time of the "To" date is 00:00, that's why it's necessary to take out the time in the date of expenses to consider it.


packages/webapp/src/containers/Finances/OtherExpense/index.js
```diff
@@ -79,9 +79,12 @@ class OtherExpense extends Component {
     let dict = {};
 
     for (let e of expenses) {
+      // Remove the time from the expense date
+      const expenseDate = moment(moment(e.expense_date).format("MM/DD/YYYY"))
+
       if (
-        moment(e.expense_date).utc().isSameOrAfter(moment(startDate)) &&
-        moment(e.expense_date).utc().isSameOrBefore(moment(endDate))
+        moment(expenseDate).utc().isSameOrAfter(moment(startDate)) &&
```
smattingly, line 86: Line 83 ensures `expenseDate` is a `moment`, so I think there's no need to call `moment()` on `expenseDate` here, or on the next line.

smattingly (changes requested):
I believe this fixes the problem in the Summary table of the Other Expenses screen, but not the Detailed History on that screen. I think the same problem exists with Labour costs, and possibly also with Revenue. So, I think a similar fix is needed in other places.

hernandcb:
I changed the way to remove the time on the dates and also applied the same changes to all the other filters.

hernandcb:
Thank you for the suggestions. Indeed, it is more clearer and cleaner in that way. 
Sorry I didn't check that out in the library before.

packages/webapp/src/containers/Finances/index.js
```diff
@@ -162,15 +162,17 @@ class Finances extends Component {
   }
 
   getTotalExpense = () => {
+    console.log('///////// Entrando')
```
smattingly, line 165: Please remove this.

smattingly (changes requested):
Requesting one small change ...

packages/webapp/src/containers/Finances/index.js
hernandcb: Done

smattingly (commented):
:+1:

smattingly (approved):

PR #1717, opened 2021-12-14
[ST-387] Setting as relative the sizes of fields in the "Add new income" interface

hernandcb:


smattingly:
@hernandcb This is fine with me, but in Jira I asked for feedback from our QA. If they interpret the stated expectations strictly, they may just bounce it back, so I thought it better to ask proactively.

JimmyRowland (approved):

JimmyRowland:
All the Financial pages need to be reworked. Code quality doesn't really matter

PR #1721, opened 2021-12-16
[ST-388]  Adding crop's name on "Actual Revenue" 

hernandcb:
Added the name of the crop as complement to the Crop variety name in the Actual revenue page.
I used the same pattern used in the estimated crop revenue.

### DIFF
--- a/packages/webapp/src/containers/Finances/ActualCropRevenue/index.js
+++ b/packages/webapp/src/containers/Finances/ActualCropRevenue/index.js
@@ -3,21 +3,27 @@ import PropTypes from 'prop-types';
 import { useDispatch, useSelector } from 'react-redux';
 import FinanceGroup from '../../../components/Finances/FinanceGroup';
 import { getManagementPlanCardDate } from '../../../util/moment';
-import { cropVarietyEntitiesSelector } from '../../cropVarietySlice';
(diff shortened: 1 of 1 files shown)


packages/webapp/src/containers/Finances/ActualCropRevenue/index.js
```diff
@@ -35,6 +41,11 @@ const ActualCropRevenue = ({ sale, history, ...props }) => {
         const convertedQuantity = roundToTwoDecimal(
           convertFromMetric(cvs.quantity.toString(), cvs.quantity_unit, 'kg').toString(),
         );
+        const cropVariety =  cropVarietyEntities[cvs.crop_variety_id];
+        const title = cropVariety.crop_variety_name
```
smattingly, line 45: The value of `title` is never used. You need to use it in the returned object.

smattingly (changes requested):
You've put together the proper string, but haven't made it display.

hernandcb:
Oh, I don't know why I sent it like that. Fixed it, also added the key prop to avoid a React warning that was appearing.

JimmyRowland:
Couldn't push onto this branch. Moved this pr here https://github.com/LiteFarmOrg/LiteFarm/pull/1733

PR #1725, opened 2021-12-18
fix non-producing in getLocationOrganicStatus

JimmyRowland:


smattingly:
Closing this because discussion of the Record A issues has clarified and extended requirements. The things addressed here will be addressed in later PR(s).

PR #1727, opened 2021-12-28
TDC-54/fix-store-racing-conditions

JimmyRowland:


packages/api/src/controllers/cropController.js
```diff
@@ -101,7 +101,7 @@ const cropController = {
           })
           : await cropController.get(farm_id);
         if (!rows.length) {
-          res.status(200).send(rows);
+          res.sendStatus(404);
```
smattingly, line 104: Should respond with 200 and empty array.

packages/api/src/controllers/cropVarietyController.js
```diff
@@ -20,7 +20,7 @@ const cropVarietyController = {
       const { farm_id } = req.params;
       try {
         const result = await CropVarietyModel.query().whereNotDeleted().where({ farm_id });
-        return res.status(200).send(result);
+        return result?.length ? res.status(200).send(result) : res.sendStatus(404);
```
smattingly, line 23: If result is zero records, should respond with 200 and empty array.

smattingly (changes requested):
Requests to RESTful APIs that result in "zero records" should get a response of 200 and an empty array. In spite of the "Not found" wording, 404 is not appropriate because 4xx tells the client it made an error and that the server did not successfully process the request. Neither is true in these scenarios.  (I've noticed LF doing this elsewhere, and have it on a list of things to address.)

smattingly (approved):

PR #1728, opened 2021-12-29
fix management plan form history stack

JimmyRowland:


packages/api/src/controllers/taskController.js
```diff
@@ -106,8 +106,8 @@ const taskController = {
           happiness,
           duration,
           wage_at_moment: override_hourly_wage ? wage_at_moment : wage.amount,
-        });
-        return result ? res.sendStatus(200) : res.status(404).send('Task not found');
+        }).returning('*');
+        return result ? res.status(200).send(result) : res.status(404).send('Task not found');
```
smattingly, line 110: Should respond with 200 and empty array.

smattingly (changes requested):
Small change: 404 -> 200

packages/api/src/controllers/taskController.js
JimmyRowland: It's a patch endpoint.

smattingly: Oh, I see. In that case I wonder if it is safe to assume that a falsy `result` is necessarily due to not finding the task? But I'll leave that up to you.

smattingly (approved):

PR #1736, opened 2021-12-30
fix isHarvest flag

JimmyRowland:


packages/webapp/src/components/Task/PureTaskDetails/index.js
```diff
@@ -31,7 +32,7 @@ export default function PureTaskDetails({
   const taskType = selectedTaskType.task_translation_key;
   const taskName = selectedTaskType.task_name;
   const isCustomType = !!selectedTaskType.farm_id;
-  const isHarvest = taskType === 'HARVEST_TASK';
+  const isHarvest = isTaskType('HARVEST_TASK');
```
smattingly, line 35: Is this what you intended? All other calls to `isTaskType` seem to pass two arguments.

smattingly (changes requested):
I marked a statement that I wasn't sure about. If you confirm it's good and resolve conversation, I'll approve PR.

smattingly (approved):

PR #1739, opened 2021-12-31
ST-391/403-handler

JimmyRowland:


packages/webapp/src/containers/ErrorHandler/saga.js
```diff
@@ -0,0 +1,35 @@
...
+import { getUserFarmsSaga } from '../ChooseFarm/saga';
+import { logout } from '../../util/jwt';
+
+
+export const handle403 = createAction(`handle403Saga`);
+
+export function* handle403Saga() {
+  yield call(getUserFarmsSaga);
+  const { has_consent } = yield select(userFarmSelector);
+  if (has_consent) {
```
smattingly, line 28: I don't understand all the details, but it seems strange to say "if we have the users consent (to the terms of use) then log them out". Maybe I'm interpreting that wrong, or it makes sense in this particular context.

smattingly (commented):
I posted a question. If you tell me it's OK as is, I'll approve.

packages/webapp/src/containers/ErrorHandler/saga.js
JimmyRowland: it first sync store with db then
It checks whether 403 is from has_consent === false. 
If status === 'inactive' or store info is incorrect fetch requests also return 403, then user will be logged out.

JimmyRowland: Should be fine

smattingly (approved):

PR #1753, opened 2022-01-06
Tdc 24 record a test

JimmyRowland:


smattingly (approved):
@JimmyRowland Just a few things, FYI. Since the tests are OK on the build server, I assume these are no problem. But I'm going to leave the merging to you just in case you want to look at anything.
1. For some reason, running only `organicCertifierController.test.js` takes a long time on my laptop-- I thought it had hung.
2. Running the entire back end test, some tests fail-- not from Record A stuff. I had this happen once in a while before, but now it happens consistently. 
3. GitHub is acting a bit weird about the tracking. It tells me 6 files were changed, but for a long time it would only identify and diff 2 of them. But it is now showing me all six.

PR #1758, opened 2022-01-07
[LF-2073] Improved translation on variety creation flow

hernandcb:


smattingly (changes requested):
If I understand correctly, this needs to modify the translation files for all four languages. Per LF-2073, put 'MISSING' in for Portuguese, and also for French and Spanish unless you can provide those translations.

smattingly (approved):
:+1:

PR #1760, opened 2022-01-07
[LF-2002] - do_not_email attribute in users

hernandcb:


hernandcb:
@smattingly 
Could you, please, guide me with following questions? 

- There is any test to send emails? 
  If so, should I update to test that the app doesn't send emails according to the user preference?

- This variable is used only in the backend, right?
  Otherwise I should include it in the variables that returned by the controller 

- Is this one of the cases where the variable should be camelCase in the backend code?
  the db column is do_not_email but I'm not sure if in the Js model it should be doNotEmail or same as in the DB

packages/api/db/migration/20220107134709_column_not_email_users.js
```diff
@@ -0,0 +1,32 @@
+/*
+ *  Copyright (C) 2007 Free Software Foundation, Inc. <https://fsf.org/>
+ *  This file (20190416183353_add_column_in_fertilizer.js) is part of LiteFarm.
```
smattingly, line 3: Remove the filename from this line. It is not needed in the copyright notice, and it is often wrong due to starting with a copy of some other file.

packages/api/db/migration/20220107134709_column_not_email_users.js
```diff
@@ -0,0 +1,32 @@
+/*
+ *  Copyright (C) 2007 Free Software Foundation, Inc. <https://fsf.org/>
```
smattingly, line 2: This line should read `Copyright 2019-2022 LiteFarm.org`

smattingly:
@hernandcb 

- There is any test to send emails? If so, should I update to test that the app doesn't send emails according to the user preference?
  - Not sure about the details of such tests, honestly. Also not sure how this is going to be used-- which is apparently out of scope for this ticket. (Since you have just created this flag, there is no code that would examine it to decide about sending email. But it's possible there won't be: the plan may be just to use this data outside the app to filter mass mailings.)

- This variable is used only in the backend, right? Otherwise I should include it in the variables that returned by the controller
  - Again, not sure how it will be used. But I presume that eventually the user will be able to modify this setting via the UI.

- Is this one of the cases where the variable should be camelCase in the backend code? the db column is do_not_email but I'm not sure if in the Js model it should be doNotEmail or same as in the DB
  - Since we last discussed this, I have learned that there is some convenience to using snake_case_naming in JS that manipulates the database, so that `{ some_name }` can be written instead of `{ some_name: someName }`. So, I suppose you may as well go ahead and stick with snake_case.

BTW, it would be better to post requirements questions on the Jira ticket so that non-devs can be in on the discussion.

hernandcb:
@smattingly Thank you for your answers. 

I adjusted the copyright text as per your comments.

JimmyRowland (approved):

PR #1761, opened 2022-01-07
[LF-2112] Added 'Soil sample results' as new document type

hernandcb:


hernandcb:
@smattingly 
I'm not able to upload files, it seems I'm missing some configuration property but I didn't find . Do you know how can I fix that ? 
I would like to test that the value is stored properly into the DB.

This is the error that appears in the backend console: 

Screenshot from 2022-01-07 15-22-59

hernandcb:
The error to upload documents was due to a missing variable in the .env file

smattingly (commented):
See the attached screenshot for details of the errors I received when trying to upload a document of the new type. (I used a PDF of a diagram I had handy.) After the error, I changed the document type to one of the old ones, and the form submitted successfully.

image

hernandcb:
Did you run the migration ? 
The pull request includes a migration to add the new type into the enum in the DB field. 
I just tried and get the same error because of that, otherwise please show me the response you get to the request to see what is happening

smattingly:
So looking at the migrations is not enough; I really should run them. Sorry about that-- this is all good.

smattingly (approved):
:+1:

PR #1762, opened 2022-01-07
[LF-2139] Link to tutorials in the menu

hernandcb:


hernandcb:
Where do you think is better to store the urls to the different youtube playlists?
I don't like the idea of having those urls in the code or the i18n files, perhaps in a config file?

JimmyRowland:
I think it's fine to store URLs in that function

hernandcb:
Thank you, I did it as you suggested.

smattingly (approved):
:+1:

PR #1763, opened 2022-01-08
Lf 2045/record a d patch

JimmyRowland:


packages/api/src/jobs/certification/record_d_generation.js
```diff
@@ -26,119 +27,160 @@ const dataTransformsMapping = {
...
-      workbook.sheet(0).range('A1:I1').style({
-        bold: true,
-        fontSize: 20,
-        fill: 'F2F2F2',
-      });
-      workbook.sheet(0).range('A6:I9').style({
-        fill: 'F2F2F2',
-      });
+    rowSix
+      .add(`2. ${t('RECORD_D.NOTE.TWO.PART_1')} `)
```
smattingly, line 47: Just wanted to be sure you knew about the spreadsheet library limitation: trailing spaces like the one here may or may not be preserved, depending on the specific spreadsheet software used to view the result. The library does the proper xml encoding on *leading* spaces so that they are always preserved. Not sure if that matters here.

smattingly (approved):

packages/api/src/jobs/certification/record_d_generation.js
JimmyRowland: Fixed

PR #1771, opened 2022-01-13
LF-2045/fix-organic-status-of-location-without-active-management plans

JimmyRowland:
Set bufferzone organic status to non-organic


smattingly (approved):
:+1:

PR #1777, opened 2022-01-17
Bugfix/bump dependencies 2022 jan

JimmyRowland:


JimmyRowland:
Tried to bump react-script, but there are too many packages breaking. It's probably better to migrate to vite

smattingly (approved):
:+1:

PR #1811, opened 2022-02-12
bugfix/harvest-task-form-arry-id

JimmyRowland:


packages/webapp/src/components/Task/TaskComplete/HarvestComplete/HarvestUses.jsx
```diff
@@ -175,7 +173,7 @@ export default function PureHarvestUses({
           </div>
         );
       })}
-      {amount_to_allocate < 0 && (
+      {amount_to_allocate < -0.1 && (
```
smattingly, line 176: This seems unsafe? I presume we want the value to be zero, but are rightly avoiding a direct floating point comparison. But couldn't the value be greater than zero? Maybe `Math.abs(amount_to_allocate - 0) < 0.1`?

smattingly (approved):

PR #1814, opened 2022-02-18
Include base properties in GETting notifications

smattingly:


packages/api/src/models/notificationUserModel.js
```diff
@@ -60,9 +60,13 @@ class NotificationUser extends baseModel {
   }
 
   static async getNotificationsForFarmUser(farm_id, user_id) {
-    return await NotificationUser.query().withGraphJoined('notification')
-      .whereRaw('notification.deleted = false AND notification_user.deleted = false AND user_id = ? AND (farm_id IS NULL OR farm_id = ?)',
-        [user_id, farm_id])
+    return await NotificationUser.query()
+      .withGraphJoined('notification')
+      .context({ showHidden: true })
```
JimmyRowland, line 65: Overwrite $afterfind for both notification and notification user model to filter out created at created by updated by, and deleted fields. Or simply don't use baseModel since it included too many irrelevant fields or functionalities. https://vincit.github.io/objection.js/guide/hooks.html

JimmyRowland (approved):

packages/api/src/models/notificationUserModel.js
smattingly: I see. I suspect that the data/format for the query may change anyway, so I'll remember this for future use.

PR #1829, opened 2022-02-25
LF-2275/assign-task-inactive-user-validation

JimmyRowland:


packages/api/src/controllers/taskController.js
```diff
@@ -83,37 +84,54 @@ const taskController = {
...
-          return res.status(403).send('A worker who is not assignee or owner of task cannot abandon it');
+          return res
+            .status(403)
+            .send('A worker who is not assignee or owner of task cannot abandon it');
         }
         // cannot abandon an unassigned task with rating or duration
         if (!hasAssignee && (happiness || duration)) {
-          return res.status(406).send('An unassigned task should not be rated or have time clocked');
+          return res
+            .status(406)
```
smattingly, line 109: 406 is about content negotiation, so it doesn't seem appropriate here. Maybe 409, or just generic 400.

smattingly (commented):
See comments about status codes.

smattingly (approved):

PR #1835, opened 2022-03-01
LF-2028 Soft delete farms where country and grid pts are NULL

smattingly:
In standup, we agreed to run this on beta as well as production. As a result, nothing here checks the environment-- the updates will just run. @JimmyRowland let me know if you think that should be handled differently.


JimmyRowland (approved):

JimmyRowland:
Need to check whether those farms are used manually on production

PR #1843, opened 2022-03-08
Change due date quick assign functionality

MwayaB:


packages/webapp/src/containers/taskSlice.js
```diff
@@ -55,6 +55,13 @@ const upsertManyTasks = (state, { payload: tasks }) => {
...
 };
 
+const upsertOneTask = (state, { payload: task }) => {
+  state.loading = false;
+  state.error = null;
+  state.loaded = true;
+  taskAdapter.upsertOne(state, task);
+};
+
 const updateOneTask = (state, { payload: task }) => {
```
kikemarquez8, line 65: should we remove updateOneTask if it's no longer being used?

MwayaB: I don't know if it's being used elsewhere in the app

packages/api/src/routes/taskRoute.js
```diff
@@ -38,6 +38,13 @@ router.patch(
   taskController.assignAllTasksOnDate(),
 );
 
+router.patch(
+  '/patch_due_date/:task_id',
+  hasFarmAccess({ params: 'task_id' }),
+  checkScope(['edit:task']),
+  taskController.patchTaskDate(),
```
smattingly, line 45: The parameters to an Express route handler like `patch` are a route string followed by a sequence of function names or definitions (not function calls); those functions will be called later when an incoming request matches the method and route. `hasFarmAccess` and `checkScope` are being called here because they return the function definitions that are passed to `patch` so that they can be called later.  hasFarmAccess and checkScope are examples of what Express calls "configurable middleware"-- they return a middleware function that is configured by the parameters passed here. `patchTaskDate` takes no parameters, which means that it always returns the same function. Since nothing is being configured, the norm with Express is to simply name the function, not call it, and simplify `patchTaskDate` so that it performs the job its name suggests instead of returning a function that performs that job.

packages/api/src/routes/taskRoute.js
```diff
@@ -38,6 +38,13 @@ router.patch(
   taskController.assignAllTasksOnDate(),
 );
 
+router.patch(
```
smattingly, line 41: Since you are adding an API route, you should add jest test cases to automatically test it. I can help you set that up if it's unfamiliar to you.

smattingly (changes requested):
Nice work. Let me know if you would like help with any of the suggested changes.

packages/api/src/controllers/taskController.js
```diff
@@ -61,13 +61,32 @@ const taskController = {
             assignee_user_id,
           })
           .whereIn('task_id', available_tasks);
-        return res.status(200).send(available_tasks);
+        return result
+          ? res.status(200).send(available_tasks)
+          : res.status(404).send('Tasks not found');
```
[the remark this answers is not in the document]
MwayaB: I actually modified this inadvertently the original code was `return res.status(400).json({ error });` I will correct it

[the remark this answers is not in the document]
MwayaB: One correction, the user arrives and the QuickAssignModal, when they click the assignee, and this request is made when the user checks the checkbox on that modal and clicks continue. Which has nothing to do with this ticket

packages/api/tests/task.test.js
```diff
@@ -7,10 +7,7 @@ jest.mock('jsdom');
 jest.mock('../src/middleware/acl/checkJwt');
 const mocks = require('./mock.factories');
 const { tableCleanup } = require('./testEnvironment');
-const moment = require('moment');
-let faker = require('faker');
-const { test } = require('../knexfile');
```
smattingly: @MwayaB I removed several unneeded `require`s in this area. The one for `test` was actually causing a problem I think. I was getting "test is not a function" when I first ran the tests. I'm not sure what `test` in the knexfile is, but it's not the `test` we need here. 
Without the `require`, eslint does complain that it does not know what `test` is. That semi-problem is addressed at the bottom of this file, where I have commented on the fix.

packages/api/tests/task.test.js
```diff
@@ -56,7 +53,7 @@ describe('Task tests', () => {
   function postTransplantTaskRequest({ user_id, farm_id }, data, callback) {
     chai
       .request(server)
-      .post(`/task/transplant_task`)
+      .post('/task/transplant_task')
```
smattingly, line 56: @MwayaB This is an example of a change made automatically by eslint when the pre-commit hook runs it. There's a style rule to prefer single quotes where they will work.
(This change isn't all that important. Just pointing out the effect of eslint's autofixing. There are many more cases of that that I won't comment on.)

packages/api/tests/task.test.js
```diff
@@ -788,7 +779,7 @@ describe('Task tests', () => {
       });
     });
 
-    xtest('should get all tasks that are related to a farm, but not from different farms of that user', async (done) => {
+    xtest('should get all tasks related to a farm, but not from different farms of that user', async (done) => {
```
smattingly, line 782: @MwayaB  `xtest` is a way of disabling an individual `test`. This is one reason that tests show as "skipped" in the results.

packages/api/tests/task.test.js
```diff
@@ -2060,17 +2009,14 @@ describe('Task tests', () => {
...
+        expect(res.status).toBe(200);
+        const updated_task = await getTask(task_id);
+        expect(updated_task.due_date.toISOString().split('T')[0]).toBe(due_date);
+        done();
+      });
     });
   });
 });
+
+/* global jest describe test xtest expect beforeEach afterAll */
```
smattingly, line 2022: @MwayaB This block comment that starts with `global` is actually a configuration instruction to the eslint tool. JavaScript doesn't have the equivalent of C/C++ header files that you can include to tell the tools that something is defined somewhere else.
This tells eslint that the listed names are defined globally (outside this file), so the tool should not complain that they aren't defined here. Lots of the test files are missing this sort of comment, and the better way is to put equivalent information in a central config file, but this was convenient for now.
My guess is that you may have `require`d (the wrong) `test` at the top of this file because eslint was complaining about it.

packages/api/tests/task.test.js
```diff
@@ -2047,11 +1996,11 @@ describe('Task tests', () => {
 
   describe('Patch task date test', () => {
     test('Owner must be able to patch task date', async (done) => {
-      const date = faker.date.future().toISOString().split('T')[0];
+      const due_date = faker.date.future().toISOString().split('T')[0];
```
smattingly, line 1999: @MwayaB I made a few fixes related to unknown variables. Here, I changed `date` to `due_date` because the (now commented out) console.log and other statements errored with `due_date` not defined.

packages/api/tests/task.test.js
```diff
@@ -2047,11 +1996,11 @@ describe('Task tests', () => {
 
   describe('Patch task date test', () => {
```
smattingly, line 1997: @MwayaB I'd suggest being specific in the descriptive text that this is about "due date". A task has other kinds of date(s) associated with it.

packages/api/tests/task.test.js
```diff
@@ -2060,17 +2009,14 @@ describe('Task tests', () => {
...
-          expect(res.status).toBe(200);
-          const updated_task = await getTask(task_id);
-          expect(updated_task.due_date).toBe(due_date);
-          done();
-        },
-      );
+      patchTaskDateRequest({ user_id, farm_id }, patchTaskDateBody, task_id, async (err, res) => {
+        expect(res.status).toBe(200);
+        const updated_task = await getTask(task_id);
+        expect(updated_task.due_date.toISOString().split('T')[0]).toBe(due_date);
```
smattingly, line 2015: @MwayaB Once I got things running, this test failed. Even though the dates were the same, the value coming back from the database had a timestamp portion to it that was removed from `due_date` when it was initialized. So I removed the timestamp to get them to match.
That feels a bit sloppy, but I think there may be a larger issue behind this. The column in the database stores only the date; I think that the JavaScript/Objection model definition is making it include a (midnight UTC default) timestamp value. This may have to do with recent changes where Jimmy changed some database columns to drop the unused timestamp portions to just be dates.

packages/api/package.json
```diff
@@ -4,7 +4,7 @@
   "description": "backend for LiteFarm",
   "main": "./api/src/server.js",
   "scripts": {
-    "test": "jest --runInBand",
```
smattingly, line 7: @MwayaB Changes to package.json often have a wide impact. Best to make them, when necessary, in consultation (or at least notification to) others.
In this particular case, I don't think you need or want to remove that. I was running only the modified test file with the command line `npx jest --runInBand ./tests/task.test.js`
That flag prevents jest from running tests in parallel; I don't know if the tests will work correctly if they are executed in parallel.

packages/api/tests/task.test.js
```diff
@@ -2013,7 +2010,14 @@ describe('Task tests', () => {
       patchTaskDateRequest({ user_id, farm_id }, patchTaskDateBody, task_id, async (err, res) => {
         expect(res.status).toBe(200);
         const updated_task = await getTask(task_id);
-        expect(updated_task.due_date.toISOString().split('T')[0]).toBe(due_date);
+        const dueDateFromDb = new Date(
```
smattingly, line 2013: @MwayaB Please see if this causes the test to pass for you. The change is based on this attempt to manually simulate the problem.
image

smattingly (approved):
Very thorough tests! I'm approving, and will let you handle the merge into `develop` (let me know if you aren't authorized). That's not our currently documented process, but I've just posted a proposal to change it.

PR #1857, opened 2022-04-12
Sse tuning

smattingly:


smattingly:
This formally commits the `X-Accel-Buffer` header that was manually applied on beta as an experiment.

JimmyRowland (approved):

PR #1860, opened 2022-04-25
Lf 2341/end to end cypress happy path tests

MwayaB:


packages/api/src/server.js
```diff
@@ -70,14 +72,21 @@ const documentRoute = require('./routes/documentRoute');
 const taskRoute = require('./routes/taskRoute');
 const productRoute = require('./routes/productRoute');
 const notificationUserRoute = require('./routes/notificationUserRoute');
+const { use } = require('chai');
 
 // register API
 const router = promiseRouter();
 
-app.get('/', (req, res) => {
+app.get('/', async (req, res) => {
```
smattingly, line 80: Why add `async`?

packages/webapp/cypress/plugins/index.js
```diff
@@ -16,7 +16,26 @@
...
+  config.env.googleClientSecret = process.env.REACT_APP_GOOGLE_CLIENT_SECRET;
+
+  // plugins code ...
+  //const testDataApiEndpoint = `${config.env.apiUrl}`;
+
+
+  on("task", {
+    async "db:tableCleanup"() {
+      // clean up the database tables
+      const { data } = await axios.post('http://localhost:5001/testData/tableCleanup');
```
smattingly, line 35: Hardcoding the port will probably be a problem. Also I'm concerned about this working in Docker containers.

smattingly (commented):
Looks very good. I'd like to run the tests, and I'm not sure how. I think I might need some env files that are not in git?

packages/api/src/server.js
MwayaB: I just checked the branch, for some reason ./cypress.env.json i'd added to the .gitignore file still appears

packages/webapp/cypress/plugins/index.js
MwayaB: I will fix the hardcoding of the port, not sure what i need to do re: making it work in docker containers

packages/api/src/server.js
MwayaB: I think i'd hard coded the clean db tables route to that route when i was testing and forgot to return it to its original state

smattingly: Adding a file to `.gitignore` won't remove it from the repo if it was committed before.

MwayaB: Understood, my assumption was that since this is a new branch those files would be ignored when i pushed my first commit

smattingly (approved):

PR #1878, opened 2022-05-05
LF-2323: error when unassigning task from task card and read only task view

fsayyid:
There was issue in validating the assignee_user_id where we were not accepting null as valid, when null is representing when a task is unassigned. So the validation was always throwing 400 when trying to make a task unassigned. 

Also added a check in the frontend such that if the task is currently unassigned, we can not assign it again to unassigned.

### DIFF
--- a/packages/api/src/middleware/validation/assignTask.js
+++ b/packages/api/src/middleware/validation/assignTask.js
@@ -25,6 +25,11 @@ async function validateAssigneeId(req, res, next) {
     return res.status(403).send('Not authorized to assign other people for this task');
   }
 
+  // if the assignee_user_id is null, this means that the task is 'Unassigned' which is a valid
+  if (assignee_user_id === null) {
+    return next();
+  }
+
   const userFarm = await userFarmModel
     .query()
     .where({
--- a/packages/webapp/src/components/Modals/QuickAssignModal/index.jsx
+++ b/packages/webapp/src/components/Modals/QuickAssignModal/index.jsx
@@ -19,13 +19,14 @@ export default function TaskQuickAssignModal({
 }) {
   const { t } = useTranslation();
   const selfOption = { label: `${user.first_name} ${user.last_name}`, value: user.user_id };
-  const unAssignedOption = { label: t('TASK.UNASSIGNED'), value: null };
+  const unAssignedOption = { label: t('TASK.UNASSIGNED'), value: null, isDisabled: false };
   const options = useMemo(() => {
     if (user.is_admin) {
       const options = users.map(({ first_name, last_name, user_id }) => ({
         label: `${first_name} ${last_name}`,
         value: user_id,
       }));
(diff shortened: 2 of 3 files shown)


packages/webapp/src/components/Modals/QuickAssignModal/index.jsx
```diff
@@ -19,13 +19,14 @@ export default function TaskQuickAssignModal({
...
   const selfOption = { label: `${user.first_name} ${user.last_name}`, value: user.user_id };
-  const unAssignedOption = { label: t('TASK.UNASSIGNED'), value: null };
+  const unAssignedOption = { label: t('TASK.UNASSIGNED'), value: null, isDisabled: false };
   const options = useMemo(() => {
     if (user.is_admin) {
       const options = users.map(({ first_name, last_name, user_id }) => ({
         label: `${first_name} ${last_name}`,
         value: user_id,
       }));
+      unAssignedOption.isDisabled = !isAssigned;
```
Cali0707, line 29: I think disabling the ability to select unassigned is a great idea, but is there any way we can visually indicate that this option is disabled? Currently when I clone this locally and then try and assign an unassigned task, I can still see "Unassigned" in the dropdown, but can't click on it. I feel like this can lead to a confusing user experience

Cali0707 (changes requested):

packages/webapp/src/components/Modals/QuickAssignModal/index.jsx
fsayyid: Yeah, I was thinking the same, I just wasn't sure what an option should look like when its disabled. Probably just make the text gray, but let's see.

fsayyid: I think this should be fine?

Screen Shot 2022-05-05 at 4 33 05 PM

Cali0707 (approved):
This looks great! Very clearly disabled

packages/api/src/middleware/validation/assignTask.js
```diff
@@ -25,6 +25,11 @@ async function validateAssigneeId(req, res, next) {
     return res.status(403).send('Not authorized to assign other people for this task');
   }
 
+  // if the assignee_user_id is null, this means that the task is 'Unassigned' which is a valid
```
smattingly, line 28: I think it's interesting that line 24 just above acknowledges the value could be `null`.

packages/webapp/src/components/Modals/QuickAssignModal/index.jsx
smattingly: I think this is good. However, I have found a place in the app that entirely removes values from the select list when they are not a valid choice. There's a chance we'll be asked to rework this for consistency. Sorry I didn't have that info for you earlier.

smattingly (approved):
:+1:

PR #1880, opened 2022-05-05
LF-2332/fix-estimated-yield-calculation-for-crops

seinnlee:
Ticket [LF-2332](https://lite-farm.atlassian.net/browse/LF-2332)

Estimated annual harvest calculation for row planting method

For example for Apricots, 
- 30 meter rows and plant spacing at 15 meters, should equal 3 plants (0, 15, 30) per row at 10 rows = 30 plants
- 30 plants at 47.5 kg/plant = 1425

To Test:
1. Select any crops and create a plan
2. Select Rows for planting method
3. Check the estimated annual yield is correct

You can find each crop's yield per plant in Crop table.

If the length isn't perfectly divisible by spacing (i.e. 16 meter spacing in a 30 meter row), you cannot plant 2.875 plants each row. You will plant 2 plants per row (at 0 meter and 16 meter) and the estimated yield is calculated accordingly.

### DIFF
--- a/packages/webapp/src/components/Crop/RowMethod/PureRowForm.jsx
+++ b/packages/webapp/src/components/Crop/RowMethod/PureRowForm.jsx
@@ -46,7 +46,16 @@ export default function PureRowForm({
 
   const IsValidNumberInput = (number) => number === 0 || number > 0;
 
-  const calculateEstimates = (length, spacing) => {
+  /**
+   * Calculates the plant count in one row given length of the row and plant spacing. 
+   * If length is perfectly divisible by spacing (i.e. 5m spacing in a 10m row),
+   * you can plant 3 crops (at 0m, 5m, and 10m).
+   * Otherwise (i.e. 6m spacing in a 10m row), you can plant only 2 crops (at 0m and 6m).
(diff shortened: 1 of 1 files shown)


jagDhillon077 (approved):
Calculations are correct

smattingly (changes requested):
Good call on creating the function. I agree that the calculations look correct as I understand the requirements.

Some suggestions:
1. `calculateEstimates` could be more descriptive, especially when there are multiple quantities being estimated in the vicinity of the code. `estimatePlantCount` or something like that?
2. As the bug ticket proves, the correct formula is not completely obvious. Please see `doc-comments` [here](https://lite-farm.atlassian.net/wiki/spaces/LITEFARM/pages/1191575576/Pre-integration+peer+review#General%2C-for-all-code) and write a JSDoc comment for the function. Let me know if that's unfamiliar to you or you have questions.
3. While you're in there, a small thing: the four calls to the function (twice in two different places) could easily be two (once in two different places) by introducing a local variable. It's not an expensive function, but still a good idea, I think.

smattingly (approved):
:100:

PR #1881, opened 2022-05-05
LF-2378/Bug fix for completed and abandoned cards showing underlined under data and assignee

jagDhillon077:
Dates and and underlines on cards do not appear if card is set to completed or abandoned

### DIFF
--- a/packages/webapp/src/components/CardWithStatus/TaskCard/styles.module.scss
+++ b/packages/webapp/src/components/CardWithStatus/TaskCard/styles.module.scss
@@ -53,6 +53,13 @@
   border-bottom: 1px solid var(--teal700);
 }
 
+.iconTextContainerDisabled {
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  gap: 3px;
+}
+
 .statusLabel {
   display: flex;
   right: 0;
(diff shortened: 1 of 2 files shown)


packages/webapp/src/components/CardWithStatus/TaskCard/styles.module.scss
```diff
@@ -53,6 +53,13 @@
   border-bottom: 1px solid var(--teal700);
 }
 
+.iconTextContainerDisabled {
```
seinnlee, line 56: I'm not sure if `iconTextContainerDisabled` is the right word here because the container is not disabled. Maybe we can change it to `iconTextContainerNoUnderline` or `iconTextContainerUneditable`?

seinnlee:
It looks good other than the naming thing! Functionally worked for me.

seinnlee (approved):

packages/webapp/src/components/CardWithStatus/TaskCard/styles.module.scss
jagDhillon077: Thanks for the feedback! Changed `iconTextContainerDisabled` to `iconTextContainerNoUnderline`

smattingly (approved):
Nice job!

PR #1882, opened 2022-05-05
Lf 2360/completing harvest task hangs

smattingly:
The api returned the updated harvest task in a form different from the front-end's expectations. This "corrupted" the Redux store.


Cali0707 (approved):
Looks good to me

PR #1884, opened 2022-05-06
Lf 1902/farm worker able to change due date reassign task

MwayaB:


packages/api/src/controllers/taskController.js
```diff
@@ -36,9 +36,15 @@ const taskController = {
           return res.status(400).send('Task has already been completed or abandoned');
         }
 
+        if(!adminRoles.includes(req.role) && checkTaskStatus.assignee_user_id != req.user.user_id && checkTaskStatus.assignee_user_id !== null){
+          return res.status(400).send('Farm workers are not allowed to reassign a task assigned to another worker');
```
Cali0707, line 40:
```suggestion
          return res.status(403).send('Farm workers are not allowed to reassign a task assigned to another worker');
```
I believe a 403 error code makes more sense here because it is an unauthorized error https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#client_error_responses

packages/api/tests/task.test.js
```diff
@@ -306,6 +306,37 @@ describe('Task tests', () => {
...
+      await mocks.location_tasksFactory({
+        promisedTask: [{ task_id }],
+        promisedField: [{ location_id }],
+      });
+      assignTaskRequest(
+        { user_id, farm_id },
+        { assignee_user_id: user_id },
+        task_id,
+        async (err, res) => {
+          expect(res.status).toBe(400);
```
Cali0707, line 334:
```suggestion
          expect(res.status).toBe(403);
```
Again, changing the error code

Cali0707 (changes requested):
Looks good - one small change on the error codes then I think this is good to go

Cali0707 (approved):
Looks great!

smattingly (approved):
Nicely done.

PR #1886, opened 2022-05-06
LF2230/allow completion of tasks for psuedo users

Cali0707:
This PR adds functionality to complete tasks for pseudo users.

To test: 
1. Create a pseudo user on your farm by adding a user with no email
2. Assign a task to the pseudo user
3. Verify that for non admin accounts e.g. farm worker they do not have the option to complete the task for the pseudo user
4. Complete the task from an admin account (Farm Owner, Farm manager, extension officer)
5. Verify that the task is completed properly
6. Repeat this for both a harvest task and a non harvest task


AbdusShaikh (approved):
Working on both harvest and non-harvest tasks

packages/api/src/controllers/taskController.js
```diff
@@ -386,15 +388,16 @@ const taskController = {
   async completeHarvestTask(req, res) {
     try {
       const nonModifiable = getNonModifiable('harvest_task');
+      const { user_id } = req.headers;
```
smattingly, line 391: Please use `req.user.user_id` instead of `req.headers.user_id`. Honestly, I'm just passing that along from past team members; I *suspect* I know why.

packages/api/src/controllers/taskController.js
```diff
@@ -619,4 +622,20 @@ async function notifyAssignee(userId, taskId, taskTranslationKey, farmId) {
   );
 }
 
+/**
+ * Checks if the current user can complete the task
```
smattingly, line 626: JSDoc comments need a period at the end of this first description sentence. (I know that's very picky, but it's meaningful to the software that processes these, should we ever use it.)

packages/api/src/models/taskModel.js
```diff
@@ -10,7 +10,7 @@
  *  LiteFarm is distributed in the hope that it will be useful,
  *  but WITHOUT ANY WARRANTY; without even the implied warranty of
  *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
- *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ *  GNU General Public License for more details, see <<https://www.gnu.org/licenses/>.>
```
smattingly, line 13: Looks like extra angle brackets?

packages/api/src/models/taskModel.js
```diff
@@ -191,6 +191,27 @@ class TaskModel extends BaseModel {
...
   }
+
+  /**
+   * Gets the assignee of a task
+   * @param {number} taskId - The ID of the task
+   * @static
+   * @async
+   * @returns {Object} - Object {assignee_user_id, assignee_role_id, wage_at_moment, override_hourly_wage}
+   */
+  static async getTaskAssignee(taskId) {
```
smattingly, line 202: Thank you for putting this in the model. :+1:

smattingly (changes requested):
Good work. Please see comments for a few needed changes (which led me to throw in some less critical suggestions, too.)

packages/api/src/models/userFarmModel.js
```diff
@@ -97,32 +97,47 @@ class userFarm extends Model {
...
+  }
+
+  /**
+   * Retrieves role for a specified user
+   * @param {uuid} userId - The specified user.
+   * @static
+   * @async
+   * @returns {number} Number corresponding to role_id.
+   */
+  static async getUserRoleId(userId) {
```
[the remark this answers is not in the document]
Cali0707: I believe that in the case of `userFarmModel.js` we are extending the `Model` class from objection, not objection-soft-delete, so as far as I can tell we do not need it in this case. Also the queries seem to be returning the expected results. Should I update this model to extend `BaseModel` instead?

Cali0707:
@smattingly I made the changes you requested and then rebased integration onto my branch to try and make merging easier. It looks like there are still some merge conflicts though.

packages/api/src/models/userFarmModel.js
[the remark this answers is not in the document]
smattingly: You're right. I wonder why user and userFarm were not set up for soft delete? Anyway, I see no reason to change that now.

smattingly (approved):
Good work-- thanks.

smattingly:
> @smattingly I made the changes you requested and then rebased integration onto my branch to try and make merging easier. I haven't done a rebase before but it looks like this added many more changes to my PR than I intended. Is there a better way of fixing merge conflicts which I should be using going forwards?

@Cali0707 I usually merge rather than rebase. I know that the only difference between the two is how the history of commits is recorded; the resulting code contents are the same. You might try merge instead of rebase next time, but what you did is fine-- and appreciated.

PR #1887, opened 2022-05-06
LF 2364 augment type filter on task filter page

Cali0707:
Updated the filter on the task page to display all active task types, regardless of whether there is a task with that type currently.

To test:
1. Make sure you have at least one default task type and one custom task type on your farm which have no tasks associated with them
2. On the tasks page, go to filter tasks
3. Try to filter the tasks by type -> you should see all task types here, not just those which currently have a task


fsayyid (approved):
I tested it on my end and it seems to be working. The logic also looks good to me.

smattingly (approved):
:+1:

PR #1888, opened 2022-05-09
Abandoned tasks now show abandonment date

AbdusShaikh:


AbdusShaikh:
To test:
1. Create a task
2. Abandon this task ensuring the abandonment date is different from the due date
3. Confirm the abandonment is appearing in place of the due date in the Tasks page.

packages/webapp/src/containers/Task/TaskCard/index.jsx
```diff
@@ -33,6 +33,13 @@ const TaskCard = ({
   const user = useSelector(userFarmSelector);
   const immutableStatus = ['completed', 'abandoned'];
 
+  let trueDate = completeOrDueDate;
+  console.log(props);
```
seinnlee, line 37: I think you forgot to remove the console.log

seinnlee (changes requested):
If `trueDate` is only going to be passed into one component, I think it's better to write this mapping in `TaskCard.jsx` component rather than in `index.jsx` file.

AbdusShaikh:
> If `trueDate` is only going to be passed into one component, I think it's better to write this mapping in `TaskCard.jsx` component rather than in `index.jsx` file.

I have made a commit making these changes

smattingly:
@AbdusShaikh Since @seinnlee is out today, I will pick up the review of this from here.
Also, merge conflicts have developed for this branch. If you know how to address that, please do. If not, reach out and we can go through that.

smattingly (approved):
:+1:

PR #1889, opened 2022-05-09
Augment the "Assignee" filter on Task filter page

AbdusShaikh:
Assignee tab in task filter page shows all users that are active, and all that are inactive with tasks

### DIFF
--- a/packages/webapp/src/containers/Filter/Tasks/index.jsx
+++ b/packages/webapp/src/containers/Filter/Tasks/index.jsx
@@ -4,6 +4,7 @@ import React, { useMemo, useRef } from 'react';
 
 import PureFilterPage from '../../../components/FilterPage';
 import { setTasksFilter, tasksFilterSelector } from '../../filterSlice';
+import { userFarmsByFarmSelector } from '../../userFarmSlice';
 
 import {
   ABANDONED,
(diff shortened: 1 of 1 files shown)


AbdusShaikh:
To Test:
1. Create a couple of users with a mix of pseudo-users and full users
2. While all users are active, all users should appear in the "Assignee" tab on the task filter page
3. Revoke user access for different users, they should no longer appear on the "Assignee" tab
4. Assign users tasks. Users with tasks will always show up regardless of whether they are active or not.

packages/webapp/src/containers/Filter/Tasks/index.jsx
```diff
@@ -29,6 +30,9 @@ const TasksFilterPage = ({ onGoBack }) => {
   const tasks = useSelector(tasksSelector);
   const dispatch = useDispatch();
   const locations = useSelector(locationsSelector);
+  const activeUsers = useSelector(userFarmsByFarmSelector)
+    .filter((user) => user.role != 'Owner')
+    .filter((user) => user.status != 'Inactive');
```
seinnlee, line 35: Are owners not considered active users?

seinnlee (commented):
There's no longer "Unassigned" option on the filter now. Is that intended? @smattingly

packages/webapp/src/containers/Filter/Tasks/index.jsx
AbdusShaikh: You are correct about the owners, they should be included and I've made that change in the latest commit.

As for the "Unassigned" option, I will let Steve respond.

smattingly (changes requested):
The assignee filter list should always include "Unassigned". Not just when there is an unassigned task.

smattingly (approved):
:+1:

PR #1891, opened 2022-05-09
Assign all tasks on date, on the task quick assign modal does not work

anant00729:
Assign all tasks on a date, on the task quick assign modal does not work

To test:
1. Go to http://localhost:3000/tasks
2. In the list of tasks, add two or more unassigned tasks for the same date.
3. click on unassigned, change it to some assignee and then click on the checkbox "Assign all unassigned tasks on this date to this person". then click on update.
4. You'll be able to see that the unassigned task will get updated with the assignee name for the same date.

For more details please check. 
https://lite-farm.atlassian.net/browse/LF-2314

### DIFF
--- a/packages/webapp/src/components/Modals/QuickAssignModal/index.jsx
+++ b/packages/webapp/src/components/Modals/QuickAssignModal/index.jsx
@@ -16,6 +16,7 @@ export default function TaskQuickAssignModal({
   onAssignTask,
   users,
   user,
+  taskCardContents,
 }) {
   const { t } = useTranslation();
   const selfOption = { label: `${user.first_name} ${user.last_name}`, value: user.user_id };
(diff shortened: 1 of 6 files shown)


jagDhillon077 (approved):
Functionality is working

packages/webapp/src/components/Modals/QuickAssignModal/index.jsx
```diff
@@ -48,6 +65,11 @@ export default function TaskQuickAssignModal({
     dismissModal();
   };
 
+  const onCheckedAll = () => {
+    checkUnassignedTaskForSameDate();
```
smattingly, line 69: Is there a reason for this statement? The return value is not used, and I don't see a side effect from the call.

smattingly (changes requested):
This looks good. I had one suggestion/question (see comment).

packages/webapp/src/components/Modals/QuickAssignModal/index.jsx
anant00729: I have removed that, I was calling that function on the checkbox click, to test out the function's results.

smattingly (approved):

PR #1893, opened 2022-05-09
LF-2383/notify user when task assigned to them has been abandoned

seinnlee:
Ticket [LF-2383](https://lite-farm.atlassian.net/browse/LF-2383)

**Description:**

- When a task has been abandoned by another user (i.e. management or owner), the worker will be notified
- @AbdusShaikh's [PR](https://github.com/LiteFarmOrg/LiteFarm/pull/1888) has been brought back

**To Test:**

1. You will need at least one worker at your farm that you have access to.
2. Create a task and assign it to the worker from step 1.
3. Abandon the task from step 2.
4. Log in to the account of the worker you initially assigned the task to from step 1.
5. You will see two notifications (initial assignment notification and abandonment notification)
6. Run 'Task Notification Tests'

Abandoner's name in the notification will be the user that performed the abandonment action (step 3).


fsayyid (approved):
It seems to be working on my end, and the code looks good to me!

packages/api/src/controllers/taskController.js
```diff
@@ -599,23 +621,48 @@ async function patchManagementPlanStartDate(trx, req, typeOfTask, task = req.bod
   }
 }
 
-async function notifyAssignee(userId, taskId, taskTranslationKey, farmId) {
-  if (!userId) return;
+async function getTaskStatus(taskId) {
+  return await TaskModel.query()
```
smattingly, line 625: We need to be sure that all queries consider [soft deletes](https://lite-farm.atlassian.net/wiki/spaces/LITEFARM/pages/1184235521/Navigating+and+querying+the+LiteFarm+database#A-note-on-%E2%80%98deleted%E2%80%99) which I will describe at tomorrow's stand up.

smattingly (changes requested):
Looks good-- especially the refactoring as more types of notifications are sent. Like LF-2376, I would like to see test(s) written if possible without disrupting sprint. Maybe you would like to apply the refactoring here to that one as well. It might make sense to merge the two, and I could help with that if you want.

Cali0707:
@seinnlee there was a change to notifications in the database that @smattingly and I made. I've retrofitted your branch so that what you did still works with the new notifications structure.

packages/api/src/controllers/taskController.js
fsayyid: @seinnlee doesn't seem like this was addressed. Add a delete: false to the where statement.

fsayyid (changes requested):
Please address the soft deletes comment, and deal with merge conflicts. Otherwise this seems to be functioning as desired 👍

fsayyid (approved):
Soft delete comment was addressed, and all tests are passing on my end.

smattingly (approved):
Nice work.

PR #1895, opened 2022-05-09
Searching for a crop name in English on the crop catalogue when app is in another language, still returns English version of the name

anant00729:
To test: 
1. go to the profile page section: http://localhost:3000/profile
2. change the language to 'Spanish'.
3. go to http://localhost:3000/crop_catalogue
4. in the search type: tomato, check if you can find 'Tomate' in the results. If not, then it's working. 
5. in the search type: Tomate, check if you can find 'Tomate' in the results. If yes, then it's working only for the Spanish language.


Cali0707 (approved):
Tested this and it works

packages/webapp/src/containers/CropCatalogue/useStringFilteredCrops.js
```diff
@@ -11,12 +11,11 @@ export default function useStringFilteredCrops(crops, filterString) {
...
-        crop?.crop_variety_name,
-        t(`crop:${crop.crop_translation_key}`),
-      ]),
-    );
+    // These fields can be used in future.
+    // Hence, keeping it in the comments
+    // use it in the check function argument.
+    // crop?.crop_common_name,
+    // crop?.crop_variety_name,
+    return crops.filter((crop) => check([t(`crop:${crop.crop_translation_key}`)]));
```
smattingly, line 19: I'm not sure which of these values we need to include in the search. Will raise the issue in Jira.

anant00729: Yes, I tested out, if we want results to be related to a specific language then we need to use only t(`crop:${crop.crop_translation_key}` which will filter out tomato whenever we are performing a search in the Spanish language. 

Please check this link: 

https://playcode.io/895689

I have created a mockup to make things more clearer

anant00729: I have added crop_common_name as discuss over the call.

smattingly (approved):
:+1:

PR #1896, opened 2022-05-09
Lf 2307 unarchive document

fsayyid:
Added the ability to unarchive documents from the read only document view. I modified the existing flows for archiving to handle unarchiving as well since they seem to go hand in hand. 

Story: [LF-2307](https://lite-farm.atlassian.net/browse/LF-2307)

Side Note: the only line in documentRoute that has changed is line 46. (I'm not sure why it's auto-formatting on commit even though I turned off Prettier.)


fsayyid:
Fixed tests to work with new changes to /document/archive/ endpoint.

AbdusShaikh:
Most of the functionality outlined in the ticket is working on my end. However, when a document fails to archive/unarchive, the screen remains the same. The archive/unarchive modal persists and the screen does not move from there. Is this intended?

fsayyid:
> Most of the functionality outlined in the ticket is working on my end. However, when a document fails to archive/unarchive, the screen remains the same. The archive/unarchive modal persists and the screen does not move from there. Is this intended?

fsayyid:
> Most of the functionality outlined in the ticket is working on my end. However, when a document fails to archive/unarchive, the screen remains the same. The archive/unarchive modal persists and the screen does not move from there. Is this intended?

When there's an error the modal should close and the read only document view should be fully visible. Plus a snack bar should pop up with an error. Let me see this on my end. Thanks.

fsayyid:
> > Most of the functionality outlined in the ticket is working on my end. However, when a document fails to archive/unarchive, the screen remains the same. The archive/unarchive modal persists and the screen does not move from there. Is this intended?
> 
> When there's an error the modal should close and the read only document view should be fully visible. Plus a snack bar should pop up with an error. Let me see this on my end. Thanks.

I think I was able to address this issue by reloading the page on error.

AbdusShaikh:
Failing to archive/unarchive now returns user to read-only view of the document as intended. Looks good to me.

AbdusShaikh (approved):
Archiving/unarchiving documents is working correctly. Succeeding and failing to archive/unarchive a document has the expected effects.

smattingly (approved):
Nice work.

PR #1899, opened 2022-05-11
Lf 2250 able to input non numerical values in numerical inputs wage year phone

AbdusShaikh:
This PR ensures that numerical input fields on the "Invite user" page cannot accept non-numeric values.

To test:
1. Go to the "Invite User" page
2. Attempt to provide the “Birth Year”, “Hourly Wage”, and “Phone Number” inputs with non-numeric values. All non-numeric inputs should not be accepted

**Note:** This bug was found on iPhone. To properly test this PR one must do it using Safari because iPhone browsers behave similarly to Safari. Testing on an actual iPhone would be even better if possible.


jagDhillon077 (approved):
Works correctly - only numerical digits are allowed in these fields

PR #1901, opened 2022-05-12
Lf 2014 update biodiversity insight to handle paginated gbif results

Cali0707:
This issue ended up touching a lot of older code, so there was a fair amount of refactoring included in this PR. To test this, you must:
1. Ensure your farm has at least 1 farm boundary (ideally more than one)
2. Go to the insights tab
3. Click on "Biodiversity"
4. If the insights for biodiversity have not yet finished loading (we start loading them on the insights page but they take a while), you should see a loading modal
5. If there is an error, you should see an error modal. (The errors occur because the GBIF API occasionally has a problem. If this does not happen for you, you can simulate it by setting the reponse code to 400 on line 258 of insightController.js)

You might have challenges testing both modals because we only re-fetch the biodiversity insight data if 30 minutes have passed since the previous fetch. To work around this, I just cleared my browser cache.


anant00729:
This link to the bug ticket is https://lite-farm.atlassian.net/browse/LF-2402

AbdusShaikh (approved):
Both modals are appearing and disappearing on my screen correctly.

smattingly (approved):
:+1:

PR #1905, opened 2022-05-13
Lf 2301/invitate existing users in proper language

Cali0707:
Fixed a bug where if an existing user was invited to a farm they were not always sent an invitation email matching their language preferences.

To test:
1. Ensure you have at least 2 accounts set up.
2. Make sure that the second account is in a different language than the first.
3. Create a new farm with the first account.
4. Invite the second account to the new farm.
5. The invite email to the second account should show up in the preferred language of the second account, instead of the preferred language of the first.
6. Invite a new user to the new farm (i.e. a user that does not have an account yet)
7. This invite email should be in the preferred language of the person who invited them to the farm.


anant00729:
The feature works fine for me. My second user's preferred language is Spanish and I reviewed the invitation in the same language.

Screen Shot 2022-05-18 at 3 18 07 PM

anant00729:
@smattingly Please review.

anant00729 (approved):

smattingly:
Then I'm changing the status for easier tracking. Thanks.

Cali0707:
@smattingly I do not think I will have time to refactor this further this sprint. I made a start, but a lot of this code calls `baseModel.post()` which is not something I am familiar with, so I did not feel comfortable refactoring it in the current timeframe.

packages/api/src/models/userFarmModel.js
```diff
@@ -140,6 +140,39 @@ class userFarm extends Model {
...
+      .join('users', 'userFarm.user_id', '=', 'users.user_id')
+      .join('farm', 'farm.farm_id', '=', 'userFarm.farm_id')
+      .join('role', 'userFarm.role_id', '=', 'role.role_id')
+      .where({ 'users.email': email, 'userFarm.farm_id': farm_id })
+      .first()
+      .select('*');
+    if (!trx) {
+      await transaction.commit();
+    }
+    return result;
```
smattingly, line 175: Seems to be a syntax error here.

smattingly (changes requested):

packages/api/src/models/userFarmModel.js
Cali0707: Thanks for catching that! I've fixed this now

smattingly (changes requested):
There are jest tests failing in the CI pipeline. They seem legit, so I'm assuming they will fail when tested locally too. If not, let me know.

Cali0707:
@smattingly looks like when I refactored the model calls out of the `userController` into the `emailTokenModel`, they were being referenced in other controllers, causing the tests to fail. I have updated the other controllers accordingly.

smattingly (approved):

PR #1909, opened 2022-05-14
Use weather API directly

smattingly:


packages/webapp/src/containers/WeatherBoard/saga.js
```diff
@@ -40,7 +45,6 @@ export function* getWeatherSaga({ payload: args }) {
     ) {
       yield put(onLoadingWeatherStart(farm_id));
       const apikey = import.meta.env.VITE_WEATHER_API_KEY;
-      const baseUri = 'https://cors-anywhere.litefarm.workers.dev/';
```
Cali0707, line 43: If we are getting rid of this and calling the endpoint directly is there any code we can get rid of which is now dead code?

Cali0707 (approved):
These changes work for me - I still get the correct weather results for my location.

PR #1910, opened 2022-05-16
Abandon rating, reason, and notes not showing on an abandoned crop plan detail page

jagDhillon077:
Create 2 crop plans from any of the plants on the crop catalogue and complete one of the crop plans and abandon the other. Make sure to fill in the details of both the completed and abandoned crop plan such as abandon notes, reason for abandoning, completion notes, rating, etc. Once each of the crop plans have been completed you can then go on the details section of each and see that the completed crop plan does not show any abandonment details and the abandoned crop plan shows abandon details and reasoning.

### DIFF
--- a/packages/webapp/src/components/Crop/ManagementDetail/ManagementPlanDetail.jsx
+++ b/packages/webapp/src/components/Crop/ManagementDetail/ManagementPlanDetail.jsx
@@ -91,7 +91,6 @@ export default function PureManagementDetail({
         )
       }
     >
-      {console.log(plan)}
       <CropHeader
         onBackClick={onBack}
         crop_translation_key={variety.crop_translation_key}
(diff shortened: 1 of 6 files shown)


Cali0707 (changes requested):
@jagDhillon077 it looks like the "Reason for abandoning" field is not having its translations handled properly: 
abandon_crop_plan_translation_error
 Could you look into this?

Cali0707 (changes requested):
@jagDhillon077 with the new changes, newly abandoned crop plans have the abandon reason displaying properly, but the one I already abandoned still has the same issue. Also, should we display "Date of status change" if there is no date of status change? Right now, it renders as "Invalid date"
Date_of_status_change

packages/webapp/src/components/Crop/ManagementDetail/ManagementPlanDetail.jsx
```diff
@@ -74,6 +91,7 @@ export default function PureManagementDetail({
         )
       }
     >
+      {console.log(plan)}
```
Cali0707, line 94: We should remove this console log

Cali0707 (changes requested):
@jagDhillon077 there is a new UI issue with these changes:
managemnet_plan_complete_date_error

Cali0707 (approved):
Looks good to me

smattingly (approved):
:+1:

PR #1911, opened 2022-05-16
Refactor location details pages

Cali0707:
This is a large refactoring of the location details pages to make them more easily extensible in the future. All shared UI is now in a single component, and the specific location type components can optionally have some extra details. As there are a lot of changes here, the easiest way to test this is probably to try creating one location of each type and ensuring that they all render properly.

PR #1914, opened 2022-05-17
LF-2199/add tasks tab and card view for locations

Cali0707:
FYI: Not worth reviewing this until #1911 is closed.

This ticket involved adding a new "Tasks" page to every location. This page is supposed to show all late and planned tasks, sorted by date. If you click "Create a task" from the location task page, when you get to the Location Selection page in the create task flow, the location should be pre-selected.

To test this:
1. Create locations of varying types on a farm
2. For each location:
     - Add tasks to some
     - Go to the map view and then click on the location
     - If the location is crop-enabled, you should see three tabs, otherwise just two tabs
     - Under the "Tasks" tab you should see all the relevant tasks for this location
3. On at least one location, click "Create a task", when you get to the page where you select a location for the task, the location should be pre-selected


fsayyid (approved):
Working on my end 👍

PR #1917, opened 2022-05-17
Fix integer data entry

smattingly:


jagDhillon077 (approved):
Checked all the numerical input fields when creating a crop plan. Works correctly

anant00729:
Checked the input value that accepts numbers. It works when we create a management plan.

PR #1920, opened 2022-05-18
LF-2401 Remove unused code

smattingly:


packages/api/src/middleware/acl/checkOwnership.js
```diff
@@ -1,73 +1,73 @@
...
-        //TODO potential bug
-      default:
-        sqlQuery = `SELECT DISTINCT u.user_id
-        FROM ? r, "users" u
-        WHERE u.user_id = ? AND r.?_id = ? AND u.farm_id = r.farm_id`;
-      }
+//       switch (relation) {
+//       case 'shift':
+//       case 'log':
+//       case 'notification':
```
Cali0707, line 33: It looks like there is something occuring here to do with notifications - are we sure this is not in use currently?

Cali0707 (dismissed):
Looks good - there was one spot where it looks like there *may* be a reference to notifications where it may be good to confirm that we do not need that. 

Also just out of curiosity, why are we commenting out the code instead of removing the files?

smattingly:
> Also just out of curiosity, why are we commenting out the code instead of removing the files?

Simply an abundance of caution-- in case we find something was broken, it's not even a git manipulation to reinstate the code. The idea is that they will truly be removed after we're confident it's safe.

packages/api/src/middleware/acl/checkOwnership.js
smattingly: Good question, but I'm sure it's unrelated to notifications as we currently know them. There were some references to "notifications" in the code/db before the recent feature development, but they were obsolete/unused prework for a somewhat different approach to the concept. There's still a bit of confusing stuff in the navbar-- one of the other icons was already referred to as "notifications" for some reason.

mickenorlen:
Hey guys, just wanted to say that I really like this initiative! It always feels good after you clean the house ;)

+1 for removing instead of commenting. Although I have no knowledge about the history of this code, I'm guessing you've checked most functions/files to see if they are actually being used or not, so I think you have good foundations for your decisions. Maybe both ways can be used if it feels like the code could become relevant again or if you have less confidence in certain cases. Anyways, I have confidence in you!

mickenorlen:
Hey guys, just wanted to say that I really like this initiative! It always feels good after you clean the house ;)

+1 for removing instead of commenting. Although I have no knowledge about the history of this code, I'm guessing you've checked most functions/files to see if they are actually being used or not, so I think you have good foundations for your decisions. Maybe both ways can be used if it feels like the code could become relevant again or if you have less confidence in certain cases. Anyways, I have confidence in your decision!

PR #1921, opened 2022-05-18
LF-2213/contextual return from tasks

Cali0707:
This PR adds contextual returns from tasks when starting on the crop plan page. To test it:
1. Make sure you have at least one crop plan
2. Add a task from the tasks view on the crop plan. When you finish adding the task, you should be returned to the crop plan, as opposed to the tasks page
3. From the crop plan, open a task and complete it. Upon completing, you should be returned to the crop plan
4. From the crop plan, open a task and abandon it. Upon abandoning, you should be returned to the crop plan


seinnlee (approved):
it lgtm!

PR #1912, opened 2022-05-17
Lf 2386 notify management of unassigned tasks due this week

fsayyid:
- Set up a route in the backend for time based notifications along with a new JWT token for the scheduler.
- Created an endpoint for sending a weekly notification about unassigned tasks due this week to farm management (FO/FM/EO) based on a farm_id
- Added logic in the frontend for filtering the tasks on the /tasks page by unassigned tasks due this week

Ticket: [LF-2386](https://lite-farm.atlassian.net/browse/LF-2386)

Linked Tickets: [LF-2397](https://lite-farm.atlassian.net/browse/LF-2397), [LF-2387]( https://lite-farm.atlassian.net/browse/LF-2387)

Right now in order to test in the frontend, we will need to manually send the request, since the scheduler itself is out of the scope of this ticket. 

How to test:
- Add this to the `packages/api/.env` file: 
  - `JWT_SCHEDULER_SECRET=Another_token_was_needed_for_the_scheduler`
- Run the backend tests in timeNotifcation.test.js, make sure they pass
- Run both the frontend and backend servers
- Create a couple users at the same farm, create a few tasks that are assigned, a few that are unassigned, and a few that are due later that one week. Make sure there is at least one task that is due with 7 days that is unassigned. 
- Grab the id of the farm where the tasks were created from the "pg-litefarm" database. You can use this query:
  - `SELECT u.farm_id FROM "userFarm" AS u JOIN users ON u.user_id=users.user_id WHERE users.first_name='<INSERT FIRST NAME OF USER>';`
- Use a tool such as [jwt.io](https://jwt.io/ ) to generate a JWT token with the same secret and payload:
  - `{ "requestTimedNotifications": true }` 
- Use a tool like Postman to send a request to:
  - `http://localhost:5001/time_notification/weekly_unassigned_tasks/<farm_id>`
  - Make sure to add a Authorization Bearer token header with the JWT token you generated
- Once the request has been sent, go back to the frontend > click the bell icon in the top right corner > make sure there is a notification with the title 'Unassigned Tasks'
- Click the notification > Click the 'Take Me There' button > make sure it takes you to the `/tasks` page and filters the tasks such that only tasks due this week that are unassigned show up (There should be three filters).

PR #1922, opened 2022-05-18
LF-2387/notify a user of tasks due today

seinnlee:
Ticket [LF-2387](https://lite-farm.atlassian.net/browse/LF-2387)
Linked Ticket [LF-2386](https://lite-farm.atlassian.net/browse/LF-2386) and [PR](https://github.com/LiteFarmOrg/LiteFarm/pull/1912)

*It is best to review [this PR](https://github.com/LiteFarmOrg/LiteFarm/pull/1912) and merge it before reviewing or merging this PR

**To Test:** (same as Faizah's PR)
- Add this to the `packages/api/.env` file: 
  - `JWT_SCHEDULER_SECRET=Another_token_was_needed_for_the_scheduler`
- Run the backend tests in timeNotifcation.test.js, make sure they pass
- Run both the frontend and backend servers
- Create a couple users at the same farm, create a few tasks that are assigned and a few that are due today.
- Grab the id of the user that the tasks were assigned to from the "pg-litefarm" database.
- Use a tool such as [jwt.io](https://jwt.io/ ) to generate a JWT token with the same secret and payload:
  - `{ "requestTimedNotifications": true }` 
- Use a tool like Postman to send a request to:
  - `http://localhost:5001/time_notification/daily_due_today_tasks/<user_id>`
  - Make sure to add a Authorization Bearer token header with the JWT token you generated
- Once the request has been sent, go back to the frontend > click the bell icon in the top right corner > make sure there is a notification with the title 'Tasks due today'
- Click the notification > Click the 'Take Me There' button > make sure it takes you to the `/tasks` page and filters the tasks such that only tasks due today that are not completed yet show up (There should be three filters).


fsayyid (commented):
Seems to be working on my end, and all the tests are passing. One thing I wanted to ask @smattingly @seinnlee did we want this to be based on user_id or farm_id?

fsayyid (approved):
Otherwise looks good to me.

seinnlee:
@smattingly @fsayyid I think it can be either `user_id` or `farm_id`. Even if a user works for multiple farms, wouldn't they still have to switch farms to access each farm's tasks? If we want to use `farm_id` then we will need to change the copy of the notifications because currently the notification doesn't indicate from which farm you received notifications

seinnlee:
@smattingly @fsayyid I updated the PR to use `farm_id` so when this endpoint is called, all the active users in this farm with tasks that are due today will receive notifications. If a user works for multiple farms, they will receive notification per farm.

packages/api/src/models/taskModel.js
```diff
@@ -232,6 +232,25 @@ class TaskModel extends BaseModel {
...
+   * @async
+   */
+  static async getTasksDueTodayFromUserId(userId) {
+    return await TaskModel.knex().raw(
+      `
+        SELECT task.task_id, task_type.task_translation_key, task.assignee_user_id, uf.farm_id
+        FROM task
+        JOIN "userFarm" AS uf ON task.assignee_user_id = uf.user_id
+        JOIN task_type ON task_type.task_type_id = task.task_type_id
+        WHERE task.assignee_user_id::text = '${userId}' AND due_date = now()::date
```
smattingly, line 250: This is potentially vulnerable to a SQL injection attack. See the knex docs on parameter binding for how to do this safely. Or let me know if you need help with that-- it's straightforward once you've seen it.

packages/api/src/controllers/timeNotificationController.js
```diff
@@ -60,6 +60,40 @@ const timeNotificationController = {
       return res.status(400).send({ error });
     }
   },
+
+  /**
+   * Notifies a user of tasks due today
```
smattingly, line 65: `For a specified farm, notifies all active users of tasks due today.`  ?

packages/api/src/models/userFarmModel.js
```diff
@@ -154,6 +154,20 @@ class userFarm extends Model {
...
+   * @param {uuid} farmId farm id
+   * @static
+   * @async
+   * @returns {Array} Array [user_id]
+   */
+  static async getActiveUsersFromFarmId(farmId) {
+    return await userFarm
+      .query()
+      .select('user_id')
+      .where('userFarm.farm_id', farmId);
```
smattingly, line 169: Not sure, but I think there should be some WHERE logic about active status?

packages/api/src/controllers/timeNotificationController.js
```diff
@@ -60,6 +60,40 @@ const timeNotificationController = {
...
+   */
+  async postDailyDueTodayTasks(req, res) {
+    const { farm_id } = req.params;
+    try {
+      const activeUsers = await UserFarmModel.getActiveUsersFromFarmId(farm_id);
+
+      if (activeUsers.length) {
+        const tasksDueTodayNotificationUsers = [];
+        for (const { user_id } of activeUsers) {
+          const { rows: tasksDueToday } = await TaskModel.getTasksDueTodayFromUserId(user_id);
```
smattingly, line 78: The function called here needs to check for tasks due today for `user_id` on `farm_id`. Otherwise line 81 may be notifying about user's tasks on some other farm.

packages/api/src/models/taskModel.js
```diff
@@ -232,6 +232,25 @@ class TaskModel extends BaseModel {
...
   }
+
+  /**
+   * Gets the tasks that are due today for a given user
+   * @param {string} userId user id
+   * @returns {Object}
+   * @static
+   * @async
+   */
+  static async getTasksDueTodayFromUserId(userId) {
```
smattingly, line 243: As mentioned, this should be for user and farm. I think that this function could return a boolean: are there any tasks due today for this user on this farm? The caller only looks at the length of the returned array.

smattingly (changes requested):
Looking good, but needs some corrected logic.

smattingly (approved):
:+1:

PR #1923, opened 2022-05-19
LF-2156 and LF-2395 Micro copy (Select...) on ghost text for MP and task abandonment reason needs to be dynamic

jagDhillon077:
Create and abandon a crop plan and create and abandon a task. When selecting the options for reasons of abandonment the placeholder select text should be correct for each language


Cali0707 (approved):
Looks good to me

smattingly (approved):
:+1:

PR #1924, opened 2022-05-19
Clicking < from variety detail page should be contextual

jagDhillon077:
Access the variety detail page using the farm map as shown in 
https://lite-farm.atlassian.net/jira/software/c/projects/LF/boards/1?modal=detail&selectedIssue=LF-2318&assignee=626c0502a6d9be006abdad16&assignee=61f0981825edab006a0f0044
and press the back button and notice it goes back to the farm map.

Secondly access the variety detail page by clicking the menu icon on the top left and crops. Click any variety and then press the back button, you should return to the crop catalogue

### DIFF
--- a/packages/webapp/src/components/Crop/Management.jsx
+++ b/packages/webapp/src/components/Crop/Management.jsx
@@ -34,7 +34,7 @@ export default function PureCropManagement({
 
   return (
     <Layout>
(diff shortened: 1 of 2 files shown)


Cali0707 (approved):
Looks good to me

smattingly (commented):
:+1:

PR #1925, opened 2022-05-19
Augment user invitation page to specify the language their invitation account should be sent in

jagDhillon077:
Click the top right icon and click 'my info' and navigate to the people tab and on the bottom it should have a button to invite a new user. When inviting a new user, you will see a new field 'Language of invitation' with a drop down with all 4 different languages a user can choose from.


Cali0707:
@smattingly I've updated this ticket, so it should be ready to merge whenever it passes PR

packages/webapp/src/components/InviteUser/index.jsx
```diff
@@ -44,11 +45,19 @@ export default function PureInviteUser({ onInvite, onGoBack, roleOptions = [] })
...
+  const languageOptions = [
+    { value: 'en', label: t('PROFILE.ACCOUNT.ENGLISH') },
+    { value: 'es', label: t('PROFILE.ACCOUNT.SPANISH') },
+    { value: 'pt', label: t('PROFILE.ACCOUNT.PORTUGUESE') },
+    { value: 'fr', label: t('PROFILE.ACCOUNT.FRENCH') },
+  ];
 
   const disabled = !isValid || !isDirty;
   const onSubmit = (data) => {
+    console.log(data);
```
seinnlee, line 57: I think this was added for testing purpose. Seems a bit unnecessary to print a lot of data

packages/webapp/src/containers/InviteUser/saga.js
```diff
@@ -14,6 +14,7 @@ export function* inviteUserToFarmSaga({ payload: user }) {
   const { user_id, farm_id } = yield select(loginSelector);
   const header = getHeader(user_id, farm_id);
   user.farm_id = farm_id;
+  console.log(user);
```
seinnlee, line 17: Same here as well. Not sure if it's intended to be here.

seinnlee (approved):
LGTM. Tested for Spanish (worked) and French (didn't work but assumed that French translation for the email invitation is not in place yet?)

smattingly (changes requested):
Works well. I think it would be good to disable the new language input until an email address is provided.

Cali0707:
@smattingly I've disabled the language input until an email is provided now - good call.

Cali0707:
@smattingly I have fixed the merge conflicts

smattingly (approved):
:+1:

PR #1927, opened 2022-05-19
LF-2392/more contextual add task flow

Cali0707:
This PR adds contextual defaults to the create task flow when it is triggered from a crop plan. To test:
1. Create a crop plan with both a planting and a transplant task
2. Create a new task from the crop plan
3. When you get to the location selection page, the location of the task should be set based off where the crop would be at that point in time. (i.e. if the crop has been planted in Field 1 and this task is occuring before it is transplanted to field 2, then field 1 should be selected)
4. On the crop plan page, the crop plan should be selected.


anant00729:
The added feature works on my end! code looks fine.

@smattingly Please review.

anant00729 (approved):

smattingly (approved):

PR #1928, opened 2022-05-19
LF-2346 Add "Invoices" and "Receipts" as Document types

anant00729:
To test: 

1. go to the document's home page: http://localhost:3000/documents
2. click on "Add a new document"
3. select any image from your local disk.
4. Now, you'll be redirected to http://localhost:3000/documents/add_document
5. look for the "Type" dropdown. 
6. expand the dropdown and look for "Invoices" and "Receipts".

Screen Shot 2022-05-19 at 12 03 48 PM


Cali0707 (changes requested):
It looks like we also need to make an update to either the model, the database, or both for this to work. When I try and add a document of type "Invoice", I get a popup telling me that it failed. Looking at the console for the api, it says `ValidationError: type: should be equal to one of the allowed values
`

fsayyid (changes requested):
The error Calum pointed out still persists, you have to add the Invoice and Receipts documents types to the backend as well.

Screen Shot 2022-05-19 at 3 32 30 PM

anant00729:
I have added enums 'Invoices' and 'Receipts' as documents "type" in the documents table. Please check again.

Cali0707 (approved):
This works for me

smattingly (changes requested):
This works well. But I think we should add the new types to the display filter, too.

anant00729:
Screen Shot 2022-05-19 at 8 00 07 PM

Added! Please check

fsayyid (approved):
All errors are gone, I am able to create and filter for both invoices and receipts. Looks good to me!

smattingly (approved):
:+1:

PR #1929, opened 2022-05-19
Lf 2322 modify survey stack export document

AbdusShaikh:
This PR exports the answers to the questions asked during a certification document as an Excel sheet.

To Test:
First we need to set up the database to pull the correct surveys. In the psql client...
1. Connect to the 'pg-litefarm' database
2. run the following command `update certifiers set survey_id='627bdbff5b5f71000118f47c'`. This will update the surveys for all certifiers

Then we need to set up the Redis service

1. Under packages/api, you have to create a directory called exports
2. Install the aws cli https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html
3. Authenticate the aws cli by running aws configure in your terminal, and using the DO_SPACES_ACCES_KEY_ID and DO_SPACES_SECRET_ACCESS_KEY values from the packages/api/.env file for the access key id and secret access key fields

Then start the service in a new terminal by typing `npm run scheduler` in the api folder

Then on localhost...
1. Navigate to the certifications page and select any certifier
4. Continue through the flow and fill in the survey
5. After completing the flow, check your email for the documents. You should see an Excel sheet with all of your answers in a format that fits the question

### DIFF
--- a/packages/api/src/jobs/certification/survey_record.js
+++ b/packages/api/src/jobs/certification/survey_record.js
@@ -87,6 +87,18 @@ module.exports = async (submission, exportId) => {
     return col, row;
   };
 
+  const writeMatrixQs = (sheet, col, row, data) => {
+    sheet.cell(`${col}${row}`).value(data['Question']).style({ fontFamily: 'Calibri', bold: true });
+    const categories = Object.keys(data['Answer'][0]);
+    row += 1;
+    for (let i = 0; i < categories.length; i++) {
+      sheet
+        .cell(`${String.fromCharCode(col.charCodeAt(0) + i)}${row}`)
+        .value(categories[i])
+        .style({ fontFamily: 'Calibri', bold: true, border: { color: '000000' } });
+    }
+  };
+
   const typeToFuncMap = {
     string: writeSimpleQs, // Text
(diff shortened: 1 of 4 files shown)


packages/api/src/jobs/certification/survey_record.js
```diff
@@ -0,0 +1,340 @@
+const XlsxPopulate = require('xlsx-populate');
+
+const rp = require('request-promise');
+const surveyStackURL = 'https://app.surveystack.io/api/';
+module.exports = async (submission, exportId) => {
+  const submissionData = await rp({
+    uri: `${surveyStackURL}/submissions/${submission}`,
+    json: true,
+    headers: { Authorization: 'ashaikh@litefarm.org' },
```
Cali0707, line 9: I feel like we should not be using someone's personal email for this. Also, if we are it should still probably be in a `.env` file as opposed to the code

Cali0707 (changes requested):
I tested this and it works locally. We should just figure out what to do with the credentials before it gets merged

Cali0707 (approved):
Works for me

smattingly (approved):
Nice job.

PR #1933, opened 2022-05-20
Refactor task assign modal to not rely on task card contents

fsayyid:
Per recent discussion on Ticket [LF-2323](https://lite-farm.atlassian.net/browse/LF-2323). 

There was issue in the task assign modal where it was relying on task card contents. This was causing an issue when the modal was being used from the read only view rather than from a task card. I have refactored the quick assign modal to grab the tasks from state instead. 

To test:
- Create a few tasks, and play around with assigning, unassigning, and assigning all unassigned tasks on date. Make sure there are no additional bugs I missed.


AbdusShaikh (approved):
Features are working and issues have been addressed. Looks good.

smattingly:
@fsayyid Does this merge mean LF-2323 should go to Verification?

fsayyid:
> @fsayyid Does this merge mean LF-2323 should go to Verification?

@smattingly There is separate related issue that I am working on that me and Kevin decided to link to this ticket, so I think I will take this back to development.

PR #1937, opened 2022-05-24
Lf 2412 user should be able to enter in wages input to delimit dollars from cents

AbdusShaikh:
This PR allows users to input periods into the wage input in the 'invite user' page.

To test:

1. Begin the process to invite a new user
2. In the wage field, you should only be able to enter a single period. Additionally, periods cannot be the first character

### DIFF
--- a/packages/webapp/src/components/Form/Input/index.jsx
+++ b/packages/webapp/src/components/Form/Input/index.jsx
@@ -193,6 +193,11 @@ export default Input;
 
 export const numberOnKeyDown = (e) => ['e', 'E', '+', '-'].includes(e.key) && e.preventDefault();
 export const integerOnKeyDown = (e) => /[^0-9]/.test(e.key) && e.preventDefault();
+export const validateWage = (e) => {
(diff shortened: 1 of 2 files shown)


smattingly (changes requested):
I think we collided a bit with our changes. In light of my changes that have already merged, I'd suggest the following for this PR/ticket:
- no changes to `Form/Input/index.jsx`; I think that is now good
- change `InviteUser/index.jsx` to call `numberOnKeyDown` for wages, not the integer version.

If that doesn't seem to make sense, let me know. Thanks.

smattingly:
@AbdusShaikh Unless I'm missing something, they will be able to use a decimal point in the approach I described. The difference between `numberOnKeyDown` and `integerOnKeyDown` is that the first accepts a decimal point but the second does not.

smattingly:
Closing this due to merge conflicts and other complications with overlapping changes. Only remaining issue is accepting multiple decimal points in a numeric input; will address that in new PR.

PR #1941, opened 2022-05-24
LF-2330 skip crop plans when non-crop locations are selected (task creation flow)

seinnlee:
**Description:**
- Ticket [LF-2330](https://lite-farm.atlassian.net/browse/LF-2330)
- When creating a task, if locations selected by the user are not referenced by active crop plans, bypass the "affect any plans" page

**To Test:**
1. Create one area with no crop plans associated with, one area with abandoned crop plans, and one area with active crop plans.
2. Create a task for the area with no crop plans. Upon selecting that location, you will be directed to the task details page rather than "affect any plans?" page.
3. Repeat for the area with abandoned crop plans (same result) and the area with active crop plans (should be redirected to "affect any plans?" page after location selection.

### DIFF
--- a/packages/webapp/src/components/Task/PureTaskCrops/index.jsx
+++ b/packages/webapp/src/components/Task/PureTaskCrops/index.jsx
@@ -26,6 +26,8 @@ const PureTaskCrops = ({
   isMulti = true,
   isRequired,
   defaultManagementPlanId,
+  history,
+  location
 }) => {
   const { t } = useTranslation();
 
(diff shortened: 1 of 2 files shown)


anant00729:
Tested the [LF-2330](https://lite-farm.atlassian.net/browse/LF-2330) ticket. It now skips the "select plan page" if there is no active plan and directly takes the user to the next step.

anant00729 (approved):

smattingly (approved):
:+1:

PR #1944, opened 2022-05-25
Made migration to add time zones to every farm with latlng

Cali0707:
To test this:
1. Copy the `VITE_GOOGLE_MAPS_API_KEY` from `packages/webapp/.env` to the `GOOGLE_MAPS_API_KEY` in `packages/api/.env`
2. Make multiple farms locally (ideally in multiple timezones)
3. Run the migrations
4. Query the `farm` table in postgres and check if the `utc_offset` values are correct. They should be the utc offset in minutes for the location of the farm

### DIFF
--- a/packages/api/src/jobs/station_sync/updateTimeZoneOffsets.js
+++ b/packages/api/src/jobs/station_sync/updateTimeZoneOffsets.js
@@ -1,0 +1,21 @@
+const { Client } = require('@googlemaps/google-maps-services-js');
+
+const client = new Client({});
+
+async function mapTimeZoneOffsetsToFarms(knex) {
+  const farmsWithLatLng = await knex('farm').select('*').whereNotNull('grid_points');
+  for (const farm of farmsWithLatLng) {
+    const timeZone = await client.timezone({
+      params: {
+        location: farm.grid_points,
+        timestamp: new Date(Date.now()),
+        key: process.env.GOOGLE_MAPS_API_KEY,
+      },
+    });
+    await knex('farm')
+      .update('utc_offset', Math.round(timeZone.data.rawOffset / 3600))
+      .where('farm_id', farm.farm_id);
+  }
+}
+
+module.exports = mapTimeZoneOffsetsToFarms;
--- a/packages/api/src/jobs/station_sync/updateTimeZoneOffsets.js
+++ b/packages/api/src/jobs/station_sync/updateTimeZoneOffsets.js
@@ -5,16 +5,47 @@ const client = new Client({});
 async function mapTimeZoneOffsetsToFarms(knex) {
   const farmsWithLatLng = await knex('farm').select('*').whereNotNull('grid_points');
   for (const farm of farmsWithLatLng) {
-    const timeZone = await client.timezone({
-      params: {
-        location: farm.grid_points,
-        timestamp: new Date(Date.now()),
-        key: process.env.GOOGLE_MAPS_API_KEY,
-      },
-    });
-    await knex('farm')
-      .update('utc_offset', Math.round(timeZone.data.rawOffset / 60))
-      .where('farm_id', farm.farm_id);
+    try {
+      const timeZone = await client.timezone({
+        params: {
+          location: farm.grid_points,
+          timestamp: new Date(Date.now()),
+          key: process.env.GOOGLE_MAPS_API_KEY,
+        },
+      });
+      await knex('farm')
+        .update('utc_offset', Math.round(timeZone.data.rawOffset / 60))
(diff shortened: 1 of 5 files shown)


anant00729:
The code looks fine to me. I have tested the migration scripts. 

Up function 
1. For farms in different time-zones, 'google-maps-services-js' is providing timeZone.data.raw offset data that gets converted to the minutes and stored in the column of utc_offset.

Down function 
1. Works fine as well.

can add try, catch in case you want to handle/catch the Promise errors. 
Screen Shot 2022-05-25 at 2 47 50 PM

anant00729 (approved):

packages/api/src/jobs/station_sync/updateTimeZoneOffsets.js
```diff
@@ -0,0 +1,21 @@
+const { Client } = require('@googlemaps/google-maps-services-js');
+
+const client = new Client({});
+
+async function mapTimeZoneOffsetsToFarms(knex) {
+  const farmsWithLatLng = await knex('farm').select('*').whereNotNull('grid_points');
+  for (const farm of farmsWithLatLng) {
+    const timeZone = await client.timezone({
```
smattingly, line 8: Please check [here](https://developers.google.com/maps/documentation/timezone/requests-timezone#TimeZoneStatus) to see if any error response handling is needed.

smattingly (changes requested):
Looks good-- I had a few questions/suggestions.

packages/api/src/jobs/station_sync/updateTimeZoneOffsets.js
```diff
@@ -0,0 +1,21 @@
...
+      params: {
+        location: farm.grid_points,
+        timestamp: new Date(Date.now()),
+        key: process.env.GOOGLE_MAPS_API_KEY,
+      },
+    });
+    await knex('farm')
+      .update('utc_offset', Math.round(timeZone.data.rawOffset / 60))
```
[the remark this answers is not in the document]
Cali0707: @smattingly the `rawOffset` is in seconds, so here we are converting it to minutes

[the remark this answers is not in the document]
smattingly: Right, sorry. Let's keep it as the API provides it -- seconds.

packages/api/src/jobs/station_sync/updateTimeZoneOffsets.js
Cali0707: I've added error handling and rate limiting checks

packages/api/src/jobs/station_sync/updateTimeZoneOffsets.js
```diff
@@ -1,20 +1,66 @@
...
-      },
-    });
-    await knex('farm')
-      .update('utc_offset', Math.round(timeZone.data.rawOffset / 60))
-      .where('farm_id', farm.farm_id);
+    try {
+      const timeZone = await client.timezone({
+        params: {
+          location: farm.grid_points,
+          timestamp: new Date(Date.now()),
```
smattingly, line 27: Suggestion: one new Date at the beginning. I doubt that the differences would matter, but might be better to have the data as of a single point in time.

packages/api/src/jobs/station_sync/updateTimeZoneOffsets.js
```diff
@@ -1,20 +1,66 @@
...
+          key: process.env.GOOGLE_MAPS_API_KEY,
+        },
+      });
+      await knex('farm')
+        .update('utc_offset', Math.round(timeZone.data.rawOffset / 60))
+        .where('farm_id', farm.farm_id);
+    } catch (e) {
+      switch (e.response?.data?.status) {
+        case 'OVER_QUERY_LIMIT':
+          console.log('Hit query limit for API: waiting for query limit to reset');
```
smattingly, line 37: In logs, it might not be clear which API. Please add some more context to the message.

smattingly (changes requested):
It looks to me like hitting the rate limit a second time will cause it to stop? Not necessarily a problem, just wonder if I read that right.

Cali0707:
> It looks to me like hitting the rate limit a second time will cause it to stop? Not necessarily a problem, just wonder if I read that right.

Not quite - because the try-catch block is on a per-request basis, we should have a one minute delay after each time we hit the rate limit. If it occurs twice in a row on the same request, then I made a judgment call that the request should fail so that we do not keep retrying indefinitely - not sure sure if this is appropriate or not.

PR #1946, opened 2022-05-25
bug fix for Clicking < from variety detail page should be contextual

jagDhillon077:
@AbdusShaikh please check to see if you are getting the same looping behaviour when trying to return to original page from crop plans


AbdusShaikh (approved):
Back button is working correctly and returns screen to correct page.

smattingly (approved):

PR #1952, opened 2022-05-26
Add time offsets to farm creation in backend

seinnlee:
**To Test:**
1. You would need to review this [PR](https://github.com/LiteFarmOrg/LiteFarm/pull/1944) first and follow the steps there.
2. Create a new farm in multiple locations
3. Check their UTC offset in the database `farm` table (should be in seconds)


seinnlee:
@anant00729 it turns out that farms cannot update their locations once created so there's no user of updating their time offsets when updating the farm

anant00729:
The update function is not updating the utc_offset on calling updateFarm function in the farmControlller. You need to call the  'const utc_offset = await this.getUTCOffsetFromGridPoints(req.body.grid_points);' to update the value of utc_offset value in the database.

packages/api/src/controllers/farmController.js
```diff
@@ -188,32 +200,77 @@ const farmController = {
...
-    const { iso, unit, id } = await knex('countries').select('*').where('country_name', country).first();
+    const { iso, unit, id } = await knex('countries')
+      .select('*')
+      .where('country_name', country)
+      .first();
     return { currency: iso, measurement: unit.toLowerCase(), id };
   },
-}
+
+  async getUTCOffsetFromGridPoints(gridPoints) {
```
anant00729, line 235: The update function is not updating the utc_offset on calling updateFarm function in the farmControlller. You need to call the 'const utc_offset = await this.getUTCOffsetFromGridPoints(req.body.grid_points);' to update the value of utc_offset value in the database.

anant00729 (changes requested):

anant00729:
The functionality of inserting the utc_offset works for me. @smattingly Please review.

anant00729 (approved):

anant00729:
Here is the snap of the entries that we tested
Screen Shot 2022-05-26 at 10 38 00 AM

smattingly (approved):

PR #1953, opened 2022-05-26
Get farms by offset range endpoint

AbdusShaikh:
This PR creates an endpoint that, when hit, pulls the farm id's of all farms within a certain UTC offset range.

To test:

1. Run `npm run migrate:testing:db` and `npm run migrate:testing:db` in the api folder
2. Go into the 'farm' table your database and modify the 'utc_offset' field of the farms. Have different values so you can properly test for range.
3. To receive a response from this endpoint we need to authenticate via a token. First go into 'farmRoute.js' and comment out lines 68 & 69. Then launch your local Litfarm environment, open developer tools, go to application, and copy the value for 'id_token'
4. Now we will make a request. Open a terminal and write `curl -H "Authorization: Bearer <YOUR-ID-TOKEN>" http://localhost:5001/farm/utc_offset_by_range/<CUSTOM-MIN-VALUE>/<CUSTOM-MAX-VALUE>` (Make sure to replace the values appropriately)

You should receive a list of the farm id's of all the farms that match your specification.


smattingly (approved):

packages/api/src/routes/farmRoute.js
```diff
@@ -61,4 +63,11 @@ router.delete(
   farmController.deleteFarm(),
 );
 
+router.get(
+  '/utc_offset_by_range/:min/:max',
+  checkSchedulerJwt,
+  hasTimeNotificationsAccess,
```
fsayyid, line 69: I'm not sure we want to use the same JWT token payload for requests to this endpoint, since we are not requesting timed notifications in this case. I think using something like`{"requestFarmsInTimeZone": true}` and making middleware to check for that would make more sense. @smattingly what do you think?

fsayyid:
I am unable to send a request because it is looking for the wrong JWT token. Add `/\/farm\/utc_offset_by_range\//i,` to the unless list in `checkJwt.js` middleware

smattingly:
I had already merged this branch into the feature branch, so it does not need to go to integration. I beileve that @fsayyid 's comments were addressed on the feature branch.

PR #1956, opened 2022-05-26
Prevent multiple decimal points in number input

smattingly:


AbdusShaikh (changes requested):
Works fine on Safari except when '.' is the first character in the input. In this case Safari allows multiple decimal points.

Also, the actual wage input (in /webapp/src/components/InviteUser/index.jsx) is not currently using the numberOnKeyDown function. I assume it should be since this fix is meant for that.

packages/webapp/src/components/Form/Input/index.jsx
```diff
@@ -201,12 +201,21 @@ const isEventOkForIntegerInput = (event) => {
...
+/**
+ * Always accepts backspace, arrow keys, and numbers.
+ * Only accepts one period and only after 1 or more numbers.
+ */
 export const numberOnKeyDown = (event) => {
-  if (isEventOkForIntegerInput(event)) return;
-  const decimalSeparator = '.'; // May want to support internationalization here.
-  if (event.key === decimalSeparator && !event.target.value.includes(decimalSeparator)) return;
-  event.preventDefault();
+  if (!['Backspace', 'ArrowUp', 'ArrowDown', 'ArrowRight', 'ArrowLeft'].includes(event.code)) {
```
smattingly, line 210: Probably need to support Home, Ins(ert), End, and possibly others (Tab to move to next input?).
That's why I avoided listing the "editing" keys.

PR #1962, opened 2022-05-27
Add utc offset to timed notification endpoints

fsayyid:
Added utc offsets to the request body of the timed notification endpoints.

NOTE: I did not add the utc offset where the requests are called in the actual job scheduler, since there is a range of timezone, so I wasn't sure if I should put the start or end.

### DIFF
--- a/packages/api/src/controllers/timeNotificationController.js
+++ b/packages/api/src/controllers/timeNotificationController.js
@@ -66,28 +66,28 @@ const timeNotificationController = {
    */
   async postDailyDueTodayTasks(req, res) {
     const { farm_id } = req.params;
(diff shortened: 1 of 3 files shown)


smattingly (commented):
I'm not sure, but I think this is more complicated than we need. I may not have been very clear, and the "requirements" have become clearer to me as I worked the other side of this.

All we really need is to tell the endpoint whether to use the server's current date as "today", or whether to move that one day forward. (When it is 6am Tuesday in UTC+7, and time to send daily notifications, the server was sending notifications for Monday, because it is still Monday in UTC+0 which the server runs on.)

In other words, the new parameter could be just a flag-- use one day forward, or don't.

PR #1972, opened 2022-05-31
F21 592: Fix: farm name character limit error handling

anant00729:
To Test:

1. Go to create a farm page in a frame name input field fill in characters more than 255.
2. check the error message.

"Scenario 1" - For the first time, it will call the post API of addFarm. 

3. Therefore, fill in a valid farm name and go to the next form page.
4. Now, come back to the same page and add a farm name having more than 255 characters.

"Scenario 2" - This time it will show the error from the update/patch API of the farm. 

match the error message from both scenarios.  

For more details: 
https://lite-farm.atlassian.net/browse/F21-592

### DIFF
--- a/packages/webapp/src/containers/AddFarm/index.jsx
+++ b/packages/webapp/src/containers/AddFarm/index.jsx
@@ -71,11 +71,19 @@ const AddFarm = () => {
     errors[COUNTRY]?.message ||
     errorMessage[errors[ADDRESS]?.type];
 
+  const showFarmNameCharacterLimitExceededError = () => {
+    setError(FARMNAME, {
+      type: 'manual',
+      message: t('ADD_FARM.FARM_NAME_ERROR'),
+    });
+  };
+
   const onSubmit = (data) => {
     const farmInfo = {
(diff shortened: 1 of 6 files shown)


packages/webapp/src/containers/AddFarm/index.jsx
```diff
@@ -71,11 +71,19 @@ const AddFarm = () => {
...
+      message: t('ADD_FARM.FARM_NAME_ERROR'),
+    });
+  };
+
   const onSubmit = (data) => {
     const farmInfo = {
       ...data,
       gridPoints,
       farm_id: farm ? farm.farm_id : undefined,
+      showFarmNameCharacterLimitExceededError: showFarmNameCharacterLimitExceededError,
```
fsayyid, line 86: I'm not sure about adding a function into a redux action payload. It seems to be working though.

anant00729: I took the reference of user sign-in when the user types a wrong password and from the API, we show the error message in the input field.

PR #1976, opened 2022-05-31
F21 608 creation flow for location (barn, greenhouse, field, farm boundary, potentially others ) loops 

AbdusShaikh:
This fix addresses an unwanted loop in the location creation flow

To test:
1. Log in to system
2. Go to “my farm” > “farm map”
3. Click “+”
4. Select a type
5. Draw shape
6. Click redraw and redraw
8. Click “Confirm”
9. Click “<“
10. Click “Confirm” again
11. Enter all required attributes for barn and click “Save”

After all of that, you should arrive at your farm map with the new addition.

**Note:** This issue was originally discovered on Barns and Greenhouses. I noticed the same issue in other area types and I extended the fix to them as well.


jagDhillon077 (approved):
LGTM - no looping issues

smattingly (approved):

PR #1977, opened 2022-05-31
Crop plan loop bug fix

jagDhillon077:
[Jira Ticket](https://lite-farm.atlassian.net/jira/software/c/projects/F21/boards/3?modal=detail&selectedIssue=F21-585)

Access the management plans page via the crop catalogue or the farm map and select a crop variety and then a plan. Go to the details page of the crop plan and press the back button to go back to the page you accesses the crop plans from.  If you accessed the crop plans from the farm map you should return to the farm map, similar behaviour should show when accesses crop plans via the crop catalogue. Also there should not be any looping behaviour when pressing the back buttons.

### DIFF
--- a/packages/webapp/src/components/Crop/Detail.jsx
+++ b/packages/webapp/src/components/Crop/Detail.jsx
@@ -54,7 +54,7 @@ function PureCropDetail({
       }
     >
       <CropHeader
-        onBackClick={onBack}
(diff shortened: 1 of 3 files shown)


AbdusShaikh (approved):
Back button is working correctly and bringing me to the correct places

PR #1978, opened 2022-05-31
F21-582/switch farms reset notification reducer state

seinnlee:
Ticket [F21-582](https://lite-farm.atlassian.net/browse/F21-582) and [F21-583](https://lite-farm.atlassian.net/browse/F21-583)

**Description:**
This PR fixes the bug where a user was seeing notifications from multiple farms on notification center when switching farms.
When a user switches farms, the notification reducer state will be reseted. [Selective reset redux state](https://webdev99.com/resetting-redux-state-with-a-root-reducer-completely-selectively/)

**To Test:**
1. You need a user with multiple farms (create one or use the one you have if you already have one).
2. Please either clear site data or log out before you test this out.
3. Log into the user's account and select one of the farms.
4. Create a notification for yourself (by assigning a task, etc.).
5. You should see the notification in the notification center.
6. Now, switch to a different farm.
7. Check the notification center and you should not be able to see the notification from another farm (step 5).

As discussed in the meeting, this PR will also solve [F21-583](https://lite-farm.atlassian.net/browse/F21-583). Users will still be able to access tasks from another farm since they have access to those tasks (by manually typing in the url) but the application's UI will not give any ways for them to navigate as such.

### DIFF
--- a/packages/webapp/src/containers/notificationSlice.js
+++ b/packages/webapp/src/containers/notificationSlice.js
@@ -43,10 +43,11 @@ export const { onLoadingNotificationStart, onLoadingNotificationFail, getNotific
   notificationSlice.actions;
 export default notificationSlice.reducer;
 
-export const notificationReducerSelector = (state) => state.entitiesReducer[notificationSlice.name];
+export const notificationReducerSelector = (state) =>
+  state.farmStateReducer[notificationSlice.name];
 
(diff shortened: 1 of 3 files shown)


fsayyid (approved):
It's working on my end, and the logic makes sense to me. Good work!

smattingly (approved):

PR #1985, opened 2022-06-01
 F21-660 finances date range doesnt filter revenues appropriately

fsayyid:
Ticket: [F21-660](https://lite-farm.atlassian.net/browse/F21-660)

FIxed finance page so that the revenue changes with the date range. 

To test:
- Make sure you have a crop variety with an active crop plan (not complete or abandoned);
- Got to Finances
- Click "Add new sale" and fill this out a couple times for different dates
- Make you are getting correct values for Actual Revenue where filtering by those date ranges
- Try making the start date after the end date, make sure the Actual Revenue is 0

### DIFF
--- a/packages/webapp/src/containers/Finances/ActualRevenue/index.jsx
+++ b/packages/webapp/src/containers/Finances/ActualRevenue/index.jsx
@@ -11,7 +11,7 @@ import { AddLink, Semibold } from '../../../components/Typography';
 import DateRangePicker from '../../../components/Form/DateRangePicker';
 import ActualCropRevenue from '../ActualCropRevenue';
 import FinanceListHeader from '../../../components/Finances/FinanceListHeader';
-import { calcRevenue, filterSalesByDateRange } from '../util';
+import { calcActualRevenue, filterSalesByDateRange } from '../util';
 
 export default function ActualRevenue({ history, match }) {
   const { t } = useTranslation();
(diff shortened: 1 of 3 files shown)


jagDhillon077 (changes requested):
LGTM -
just remove console log on packages/webapp/src/containers/Finances/index.jsx line 97

fsayyid:
@Cali0707 I estimated revenue is working on my end, and I fixed the issue with the Actual Revenue page. I'm unsure about date range mirroring issue and what the expected behaviour is.

Cali0707 (changes requested):
@fsayyid it seems on my end like something is not working now. When I go to the "Real revenue" page,  I can see the total revenue, but it is not showing the sale that should be in the range of dates selected. My current farm has a sale worth $50,000 that occured June 1st 2022, but with the following range filter nothing is showing up:
expenses_seem_wrong

fsayyid:
@Cali0707 I think I fixed the issue.

Cali0707 (changes requested):
@fsayyid I am still having the same error - to reproduce, add a new sale on your farm. For me it does not show up on the "Actual revenue" page

Cali0707 (approved):
This works for me, looks like I had not pulled the latest change when I said that it was still not working - sorry for the confusion!

PR #1989, opened 2022-06-01
Fixed filtering on crop management plans

Cali0707:
This was fixing https://lite-farm.atlassian.net/browse/F21-651 

This sentry error was caused when someone tried to search on the "crop" page associated with a location. I fixed the error, and also updated the filtering to filter the crops by the appropriate language.

To test:
1. Have at least one crop plan associated with a location. 
2. Go to map view, and click on that location
3. Go to the crops tab, and search. It should correctly filter out the crops based on the search term, and also on the language you are currently using


fsayyid (approved):
Looks good, its working on my end.

smattingly (approved):

PR #1992, opened 2022-06-01
Refined to fit large names

jagDhillon077:


fsayyid (approved):
Looks good, working on my end when I try both small and large screen sizes with long strings.

smattingly (approved):

PR #1996, opened 2022-06-02
F21-650 french export email not translating 

fsayyid:
Ticket: [F21-650](https://lite-farm.atlassian.net/browse/F21-650)

It looks like it is translating correctly, but I think someone may have gotten confused while translating.

I don't think this one can be tested properly until the translations are fixed.


Cali0707:
@fsayyid I think our work got a little confused here - I added the missing translations into a separate PR so I think we can close this one

PR #2009, opened 2022-06-03
Adding translations to support values on the help request template

kikemarquez8:


packages/api/src/templates/locales/fr.json
```diff
@@ -97,5 +101,6 @@
     "SUBJECT": "EXPORT.SUBJET",
     "PRIVACY": "EXPORT.CONFIDENTIALITÉ",
     "LANGUAGE_DELAY": "EXPORT.DÉLAIS.LANGUE"
```
Cali0707, line 103: These three look like they were not translated properly in the past, can you mark them as "MISSING"  and I'll add them later when I do another pass of FR translations?

Cali0707 (changes requested):

Cali0707 (approved):
This looks good to me 👍

PR #2010, opened 2022-06-03
Fixed complete harvest task leading to blank page

Cali0707:
I was able to recreate the error Sarah-Louise reported in https://lite-farm.atlassian.net/browse/F21-667 by using her credentials for her farm. However, the error was no longer occuring for custom task types (I imagine it got fixed somewhere else). So, the only error I was able to recreate on her farm was the blank page that occurred while completing a harvest task. This PR fixes that blank page bug.

To test:
1. Create a harvest task
2. Complete the task and only select one harvest use
3. You should not see a blank page once you complete the task


fsayyid (approved):
Looks good, its working on my end.

PR #2011, opened 2022-06-03
Remove redundant env var

smattingly:
We recently added `GOOGLE_MAPS_API_KEY` to the api package. That is the name used by the webapp package. api already had `GOOGLE_API_KEY`, which is the same value. This removes the new, redundant environment variable.


Cali0707 (approved):
Looks good to me, sorry for introducing the redundant env variable - I didn't realize that it was already in the .env file.

anant00729 (approved):
This looks fine to me. Made .evn name changes in the controller and migration script.

PR #2013, opened 2022-06-03
Need to ensure after a user switches to a new farm that they receive alerts for that farm

jagDhillon077:
- Setup:

Make sure to have 2 different farms
Create X number of tasks for farm A (I just created 2) and assign these tasks to yourself. Any tasks is fine I just did cleaning task as it is the easiest to work with. You will notice the notification bell update when you have created the task.

Create Y number of tasks for farm B (I just created 3) and do the same thing above, assign to self and make sure the notification bell on this second farm has 3 tasks. 

- Test

Switch between the two farms and the notification bell alerts number should update based on the amount of alerts in each farm.

### DIFF
--- a/packages/webapp/src/containers/Home/index.jsx
+++ b/packages/webapp/src/containers/Home/index.jsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import { React, useEffect } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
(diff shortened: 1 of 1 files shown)


seinnlee (approved):
Everything looks good to me!

smattingly (approved):

PR #2017, opened 2022-06-03
fix: hide take me there when notification.ref is null

fsayyid:
Ticket: [F21-629](https://lite-farm.atlassian.net/browse/F21-629)

To test:
- Make sure you have at least on user in your database
- Open `notifyAllUsers.sql` in your code editor.
- Replace the value for ref in the `INSERT` command to either `NULL` or `'{}'::jsonb`
- Copy and paste commands one by one in postgres
- Login and go to notifications
- You should see the notification you just created with no "Take Me There" link


Cali0707 (approved):
This works on my end!

smattingly (approved):

PR #2023, opened 2022-06-06
Lf 2342/cypress tests for the login flow

MwayaB:


smattingly (changes requested):
I"m not able to run the tests as they are currently configured. I think that the mail server is running, but that Cypress does not have access to the port?

smattingly (approved):

PR #2027, opened 2022-06-06
Update searchbar and filter on notifications listview

AbdusShaikh:
This PR fixes the search functionality in the notifications centre. It also allows users to receive accented character outputs from non accented search inputs. Additionally, the "filter" icon is no longer visible beside the search bar because filtering is not applicable here.

Ticket: [F21-695](https://lite-farm.atlassian.net/browse/F21-695)

To test:
1. Navigate to the notifications centre where all notifications are displayed
2. Ensure the "filter" icon is not appearing beside the search bar
3. Enter a search term into the search bar and ensure the appropriate notifications are appearing. Try testing searches against both the body and the title of the notification

To thoroughly test this functionality, try changing the language and searching for terms with accented characters in them. For example, you should find that searching for "a" will match with notifications that have "à" or "â" in their title/body.

### DIFF
--- a/packages/webapp/src/components/PopupFilter/PureSearchbarAndFilter/index.jsx
+++ b/packages/webapp/src/components/PopupFilter/PureSearchbarAndFilter/index.jsx
@@ -34,7 +34,13 @@ const useStyles = makeStyles({
   },
 });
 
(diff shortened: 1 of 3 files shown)


jagDhillon077 (approved):
LGTM

PR #2028, opened 2022-06-06
Removed download button for iOS users

Cali0707:
This PR makes the changes requested by Kevin on https://lite-farm.atlassian.net/browse/F21-639

This PR removes the option for iOS users to directly download the map as different methods of downloading the map work on different browsers/iOS versions. It also reverts to the previous method of downloading the map image because that did not have need of a "Loading..." text on the download button before the user had clicked on it

### DIFF
--- a/packages/webapp/src/components/Modals/ExportMapModal/index.jsx
+++ b/packages/webapp/src/components/Modals/ExportMapModal/index.jsx
@@ -6,19 +6,23 @@ import { AiOutlineMail } from 'react-icons/all';
 import styles from './styles.module.scss';
 import { Main, Title } from '../../Typography';
 import Button from '../../Form/Button';
-import html2canvas from 'html2canvas';
+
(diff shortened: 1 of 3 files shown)


Cali0707:
@smattingly do you have any ideas why the team city build is failing here?

smattingly:
@Cali0707 There is a common category of "false problems" reported by TeamCity because it says it can no longer reach the commit for the build that is now at the head of the queue. Appears to happen when more changes are made while the build is in the queue. As far as I can tell, it is always a "non issue"-- just one more reason I'll be happy to move to Actions.

fsayyid (approved):
The loading issue Kevin mentioned seems to be gone now, and both buttons are still showing on Mac (and likely other non iOS systems) as desired. We can verify that this is hiding the download on iOS after it is merged into beta.

smattingly (approved):

PR #2029, opened 2022-06-06
Prevent server crash

smattingly:
This fixes [F21-687](https://lite-farm.atlassian.net/browse/F21-687). See details there.


Cali0707 (approved):
This looks like a good way to avoid the server crashing in the event that an undefined value somehow gets through all of our validation.

PR #2030, opened 2022-06-06
Added contextual navigation for wild crop plans

Cali0707:
This resolves https://lite-farm.atlassian.net/browse/F21-579

To test:
1. Create a wild crop plan. (When making the plan ensure you select "in ground" to be able to make it a wild crop)
2. Create a new task from the crop management plan
3. As you go through the task creation flow, the "This task targets a wild crop" checkbox should be checked automatically. Additionally, the crop plan should be pre-selected when you reach that page
4. When you finish creating the task, you should be returned to the management plan, not the tasks page


jagDhillon077 (approved):
LGTM - the checkbox is automatically selected and I am returning to the management plans after creating the task for the wild crop

PR #2031, opened 2022-06-06
Notifications-are-sorted-in-the-wrong-order

jagDhillon077:
[Jira Ticket](https://lite-farm.atlassian.net/jira/software/c/projects/F21/boards/3?modal=detail&selectedIssue=F21-697)

To test: 
Go to the notifications page and check to see if the notifications are sorted in descending order. (ie. June 6 should be on top while June 5th notification should be below the June 6th notification)

### DIFF
--- a/packages/webapp/src/containers/Notification/index.jsx
+++ b/packages/webapp/src/containers/Notification/index.jsx
@@ -66,6 +66,7 @@ export default function NotificationPage() {
               />
             );
           })
+          .reverse()
       ) : (
         <Semibold style={{ color: 'var(--teal700)', marginLeft: '24px' }}>
           {t('NOTIFICATION.NONE_TO_DISPLAY')}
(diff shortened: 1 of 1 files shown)


anant00729:
This works for me.

anant00729 (approved):

packages/webapp/src/containers/Notification/index.jsx
```diff
@@ -66,6 +66,7 @@ export default function NotificationPage() {
               />
             );
           })
+          .reverse()
```
smattingly, line 69: I don't think the Jira ticket gave a full description of the requirements. As a result, I don't think this will fix the problem. I've asked for confirmation in Jira.

smattingly (changes requested):

anant00729 (changes requested):
Will follow up with the new changes

jagDhillon077:
Removed functionality where alerts are taken into account when sorting notifications

anant00729 (approved):
Looks good to me!

smattingly (approved):

PR #2034, opened 2022-06-06
Updated dates to mirror those on the other finance pages

Cali0707:
This PR resolves https://lite-farm.atlassian.net/browse/F21-669

To test:
1. Go to the "Finances" page
2. Set the date range to some initial value that is different from the default value
3. Go to both the "Actual Revenue" and "Estimated Revenue" pages and see if the date range there matches the value on the main finance page (it should)
4. Change the date range in the actual and estimated revenue pages and see if it updates on the main finances page (it should)

### DIFF
--- a/packages/webapp/src/containers/Finances/ActualRevenue/index.jsx
+++ b/packages/webapp/src/containers/Finances/ActualRevenue/index.jsx
@@ -32,8 +32,8 @@ export default function ActualRevenue({ history, match }) {
     mode: 'onBlur',
     shouldUnregister: true,
     defaultValues: {
-      from_date: new Date(dateRange.startDate).toISOString().split('T')[0],
-      to_date: new Date(dateRange.endDate).toISOString().split('T')[0],
+      from_date: dateRange.startDate,
+      to_date: dateRange.endDate,
     },
   });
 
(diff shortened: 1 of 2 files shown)


fsayyid (approved):
Logic makes sense, and the date ranges are all mirroring each other on my end 👍

packages/webapp/src/containers/Finances/ActualRevenue/index.jsx
```diff
@@ -30,8 +32,8 @@ export default function ActualRevenue({ history, match }) {
     mode: 'onBlur',
     shouldUnregister: true,
     defaultValues: {
-      from_date: moment().startOf('year').format('YYYY-MM-DD'),
-      to_date: moment().endOf('year').format('YYYY-MM-DD'),
+      from_date: new Date(dateRange.startDate).toISOString().split('T')[0],
```
smattingly, line 35: I believe that `dateRange.startDate` is equivalent to `new Date(dateRange.startDate).toISOString().split('T')[0]`. 
Same comment applies elsewhere in this commit.

smattingly (changes requested):

packages/webapp/src/containers/Finances/ActualRevenue/index.jsx
Cali0707: Good point, I've resolved this now

PR #2035, opened 2022-06-07
Added default map pin

Cali0707:
This PR resolves https://lite-farm.atlassian.net/browse/F21-698

To test:
1. Either retire all existing locations on one of your farms, or create a new farm without any locations
2. Go to the farm map
3. You should see a pin with the name of your farm beneath it at the latlong coordinates of your farm


fsayyid (approved):
This is working well on my end.

smattingly (approved):

PR #2036, opened 2022-06-07
Track and limit subscriptions per user, browser

smattingly:
This fixes a problem related to alert tracking opening too many subscriptions.
See [LF-688](https://lite-farm.atlassian.net/browse/F21-688) for details.

### DIFF
--- a/packages/api/src/controllers/notificationUserController.js
+++ b/packages/api/src/controllers/notificationUserController.js
@@ -25,7 +25,30 @@ module.exports = {
    * @param {Response} res - The HTTP response object.
    */
   subscribeToAlerts(req, res) {
-    const { user_id, farm_id } = req.query;
+    const { user_id, farm_id, subscriber_id } = req.query;
+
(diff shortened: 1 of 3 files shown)


Cali0707 (commented):
@smattingly this works great for users in the same browser (i.e. multiple tabs in Chrome). However, when I logged in on Firefox while keeping 5 Chrome tabs open, I was able to open another subscription:
```
Opening subscription: user 202b8ec8-e66f-11ec-ac45-acde48001122 as subscriber 9bb595be-ee67-4e9f-8403-99e3526d1bce at 2022-06-07T16:17:54.609Z
  1 users have 6 active subscriptions.
  ```

smattingly:
@Cali0707 Yes, that's intended. You can use as many browsers/devices as you want, but a given browser is limited to 5 active alerting tabs.

smattingly:
This breaks the fix to ensure correct alerts when user changes farms. Going to Draft status; more commits to come.

Cali0707 (approved):
It seems like this is working to me - there is never more than 5 subscriptions per user per browser

PR #2044, opened 2022-06-07
F21 698 sticky pin based on gps location from farm creation flow on map until first area drawn

Cali0707:
Following up on the new desired behaviour described in the comments on https://lite-farm.atlassian.net/browse/F21-698

To test:
1. Create a new farm or retire all the existing locations on your current farm
2. You should see a pin (but no title) at the location of your farm
3. Add a location
4. As soon as you begin adding the location, the pin for your farm should go away
5. When there are 1+ locations on the farm, the pin should not appear

### DIFF
--- a/packages/webapp/src/containers/Map/index.jsx
+++ b/packages/webapp/src/containers/Map/index.jsx
@@ -66,6 +66,8 @@ export default function Map({ history }) {
   const [showZeroAreaWarning, setZeroAreaWarning] = useState(false);
   const successMessage = useSelector(setSuccessMessageSelector);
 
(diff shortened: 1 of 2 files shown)


AbdusShaikh (approved):
Looks good. Pin was there when there were no locations, and disappeared when I added a location. Pin also reappears when all previously existing locations are retired.

jagDhillon077 (commented):
@Cali0707 When you said 'As soon as you begin adding the location, the pin for your farm should go away' does this mean as soon as I start to draw or once I have confirmed a location? The pin does not go away while I am drawing only when the location has been confirmed as @AbdusShaikh mentioned

smattingly (approved):

PR #2046, opened 2022-06-07
Fixed various weird date problems

Cali0707:
This is a relatively messy bit of code to create the default values for the revenue date range, but the issue is that the `dateRange` object can have dates that are either a moment.js object, a string, or undefined. I did not have time to remove all of the moment.js instances throughout this part of the codebase, so I implemented a temporary fix here.

To test: 

1. Go to the "Finances" page
2. Set the date range to some initial value that is different from the default value
3. Go to both the "Actual Revenue" and "Estimated Revenue" pages and see if the date range there matches the value on the main finance page (it should)
4. Change the date range in the actual and estimated revenue pages and see if it updates on the main finances page (it should)

### DIFF
--- a/packages/webapp/src/containers/Finances/ActualRevenue/index.jsx
+++ b/packages/webapp/src/containers/Finances/ActualRevenue/index.jsx
@@ -37,10 +37,22 @@ export default function ActualRevenue({ history, match }) {
     shouldUnregister: true,
     defaultValues: {
       from_date: dateRange?.startDate
-        ? new Date(dateRange.startDate.split('T')[0] + 'T00:00:00.000Z').toISOString().split('T')[0]
+        ? new Date(
+            typeof dateRange.startDate === 'string'
+              ? dateRange.startDate.split('T')[0] + 'T00:00:00.000Z'
+              : dateRange.startDate,
+          )
+            .toISOString()
+            .split('T')[0]
         : `${year}-01-01`,
       to_date: dateRange?.endDate
-        ? new Date(dateRange.endDate.split('T')[0] + 'T00:00:00.000Z').toISOString().split('T')[0]
+        ? new Date(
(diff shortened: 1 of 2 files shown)


fsayyid (changes requested):
Screen Shot 2022-06-07 at 3 26 33 PM
Getting these errors on Actual Revenue

packages/webapp/src/containers/Finances/ActualRevenue/index.jsx
```diff
@@ -22,6 +22,10 @@ export default function ActualRevenue({ history, match }) {
   const dateRange = useSelector(dateRangeSelector);
   const dispatch = useDispatch();
 
+  console.log(dateRange);
```
fsayyid, line 25: Remove console.log

packages/webapp/src/containers/Finances/ActualRevenue/index.jsx
```diff
@@ -32,13 +36,18 @@ export default function ActualRevenue({ history, match }) {
     mode: 'onBlur',
     shouldUnregister: true,
     defaultValues: {
-      from_date: dateRange.startDate,
-      to_date: dateRange.endDate,
+      from_date: dateRange?.startDate
+        ? new Date(dateRange.startDate.split('T')[0] + 'T00:00:00.000Z').toISOString().split('T')[0]
```
fsayyid, line 40: Not sure exactly what your trying to do, but are you trying to prevent the date being changed by the timezone?

Something like this has worked for me in the past: `new Date(startDate.getUTCFullYear(), startDate.getUTCMonth(), startDate.getUTCDate())`

fsayyid (approved):
This is working on my end.

smattingly (approved):

PR #2049, opened 2022-06-07
fix: select and create language fix for all drop downs - F21-696, F21-701

fsayyid:
Tickets: [F21-701](https://lite-farm.atlassian.net/browse/F21-701), [F21-696](https://lite-farm.atlassian.net/browse/F21-696)


Cali0707 (approved):
Looks good to me, not sure if we should add the language tags now or fix them after this is merged

smattingly (approved):

PR #2052, opened 2022-06-07
Prevent server crash

smattingly:
The permissions check server would crash when the farm_id came through as the string literal `'undefined'`.
You can test the data screening and the query error handling by explicitly assigning that string to `farm_id` at checkScope.js line 20 and/or line 58.


AbdusShaikh (approved):
Server continues running when farm_id='undefined'

PR #2054, opened 2022-06-07
F21-575/spotlight twice bug fix

seinnlee:
Ticket [F21-575](https://lite-farm.atlassian.net/browse/F21-575)

Please look at the discussion.

The bug:
1. User creates a new account
2. They see the navigation bar spotlight
3. They see the notification spotlight

To Test:
1. Create a new account
2. Go through the navigation bar spotlight
3. There shouldn't be notification spotlight anymore


smattingly (changes requested):
I create a new account and a farm, and I see the spotlights again every time I log in.

seinnlee:
@smattingly it seems my way was only dispatching one action. I updated it so that multiple flags could be updated in one dispatch.

smattingly (approved):

PR #2057, opened 2022-06-09
Lf 2486 setup testing of html email

MwayaB:


Cali0707 (changes requested):
Could you remove the `package-lock.json` file? It looks like this was made by using `npm install` instead of `pnpm install`, and I believe we want to be using `pnpm` in the webapp.

packages/webapp/package.json
```diff
@@ -63,6 +63,7 @@
     "immer": "^9.0.12",
     "moment": "^2.29.1",
     "moment-range": "^4.0.2",
+    "nodemailer": "^6.7.5",
```
smattingly, line 66: Since this is for Cypress testing, nodemailer should be a devDependency.
Probably okay to just move the line to the other section, but please check results by completely deleting `webapp/node_modules` then doing `pnpm install`.

smattingly (changes requested):
dependency -> dev dependency; see comment

PR #2059, opened 2022-06-09
LF-2478/pseudo user conversion ui change

seinnlee:
Ticket [LF-2478](https://lite-farm.atlassian.net/browse/LF-2478)

**To Test:**
1. Create a pseudo user (Go to people tab of the farm and invite a user without an email)
2. Click the pseudo user and click "Convert this worker to a user with account"
3. You should see all the fields same as regular invitation UI
4. Fill out the form and invite the user
5. Check in the users table from the database that the information has been correctly updated

### DIFF
--- a/packages/api/src/controllers/userFarmController.js
+++ b/packages/api/src/controllers/userFarmController.js
@@ -497,7 +497,7 @@ const userFarmController = {
     };
   },
 
-  patchPseudoUserEmail() {
+  upgradePseudoUser() {
     return async (req, res) => {
       const { user_id, farm_id } = req.params;
       const { email, gender, birth_year, language, phone_number } = req.body;
--- a/packages/api/src/controllers/userFarmController.js
+++ b/packages/api/src/controllers/userFarmController.js
@@ -598,7 +598,16 @@ const userFarmController = {
       }
       try {
         const { farm_name } = userFarm;
-        await emailModel.createTokenSendEmail(userFarm, userFarm, farm_name);
+        await emailModel.createTokenSendEmail(
+          {
+            email,
+            gender,
+            birth_year,
+            language,
+          },
+          userFarm,
+          farm_name,
+        );
       } catch (e) {
         console.log(e);
       }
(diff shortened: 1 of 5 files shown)


packages/api/src/controllers/userFarmController.js
```diff
@@ -500,7 +500,7 @@ const userFarmController = {
   patchPseudoUserEmail() {
```
fsayyid, line 500: Refactor this name to reflect that not only the email is being patched. Maybe `upgradePseudoUser` or something.

packages/api/src/controllers/userFarmController.js
```diff
@@ -565,6 +565,10 @@ const userFarmController = {
               .patch({
                 email,
                 status_id: 2,
```
fsayyid, line 567: Can a pseudo user only converted to a worker? I think the frontend should reflect this if that's the case.

fsayyid: I think I understood wrong, I believe you can ignore the above comment.

fsayyid:
When I open the invite link, the gender and info fields are not autofilled in as they should be.

fsayyid (changes requested):
I think you should go through the invite flow after the pseudo user is upgraded, make sure all the info that the FO/FM/EO filled in before inviting the user, is filled in when they click the invite link.

fsayyid:
Also all the info filled in when the pseudo user is first created (phone number, gender, etc) should be filled in the Edit User form when "Convert this worker to a user with an account". And I think I am still unclear on whether a pseudo user can be converted to a FO/FM/EO.

seinnlee:
@fsayyid The invite link should now have the pre-existing information filled out and edit user will also fill out the form (except for "gender", which I haven't figured out why) Gender and Language use the same logic but for some reason Gender isn't filled out when "convert this worker to a user with an account". Maybe you can find the bug while reviewing. Also Kevin says that the pseudo user can be converted to FO/FM/EO.

fsayyid (approved):
This working well for me now. Good work 👍.

smattingly (approved):

PR #2060, opened 2022-06-09
LF-2462 - feat: created notification timeline

fsayyid:
Ticket: [LF-2462](https://lite-farm.atlassian.net/browse/LF-2462)

Created a notification timeline for viewing notifications related to the same entity (eg. a task).

To test:
- See [Figma](https://www.figma.com/file/F3f6DpGcxAaEsxfl7OZkI2/Litefarm?node-id=27547%3A88602&properties-panel-tab=code) for design expectations.
- Assign a task to yourself, then reassign/unassign it (we want to trigger two notifications);
- Go to the notification related to the above event, make sure that we see the notification timeline with the two notifications
- If you are viewing the "Assigned task" notification, make sure that the warning box shows up
- In the timeline, you should be able to click the different items to switch between related notifications

### DIFF
--- a/packages/webapp/public/locales/en/translation.json
+++ b/packages/webapp/public/locales/en/translation.json
@@ -1073,7 +1073,7 @@
   },
   "ENTITY_TYPES": {
     "TASK": "task",
-    "location": "location"
+    "LOCATION": "location"
   },
   "NOTIFICATION": {
     "TIMELINE": {
--- a/packages/webapp/public/locales/en/translation.json
+++ b/packages/webapp/public/locales/en/translation.json
@@ -1071,7 +1071,16 @@
       "YOU_WILL_FIND": "Here you will find:"
     }
   },
+  "ENTITY_TYPES": {
+    "TASK": "task",
+    "location": "location"
+  },
   "NOTIFICATION": {
+    "TIMELINE": {
+      "HEADING": "Notification timeline",
+      "VIEW_NOW": "View now",
+      "MORE_RECENT_NOTIFICATION": "There are more recent notifications about this {{entityType}}."
+    },
     "DAILY_TASKS_DUE_TODAY": {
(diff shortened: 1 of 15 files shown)


Cali0707 (approved):
Looks great!!

packages/webapp/src/components/Notifications/NotificationTimeline/index.jsx
```diff
@@ -0,0 +1,73 @@
...
+  return (
+    <div className={styles.container} style={style}>
+      <header className={styles.header}>
+        <h2 className={styles.heading}>{t('NOTIFICATION.TIMELINE.HEADING')}</h2>
+        <hr className={styles.headerLine} />
+      </header>
+      {mostRecentNotification && (
+        <MoreRecentNotificationWarning
+          notificationId={mostRecentNotification?.notification_id}
+          entityType={mostRecentNotification?.ref?.entity?.type}
```
smattingly, line 44: I think the first optional chaining on this and the previous line are redundant with the check on lint 41?

packages/webapp/public/locales/en/translation.json
```diff
@@ -1071,7 +1071,16 @@
       "YOU_WILL_FIND": "Here you will find:"
     }
   },
+  "ENTITY_TYPES": {
+    "TASK": "task",
+    "location": "location"
```
smattingly, line 1076: Is there a reason this key name is in lower case? It seems inconsistent, possibly incorrect (I know sometimes strings are converted to upper case for uses like these).

smattingly (changes requested):
I'm impressed! This looks really good.
In the comments I had one question and one possible nitpicky change.
No changes necessary if the `location` key capitalization is intentional.

smattingly (approved):

PR #2061, opened 2022-06-10
Unit Indicators are not dynamic

jagDhillon077:
[Jira Ticket:](https://lite-farm.atlassian.net/jira/software/c/projects/LF/boards/1?modal=detail&selectedIssue=LF-2491)
Create a crop plan and input values higher than the threshold that would convert the numbers to a smaller value. 
For example, input at least 100cm in plant spacing/depth etc in order for the values to be converted into meters. Do the same thing for estimated annual harvest (At least 1000kg to convert to mt). 

Once the crop plan has created open the details page of the plan and make sure the conversion occurred and the units are now in mt. Similarly check the plan spacing/depth etc in the planting task and make sure the cm have converted to meters.

### DIFF
--- a/packages/webapp/src/components/Form/Unit/index.jsx
+++ b/packages/webapp/src/components/Form/Unit/index.jsx
@@ -226,6 +226,8 @@ const Unit = ({
   }, [hookFormUnit]);
 
   useEffect(() => {
+    !hookFormGetValue(displayUnitName) &&
+      hookFormSetValue(displayUnitName, getUnitOptionMap()[displayUnit]);
     if (hookFormGetValue(displayUnitName)) {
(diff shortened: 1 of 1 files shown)


Cali0707 (changes requested):
@jagDhillon077 it looks like there is now no default unit displaying. I feel like we should be displaying some form of unit by default.
missing_units

jagDhillon077:
> @jagDhillon077 it looks like there is now no default unit displaying. I feel like we should be displaying some form of unit by default. missing_units

@Cali0707  Added back default values, thanks for catching that!

Cali0707 (approved):
This is working now 👍

kcussen:
Merging this so we can test on beta based on the fact that:
1. This is an urgent fix
2. Limited code changes
Please review when you're back @smattingly.

PR #2062, opened 2022-06-13
Feature/ensemble communications LF - 2453: Bulk sensor claim helper function

fsayyid:
Ticket: [LF-2453](https://lite-farm.atlassian.net/browse/LF-2453)

Bulk sensor claim helper function for the ensemble communications flow.

To test:
- Add a function call to `bulkSensorClaim` at the end of the file with valid arguments
- Make sure the response is appropriate (you can add a break point or use a console log to check, since we do not have the endpoint that would call this set up yet, so `res.send` will not be available).

### DIFF
--- a/packages/api/src/util/ensemble.js
+++ b/packages/api/src/util/ensemble.js
@@ -27,14 +27,15 @@ const { ensembleAPI } = require('../endPoints');
  */
 async function bulkSensorClaim(organizationId, esids, accessToken) {
   try {
-    return await axios.post(
+    const response = await axios.post(
       `${ensembleAPI}/organizations/${organizationId}/devices/bulkclaim/`,
       { esids },
       { headers: getHeaders(accessToken) },
     );
+    return { ...response.data, status: response.status };
   } catch (error) {
     if (error.response?.data && error.response?.status) {
-      return error.response;
+      return { ...error.response.data, status: error.response.status };
     } else {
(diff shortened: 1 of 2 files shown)


packages/api/src/util/ensemble.js
```diff
@@ -12,3 +12,39 @@
...
+ * @param {Response} res - The HTTP response object.
+ * @param {uuid} organizationId - a uuid for an Ensemble organization
+ * @param {Array} esids - an array of ids for Ensemble devices
+ * @param {String} accessToken - a JWT token for accessing the Ensemble API
+ * @returns {Object} - the response from the Ensemble API
+ * @async
+ */
+async function bulkSensorClaim(res, organizationId, esids, accessToken) {
+  try {
+    return await axios.post(
```
Cali0707, line 30: could you have this return an object containing the `success`, `does_not_exist`, `already_owned`, and `occupied` lists instead of just the API response? I feel like that would be easier to work with elsewhere in the codebase

Cali0707 (changes requested):
Looks good and seems to be working for me - just a small request to change the format of what the function returns

fsayyid:
Actually there's an error here, wait a moment before merging

fsayyid:
Okay it is ready for review

Cali0707 (approved):
Looks great!

PR #2063, opened 2022-06-13
Lf 2456 add sensor to points section

MwayaB:


fsayyid:
@MwayaB I think we can close this PR, I believe sensor is in the point section on the feature branch.

PR #2064, opened 2022-06-13
LF-2482/structures to support ensemble integration

seinnlee:
Ticket [LF-2482](https://lite-farm.atlassian.net/browse/LF-2482)

**integratingPartners**
- partner_id
- partner_name
- access_token
- refresh_token
- root_url
- deactivated

**farmExternalIntegrations**
- farm_id
- partner_id
- organization_uuid
- webhook_address

### DIFF
--- a/packages/api/src/models/farmExternalIntegrationsModel.js
+++ b/packages/api/src/models/farmExternalIntegrationsModel.js
@@ -1,0 +1,84 @@
+/*
+ *  Copyright 2019, 2020, 2021, 2022 LiteFarm.org
+ *  This file is part of LiteFarm.
+ *
+ *  LiteFarm is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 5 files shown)


packages/api/src/models/farmExternalIntegrationsModel.js
```diff
@@ -0,0 +1,84 @@
...
+   * Supports validating instances of this Model class.
+   * @static
+   * @returns {Object} A description of valid instances.
+   */
+  static get jsonSchema() {
+    return {
+      type: 'object',
+      properties: {
+        farm_id: { type: 'string' },
+        partner_id: { type: 'string' },
```
Cali0707, line 49: I believe this should be an integer, not a string

Cali0707 (changes requested):
One small change to make and then this should be good to go

Cali0707 (approved):
Looks great!

PR #2065, opened 2022-06-13
Lf 2451 register organization callout with ensemble api

Cali0707:
To test, add a function call to the bottom of `ensemble.js` and import `ensemble.js` into `server.js`


AbdusShaikh (approved):
Farm is being registered with the api correctly. Looks good.

PR #2066, opened 2022-06-13
[Lf 2458, LF-2459] : Feature - building bulk sensor upload modal

anant00729:
This PR is dependent on [LF-2460-bulk-create-sensors](https://github.com/LiteFarmOrg/LiteFarm/tree/LF-2460-bulk-create-sensors). Please review that one before starting with this.

Also, 
1. run pnpm i
2. review the files only related to webapp/frontend.

To Test:
1. It's good to check the details in the ticket. 
https://lite-farm.atlassian.net/browse/LF-2458
https://lite-farm.atlassian.net/browse/LF-2459


packages/api/db/migration/20220613141913_create_sensors_tables.js
```diff
@@ -0,0 +1,31 @@
+exports.up = function (knex) {
+  return Promise.all([
+    knex.schema.createTable('sensors', function (table) {
```
fsayyid, line 3: @AbdusShaikh I should we merge your PR first, since we talked about how we wanted the table names to be singuler?

packages/api/src/controllers/sensorController.js
```diff
@@ -0,0 +1,208 @@
...
+
+/**
+ * Parses the csv string into an array of objects and an array of any lines that experienced errors.
+ * @param {String} csvString
+ * @param {Object} mapping - a mapping from csv column headers to object keys, as well as the validators for the data in the columns
+ * @param {String} delimiter
+ * @returns {Object<data: Array<Object>, errors: Array<Object>>}
+ */
+
+const parseCsvString = (csvString, mapping, delimiter = ',') => {
```
fsayyid, line 176: Are we doing this validation on the backend because it may take time? I'm just interested in the rationale behind doing on backend vs the frontend.

packages/api/src/routes/sensorRoute.js
```diff
@@ -0,0 +1,33 @@
...
+const multer = require('multer');
+
+const SensorController = require('../controllers/sensorController');
+
+const storage = multer.memoryStorage();
+const upload = multer({ storage });
+
+const router = express.Router();
+
+router.get('/get_sensors', SensorController.getSensorsByFarmId());
```
fsayyid, line 26: I think at the moment we only need an add sensors endpoint

fsayyid: We can keep the functions in the controllers with some todo comments if we want, but I think we should get rid of routes so users can not access them.

packages/api/src/controllers/sensorController.js
fsayyid: Oh I see, this is just parsing the data, never mind

fsayyid: Wait no, I'm still confused. Are we validating the data twice?

packages/api/db/migration/20220613141913_create_sensors_tables.js
AbdusShaikh: I'd avoid merging the migrations at this point since changes are still being made here and there. My part is not yet ready so I cannot merge it first.

packages/webapp/src/components/Map/Modals/BulkSensorUploadModal/useValidateBulkSensorData.js
```diff
@@ -20,73 +21,72 @@ const requiredFields = [
...
-  const [disabled, setDisabled] = useState(true);
+export function useValidateBulkSensorData(onUpload, t) {
+  const bulkSensorsUploadResponse = useSelector(bulkSensorsUploadSliceSelector);
+  const [disabled, setDisabled] = useState(null);
   const [selectedFileName, setSelectedFileName] = useState('');
   const [sheetErrors, setSheetErrors] = useState([]);
   const [errorCount, setErrorCount] = useState(0);
   const fileInputRef = useRef(null);
 
+  const validationFields = [
```
fsayyid, line 34: Yeah, it feels like we are validating twice, unless I am confused

fsayyid (changes requested):
@anant00729 please review the comments

packages/webapp/src/containers/Map/index.jsx
```diff
@@ -488,13 +493,15 @@ export default function Map({ history }) {
...
-              setShowBulkSensorUploadModal(false);
-              handleClickAdd();
-            }}
+            dismissModal={dismissBulkSensorsUploadModal}
             onUpload={(file) => {
-              console.log('file', file);
-              // call saga function here
+              const payload = {
+                file,
+                dismissBulkSensorsUploadModal: () => {
```
fsayyid, line 500: Instead of this, can we use the loading state to figure out when to hide the modal?

fsayyid: To keep are redux logic separate from our components.

packages/api/src/controllers/sensorController.js
anant00729: Yes we are doing it on both backend and frontend

packages/webapp/src/containers/Map/index.jsx
anant00729: sure

packages/webapp/src/containers/Map/saga.js
```diff
@@ -74,7 +82,45 @@ export function* setSpotlightToShownSaga({ payload: spotlights }) {
...
+    formData.append('sensors', file);
+    const fileUploadResponse = yield call(axios.post, bulkUploadSensorsInfoUrl(), formData, {
+      headers: {
+        'Content-Type': 'multipart/form-data',
+        Authorization: 'Bearer ' + localStorage.getItem('id_token'),
+        farm_id: farm_id,
+      },
+    });
+
+    if (fileUploadResponse.status === 200) {
```
fsayyid, line 101: @Cali0707 I think you and @anant00729 should coordinate about how you want to handle errors. Since right now its showing an error on the frontend even when sensors are being added to the database.

Cali0707: @fsayyid I think this may have been resolved by the commit I just pushed - would you be able to test this and confirm?

fsayyid: I think for the most part this is fine now. It's not likely that users will be claiming sensors owned by others or non existing sensors. Here's what we can do for now:

@Cali0707 Send a 400 error in the case of occupied or non existing sensor.  
@anant00729 on the frontend if there are any successful sensors added, we can enqueue the success snackbar. If there is a 400 also enqueue as an error "Non existing or occupied sensors could not be registered". Or something like that, so users see both snackbars.

fsayyid (changes requested):
See my above comments

fsayyid (approved):
I am going to merge this PR, and let the error handling for the sensor does not exist and sensor occupied cases be handled in another PR.

PR #2071, opened 2022-06-14
Lf 2495 supplier displaying for fr pt and sp when editing an existing crop

MwayaB:


Cali0707 (changes requested):
@MwayaB it looks like some of the changes for your ensemble ticket are on this branch. I think we should either wait for the ensemble feature branch to be merged before merging this PR, or remove those changes from this PR before merging it

smattingly (approved):

smattingly (approved):

PR #2073, opened 2022-06-15
Lf 2484 structures to support sensors

AbdusShaikh:
To test:

1. Run migrations from api folder, you should see two new db tables `sensor` and `sensor_reading`
2. Make calls to the sensor endpoints and ensure the data is stored in the db correctly.

Here is some general information to help you make the api calls
- To add a reading, the sensor that reading belongs to must exist in the db. I suggest creating sensors manually and adding readings through the api since the `add_sensor` endpoint is part of a different ticket
- Here is an example POST that ensemble may send to the `add_reading` endpoint
`[
  {
    "node_esid": "123",
    "sensor_esid": "124",
    "parameter_number": "Temperature",
    "unit": "ºC",
    "time": [
      "02-09-2022, 22:53:52",
      "02-09-2022, 22:54:07"
    ],
    "value": [
      1,
      2
    ],
    "validated": [
      false,
      false
    ]
  },
  {
    "node_esid": "123",
    "sensor_esid": "123",
    "parameter_number": "Soil Moisture",
    "unit": "kPa",
    "time": [
      "02-09-2022, 22:54:23"
    ],
    "value": [
      3
    ],
    "validated": [
      false
    ]
  }
]`
- Use '1' as the partner ID when you add a reading as this is the only partner ID we have at the moment
- When sending a request, make sure you include the appropriate JSON body

### DIFF
--- a/packages/api/src/controllers/sensorController.js
+++ b/packages/api/src/controllers/sensorController.js
@@ -80,9 +80,7 @@ const sensorController = {
       const trx = await transaction.start(Model.knex());
       try {
         const infoBody = {
-          reading_id: req.body.reading_id,
           read_time: req.body.read_time,
-          //   transmit_time: req.body.transmit_time,
           sensor_id: req.body.sensor_id,
           reading_type: req.body.reading_type,
           value: req.body.value,
--- a/packages/api/src/controllers/sensorController.js
+++ b/packages/api/src/controllers/sensorController.js
@@ -79,15 +79,29 @@ const sensorController = {
     return async (req, res) => {
       const trx = await transaction.start(Model.knex());
       try {
-        const infoBody = {
-          read_time: req.body.read_time,
-          sensor_id: req.body.sensor_id,
-          reading_type: req.body.reading_type,
-          value: req.body.value,
-          unit: req.body.unit,
-        };
+        // const infoBody = {
+        //   read_time: req.body.read_time,
+        //   sensor_id: req.body.sensor_id,
+        //   reading_type: req.body.reading_type,
+        //   value: req.body.value,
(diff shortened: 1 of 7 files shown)


packages/api/src/controllers/sensorController.js
```diff
@@ -188,40 +187,45 @@ const sensorController = {
...
-        // const result = await baseController.postWithResponse(sensorReadingModel, infoBody, req, {
-        //   trx,
-        // });
-        // await trx.commit();
-        // res.status(200).send(result);
-        res.sendStatus(200);
+        const result = await baseController.postWithResponse(sensorReadingModel, infoBody, req, {
+          trx,
+        });
+        await trx.commit();
```
Cali0707, line 226: You'll need to declare `trx` somewhere earlier, because I am currently getting errors that it is undefined

Cali0707 (changes requested):

packages/api/src/util/ensemble.js
```diff
@@ -94,7 +94,7 @@ async function registerOrganizationWebhook(farmId, organizationId, accessToken)
   const onResponse = async (response) => {
     await FarmExternalIntegrationsModel.updateWebhookAddress(
       farmId,
-      `${baseUrl}/sensors/add_reading/`,
+      `${baseUrl}/sensors/add_reading/:partner_id`,
```
fsayyid, line 97: @AbdusShaikh I believe we need the actual partner id of 'Ensemble Scientific' here

fsayyid, line 97: @AbdusShaikh I believe we need the actual partner id of 'Ensemble Scientific' here

PR #2074, opened 2022-06-15
Patch/unit indicators are not dynamic

jagDhillon077:


anant00729:
This works for me.

anant00729 (approved):

smattingly (approved):
It's not clear to me that this is the intended behavior, but honestly I don't fully understand the intended behavior. I will approve and merge this so that others can evaluate from QA perspective.

PR #2078, opened 2022-06-17
Bug fix for map and crop variety detail page

AbdusShaikh:
To test:

1. If you don't already have one, create a crop plan on your farm
2. Invite a farm worker from your current farm. Make sure the invitee would be making a fresh account when they receive the invitation
3. Sign in from this new worker and navigate to the crop plan you made.
4. You should see the appropriate spotlight only once. Click the details tab and ensure you do not see the spotlight again. You can also re-navigate to this same page to ensure you do not see the spotlight again
5. Navigate to the map and ensure you only see the spotlight once. Try creating an area or re-navigating to the map to make sure of this as well.

### DIFF
--- a/packages/webapp/src/containers/Crop/CropVarietySpotlight/index.jsx
+++ b/packages/webapp/src/containers/Crop/CropVarietySpotlight/index.jsx
@@ -9,7 +9,7 @@ export default function CropVarietySpotlight({ children }) {
   const { t } = useTranslation();
   const dispatch = useDispatch();
   const { crop_variety_detail } = useSelector(showedSpotlightSelector);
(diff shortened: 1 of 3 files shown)


seinnlee:
@Cali0707 I think you were going to review this PR if I remember correctly. You can ignore my self review request and review this PR.

fsayyid (changes requested):
It looks like there a few other places in the app that have this same problem. 
Screen Shot 2022-06-20 at 12 36 01 PM

If you can either 
a) refactor all of them
b) change setSpotlightToShown to accept both strings and arrays
then we can have a global fix for this issue.

fsayyid (approved):
This is working for me 👍

PR #2080, opened 2022-06-17
Fix code, English for incorrect wording

smattingly:
Note that this intentionally inserts `MISSING` as a substring in a key for non-English languages where words are missing from the otherwise correct translation.


Cali0707 (approved):
This looks good to me. I received a notification with the proper text and the correct user name filled in.

PR #2081, opened 2022-06-17
LF-2478/pseudo user invite email language bug fix

seinnlee:
Ticket [LF-2478 comments](https://lite-farm.atlassian.net/browse/LF-2478?focusedCommentId=13522)

To Test:
1. Create a pseudo user without an email account.
2. Convert the user to an account with email.
3. Try different language of invitation.
4. The invitation email must be in the correct language and when you click the link, the app should be in the correct language.

### DIFF
--- a/packages/webapp/src/containers/Profile/People/saga.js
+++ b/packages/webapp/src/containers/Profile/People/saga.js
@@ -125,6 +125,7 @@ export function* invitePseudoUserSaga({ payload: user }) {
       user,
       header,
     );
+    localStorage.setItem('litefarm_lang', user.language);
     yield put(
       invitePseudoUserSuccess({
         newUserFarm: result.data,
--- a/packages/api/src/controllers/userFarmController.js
+++ b/packages/api/src/controllers/userFarmController.js
@@ -603,7 +603,7 @@ const userFarmController = {
             email,
             gender,
             birth_year,
-            language,
+            language_preference: language,
           },
           userFarm,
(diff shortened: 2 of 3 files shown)


fsayyid (approved):
This is working for me 👍

seinnlee:
@smattingly I just tried again in French and the app was in French for me. My steps were:
1. Create a pseudo user in English
2. Convert that user to a user with an account (invitation language to French)
3. The email was in French and the app was in French (my browser language is English)

smattingly:
@seinnlee Update: after describing the issue I saw in my review comment, I left that second browser open. Returning to it now, much later, the LF page is showing in French. I'm not sure what to make of that.

packages/webapp/src/containers/Profile/People/saga.js
```diff
@@ -125,6 +125,7 @@ export function* invitePseudoUserSaga({ payload: user }) {
       user,
       header,
     );
+    localStorage.setItem('litefarm_lang', user.language);
```
smattingly, line 128: @seinnlee Can you explain why we set the user's language here?

seinnlee: You're right. This shouldn't be here. This only work for local environment.

I made a new change and pushed it. Can you review again? (I'm using url param from email invite link to change the app language)

smattingly (approved):
Works very smoothly.

PR #2084, opened 2022-06-20
Created migration to update user name variable to match new notification strings

Cali0707:


smattingly (changes requested):
This looks good. However, I'm realizing there is a larger scope if we are going to fully address this. When this fixes the tag, the username will still be wrong. 
Could you research the complexity/effort of migration queries to: 
1. use the embedded ref to the task record to find out who the assigner is, and
2. update that information in the notification record as well?
I say "research" because that may not be worth the effort. 
Thanks!

smattingly (approved):

PR #2091, opened 2022-06-22
Bugfix - Farm worker no longer sees 'add to map' spotlight

AbdusShaikh:
Sensor controllers refactored avoid functions that always and only return the same function.

"integratingPartners" and "farmExternalIntegrations" migrations refactor

### DIFF
--- a/packages/api/src/models/integratingPartnersModel.js
+++ b/packages/api/src/models/integratingPartnersModel.js
@@ -22,7 +22,7 @@ class IntegratingPartners extends Model {
    * @returns {string} Names of the database table.
    */
   static get tableName() {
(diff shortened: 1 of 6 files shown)


packages/api/src/models/integratingPartnersModel.js
```diff
@@ -22,7 +22,7 @@ class IntegratingPartners extends Model {
    * @returns {string} Names of the database table.
    */
   static get tableName() {
-    return 'integratingPartners';
+    return 'integrating_partner';
```
fsayyid, line 25: We should do the same for the farm external integrations model

PR #2093, opened 2022-06-22
Feature/ensemble communications webhook auth

jagDhillon077:
Added sensor secret to environment and added it to webhook authorization header. Updated .env default

### DIFF
--- a/packages/api/.env.default
+++ b/packages/api/.env.default
@@ -47,6 +47,8 @@ REDIS_HOST=localhost
 REDIS_PORT=6379
 REDIS_PASSWORD=test
 
+SENSOR_SECRET=?
+
 REPORT_URL=http://localhost:3000/render_survey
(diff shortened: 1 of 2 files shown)


fsayyid (approved):
The webhook registration is still working on my end.

fsayyid (changes requested):
Actually, on further thinking, I am realizing that we need to handle this authorization at the add_reading endpoint. Because right now the header is being added, but we are not checking for it.

smattingly:
@jagDhillon077 @fsayyid I'm confused about the plans for this PR. I'm ready to merge it as is. Should I do that, or are more changes expected?

fsayyid:
I think that's up to how Jag wants to handle my above comment. @jagDhillon077 do you want to add those changes in this PR or another PR?

smattingly (approved):

PR #2096, opened 2022-06-27
Lf 2533 refresh store with last x days sensor readings when opening farm map

seinnlee:
Ticket [LF-2533](https://lite-farm.atlassian.net/browse/LF-2533)

7 last days sensor readings of a farm will be populated into the redux store when the farm map is opened.

To Test:
1. Create some dummy sensor reading data (you can manually create them into the db)
2. On `packages/webapp/src/containers/Map/index.jsx` you can console log `mapSensorReadings` from line 333 to see the data when the farm map is opened.

### DIFF
--- a/packages/api/src/controllers/sensorController.js
+++ b/packages/api/src/controllers/sensorController.js
@@ -237,19 +237,7 @@ const sensorController = {
       if (!farm_id) {
         return res.status(400).send('Invalid farm id');
       }
-      let result;
-      if (!days) {
-        result = await SensorReadingModel.query()
-          .joinRaw('JOIN sensor ON sensor_reading.sensor_id::uuid = sensor.sensor_id')
-          .where('farm_id', farm_id);
-      } else {
-        const pastDate = new Date();
-        pastDate.setDate(pastDate.getDate() - days);
-        result = await SensorReadingModel.query()
-          .joinRaw('JOIN sensor ON sensor_reading.sensor_id::uuid = sensor.sensor_id')
-          .where('farm_id', farm_id)
-          .andWhere('created_at', '>=', pastDate);
-      }
+      const result = await SensorReadingModel.getSensorReadingsByFarmId(farm_id, days);
       res.status(200).send(result);
     } catch (error) {
       res.status(400).send(error);
--- a/packages/api/src/controllers/sensorController.js
+++ b/packages/api/src/controllers/sensorController.js
@@ -237,7 +237,7 @@ const sensorController = {
       if (!farm_id) {
         return res.status(400).send('Invalid farm id');
       }
-      const result = await SensorReadingModel.getSensorReadingsByFarmId(farm_id, days);
+      const result = await SensorReadingModel.getSensorReadingsInDaysByFarmId(farm_id, days);
       res.status(200).send(result);
     } catch (error) {
       res.status(400).send(error);
--- a/packages/api/src/controllers/sensorController.js
+++ b/packages/api/src/controllers/sensorController.js
@@ -235,14 +235,13 @@ const sensorController = {
     try {
       const { farm_id, days } = req.params;
       if (!farm_id) {
-        res.status(400).send('Invalid farm id');
+        return res.status(400).send('Invalid farm id');
       }
       let result;
       if (!days) {
         result = await SensorReadingModel.query()
           .joinRaw('JOIN sensor ON sensor_reading.sensor_id::uuid = sensor.sensor_id')
-          .where('farm_id', farm_id)
-          .debug();
+          .where('farm_id', farm_id);
       } else {
         const pastDate = new Date();
         pastDate.setDate(pastDate.getDate() - days);
@@ -253,9 +252,7 @@ const sensorController = {
       }
       res.status(200).send(result);
     } catch (error) {
-      res.status(400).json({
-        error,
-      });
+      res.status(400).send(error);
     }
   },
 
--- a/packages/api/src/controllers/sensorController.js
+++ b/packages/api/src/controllers/sensorController.js
@@ -250,14 +250,20 @@ const sensorController = {
 
   async getReadingsByFarmId(req, res) {
     try {
-      const { farm_id } = req.params;
+      const { farm_id, days } = req.params;
       if (!farm_id) {
         res.status(400).send('Invalid farm id');
       }
+      if (!days) {
+        res.status(400).send('Invalid days');
+      }
+      const pastDate = new Date();
+      pastDate.setDate(pastDate.getDate() - days);
       const result = await sensorReadingModel
         .query()
         .join('sensor', 'sensor_reading.sensor_id', 'sensor.sensor_id')
-        .where('farm_id', farm_id);
+        .where('farm_id', farm_id)
+        .andWhere('created_at', '>=', pastDate);
       res.status(200).send(result);
     } catch (error) {
       res.status(400).json({
--- a/packages/api/src/controllers/sensorController.js
+++ b/packages/api/src/controllers/sensorController.js
@@ -258,14 +258,14 @@ const sensorController = {
       if (!days) {
         result = await sensorReadingModel
           .query()
-          .join('sensor', 'sensor_reading.sensor_id', 'sensor.sensor_id')
-          .where('farm_id', farm_id);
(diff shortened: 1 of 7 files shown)


packages/api/src/controllers/sensorController.js
```diff
@@ -231,6 +231,34 @@ const sensorController = {
...
+      const { farm_id, days } = req.params;
+      if (!farm_id) {
+        res.status(400).send('Invalid farm id');
+      }
+      let result;
+      if (!days) {
+        result = await SensorReadingModel.query()
+          .joinRaw('JOIN sensor ON sensor_reading.sensor_id::uuid = sensor.sensor_id')
+          .where('farm_id', farm_id)
+          .debug();
```
anant00729, line 245: Is this required? cannot find it anywhere else in the project

packages/webapp/src/containers/Map/mapSensorSlice.js
```diff
@@ -0,0 +1,68 @@
...
+    onLoadingSensorReadingStart: (state) => {
+      state.loading = true;
+    },
+    onLoadingSensorReadingFail: (state, { payload: error }) => {
+      state.loading = false;
+      state.error = error;
+      state.loaded = true;
+    },
+    getSensorReadingSuccess: (state, { payload: sensorReadings }) => {
+      addManySensorReadings(state, { payload: sensorReadings });
```
anant00729, line 45: can directly pass the object on line 44
getSensorReadingSuccess: (state, action) => {
addManySensorReadings(state, action); .....

packages/webapp/src/containers/Map/saga.js
```diff
@@ -162,10 +163,25 @@ export function* bulkUploadSensorsInfoFileSaga({ payload: { file } }) {
...
+
+export function* getSensorReadingsSaga() {
+  const { user_id, farm_id } = yield select(userFarmSelector);
+  const header = getHeader(user_id, farm_id);
+  try {
+    yield put(onLoadingSensorReadingStart(user_id, farm_id));
+    const result = yield call(axios.get, `${sensorUrl}/sensor_readings/${farm_id}/7`, header);
+    yield put(getSensorReadingSuccess(result.data));
+  } catch (e) {
+    console.error(e);
```
anant00729, line 176: can clear the state in case the API fails. It's better to clear the data so that If someone is using the same data in another component will not be showing wrong information using the selector function.

packages/webapp/src/containers/Map/saga.js
```diff
@@ -162,10 +163,25 @@ export function* bulkUploadSensorsInfoFileSaga({ payload: { file } }) {
...
 
+export const getSensorReadings = createAction('getSensorReadingsSaga');
+
+export function* getSensorReadingsSaga() {
+  const { user_id, farm_id } = yield select(userFarmSelector);
+  const header = getHeader(user_id, farm_id);
+  try {
+    yield put(onLoadingSensorReadingStart(user_id, farm_id));
+    const result = yield call(axios.get, `${sensorUrl}/sensor_readings/${farm_id}/7`, header);
+    yield put(getSensorReadingSuccess(result.data));
```
anant00729, line 174: can add 200, 500 and 400 status code conditions

packages/api/src/controllers/sensorController.js
```diff
@@ -231,6 +231,34 @@ const sensorController = {
...
+        const pastDate = new Date();
+        pastDate.setDate(pastDate.getDate() - days);
+        result = await SensorReadingModel.query()
+          .joinRaw('JOIN sensor ON sensor_reading.sensor_id::uuid = sensor.sensor_id')
+          .where('farm_id', farm_id)
+          .andWhere('created_at', '>=', pastDate);
+      }
+      res.status(200).send(result);
+    } catch (error) {
+      res.status(400).json({
```
anant00729, line 256: can make it consistent. either use .send or .json

packages/api/src/controllers/sensorController.js
```diff
@@ -231,6 +231,34 @@ const sensorController = {
...
+          .debug();
+      } else {
+        const pastDate = new Date();
+        pastDate.setDate(pastDate.getDate() - days);
+        result = await SensorReadingModel.query()
+          .joinRaw('JOIN sensor ON sensor_reading.sensor_id::uuid = sensor.sensor_id')
+          .where('farm_id', farm_id)
+          .andWhere('created_at', '>=', pastDate);
+      }
+      res.status(200).send(result);
```
anant00729, line 254: not sure if this should be send or json?

packages/api/src/controllers/sensorController.js
```diff
@@ -231,6 +231,34 @@ const sensorController = {
     }
   },
 
+  async getReadingsByFarmId(req, res) {
+    try {
+      const { farm_id, days } = req.params;
+      if (!farm_id) {
+        res.status(400).send('Invalid farm id');
```
anant00729, line 238: add return statement

anant00729 (changes requested):

packages/api/src/controllers/sensorController.js
seinnlee: sensor_id is uuid in sensor_reading table but it's string in sensor table so we need to type cast

packages/webapp/src/containers/Map/saga.js
seinnlee: Good idea! I only put the data in the store when the api returned correct data

anant00729:
Looks good to me. @smattingly Please review.

packages/api/src/routes/sensorRoute.js
```diff
@@ -34,6 +34,7 @@ router.post(
 router.delete('/delete_sensor/:sensor_id', SensorController.deleteSensor);
 router.post('/add_reading/:partner_id', SensorController.addReading);
 router.post('/get_readings', SensorController.getAllReadingsBySensorId);
+router.get('/sensor_readings/:farm_id/:days?', SensorController.getReadingsByFarmId);
```
smattingly, line 37: Have you tested this without a days value in the URL? I suspect you mean the ? to indicate that days is optional, but that's not how it usually works with Express. From the documentation, it's not clear how ? would combine with a route param like `:days`

packages/api/src/controllers/sensorController.js
```diff
@@ -231,6 +231,31 @@ const sensorController = {
...
 
+  async getReadingsByFarmId(req, res) {
+    try {
+      const { farm_id, days } = req.params;
+      if (!farm_id) {
+        return res.status(400).send('Invalid farm id');
+      }
+      let result;
+      if (!days) {
+        result = await SensorReadingModel.query()
```
smattingly, line 242: Please keep all knex and Objection API calls in the model classes. See [db-models](https://lite-farm.atlassian.net/wiki/spaces/LITEFARM/pages/1191575576/Pre-integration+peer+review#Backend-code)

smattingly (changes requested):
See line comments.

packages/api/src/routes/sensorRoute.js
seinnlee: When I tested with Postman, it worked with days as optional parameter. I added it just in case some other functions need to use the endpoint without a fixed amount of readings.

packages/api/src/controllers/sensorController.js
seinnlee: Moved them to model classes!

packages/api/src/routes/sensorRoute.js
smattingly: I was worried about that because I could not find it documented as meaning an optional parameter. The closest I can find is quoted below. Could you please test if `(?)` works and use that if it does? I'd feel safer.

> To have more control over the exact string that can be matched by a route parameter, you can append a regular expression in parentheses (()):
> 
> `Route path: /user/:userId(\d+)`
> `Request URL: http://localhost:3000/user/42`
> `req.params: {"userId": "42"}`

smattingly: Also, if you haven't, test with multidigit values for days. The question mark may only match one character.

seinnlee: The endpoint seems to work with multidigit values for days as well. We could remove `?` since currently the endpoint isn't used by other features to be safe.

smattingly (approved):

PR #2098, opened 2022-06-27
Fixed which fields display on a completed crop plan

Cali0707:
This PR resolves https://lite-farm.atlassian.net/browse/LF-2519

To test:
1. Create a crop plan
2. Ensure that the only fields you can see in the "detail" tab are the plan name above the tabs, as well as the Plan notes, and Estimated annual harvest
3. Complete the crop plan. You should now be able to see all the same fields before in the details page, as well as: Completed date, Plan rating, and Completion notes

### DIFF
--- a/packages/webapp/src/components/Crop/ManagementDetail/ManagementPlanDetail.jsx
+++ b/packages/webapp/src/components/Crop/ManagementDetail/ManagementPlanDetail.jsx
@@ -62,7 +62,6 @@ export default function PureManagementDetail({
     shouldUnregister: false,
     mode: 'onChange',
   });
-  console.log(plan);
 
   const isAbandoned = plan.abandon_date ? true : false;
   const isCompleted = plan.complete_date ? true : false;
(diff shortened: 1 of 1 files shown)


packages/webapp/src/components/Crop/ManagementDetail/ManagementPlanDetail.jsx
```diff
@@ -62,8 +62,10 @@ export default function PureManagementDetail({
     shouldUnregister: false,
     mode: 'onChange',
   });
+  console.log(plan);
```
seinnlee, line 65: I think you forgot to delete this?

seinnlee (approved):
LGTM! Just one console statement but pre approving

packages/webapp/src/components/Crop/ManagementDetail/ManagementPlanDetail.jsx
Cali0707: Good catch!

smattingly (approved):

PR #2099, opened 2022-06-27
Handling of async bulk upload on frontend.

anant00729:
To test: 

1. mock the API response of bulk upload as 202. (for me the add sensors is giving 500)
2. check after bulk upload, if you can see the following modal.

Screen Shot 2022-06-21 at 2 19 43 PM

for more details 
https://lite-farm.atlassian.net/browse/LF-2524

### DIFF
--- a/packages/webapp/src/containers/Map/saga.js
+++ b/packages/webapp/src/containers/Map/saga.js
@@ -113,50 +113,53 @@ export function* bulkUploadSensorsInfoFileSaga({ payload: { file } }) {
         Authorization: 'Bearer ' + localStorage.getItem('id_token'),
         farm_id: farm_id,
       },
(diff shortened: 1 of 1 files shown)


Cali0707 (changes requested):
@anant00729 this does not seem to be handling the partial success in any way - when I get a partial success the UI does not update at all. It does not show an error message, or a success message, and it also does not provide a text file download

Cali0707:
> This is only PR is handling only the async bulk sensor's upload i.e LF-2524.
> 
> I am working on the new request changes given in LF-2458. I LL be creating a separate PR for that.

If this is only handling the async modal, then that is working on my end.

Cali0707 (approved):

PR #2101, opened 2022-06-28
LF-2480 home page background displays the wrong season

fsayyid:
Ticket: [LF-2480)](https://lite-farm.atlassian.net/browse/LF-2480)

Switched from using moment js to javascript date object to fix season image issue. (Note, the month value in the javascript date object is from 0-11. So June is 5 not 6 as we would expect).

To test:
- create a farm in the northern hemisphere (ex. Canada) and farm is the southern hemisphere (ex. Australia)
- Make sure the image being shown in the north is summer (since it is currently June 28).
- Make sure the image being shown in the south is winter

### DIFF
--- a/packages/webapp/src/containers/Home/utils/season.js
+++ b/packages/webapp/src/containers/Home/utils/season.js
@@ -1,4 +1,3 @@
-import moment from 'moment';
 import { DO_CDN_URL } from '../../../util/constants';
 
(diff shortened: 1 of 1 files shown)


Cali0707 (approved):
This works for me!

smattingly:
@fsayyid Thanks. I was glad to see the move away from moment-- just curious about what the issue was since the change looked like a straight "translation".

PR #2102, opened 2022-06-28
Lf 2525 handle asynchronous add sensor outcome

Cali0707:
This PR handles the sync/async outcomes based off how long the flow is taking.

To test:
1. In `sensorController.js` on line 42, set the timeout to something small, like `50`
2. Comment out line 134 in `sensorController.js` as the webhook won't register properly until we are on beta/avocado
3. Try uploading a test csv file - you should see an error snackbar appear, this is because the 202 response code is not currently handled on the frontend in this branch - that is being handled in #2099 
4. Go to the notification that has been generated, it should have a "take me there" button which either a) downloads a text file if there were errors, or b) takes you to the map if it was successful

A useful testing file is: 
[testBulkUpload.csv](https://github.com/LiteFarmOrg/LiteFarm/files/9002688/testBulkUpload.csv)

### DIFF
--- a/packages/api/src/controllers/sensorController.js
+++ b/packages/api/src/controllers/sensorController.js
@@ -28,8 +28,6 @@ const {
 
 const sensorErrors = require('../util/sensorErrors');
 
-// TODO: make function which returns a sendReturn() function that depending on the timer status sends the appropriate response
-
 const sensorController = {
   async addSensors(req, res) {
     let hasTimedOut = false;
@@ -120,7 +118,7 @@ const sensorController = {
               user_id,
               farm_id,
               SensorNotificationTypes.SENSOR_BULK_UPLOAD_FAIL,
-              { error_download: { errors, file_name: 'sensor-upload-outcomes.txt' } }, //TODO: update to have proper file name
+              { error_download: { errors, file_name: 'sensor-upload-outcomes.txt' } },
             )
           : res
               .status(400)
--- a/packages/api/src/controllers/sensorController.js
+++ b/packages/api/src/controllers/sensorController.js
@@ -38,9 +38,9 @@ const sensorController = {
           'Processing your upload is taking longer than expected. We will send you a notification when this finished processing.',
       });
     }, 5000);
+    const { farm_id } = req.headers;
+    const { user_id } = req.user;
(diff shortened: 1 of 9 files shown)


fsayyid (approved):
From a backend perspective, this seems to be working well 👍

packages/api/src/controllers/sensorController.js
```diff
@@ -139,17 +179,49 @@ const sensorController = {
...
+            : res
+                .status(200)
+                .send({ message: 'Successfully uploaded!', sensors: sensorLocations }) &&
+                clearTimeout(timer);
         }
       }
     } catch (e) {
       console.log(e);
-      res.status(500).send(e.message);
+      const { user_id } = req.user;
```
smattingly, line 216: `user_id` and `farm_id` are already defined, just not in this scope. Move the definitions out of the `try` block and eliminate the repetition.

packages/api/src/controllers/sensorController.js
```diff
@@ -30,6 +30,14 @@ const sensorErrors = require('../util/sensorErrors');
 
 const sensorController = {
   async addSensors(req, res) {
+    let hasTimedOut = false;
```
smattingly, line 33: Technically this comment is for the line above the one marked:
See [simplify-routing](https://lite-farm.atlassian.net/wiki/spaces/LITEFARM/pages/1191575576/Pre-integration+peer+review#Backend-code): avoid functions that return constant functions.

smattingly (changes requested):
Requesting a couple of improvements.

packages/api/src/controllers/sensorController.js
Cali0707: Resolved

packages/api/src/controllers/sensorController.js
Cali0707: @smattingly I'm not sure what you are referring to here - as far as I can tell `addSensors` is not returning a constant function?

smattingly: Sorry, you're right. I guess I just entirely misread it.

smattingly (approved):

PR #2111, opened 2022-06-29
Fix validation condition to show errors for selecting invalid CSV file

anant00729:
To Test.

1. select an invalid file format and check the validation errors.

https://lite-farm.atlassian.net/browse/LF-2616

### DIFF
--- a/packages/api/src/controllers/sensorController.js
+++ b/packages/api/src/controllers/sensorController.js
@@ -112,6 +112,15 @@ const sensorController = {
           errorTranslationKey: sensorErrors.SENSOR_HARDWARE_VERSION,
         },
       });
+      if (!data.length > 0) {
+        return hasTimedOut
+          ? await sendSensorNotification(
+              user_id,
+              farm_id,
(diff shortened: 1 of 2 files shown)


Cali0707 (changes requested):
There is still a need to check if the file has any rows on the backend @anant00729 

If you want, I can push the necessary changes for that onto this branch. I do think that we should have that in place before we merge this however

anant00729:
Yes, you can add that validation on the backend.

Cali0707:
@anant00729 I've updated the backend code, but you will need to handle the new `error_type` on the frontend before this is ready to go. The new `error_type` is `empty_file`

anant00729:
I am doing all the validation that includes an empty file on the frontend before making the API call. I don't think there will be a situation where I need to handle that,

anant00729:
I have tested the scenario of empty file upload for the backend. It works for me. @smattingly Please review.

Cali0707 (approved):
Looks good

smattingly (approved):

PR #2112, opened 2022-06-30
Add try/catch

smattingly:
I believe this fixes a crash experienced with beta api. Operations using db transactions should be wrapped with `try`, and `catch` block should rollback transaction. Otherwise knex connection pool becomes exhausted.

### DIFF
--- a/packages/api/src/models/sensorModel.js
+++ b/packages/api/src/models/sensorModel.js
@@ -1,6 +1,6 @@
 /*
- *  Copyright (C) 2007 Free Software Foundation, Inc. <https://fsf.org/>
- *  This file (farmModel.js) is part of LiteFarm.
+ *  Copyright 2019, 2020, 2021, 2022 LiteFarm.org
+ *  This file is part of LiteFarm.
  *
  *  LiteFarm is free software: you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


fsayyid:
This seems to be working, but we are not handling the `SensorModel.createSensor()` error case correctly, it sends a 200 when it should be a 500. I think maybe Calum can look to fix this?

fsayyid (approved):

smattingly:
@Cali0707 Could you look at this with Faizah's comment in mind? Let me know if you think we should do this differently somehow. Thanks.

Cali0707:
@smattingly I believe the changes to address @fsayyid's comment would have to be done in the sensor controller. I have a few open PRs with changes to that file to fix various bugs that have come up, so my preference would be to merge this and those other PRs first and then deal with this issue afterwards. Does that make sense as an approach?

smattingly:
@Cali0707 Ok, then I will merge this. I'm trusting you to handle the change that Faizah describes-- either directly or by creating a Jira ticket if you think that's better. Thanks.

Cali0707:
@smattingly I've fixed this in #2118 - this should be good to merge now I think

PR #2113, opened 2022-06-30
Allow custom timer via localStorage/query string

smattingly:
This supports testing of asynchronous sensor upload mode without code changes.
1. The backend has been modified so that `sensorUploadTimer=N` in the query string will use `N` as the timer length.
2. The frontend has been modified so that a localStorage value for `sensorUploadTimer` will be added to the request query string.
(Note that branch name is misleading as I abandoned the original cookie-based idea.)


Cali0707 (approved):
This is working for me!

PR #2114, opened 2022-06-30
LF-2478 conversion of pseudo user to use the existing user's language preference

seinnlee:
This PR addresses the comment to this ticket [LF-2478](https://lite-farm.atlassian.net/browse/LF-2478)

To Test:
1. Create a user account (in a different language than English)
2. Log into a farm that you already have and create a pseudo user
3. Convert that pseudo user to a user with an account but invite the user from step 1 (select English or a language different from step 1 as the language of invitation)
4. The invite email and the app should be in the language of the existing user rather than the language from step 3

### DIFF
--- a/packages/api/src/controllers/userFarmController.js
+++ b/packages/api/src/controllers/userFarmController.js
@@ -425,13 +425,11 @@ const userFarmController = {
     return async (req, res) => {
       let result;
       const { user_id, farm_id } = req.user;
-      const { language_preference } = req.body;
       if (!/^\d+$/.test(user_id)) {
         const user = await userModel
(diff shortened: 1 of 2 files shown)


AbdusShaikh (approved):
Tried with several different languages and accounts and they all behaved correctly.

smattingly:
@seinnlee As I read them, your description of expected behavior above does not match description in the Jira comment. I'll raise that in discussion on the ticket.

smattingly (changes requested):
See earlier comment.

smattingly (approved):

PR #2116, opened 2022-07-04
Fixed error where bulk upload crashes

Cali0707:
This PR addresses the bulk upload failure that was occurring for one of the csv files we uploaded during last sprint review. The issue seems to have been to do with the delimiters we were splitting the file on not always being handled, as well as a trailing delimiter leading to an empty string in the array of rows.

To test: 
1. Comment out line 132 in `sensorController.js`
2. Upload [unclaimedSensorsForReview-2.csv](https://github.com/LiteFarmOrg/LiteFarm/files/9040393/unclaimedSensorsForReview-2.csv)


fsayyid (approved):
This seems to be working on my end.

smattingly (approved):

PR #2118, opened 2022-07-04
Bugfix/sensor upload problems

Cali0707:
This PR addresses various issues with the bulk upload:
1. You can now upload a sensor without an external id (so long as the brand is not "Ensemble Scientific")
2. If some sensors are claimed but not stored in the database on an initial upload for any reason, they will be stored on a subsequent upload now
3. Sensors that do not have a brand name of "Ensemble Scientific" are now saved properly

### DIFF
--- a/packages/api/src/controllers/sensorController.js
+++ b/packages/api/src/controllers/sensorController.js
@@ -352,36 +352,33 @@ const sensorController = {
 const parseCsvString = (csvString, mapping, delimiter = ',') => {
   // regex checks for delimiters that are not contained within quotation marks
   const regex = new RegExp(`(?!\\B"[^"]*)${delimiter}(?![^"]*"\\B)`);
-  const headers = csvString.substring(0, csvString.indexOf('\n')).split(regex);
+  const rows = csvString.split(/\r\n|\r|\n/).filter((elem) => elem !== '');
+  const headers = rows[0].split(regex);
   const allowedHeaders = Object.keys(mapping);
-  const { data, errors } = csvString
-    .substring(csvString.indexOf('\n') + 1)
-    .split('\n')
-    .reduce(
-      (previous, row, rowIndex) => {
-        const values = row.split(regex);
-        const parsedRow = headers.reduce((previousObj, current, index) => {
-          if (allowedHeaders.includes(current)) {
-            const val = mapping[current].parseFunction(
-              values[index].replace(/^(["'])(.*)\1$/, '$2'),
-            ); // removes any surrounding quotation marks
-            if (mapping[current].validator(val)) {
-              previousObj[mapping[current].key] = val;
-            } else {
-              previous.errors.push({
-                row: rowIndex + 2,
-                column: current,
-                translation_key: mapping[current].errorTranslationKey,
-              });
-            }
+  const dataRows = rows.slice(1);
+  const { data, errors } = dataRows.reduce(
+    (previous, row, rowIndex) => {
+      const values = row.split(regex);
+      const parsedRow = headers.reduce((previousObj, current, index) => {
+        if (allowedHeaders.includes(current)) {
+          const val = mapping[current].parseFunction(values[index].replace(/^(["'])(.*)\1$/, '$2')); // removes any surrounding quotation marks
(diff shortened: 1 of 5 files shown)


Cali0707:
@smattingly I ran into a problem while trying to rollback the migration here, as it was deleting a record that was being used as a foreign key in the `sensor` table. Should I be deleting any sensors in that table which have the `partner_id` of 0 (default partner which was added in this PR), or should I be setting the `partner_id` of those sensors to 1 (Ensemble) ?

smattingly:
@Cali0707 I think you probably want to delete the sensors. I don't see a scenario where we would ever rollback a migration in production, so I think of the `down` functions as being for dev use. And I never really run them, preferring to nuke the db and rebuild clean.

fsayyid:
When I upload [sensors060722.csv](https://github.com/LiteFarmOrg/LiteFarm/files/9057209/sensors060722.csv)

I get the following in the outcomes txt: 
Screen Shot 2022-07-06 at 2 26 05 PM

Cali0707:
@fsayyid yeah, that bug got introduced in #2111 there is a fix for it in #2119

packages/api/src/controllers/sensorController.js
```diff
@@ -174,7 +176,12 @@ const sensorController = {
         // Save sensors in database
         const sensorLocations = await Promise.all(
           registeredSensors.map(async (sensor) => {
-            return await SensorModel.createSensor(sensor, farm_id, user_id);
+            return await SensorModel.createSensor(
```
fsayyid, line 179: If this fails, but all sensors are registered with ensemble, we are still sending a 200 response

fsayyid:
@Cali0707 I'm not sure how to test the first and third issue without https://github.com/LiteFarmOrg/LiteFarm/pull/2119 being merged

Cali0707:
@fsayyid that makes sense, we can hold off here as we wait for reviews and merge on #2119

Cali0707:
@fsayyid I believe I have now fixed the issue you reported about the response codes

packages/api/src/controllers/sensorController.js
```diff
@@ -172,13 +174,32 @@ const sensorController = {
...
           }),
         );
 
-        if (success.length + already_owned.length < esids.length) {
+        const successSensors = sensorLocations.reduce((prev, curr) => {
+          if (curr.status === 'fulfilled') {
+            prev.push(curr.value.sensor.external_id);
+          } else {
+            errorSensors.push({
+              row: data.findIndex((elem) => elem === curr.value.sensor.external_id) + 2,
```
fsayyid, line 193: `curr.value.sensor` may be undefined in this case because a rejected promise will return an error instead of the sensor. maybe you can do this by index instead since `registeredSensors` and `sensorLocations` should be the same length.

packages/api/src/util/sensorErrors.js
```diff
@@ -27,4 +27,5 @@ module.exports = {
   SENSOR_HARDWARE_VERSION: 'FARM_MAP.BULK_UPLOAD_SENSORS.VALIDATION.SENSOR_HARDWARE_VERSION',
   SENSOR_ALREADY_OCCUPIED: 'FARM_MAP.BULK_UPLOAD_SENSORS.SENSOR_CLAIM_ERROR.ALREADY_OCCUPIED',
   SENSOR_DOES_NOT_EXIST: 'FARM_MAP.BULK_UPLOAD_SENSORS.SENSOR_CLAIM_ERROR.DOES_NOT_EXIST',
+  INTERNAL_ERROR: 'FARM_MAP.BULK_UPLOAD_SENSORS.SENSOR_CLAIM_ERROR.INTERNAL_ERROR',
```
fsayyid, line 30: We will need this key in the frontend translations.json.

fsayyid: But maybe we can have a separate PR to handle that

packages/api/src/controllers/sensorController.js
Cali0707: Good catch, and great idea for the solution. Thanks!

smattingly (approved):

PR #2119, opened 2022-07-05
Got the sync/async response to work, fixed a bug on sync partial success

Cali0707:
This PR refactors the sync/async logic for the bulk sensor upload into a separate file to make it a more usable pattern going forwards(as described in https://lite-farm.atlassian.net/browse/LF-2610). It also adds a small bugfix to the frontend for the synchronous bulk sensor upload partial success flow as I discovered the bug while testing my code. As such, this PR also resolves https://lite-farm.atlassian.net/browse/LF-2630

To test:
1. Add the optional second argument to the `syncAsyncResponse` function on line 34 of `sensorController.js` to make the timer something like 50 milliseconds, instead of 5000
2. Comment out line 144 (registering the webhook) or set up an ngrok tunnel and set the appropriate webhook callback address
3. Test that the asynchronous flow is working for the success, partial success, and failure outcomes
4. Remove the optional second argument from the `syncAsyncResponse` function, so that it defaults back to 5000 milliseconds
5. Test that the synchronous flow is working for the success, partial success, and failure outcomes.

Note that this branch does not have the fix from #2116 , so if your csv file uses a newline character that is not `\n`, it will fail when you try and upload it

### DIFF
--- a/packages/api/src/controllers/sensorController.js
+++ b/packages/api/src/controllers/sensorController.js
@@ -27,17 +27,11 @@ const {
 } = require('../util/ensemble');
 
 const sensorErrors = require('../util/sensorErrors');
+const syncAsyncResponse = require('../util/syncAsyncResponse');
 
 const sensorController = {
   async addSensors(req, res) {
-    let hasTimedOut = false;
-    const timer = setTimeout(() => {
-      hasTimedOut = true;
-      res.status(202).send({
-        message:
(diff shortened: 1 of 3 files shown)


anant00729 (approved):
The API works fine on my end. @smattingly Please check.

Cali0707:
@smattingly I've fixed the merge conflicts. For the translations I believe Faizah and Sein already did those in a previous PR

smattingly (approved):

PR #2122, opened 2022-07-06
Fixed user role bug

Cali0707:
This PR addresses https://lite-farm.atlassian.net/browse/LF-2554 

To test:
1. Invite a user to your farm
2. Edit the user and change their role
3. Their role should change, and their new role should be visible


AbdusShaikh (approved):
Looks good

PR #2123, opened 2022-07-06
Added sensor task tab

Cali0707:
This resolves https://lite-farm.atlassian.net/browse/LF-2505

To test:
1. Get the location id of a sensor on your farm:
```sql 
SELECT l.location_id FROM sensor s INNER JOIN location l ON s.location_id = l.location_id WHERE l.location_id = '{your_farm_id}'
```
2. Using one of the location ids from the above query, go to `localhost:3000/sensor/{location_id}/tasks`
3. You should see the tasks page


AbdusShaikh (approved):
Looks good. Tasks are appearing and behaving correctly for me

PR #2124, opened 2022-07-06
Webhook verification

jagDhillon077:
Only thing to test here is to upload a sensor and make sure that there is no webhook error in the API server. 
This code only validates the request coming from ensemble so if the sensor successfully uploads it means it has passed verification
Make sure to run a ngrok tunnel when doing this otherwise the webhook will fail


packages/api/src/routes/sensorRoute.js
```diff
@@ -32,9 +34,20 @@ router.post(
   SensorController.addSensors,
 );
 router.delete('/delete_sensor/:sensor_id', SensorController.deleteSensor);
-router.post('/add_reading/:partner_id', SensorController.addReading);
+router.post('/add_reading/:partner_id/:farm_id', validateRequest, SensorController.addReading);
 router.post('/get_readings', SensorController.getAllReadingsBySensorId);
 router.get('/sensor_readings/:farm_id/:days', SensorController.getReadingsByFarmId);
 router.post('/invalidate_readings', SensorController.invalidateReadings);
 
+function validateRequest(req, res, next) {
```
fsayyid, line 42: I think to be consistent this function should be in its own file in the middleware folder.

packages/api/src/routes/sensorRoute.js
```diff
@@ -32,9 +34,20 @@ router.post(
...
+  const farmId = req.params.farm_id;
```
fsayyid, line 43: We also want to check that the partner id is 1, so that we are only apply this authorization to requests for Ensemble readings.

packages/api/src/routes/sensorRoute.js
```diff
@@ -32,9 +34,20 @@ router.post(
...
+  const authKey = `${farmId}${process.env.SENSOR_SECRET}`;
+  if (req.headers.authorization === authKey) {
+    next();
+  } else {
+    console.error('forbidden');
+    res.status(403);
```
fsayyid, line 49: There is a bug here, I believe we want this to be `res.status(403).send('Not authorized to add readings')`. Otherwise the request will never send a response.

packages/api/src/middleware/validation/validateWebhook.js
```diff
@@ -0,0 +1,17 @@
+async function validateRequest(req, res, next) {
+  const partnerId = req.params.partner_id;
+  if (partnerId !== 1) {
+    next();
+  } else {
+    const farmId = req.params.farm_id;
```
fsayyid, line 6: I think you removed the farm_id from the params, or maybe this happened when you merged collab into this branch

packages/api/src/routes/sensorRoute.js
```diff
@@ -34,7 +38,7 @@ router.post(
 
 router.delete('/delete_sensor/:sensor_id', SensorController.deleteSensor);
 router.patch('/:sensor_esid', SensorController.updateSensorbyID);
-router.post('/add_reading/:partner_id', SensorController.addReading);
+router.post('/add_reading/:partner_id', validateRequest, SensorController.addReading);
```
fsayyid, line 41: I think we had /:farm_id here before

fsayyid (approved):

packages/api/src/middleware/validation/validateWebhook.js
```diff
@@ -0,0 +1,17 @@
...
+  const partnerId = parseInt(req.params.partner_id);
+  if (partnerId !== 1) {
+    next();
+  } else {
+    const farmId = req.params.farm_id;
+    const authKey = `${farmId}${process.env.SENSOR_SECRET}`;
+    if (req.headers.authorization === authKey) {
+      next();
+    } else {
+      console.error('forbidden');
```
fsayyid, line 11: Maybe get rid of this console log though

packages/api/src/routes/sensorRoute.js
```diff
@@ -16,6 +16,10 @@
 const express = require('express');
 const multer = require('multer');
 const checkScope = require('../middleware/acl/checkScope');
+const validateRequest = require('../middleware/validation/validateWebhook');
+
+const path = require('path');
+require('dotenv').config({ path: path.resolve(__dirname, '..', '..', '..', '.env') });
```
smattingly, line 22: Why is this needed?

smattingly (approved):

PR #2125, opened 2022-07-06
LF-2614: api crashes when user ignores invite and continues with google

fsayyid:
Ticket: [LF-2614](https://lite-farm.atlassian.net/browse/LF-2614)

To test:
- Invite a user whose email is @gmail.com or backed by Google, such as a litefarm org user 
- Logout, or use a different browser to …
- Visit the app homepage. Click “Continue with Google”. Log in using the invited user’s email.
- You should see the inline error: `We've updated our infrastructure and you'll need to check your inbox for a farm invitation to proceed.`


Cali0707 (approved):
This is working for me

smattingly (approved):

PR #2128, opened 2022-07-07
Fixed bug on map

Cali0707:
This PR resolves https://lite-farm.atlassian.net/browse/LF-2576

To test:
1. Go to the farm map
2. Try to create a watercourse or a buffer zone with no length (i.e. just a point)
3. You should get a popup prompting you to redraw the line


AbdusShaikh (approved):
Looks good to me.

smattingly (approved):

PR #2129, opened 2022-07-08
LF-2623 tasks not showing up

Cali0707:
This PR fixes a class of bugs that were occurring on the tasks page (documented here: https://lite-farm.atlassian.net/browse/LF-2623). Whenever a user created or updated a task that referenced a resource (assignee, task type, location) that was created after the filter on the task page was initialized, the task would disappear. 

To test:
1. Go to the tasks tab
2. Create a new user
3. Go back to the tasks tab and assign a task to the new user
4. The task should still be there
5. Create a new task type
6. Create a task of the new task type
7. That task should show up
8. Create a new location
9. Create a new task with that location set as its location
10. That task should show up

### DIFF
--- a/packages/webapp/src/containers/filterSlice.js
+++ b/packages/webapp/src/containers/filterSlice.js
@@ -212,7 +212,6 @@ export const documentsFilterSelector = createSelector(
   (filterReducer) => filterReducer.documents,
 );
 export const tasksFilterSelector = createSelector([filterReducerSelector], (filterReducer) => {
-  console.log(filterReducer);
   return filterReducer.tasks;
 });
 export const cropCatalogueFilterDateSelector = createSelector(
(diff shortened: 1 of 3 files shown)


packages/webapp/src/containers/filterSlice.js
```diff
@@ -175,10 +211,10 @@ export const documentsFilterSelector = createSelector(
   [filterReducerSelector],
   (filterReducer) => filterReducer.documents,
 );
-export const tasksFilterSelector = createSelector(
-  [filterReducerSelector],
-  (filterReducer) => filterReducer.tasks,
-);
+export const tasksFilterSelector = createSelector([filterReducerSelector], (filterReducer) => {
+  console.log(filterReducer);
```
fsayyid, line 215: remove this console log

fsayyid:
If I create a task that is unassigned it doesn't appear until I mess with the filters, even though it is initially set to show all tasks.

Cali0707:
> If I create a task that is unassigned it doesn't appear until I mess with the filters, even though it is initially set to show all tasks.

@fsayyid good catch - I forgot to set the unassigned option in one place, I've updated this PR to fix that

fsayyid (approved):
Seems to be working now 👍

smattingly (approved):

PR #2130, opened 2022-07-08
LF 2502 - Temperature reading component

anant00729:
To Test:

1. Add data sensor readings from these two CSV files. make changes in the sensor_id column of sensor_readings to map your sensor_ids present in the sensor table. 

[sensor.csv](https://github.com/LiteFarmOrg/LiteFarm/files/9074325/sensor.csv)
[sensor_reading.csv](https://github.com/LiteFarmOrg/LiteFarm/files/9074326/sensor_reading.csv)

2. Navigate to http://localhost:3000/sensor/<location_id_of_sensor>/readings
3. check if you can find this page.

Screen Shot 2022-07-08 at 10 42 15 AM

4. For the Requirements to test. please check the description under.
https://lite-farm.atlassian.net/browse/LF-2502


Cali0707 (changes requested):
@anant00729 whenever I try to load in the page when I have a sensor but no readings yet, the empty graph will load for about one second and then I am redirected to the LiteFarm login page

anant00729:
@Cali0707 I have added the sensor reading CSV file. you can load the CSV file into the sensor readings table.

Cali0707:
@anant00729 I know that. However, this component should not be logging someone out of LiteFarm if there is no data for that sensor in the table, which is the behaviour I am currently seeing

anant00729:
@Cali0707 thanks for letting me know. I ll look into this.

Cali0707 (changes requested):
@anant00729 it looks like the wrong set of dates is being displayed - I am getting a graph for Wed June 22 to Sat June 25. I would expect that this should show the previous 3 days, which would be Fri July 8 to Mon July 11

anant00729:
Handled NO Data scenario for sensor readings.

Screen Shot 2022-07-11 at 9 50 09 PM

As per,

Screen Shot 2022-07-11 at 9 52 52 PM

anant00729:
Also, handled partial data availability of sensor readings.

Screen Shot 2022-07-11 at 9 51 51 PM

Cali0707 (approved):
Besides using hard-coded dates, this seems to be working. I'd be fine with merging this into integration so long as we document that we need to update the code to not use hard-coded dates.

@smattingly what do you think?

smattingly:
@Cali0707 I agree. With @anant00729 and maybe others, let's make sure the work that's blocked is appropriately captured by the end of today's standup.

smattingly (approved):

PR #2131, opened 2022-07-08
Changed default error behaviour to be more consistent with other errors, translated error files

Cali0707:
This resolves https://lite-farm.atlassian.net/browse/LF-2645 and also translates all the error downloads

To test:
1. Without setting up an ngrok tunnel or commenting out the code to register a webhook with ensemble, upload sensors
2. You should get an error file instead of a snackbar
3. Set up an ngrok tunnel or comment out the code to register a webhook with ensemble
4. Upload sensors
5. See if the error download is translated (should be full of "MISSING" in other languages)


fsayyid (approved):
This seems to be working 👍

smattingly (approved):

PR #2132, opened 2022-07-11
LF 2511 preview sensors on farm map

seinnlee:
Ticket [LF 2511](https://lite-farm.atlassian.net/browse/LF-2511)

To Test:
1. Have sensors uploaded to your farm map (at least two in similar locations so they can be clustered and at least one on its own).
2. On desktop, long press a sensor icon on the map. Preview component should be populated (the child components currently don't exist so it will just say "preview body"). 
3. For sensor clusters, click them to see the list of sensors. Right click one of the listed sensors to see the preview (haven't been able to make it long press. this PR will be updated once the solution is found)
4. On mobile, you can long press both individual sensor locations or cluster sensors to see the preview.


Cali0707 (approved):
This works on my end, minus the issues described in the PR body

@seinnlee is the reason the long press won't work for the sensor clusters that you are already listening for that elsewhere?

smattingly (approved):

PR #2135, opened 2022-07-11
Farm worker no longer sees add item to map spotlight

AbdusShaikh:
To test:
1. Invite a new farm worker with an email
2. Log into Litefarm with the email you used to invite the farm worker
3. Navigate to the farm map. You should only see the "filter map" and "export map" spotlights. You should not see the "add to map" spotlight


Cali0707 (approved):
This works for me

smattingly (approved):

PR #2136, opened 2022-07-11
Lf 2577 shifting issue for watercourse and buffer zone

AbdusShaikh:
To test:
1. Navigate to your farm map
2. Begin the process to add a watercourse
3. Play around with the values in the input fields. If you surpass the maximum acceptable value, you should see a warning. Any values greater than the acceptable max should default to the maximum (not in the input field but visually) and not be accepted.
4. Repeat steps 2-3 for a buffer zone

**NOTE:** If you have any, please make suggestions on how to better deal with the constant defined on line 61 of packages/webapp/src/components/Map/LineMapBoxes/index.jsx.

### DIFF
--- a/packages/webapp/src/components/Map/LineMapBoxes/index.jsx
+++ b/packages/webapp/src/components/Map/LineMapBoxes/index.jsx
@@ -13,6 +13,7 @@ import { line_width } from '../../../util/convert-units/unit';
 import { cloneObject } from '../../../util';
 import { useSelector } from 'react-redux';
 import { measurementSelector } from '../../../containers/userFarmSlice';
+import { useMemo } from 'react';
 
 export default function PureLineBox({
   typeOfLine,
@@ -41,8 +42,8 @@ export default function PureLineBox({
   });
 
   const { t } = useTranslation();
-  const measurmentUnit = useSelector(measurementSelector);
-  const maxValue = measurmentUnit == 'metric' ? 1600 : 5280;
+  const measurementUnit = useSelector(measurementSelector);
+  const maxValue = useMemo(() => (measurementUnit === 'metric' ? 1600 : 5280), [measurementUnit]);
 
   const widthLabel =
     typeOfLine === locationEnum.watercourse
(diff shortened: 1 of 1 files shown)


packages/webapp/src/components/Map/LineMapBoxes/index.jsx
```diff
@@ -39,6 +41,8 @@ export default function PureLineBox({
   });
 
   const { t } = useTranslation();
+  const measurmentUnit = useSelector(measurementSelector);
+  const maxValue = measurmentUnit == 'metric' ? 1600 : 5280;
```
fsayyid, line 45: This should be `===`. Also, I would put this in a `useMemo` like `const maxValue = useMemo(() => measurmentUnit === 'metric' ? 1600 : 5280, [measurmentUnit]);`. Otherwise this value will be recalculated everytime any state changes.

packages/webapp/src/components/Map/LineMapBoxes/index.jsx
```diff
@@ -58,7 +62,10 @@ export default function PureLineBox({
 
   useEffect(() => {
     trigger();
-    updateWidth((widthValue || 0) + (bufferWidthValue || 0));
+    updateWidth(
```
fsayyid, line 65: On my end, the max value seems to be working without this code. I believe this can be reverted to `updateWidth((widthValue || 0) + (bufferWidthValue || 0));`

fsayyid: Though I guess this code would prevent large values from actually showing on google maps.

fsayyid: Nevermind, this looks good to me.

packages/webapp/src/components/Map/LineMapBoxes/index.jsx
fsayyid: Also measurement is spelled wrong here, missing an 'e'

fsayyid (approved):

smattingly (approved):

PR #2139, opened 2022-07-12
LF-2528 sensor map preview compact temperature component

seinnlee:
Ticket [LF-2528](https://lite-farm.atlassian.net/browse/LF-2528)

**This PR also resolves the issue from #2132  where sensor cluster didn't have long press**

To Test:
1. Ideal for #2132 to be merged first before reviewing this PR.
2. Have sensor temperature reading data (could be dummy data in the database as long as they're mapped correctly to a sensor in your farm map). The creation date must be within the last 24 hrs (the map preview will only show the most recent 24hr data)
3. Long press a sensor location on the map to see the preview of the sensor data.
4. Try long press a sensor from sensor cluster on the map to see the preview of the sensor data.
5. For the same sensor reading type (i.e. temperature), you should only see the most recent reading.


Cali0707:
@seinnlee do you want to merge this branch into LF-2511 (which is now in integration), or would it be better to switch the base branch to integration for this PR?

seinnlee:
@Cali0707 I changed the base branch to integration

seinnlee:
@Cali0707 The reading data has to be from within the last 24 hours. I updated the PR description (forgot to mention it).

Cali0707:
@seinnlee from the requirements it looks like it should still display the component, but it would say that the value in unknown:
"If the most recent reading is more than 6 hours old, display in error red “Unknown”"

Cali0707 (approved):
Looks good!

packages/webapp/src/containers/Map/LocationSelectionModal/index.jsx
```diff
@@ -5,16 +5,24 @@ import React from 'react';
...
 import PurePreviewPopup from '../../../components/Map/PreviewPopup';
+import { SENSOR } from '../../SensorReadings/constants';
 
-export default function LocationSelectionModal({ history, selectingOnly }) {
+export default function LocationSelectionModal({ history, sensorReadings, selectingOnly }) {
   const { dismissSelectionModal } = useSelectionHandler();
   const showSelection = useSelector(canShowSelectionSelector);
   const locations = useSelector(mapLocationsSelector);
-  if (showSelection && locations.length === 1 && locations[0].type === 'sensor') {
+  const locationSensorReadings = Object.values(sensorReadings).filter(
```
fsayyid, line 14: You might want to wrap this in a `useMemo`. Also why are we only considering the first location in the list?

packages/webapp/src/containers/Map/LocationSelectionModal/index.jsx
```diff
@@ -5,16 +5,24 @@ import React from 'react';
...
+    (reading) => reading.location_id === locations[0]?.id,
+  );
+  if (showSelection && locations.length === 1 && locations[0].type === SENSOR) {
     return (
       <div className={styles.selectionModal} onClick={dismissSelectionModal}>
         <div className={styles.selectionContainer}>
-          <PurePreviewPopup location={locations[0]} history={history} />
+          <PurePreviewPopup
```
fsayyid, line 21: I'm not sure what the practice for this would be, but would it make more sense to create a container PurePreviewPopup, create a selector for getting readings by location_id from the redux store, and have the container retrieve the data from the store rather than passing the sensor readings from component to component like this. I feel that would be cleaner, but I'll leave it up to you.

fsayyid (changes requested):
Please see the comments I made above. In addition:
- I think we can add the degree symbol for the temperature by using the HTML symbols for degrees celcius and degrees farenheight. You can try this and try changing the locale of chrome to see if it still appears correctly. 
- When my farm is using the imperial system, the readings still show in degrees celcius rather than farenheight. I'm not sure, I would check with Kevin on this but I believe ideally the readings will all be saved in celcius and showed to the user based on their preference (metric or imperial)

fsayyid:
Actually, I think we already have a way of getting the degrees symbol, see `packages/webapp/src/containers/WeatherBoard/utils/index.jsx`

packages/webapp/src/containers/Map/LocationSelectionModal/index.jsx
seinnlee: The array only has one location unless it's a cluster, in which case, we use all the locations. So for previewing one sensor, the locations will have one item (it would make sense more if it wasn't an array but we need it to be an array for the clusters).

packages/webapp/src/containers/Map/LocationSelectionModal/index.jsx
seinnlee: I updated so that LocationSelectionModal will retrieve readings by location_id and then pass them into the popup components.

seinnlee:
@fsayyid I asked Kevin about the temperature units (might get a response on Monday). I think we can update that in a separate PR when we have actual reading data. (it seems like Ensemble will send unit as "celsius" rather than "metric") Also, not sure how we will update the temperature once a user updates their preferred unit of the farm. For now, they use the database data.

fsayyid (changes requested):
Priority: 
- When viewing the temperature preview from a cluster, it always shows data for the first location in the list. 
- When I click on the compact temperature component it navigates to the details page, but I believe according to the ticket description it should be the readings page. 

Other comments (you can handle these in another PR if you want):
- The ticket also mentions that when you click on the compact temperature component it should highlight grey for 0.25 seconds before navigating to the readings page
- And when it navigates to the details page it should be scrolled to the temperature readings section

seinnlee:
@fsayyid I don't quite understand your first comment. When a user wants to select a location from a cluster, they will have to click the cluster first to see what locations are clustered. After clicking the cluster and seeing the list of locations, then the user can long-press which location they want to see the preview on. It doesn't make sense for user to long-press a cluster since they can be a mixture of different locations (sensors, gate, etc.). I addressed your second and third comment.

The last comment, I will make a separate PR. I hope to get this PR in today since it's blocking a couple of my other tickets.

fsayyid (approved):

PR #2141, opened 2022-07-12
User is returned to map after creating buffer zone

AbdusShaikh:
To test:
1. Begin creating a buffer zone
2. Enter a very large value for the width, enough that it start shifting.
3. Click confirm, then enter random information in all fields of the form. The form doesn't let you continue when the width value is too big so adjust for that
4. Click the '<' button, click confirm again, then proceed to saving the buffer zone while entering all the info

You should be returned to the map with your new buffer zone


Cali0707 (approved):
The change makes sense, and it is working on my end.

smattingly (approved):

PR #2143, opened 2022-07-12
Removal of part number and brand from the bulk sensor template

anant00729:
To test: 

1. Open bulk upload sensor modal.
2. click on "Download this template"
3. check if part number and brand are not present as the columns in the downloaded CSV file.

Screen Shot 2022-07-12 at 12 13 10 PM


Cali0707 (approved):
Works on my end.

PR #2145, opened 2022-07-12
LF-2527 - fix: sensor error download issue for sensor fail notification

fsayyid:
Due to recent changes to the `createSensorErrorDownload` function some issues resulted for the sensor fail notification which are dealt with in this PR.

To test:
- Add an item to local storage: `sensorUploadTimer : 10` to trigger the async flow
- Upload sensors to trigger failures and partial successes. See that the error download is working as expected in the notification when clicking "Take Me There"
- Try this in different languages as well

### DIFF
--- a/packages/api/src/controllers/sensorController.js
+++ b/packages/api/src/controllers/sensorController.js
@@ -116,13 +116,20 @@ const sensorController = {
       if (!data.length > 0) {
         return await sendResponse(
           () => {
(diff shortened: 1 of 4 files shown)


Cali0707 (approved):
@fsayyid looks good, thanks for fixing this!

smattingly (approved):

PR #2148, opened 2022-07-13
Lf 2637 ngrok setup

fsayyid:
Ticket: [LF-2637](https://lite-farm.atlassian.net/browse/LF-2637)

This PR is to allow simultaneous backend and frontend forwarding with ngrok, and easier set up. 

**New Commands (run from root of repo):**
- `npm run ngrok` to forward both backend and frontend ports with ngrok
- `npm run ngrok:setup` to add the ngrok urls to the file `.env` files (always run after forwarding a port to ngrok)
- `npm run ngrok:api` to forward the backend port with ngrok
- `npm run ngrok:webapp` to forward the frontend port with ngrok

To test: 
- First, run the app as you would normally to ensure changes in this PR do not cause issues in the app otherwise. 
- Create an ngrok account at https://ngrok.com/
- Add the auth token to the `ngrok/ngrok.yml`
- run `npm run ngrok`
- run `npm run ngrok:setup` in a new terminal
- Run both the frontend and backend servers (with `npm run nodmon` from `packages/api` and `pnpm dev` from `packages/webapp`)
- Got to ngrok webapp url on your phone and see if everything is working normally (you may need to test in incognito, and it may run slowly)

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ Since this is a mobile web application, webapp should be viewed in a mobile view
 
 # ngrok
 
-# Use cases for ngrok
+## Use cases for ngrok
 
 Please see https://ngrok.com/ for more general information about ngrok. 
 
@@ -106,14 +106,14 @@ Use cases in which we currently utilize ngrok at LiteFarm include:
 - Testing local changes on phones or different devices
 - Testing local changes when working with other APIs and integrations
 
-# Set up
+## Set up
 
 - Go to https://ngrok.com/ and sign up for an account
 - Create a copy of `ngrok/ngrok.default.yml` and call it `ngrok.yml`. Make sure this file is in the `ngrok` folder at the root of the repo
 - Login to your ngrok account and go to https://dashboard.ngrok.com/get-started/your-authtoken
 - Add the auth token from there in `ngrok.yml` by replacing the `?`
 
-# Commands
+## Commands
 These commands can be run from the root of the repo.
 - `npm run ngrok` to forward both backend and frontend ports with ngrok
 - `npm run ngrok:setup` to add the ngrok urls to the file .env files (always run after forwarding a port to ngrok)


package.json
```diff
@@ -9,7 +9,11 @@
     "lint": "lerna run lint",
     "start": "lerna run start",
     "bootstrap": "lerna bootstrap",
-    "prepare": "husky install"
+    "prepare": "husky install",
+    "ngrok": "ngrok start --config=./ngrok/ngrok.yml --all",
+    "ngrok:setup": "python3 ./ngrok/ngrok-setup.py",
+    "ngrok:api": "ngrok start --config=./ngrok/ngrok.yml api",
+    "ngrok:webapp": "ngrok start --config=./ngrok/ngrok.yml webapp"
```
Cali0707, line 16: These new commands should be documented in the README, as well as the need to have python 3 installed on your machine to run them

packages/api/src/util/ensemble.js
```diff
@@ -27,9 +27,14 @@ if (process.env.NODE_ENV === 'integration') {
   baseUrl = 'https://api.beta.litefarm.org';
 } else if (process.env.NODE_ENV === 'production') {
   baseUrl = 'https://api.app.litefarm.org';
+} else if (process.env.NODE_ENV === 'development') {
```
Cali0707, line 30: As far as I know `process.env.NODE_ENV` is always defaulting to `development` during local development (e.g. if you run `npm run nodemon`). We should probably have some sort of separate mode which we only set if the ngrok commands are run

Cali0707 (changes requested):
When running `ngrok-setup.py` I am getting the following error:
```Traceback (most recent call last):
  File "./ngrok/ngrok-setup.py", line 4, in <module>
    def getURLDict() -> dict[str, str]:
TypeError: 'type' object is not subscriptable
```

packages/api/src/util/ensemble.js
fsayyid: The github tests set the NODE_ENV to 'pipeline'. I think in this case its fine since the webhook will always have an error unless we use ngrok or something of the sort. I also don't think we want to have a separate mode for ngrok since almost everything will be the same as development when using ngrok. I think the only the difference would be the api URL used in to make calls from the frontend.

Cali0707 (approved):
This is working on my end now

smattingly (commented):
Some questions/comments:
1. I know this PR does not introduce it, but line 38 in ensemble.js has a hard coded port number. That should be avoided and now seems like a good time to address it. Ideally there should not be hostnames either, but that may be too much to address now.
2. I think that introducing a new programming language and runtime to the project should get a skeptical cost/benefit analysis. On the one hand, the Python here is not actually part of the application code. On the other, its presence increases the likelihood that Python code will become mission critical without careful analysis. It also complicates the development setup for new/prospective contributors, which I think is too complicated already.
So, I'm encouraging discussion of cost/benefit. Maybe this is too valuable for the team to pass up. One specific point-- could anyone easily replicate the Python as a node or bash script?

smattingly:
@fsayyid Do you think that the effort to do the conversion is reasonable? If not, let me know-- I will approve and merge this as is if that seems to make the most sense. A hybrid approach would be to merge this and create a ticket to replace the Python later; that has the advantage of letting Kevin and others weigh in on the cost/benefit. Perhaps that ticket would never be high enough priority to do.

fsayyid:
@smattingly I don't think it will take too much effort to convert since its a fairly short script.

smattingly (approved):

PR #2150, opened 2022-07-13
Fixed some csv reading types

Cali0707:
Fixes enums in the csv to match the spec

### DIFF
--- a/packages/api/src/controllers/sensorController.js
+++ b/packages/api/src/controllers/sensorController.js
@@ -102,10 +102,23 @@ const sensorController = {
         Reading_types: {
           key: 'reading_types',
           parseFunction: (val) => val.replaceAll(' ', '').split(','),
-          validator: (val) =>
-            val.includes('soil_moisture_content') ||
-            val.includes('water_potential') ||
-            val.includes('temperature'),
+          validator: (val) => {
+            if (!val.length) {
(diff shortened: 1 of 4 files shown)


anant00729:
This works for me. @smattingly please check.

anant00729 (approved):

smattingly (changes requested):
It looks like sensorController is still using the old strings.

packages/api/src/controllers/sensorController.js
```diff
@@ -102,10 +102,23 @@ const sensorController = {
...
+            }
+            const allowedReadingTypes = [
+              'soil_water_content',
+              'soil_water_potential',
+              'temperature',
+            ];
+            let isValid = true;
+            val.forEach((readingType) => {
+              if (!allowedReadingTypes.includes(readingType)) {
+                isValid = false;
```
smattingly, line 117: Suggestion: return false here, return true on line 120, and eliminate the variable.

smattingly (changes requested):

packages/api/src/controllers/sensorController.js
Cali0707: Fixed

smattingly (approved):

PR #2153, opened 2022-07-15
Sensor details only makes necessary callouts

AbdusShaikh:
To test:
1. Open the 'Network' tab in your browsers developer tools.
2. Navigate to a sensor's details view. In the network tab,  ensure the necessary callouts are being made once and not redundantly


Cali0707 (approved):
This looks good to me!

PR #2154, opened 2022-07-15
Fix immuntable gender when changing languages bug

jagDhillon077:
Create a new lite farm account and you should be able to change the gender field to any value instead of the field being stuck to one fixed value. Make sure to also do this in different languages as well.


Cali0707 (approved):
This is working on my end

smattingly (approved):

PR #2155, opened 2022-07-15
Feature/fix sensor as standard location

fsayyid:
Ticket: [LF-2683](https://lite-farm.atlassian.net/browse/LF-2683)

Summary: 
- Refactors code so that `sensor_id, name, farm_id, deleted` are no longer in the sensor table
- Refactors sensor endpoints so that the URLs are more "RESTful"

To test: 
- Run the latest migrations
- Comment out line 199 in `sensorController` to get rid of webhook error. 
- Test all current sensor flow:
  - Add sensor
  - Go to sensor details page
  - Edit sensor details
  - Sensor readings tab
  - Sensor tasks tab
  - Retire sensor
- If a flow isn't working check first if its working on integration / if there's already a bug ticket for it


packages/api/db/deleteSensorData.js
```diff
@@ -0,0 +1,69 @@
+async function deleteSensorData(knex) {
```
fsayyid, line 1: @smattingly can you review this migration?

packages/api/db/migration/20220715214935_sensor_as_standard_location.js
```diff
@@ -0,0 +1,56 @@
+const deleteSensorData = require('../deleteSensorData');
+
```
fsayyid, line 2: @smattingly this file as well.

packages/api/db/deleteSensorData.js
smattingly: @fsayyid This seems good to me, except that it deletes from `sales_task` repeatedly. I don't know the full list of task oriented tables without looking. 

I know that the DROP TABLE statements are repeated because the first appearance is inside a transaction. But I would try to avoid that duplication by using a transaction in all cases.

packages/api/db/migration/20220715214935_sensor_as_standard_location.js
smattingly: This does not autopopulate the uuid for the sensor table-- not sure if that's intentional. Otherwise, this look fine to me.

fsayyid: @smattingly I imagine that the primary key for the sensor table (`location_id`) will be generated by the location table. I believe the other locations (ex. field, gate) use the same logic for their primary keys.

Also, should we keep the farm_id in the sensor table?

smattingly: I didn't think of that, and it makes sense, although I checked some other location "subtype" tables that do autogenerate.

Good question about farm_id. For consistency with the rest of the design, I would lean toward "no".

jagDhillon077:
LGTM:

Did not have to comment out line 199 of sensor controller, webhook initialized and populated correctly

Add sensor - working
Go to sensor details page - details populated including model (model was known to not populate before but it does now)
Edit sensor details - able to edit all details and the patched data then displayed correctly on details 
Sensor readings tab - readings populated on graph correctly, some odd behaviour has been brought up to @anant00729 
Sensor tasks tab - tasks have been created /completed correctly
Retire sensor - able to retire a sensor and then re add the sensor after retiring it

jagDhillon077 (approved):

packages/api/src/routes/sensorRoute.js
```diff
@@ -25,25 +25,26 @@ const upload = multer({ storage });
...
+  validateRequest,
+  SensorController.addReading,
 );
+router.get('/:location_id/reading', SensorController.getAllReadingsByLocationId);
+router.get('/reading/farm/:farm_id', SensorController.getReadingsByFarmId);
+router.post('/reading/invalidate', SensorController.invalidateReadings);
+router.post('/unclaim', SensorController.retireSensor);
+router.get('/:location_id/reading_type', SensorController.getSensorReadingTypes);
+router.get('/partner/:partner_id/brand_name', SensorController.getBrandName);
+router.post('/reading/visualization', SensorController.getAllSensorReadingsByLocationIds);
```
smattingly, line 49: I think this should be GET rather than POST.

smattingly (changes requested):
Some of the comments are just questions for you to consider, if you haven't already. Use your judgment.

packages/api/db/migration/20220715214935_sensor_as_standard_location.js
```diff
@@ -0,0 +1,56 @@
...
+    table
+      .uuid('partner_reading_type_id')
+      .references('partner_reading_type_id')
+      .inTable('partner_reading_type');
+    table.uuid('location_id').references('location_id').inTable('sensor').onDelete('CASCADE');
+  });
+};
+
+exports.down = async function (knex) {
+  deleteSensorData(knex);
```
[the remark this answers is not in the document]
fsayyid: `deleteSensorData` will drop the tables as well

smattingly (approved):

PR #2156, opened 2022-07-18
Fixed copy for inline error

Cali0707:
This PR fixes a small error with the copy text for the inline errors in the bulk upload modal (reported in the comments here: https://lite-farm.atlassian.net/browse/LF-2645)

To test:
1. Go to the farm map and try to upload an invalid file.
2. You should see an inline error saying "We found some errors with your upload. Click here to view them." and "Click here to view them" should be underlined & clickable

The translations are missing in ES, PT, and FR so it probably is not worth testing this change in other languages


anant00729:
This looks good to me

anant00729 (approved):

smattingly (approved):

PR #2157, opened 2022-07-18
Added undelete functionality

Cali0707:
This PR fixes issues with sensors not being undeleted on subsequent uploads (see comments on https://lite-farm.atlassian.net/browse/LF-2646)

To test:
1. Set deleted = true on a sensor's location
2. Re-upload that sensor
3. The sensor should now have deleted = false on its location


fsayyid (approved):
I was able to retire and reupload a sensor properly, so this seems to be working 👍

smattingly (approved):

PR #2160, opened 2022-07-18
Made cherry pick

Cali0707:
This PR cherry picks the changes for LF-2576.

To test:
1. Try to add a watercourse with zero length to the farm map - you should get a popup prompting you to redraw it
2. Try to add a fence with zero length to the farm map - you should get a popup prompting you to redraw it

### DIFF
--- a/packages/webapp/public/locales/en/translation.json
+++ b/packages/webapp/public/locales/en/translation.json
@@ -460,7 +460,8 @@
     },
     "DRAWING_MANAGER": {
       "REDRAW": "Redraw",
-      "ZERO_AREA_DETECTED": "Field with no area detected. Please add more points to current drawing or draw again."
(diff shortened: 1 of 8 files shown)


smattingly:
@Cali0707 This works as described for drawing, but I am able to set the lengths to zero in the form. Can you confirm that is okay?

Cali0707:
@smattingly - I believe so, this change was moved to release ready with the same behaviour as you are describing. We needed to make this change to the drawing to work around some bugs that come up with zero-length polylines that have a weidth in google maps. But, since the length attribute doesn't affect the points we store, I think this behaviour or okay. Let me know if that makes sense, or if you would like me to change the behaviour.

smattingly (approved):

PR #2163, opened 2022-07-18
Addition of empty file error message on the FileUpload component

anant00729:
To test,

1. open http://localhost:3000/map
2. select an empty CSV file and check the error message.

For more details: 
https://lite-farm.atlassian.net/browse/LF-2616


anant00729:
Please check this screenshot

Screen Shot 2022-07-18 at 2 55 13 PM

packages/webapp/src/components/Map/Modals/BulkSensorUploadModal/useValidateBulkSensorData.js
```diff
@@ -54,6 +54,8 @@ export function useValidateBulkSensorData(onUpload, t) {
   const [errorCount, setErrorCount] = useState(0);
   const fileInputRef = useRef(null);
   const [translatedUploadErrors, setTranslatedUploadErrors] = useState([]);
+  const [uploadErrorMessage, setUploadErrorMessage] = useState('');
+  const [errorTypeCode, setErrorTypeCode] = useState(-1);
```
Cali0707, line 58: Could we maybe map these `errorTypeCode`s to some sort of object of enums, so that other devs can understand what the different codes mean, and it is easier to work with? E.g.
```javascript
const ErrorTypes = {
    EMPTY_FILE: 1,
    INVALID_CSV: 0,
}
```

packages/webapp/src/components/Modals/BulkSensorUploadModal/index.jsx
```diff
@@ -23,6 +23,7 @@ export default function BulkSensorUploadModal({
   onTemplateDownloadClick,
   uploadErrorMessage,
   uploadErrorLink,
+  errorTypeCode,
```
Cali0707, line 26: This should also use the same object of enums

Cali0707 (changes requested):
Some small changes to improve maintainability in the future, otherwise looks good!

packages/webapp/src/components/Modals/BulkSensorUploadModal/index.jsx
anant00729: Done

Cali0707 (approved):
Looks good to me!

PR #2168, opened 2022-07-19
Refactored get reading types, patch sensor, and get brand name for sensor

jagDhillon077:
To make sure this is working:

Upload a sensor if you have not already
Check the sensor details and make sure all it's fields are populated, mainly the brand and reading types.

I think there is a bug where the model does not get inputted from the bulk sensor upload so if model has not been populated it is likely because it has not been uploaded to the database in the first place. Without having the model the sensor will not update.

Try to update the sensor fields and make sure there are no errors.

### DIFF
--- a/packages/api/src/models/sensorModel.js
+++ b/packages/api/src/models/sensorModel.js
@@ -125,7 +125,7 @@ class Sensor extends Model {
   static async getSensorLocationBySensorIds(locationIds = []) {
     return await knex.raw(
       `SELECT 
-      s.sensor_id, 
+      s.location_id, 
       s.name, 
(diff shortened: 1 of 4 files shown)


packages/api/src/models/sensorModel.js
```diff
@@ -125,7 +125,7 @@ class Sensor extends Model {
   static async getSensorLocationBySensorIds(locationIds = []) {
```
fsayyid, line 125: @jagDhillon077 can you refactor this function name as well? Maybe `getSensorLocationById` or `getSensorLocationByLocationId` ?

PR #2169, opened 2022-07-20
Retiring a non-integrated sensor no longer fails

AbdusShaikh:
To test:
1. Upload a non-integrated sensor. A non-integrated sensor is one without a brand name or external id
2. After that sensor is successfully uploaded, go to that sensors details view by clicking on its icon in the map
3. Retire that sensor
4. You should see a success message snackbar. As an extra check, you can check to see if that sensor is deleted in the location table in your db.

### DIFF
--- a/packages/api/src/controllers/sensorController.js
+++ b/packages/api/src/controllers/sensorController.js
@@ -552,22 +552,29 @@ const sensorController = {
   async retireSensor(req, res) {
     const trx = await transaction.start(Model.knex());
     try {
-      const { external_id, location_id, farm_id, partner_id } = req.body.sensorInfo;
+      const { external_id, location_id, farm_id, partner_id, brand_name } = req.body.sensorInfo;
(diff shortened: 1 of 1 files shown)


packages/api/src/controllers/sensorController.js
```diff
@@ -552,22 +552,29 @@ const sensorController = {
   async retireSensor(req, res) {
     const trx = await transaction.start(Model.knex());
     try {
-      const { external_id, location_id, farm_id, partner_id } = req.body.sensorInfo;
+      const { external_id, location_id, farm_id, partner_id, brand_name } = req.body.sensorInfo;
+      console.log(req.body.sensorInfo);
```
Cali0707, line 556: Probably want to get rid of this console log

Cali0707 (approved):
This looks good to me, and it seems to work well. I saw one console log we probably want to remove, but otherwise this looks good.

smattingly (approved):

PR #2171, opened 2022-07-20
Generate api docs

smattingly:
Adds tooling to auto-generate OpenAPI documentation, and initial versions of the documentation.
Adds a new dependency for api, so run `npm install` there.
See comments in server.js for context.


Cali0707 (approved):
Very cool, looks good to me!

PR #2174, opened 2022-07-21
Fixed bug where webhook is registered when there are no ESCI devices in the upload

Cali0707:
This PR addresses a bug where a new webhook is still registered with Ensemble when there are no ESCI sensors in the upload.

To test:
1. Create a new farm
2. Upload a file with no Ensemble sensors
3. There should be no webhook for Ensemble in the database for that farm

### DIFF
--- a/packages/api/src/controllers/sensorController.js
+++ b/packages/api/src/controllers/sensorController.js
@@ -192,26 +192,30 @@ const sensorController = {
           },
         );
       } else {
-        // register organization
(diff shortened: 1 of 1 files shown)


jagDhillon077 (approved):
LGTM - 
I uploaded 1 sensor with no brand and one with ESCI and there was only one webhook address registered

smattingly (approved):

PR #2177, opened 2022-07-21
Switched tasks to keep dates in UTC

Cali0707:
This PR fixes the bug described here by switching all the task dates to be handled in UTC

To test:
1. Create a task
2. Update the task's due date to be today
3. This should work

### DIFF
--- a/packages/api/package-lock.json
+++ b/packages/api/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "litefarm-api",
-  "version": "3.2.2",
+  "version": "3.2.3",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "litefarm-api",
-      "version": "3.2.2",
+      "version": "3.2.3",
       "dependencies": {
         "@googlemaps/google-maps-services-js": "^3.3.14",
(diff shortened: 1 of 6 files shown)


jagDhillon077 (changes requested):
Screen Shot 2022-07-21 at 3 12 37 PM
In this view if I change the due date to today's date or any other date it works as intended.

When I click on the task card and check it's due date the value is pre populated correctly.

Screen Shot 2022-07-21 at 3 17 39 PM

However if I try to change the due date in the task details view it always changes the date to one behind the selected date. ex: setting the due date to today will give the task a due date of yesterday. But it looks like it is only a frontend updating issue because If I click off the task and re click the details tab it populates correctly.
Available on Slack if you require any further clarifications

Cali0707:
@jagDhillon077 good catch - looks like I hadn't updated it everywhere!

jagDhillon077 (approved):
LGTM - works perfectly now

smattingly (commented):
See discussion in Jira [LF-2634](https://lite-farm.atlassian.net/browse/LF-2634)

smattingly:
I don't expect any further work on this branch/PR, but we can reopen if needed.

PR #2180, opened 2022-07-22
Send local timezone for server check

smattingly:
See discussion in Jira [LF-2634](https://lite-farm.atlassian.net/browse/LF-2634)


smattingly:
Since this bug fix has been removed from what will be v3.2.4, moving this PR onto integration.

Cali0707:
@smattingly is this PR meant for the collab branch?

smattingly:
@Cali0707 No, that was the point of my previous comment. We removed this from the fix release because I am concerned about potential ripple effects-- I want us to live with it for a while on beta before it is released.

smattingly:
Closing this due to change in requirements.

PR #2181, opened 2022-07-22
Collab/fixes jul 2022

smattingly:
Bring high-priority fixes to main to move toward release.


smattingly:
For unknown reasons, checking out this PR did not give the expected combination of code. When I locally branched from `main` and merged collab branch, I got expected results.

PR #2184, opened 2022-07-25
LF-2675: after selecting a sensor on the farm map it becomes a gate

fsayyid:
Ticket: [LF-2675](https://lite-farm.atlassian.net/browse/LF-2675)

To test:
- Make sure you have some sensors on you map
- Create a task - select one of the sensors as the location
- Make sure that when you select the sensor, the icon is still a sensor icon and not a gate.


AbdusShaikh (approved):
Works well and code looks good.

smattingly (approved):

PR #2185, opened 2022-07-25
Added extra case in handling the ensemble sensor claim response

Cali0707:
This PR just adds a little bit of extra safety into our sensor upload flow, and sends a generic error message if the ensemble API returns data that is not in the expected format for some/all of the sensors.


fsayyid (approved):
I tested on my end by manually assigning `{ success, already_owned, does_not_exist, occupied } = { hello: 'world' }`. Then also testing that uploading non-esci sensors still works. 

So it seems to be working on my end.

anant00729:
This looks good to me.

anant00729 (approved):

PR #2192, opened 2022-07-26
fix: bulk sensor upload model should be optional

fsayyid:
To test:
- Upload sensors where the model column in the csv is empty


Cali0707 (approved):
LGTM 👍

smattingly (approved):

PR #2193, opened 2022-07-26
Bugfix/save button disabled

Cali0707:
This PR fixes a bug where the submit button remained disabled after the email was corrected on the upgrade pseudo user flow.

To test:
1. Try to upgrade a pseudo user to a full user using an email already existing on that user farm
2. This should not work
3. Try switching the email to a new email
4. You should now be able to save the users info


Cali0707:
@smattingly I believe I have fixed the issues you described with my newest commit

smattingly (approved):

PR #2199, opened 2022-07-26
Lf 2662 remove webhook column

fsayyid:
Ticket: [LF-2662](https://lite-farm.atlassian.net/browse/LF-2662)

To test:
- run the latest migrations
- run `npm run ngrok`
- run `npm run ngrok:setup`
- then start both the frontend and backend servers
- Create a new user
- Add an ESCI sensor to the map
- Make sure the webhook_id is added correctly


seinnlee (approved):
lgtm. webhook_id was added for the newly created farm.

packages/api/db/migration/20220725200254_remove_webhook_address_from_integrations.js
```diff
@@ -0,0 +1,11 @@
+exports.up = async function (knex) {
+  return await knex.schema.alterTable('farm_external_integration', (table) => {
+    table.dropColumn('webhook_address');
+  });
+};
+
+exports.down = async function (knex) {
+  return await knex.schema.alterTable('farm_external_integration', (table) => {
+    table.integer('webhook_address');
```
Cali0707, line 9: To make sure the migration can be rolled back, we should be recreating the column to have the same type as before
```suggestion
    table.string('webhook_address');
```

Cali0707 (changes requested):
@fsayyid one small change to make sure we can revert the migration and then this is good to go - everything seems to be working on my end

fsayyid:
Good catch @Cali0707 !

Cali0707 (approved):
Looks great @fsayyid !

PR #2202, opened 2022-07-27
Fixed 400 error when updating sensor

jagDhillon077:
To test:

Drop database and recreate with new updated sensor migrations

Upload a sensor 
Update the sensor values
Ensure the sensor successfully updates and doesn't throw a 400 (bad request) error

### DIFF
--- a/packages/api/db/migration/20220613141913_create_sensors_tables.js
+++ b/packages/api/db/migration/20220613141913_create_sensors_tables.js
@@ -12,7 +12,6 @@ exports.up = function (knex) {
       table.string('external_id').notNullable();
       table.uuid('location_id').references('location_id').inTable('location').notNullable();
       table.float('depth');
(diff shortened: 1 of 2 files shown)


AbdusShaikh (changes requested):
Tested by updating name, lat/lng, reading types, and depth. All are working except depth

anant00729:
Tested the update flow: name, lat, long, and depth are getting updated on my end.

Step: 1: Before Update
Screen Shot 2022-07-28 at 10 41 47 AM

Step 2: On Update
Screen Shot 2022-07-28 at 10 42 16 AM

Step 3: After Update
Screen Shot 2022-07-28 at 10 42 28 AM

anant00729 (approved):

fsayyid:
I'm not sure if editing existing migration is the best approach since these migrations have already been run on beta. We could do it this way, but we would also need an sql script or a knex migration that we can run on the beta database to make these changes.

PR #2203, opened 2022-07-27
LF-2640 add helptext to abandon task screen informing that hours still count toward payroll

seinnlee:
Ticket [LF-2640](https://lite-farm.atlassian.net/browse/LF-2640)

To Test:

1. Create a task and assign it to someone
2. Click "Abandon this task"
3. There should be an info tooltip next to "Was any work completed for this task?"
4. When hovering over, an info text should appear (click when on mobile)

### DIFF
--- a/packages/webapp/src/components/Task/AbandonTask/index.jsx
+++ b/packages/webapp/src/components/Task/AbandonTask/index.jsx
@@ -6,6 +6,7 @@ import { Info, Main } from '../../Typography';
 import { useTranslation } from 'react-i18next';
 import Button from '../../Form/Button';
 import ReactSelect from '../../Form/ReactSelect';
+import { BsInfoCircleFill } from 'react-icons/bs';
 import { Controller, useForm } from 'react-hook-form';
 import InputAutoSize from '../../Form/InputAutoSize';
 import Input from '../../Form/Input';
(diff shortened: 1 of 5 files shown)


packages/webapp/src/components/Task/AbandonTask/index.jsx
```diff
@@ -113,7 +115,21 @@ const PureAbandonTask = ({
...
+                  paddingRight: '10px',
+                  fontSize: '16px',
+                  lineHeight: '20px',
+                  display: 'inline-block',
+                }}
+              >
+                {<Infoi content={t('TASK.ABANDON_TASK_HELPTEXT')} />}
+              </Label>
+            </div>
+          </div>
```
AbdusShaikh, line 132: This works but there is a much simpler way of doing the same thing.

The 'Main' component has a 'toolTipContent' prop. We can just pass t('TASK.ABANDON_TASK_HELPTEXT') into 'Main' as a prop and get the same result

packages/webapp/src/components/Task/AbandonTask/index.jsx
```diff
@@ -113,7 +115,21 @@ const PureAbandonTask = ({
 
       {hasAssignee && (
         <>
-          <Main style={{ marginBottom: '24px' }}>{t('TASK.ABANDON_TASK_DURATION')}</Main>
```
AbdusShaikh, line 116: We can make this:

`<Main style={{ marginBottom: '24px' }} tooltipContent={t('TASK.ABANDON_TASK_DURATION')}>{t('TASK.ABANDON_TASK_DURATION')}</Main>`

AbdusShaikh (changes requested):
Works but there is a simpler and cleaner method of doing the same thing

packages/webapp/src/components/Task/AbandonTask/index.jsx
seinnlee: That is much simpler and cleaner. Thanks!

AbdusShaikh (approved):
Looks good!

Cali0707 (approved):
Looks good to me!

PR #2205, opened 2022-07-28
Prevented api from crashing

Cali0707:
This PR fixes the API crashes from the login flow, although it doesn't address why users without a password in the password table were able to log in with a password through the UI as I was unable to recreate that.

To test:
1. Create a user with a password
2. Log out and verify that you can log back in with the password
3. Delete the password from the password table
4. Try to log in again - you shouldn't be allowed to login, but the API shouldn't crash anymore


fsayyid:
@smattingly Kevin mentioned placing this fix on Avocado

Cali0707:
@smattingly I have updated the code to be a little safer

PR #2206, opened 2022-07-28
Lf 2502/hotfix in temperature component due to database migration changes

anant00729:
To test,

1. Add sensor readings to the sensor_reading table. 
2. check if you can see sensor readings.

### DIFF
--- a/packages/api/src/models/sensorReadingModel.js
+++ b/packages/api/src/models/sensorReadingModel.js
@@ -87,6 +87,7 @@ class SensorReading extends Model {
     return await SensorReading.query()
       .select('*')
       .joinRaw('JOIN sensor ON sensor_reading.location_id::uuid = sensor.location_id')
+      .joinRaw('JOIN location ON location.location_id::uuid = sensor.location_id::uuid')
       .whereIn('sensor.location_id', locationIds)
       .andWhere('reading_type', '=', readingType)
       .andWhere('valid', '=', true)
--- a/packages/api/src/models/sensorReadingModel.js
+++ b/packages/api/src/models/sensorReadingModel.js
@@ -87,7 +87,7 @@ class SensorReading extends Model {
     return await SensorReading.query()
       .select('*')
       .joinRaw('JOIN sensor ON sensor_reading.location_id::uuid = sensor.location_id')
-      .joinRaw('JOIN location ON location.location_id::uuid = sensor.location_id::uuid')
+      .joinRaw('JOIN location ON location.location_id = sensor.location_id')
       .whereIn('sensor.location_id', locationIds)
       .andWhere('reading_type', '=', readingType)
       .andWhere('valid', '=', true)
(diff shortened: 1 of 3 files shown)


jagDhillon077 (approved):
LGTM - temperature showing up in both metric and imperial without any errors

packages/api/src/models/sensorReadingModel.js
```diff
@@ -87,6 +87,7 @@ class SensorReading extends Model {
     return await SensorReading.query()
       .select('*')
       .joinRaw('JOIN sensor ON sensor_reading.location_id::uuid = sensor.location_id')
+      .joinRaw('JOIN location ON location.location_id::uuid = sensor.location_id::uuid')
```
Cali0707, line 90: As far as I know, all of the columns for location_id are already type `uuid`, so I believe we do not need to cast them to type `uuid`

packages/api/src/routes/sensorRoute.js
```diff
@@ -46,5 +46,5 @@ router.post('/reading/invalidate', SensorController.invalidateReadings);
 router.post('/unclaim', SensorController.retireSensor);
 router.get('/:location_id/reading_type', SensorController.getSensorReadingTypes);
 router.get('/partner/:partner_id/brand_name', SensorController.getBrandName);
-router.get('/reading/visualization', SensorController.getAllSensorReadingsByLocationIds);
+router.post('/reading/visualization', SensorController.getAllSensorReadingsByLocationIds);
```
Cali0707, line 49: Is there a reason why this is being switched to a POST request? If I'm understanding what's going on, this endpoint retrieves the sensor readings, so it should be a GET request?

Cali0707 (changes requested):

packages/api/src/routes/sensorRoute.js
anant00729: it was POST initially, I am using the post method to send the data in the request body. I got this new change from database migration PR that I had to again change it from GET to POST in this PR.

packages/api/src/models/sensorReadingModel.js
anant00729: Done

packages/api/src/routes/sensorRoute.js
Cali0707: 👍

packages/api/src/models/sensorReadingModel.js
```diff
@@ -87,6 +87,7 @@ class SensorReading extends Model {
     return await SensorReading.query()
       .select('*')
       .joinRaw('JOIN sensor ON sensor_reading.location_id::uuid = sensor.location_id')
```
Cali0707, line 89:
```suggestion
      .joinRaw('JOIN sensor ON sensor_reading.location_id = sensor.location_id')
```
`sensor_reading.location_id` is type `uuid`, no reason for this to be type casted either

Cali0707 (changes requested):

packages/api/src/models/sensorReadingModel.js
anant00729: corrected that as well. Thanks!

Cali0707 (approved):
Looks good to me!

PR #2209, opened 2022-07-29
Merge 3.2.4 integration

smattingly:


packages/api/src/routes/userFarmRoute.js
```diff
@@ -98,7 +98,7 @@ router.post(
   '/invite/farm/:farm_id/user/:user_id',
   hasFarmAccess({ params: 'farm_id' }),
   checkScope(['edit:users']),
-  userFarmController.upgradePseudoUser(),
+  userFarmController.patchPseudoUserEmail(),
```
fsayyid, line 101: I think the name here was changed to upgradePseudoUser. @seinnlee can you take a look at the code here that involves your pseudo user fix?

packages/api/src/controllers/userFarmController.js
```diff
@@ -587,25 +575,20 @@ const userFarmController = {
           .first()
           .select('*');
         res.status(201).send(userFarm);
+        try {
+          const { farm_name } = userFarm;
+          const user = await userModel.getUserByEmail(email);
+          await emailModel.createTokenSendEmail(
+            { email, gender, birth_year, language: user ? user.language_preference : language },
```
seinnlee, line 582: `language_preference` should be used here since emailModel tries to access the property name of `language_preference` rather than `language`

fsayyid (commented):
@seinnlee I think something is still wrong with the upgrade pseudo user flow, because now the values for gender and birth year are not being autofilled in the sign up flow.

fsayyid:
Actually, I've tested it two more times and I think I may have just made a mistake the first time I tested it. It seems to be working but maybe Sein can verify on his end.

fsayyid (approved):
I tested flows that might have been affected by this PR, and those seem to be working. So I think its safe to merge this.

seinnlee (approved):
LGTM. The new functionalities seemed to work for me. It looks like some code changes were mostly formatting changes presumably from lint rules so they look safe to merge.

PR #2210, opened 2022-08-01
Correct default units for sensor depth

AbdusShaikh:
To test:
1. Navigate to the details page of a sensor
2. If you are using the imperial system, the depth field should be using inches. If you are using metric, the depth field should be using cm


fsayyid (commented):
When I go to edit the depth field, should cm be the default or m? Right now its m

jagDhillon077:
> When I go to edit the depth field (for metric), should cm be the default or m? Right now its m

@fsayyid I have populated the database to have a default value of cm.

fsayyid:
@AbdusShaikh @jagDhillon077 after merging Jag's PR the default unit when editing a sensor is still metres: 
Screen Shot 2022-08-02 at 4 51 30 PM

PR #2211, opened 2022-08-02
LF-2618 Entering in an abandon/completion future date for a plan, doesn't change status

seinnlee:
Ticket [LF-2618](https://lite-farm.atlassian.net/browse/LF-2618)

To Test:
1. Go to an existing crop plan (or create one if not).
2. Go through the abandon crop plan flow.
3. Change the "Date of status change" to a future date by clicking the input form and manually typing in a future date (calendar selection is already disabled for future dates).
4. There should be an in-line error message and "Abandon" button should be disabled.

### DIFF
--- a/packages/webapp/public/locales/en/translation.json
+++ b/packages/webapp/public/locales/en/translation.json
@@ -976,6 +976,7 @@
       "COMPLETE_PLAN": "Complete plan",
       "DATE_OF_CHANGE": "Date of status change",
       "NOTES_CHAR_LIMIT": "Notes must be less than 10,000 characters",
+      "FUTURE_DATE_INVALID": "May not be in the future",
       "RATING": "Rate this crop plan",
       "REASON": {
         "CROP_FAILURE": "Crop failure",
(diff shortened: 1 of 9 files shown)


Cali0707 (changes requested):
@seinnlee looking at the ticket, it looks like completed dates should also not be allowed to be in the future for crop plans. Would you be able to add that into this PR?

seinnlee:
@Cali0707 Good catch! Just updated it to include the completion flow.

Cali0707 (approved):
This works for me!

fsayyid:
@seinnlee This is working well, but we would also want this validation in other parts of the app, for example on the task abandonment/completion. That is out of the scope of this ticket, but do you think you can add this as optional validation to the date input component itself so that this functionality can be reused?

seinnlee:
@fsayyid I updated the Input component to handle future date validation, but since the abandon or complete button lives in the parent components (abandon or complete tasks/plans pages), the Input component will have to send back the validation result to the parent components. If we want to disable future dates for certain components, we can now add `disableFuture={true}`. The default is false so all the other date input components (harvest date, etc.) I haven't updated should still function as expected.

fsayyid:
I feel like you can make use of hookform custom validation to greatly simplify this. I apologize for not mentioning this before. But you can refactor date validation into a utility function that returns the error message, `isNotInFuture` for example, and then assign the prop`hookFormRegister={ register(DATE, { required: true, validate: isNotInFuture  }) }` . 

Here's the [documentation](https://react-hook-form.com/api/useform/register)

This way we still have reusable code without adding complexity to the date input component. 

Sorry again for the back and forth.

seinnlee:
@fsayyid I updated the PR to use the hookform custom validation. I agree that it's much simpler and safer than refactoring a very commonly used component. Thanks for the idea!!

PR #2212, opened 2022-08-02
Added sensor depth unit to sensor table

jagDhillon077:


fsayyid (commented):
@jagDhillon077 It seems to still be defaulting to metres in the UI rather than cm.

fsayyid:
But otherwise seems to be working.

fsayyid (approved):

PR #2215, opened 2022-08-03
Maximum depth of 10,000 and able to update sensor properties on a non ensemble sensor

jagDhillon077:
To test:
Upload a Non Ensemble Scientific Sensor
When updating a sensor - make sure to input a value > 10,000 in the depth field and see if it throws an error
Update the name depth long, lat, etc and make sure it updates correctly

### DIFF
--- a/packages/api/src/controllers/sensorController.js
+++ b/packages/api/src/controllers/sensorController.js
@@ -394,10 +394,7 @@ const sensorController = {
         .patch({ name: sensor_name })
         .where('location_id', location_id);
 
-      await SensorModel.query()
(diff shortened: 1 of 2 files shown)


AbdusShaikh (approved):
All the functionality is working for me and the code changes make sense.
Looks good

fsayyid (commented):
This looks good seems to be working. When I update the name though, it doesn't immediately update in the UI, can that be something that's fixed in this PR or would you rather it be in another?

fsayyid (approved):

PR #2216, opened 2022-08-03
Added date to emails

Cali0707:
This PR adds a date to the subject line of the support emails (see ticket here: https://lite-farm.atlassian.net/browse/LF-2362)

To test:
1. Click on the Profile icon in the top right-hand corner
2. Click on the Help button
3. Submit a help request email
4. You should have an email in your inbox with the date and time included in the subject line


seinnlee (approved):
lgtm. It also worked in French. 

Just an idea is that we can maybe replace `Z` with `(UTC)` in subject. Not sure which is more intuitive for users though.

fsayyid (approved):
Looks good

PR #2218, opened 2022-08-04
fix: crop plan 400 error on setting estimated seeding to mt

fsayyid:
Ticket (See latest comment):  [LF-2521](https://lite-farm.atlassian.net/browse/LF-2521)

To test:
- Run the latest migrations
- Start the flow to creating a planting crop plan
-  Choose rows as planting method
- Enter an amount for 'Estimated seeding' and change the unit to `mt`
- Complete plan creation flow


jagDhillon077 (approved):
Works for me - not seeing a 400 error and crop plan is created

seinnlee (approved):
I tested with both metric and imperial and it worked.

PR #2220, opened 2022-08-04
LF-2253 when revoking a users access ensure at least one fo fm or eo retains access to the farm

seinnlee:
Ticket [LF-2253](https://lite-farm.atlassian.net/browse/LF-2253)

To Test:
1. Create a farm with one admin account (try one admin and no workers, try one inactive admin and one active admin, try two active admins, etc.).
2. Go edit the last admin's profile.
3. Click "Revoke User Access".
4. "Cannot revoke access" modal should pop up instead of "Revoke User Access" modal.
5. If you click "here" in the modal, it should redirect you to Request Help page.

### DIFF
--- a/packages/api/src/controllers/userFarmController.js
+++ b/packages/api/src/controllers/userFarmController.js
@@ -391,8 +391,10 @@ const userFarmController = {
 
         // check if access is revoked or restored: update email info based on this
         if (currentStatus === 'Active' || currentStatus === 'Invited') {
-          if (admins.length <= 1 && admins[0].user_id === user_id) {
-            return res.status(400).send("You cannot revoke the last admin's access on this farm");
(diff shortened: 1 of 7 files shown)


Cali0707 (changes requested):
@seinnlee if I invite a second manager from the people tab and then try to revoke access to either the second manager or my current user, it says that I can't revoke access for the last admin on the farm. This probably shouldn't happen, because as I understand it there would be two admins on the farm in this scenario

seinnlee:
@Cali0707 I updated it so that you can revoke invited managers as long as there is an active admin. But I don't think you should be able to revoke access of yourself when the invited manager hasn't created an account yet (they could have made a typo in the invite email, etc.).

Cali0707 (approved):
This looks good to me!

fsayyid (approved):

PR #2221, opened 2022-08-05
Lf 2632 point locations are being inappropriately clustered

Cali0707:
This PR resolves an issue where some points were being clustered when they shouldn't have been clustered.

To test:
1. Add two gates to your map, which are very close but not touching each other when the map is on max zoom
2. When you zoom out, they should be grouped into a marker
3. When you zoom in, they should separate out into two separate icons

### DIFF
--- a/packages/webapp/package.json
+++ b/packages/webapp/package.json
@@ -39,7 +39,7 @@
   },
   "dependencies": {
     "@emotion/react": "^11.7.1",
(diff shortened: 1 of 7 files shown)


Cali0707:
FYI you will need to run `pnpm install` in the webapp for this PR

fsayyid (approved):
Looks good to me 👍

AbdusShaikh (approved):
Looks good!

PR #2222, opened 2022-08-05
PATCH: fix: crop plan 400 error on setting estimated seeding to mt

fsayyid:
Patch fix issue on production.

Related PR: #2218

Related ticket: https://lite-farm.atlassian.net/browse/LF-2521

### DIFF
--- a/packages/api/db/migration/20220804164229_estimated_seeds_unit_add_tonnes.js
+++ b/packages/api/db/migration/20220804164229_estimated_seeds_unit_add_tonnes.js
@@ -9,6 +9,8 @@ exports.up = function (knex) {
 
 exports.down = function (knex) {
   return knex.schema.raw(`
+        UPDATE planting_management_plan SET estimated_seeds_unit = 'lb' WHERE estimated_seeds_unit = 't';
+        UPDATE planting_management_plan SET estimated_seeds_unit = 'kg' WHERE estimated_seeds_unit = 'mt';
         ALTER TABLE planting_management_plan
         DROP CONSTRAINT planting_management_plan_estimated_seeds_unit_check,
         ADD CONSTRAINT planting_management_plan_estimated_seeds_unit_check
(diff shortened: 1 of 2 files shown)


packages/api/db/migration/20220804164229_estimated_seeds_unit_add_tonnes.js
```diff
@@ -0,0 +1,17 @@
...
+        CHECK (estimated_seeds_unit = ANY (ARRAY['g'::text, 'kg'::text, 'mt'::text, 'oz'::text, 'lb'::text, 't'::text]));
+    `);
+};
+
+exports.down = function (knex) {
+  return knex.schema.raw(`
+        ALTER TABLE planting_management_plan
+        DROP CONSTRAINT planting_management_plan_estimated_seeds_unit_check,
+        ADD CONSTRAINT planting_management_plan_estimated_seeds_unit_check
+        CHECK (estimated_seeds_unit = ANY (ARRAY['g'::text, 'kg'::text, 'oz'::text, 'lb'::text]));
```
Cali0707, line 15: This will fail if there is at least one row in the table that is using 'mt' - can you update it to maybe switch all of those to be 't' before adding the constraint?

Cali0707 (changes requested):

packages/api/db/migration/20220804164229_estimated_seeds_unit_add_tonnes.js
fsayyid: I think we would want it to fail if there is at least one row in the database that has `mt` or `t` right? Since we wouldn't want to invalidate or change the existing data. We don't usually run these down functions from what I understand.

Cali0707: @fsayyid I don't think this would really be invalidating the data - wouldn't this just be updating the default display unit because everything in the database is stored in kg?

Cali0707 (approved):

PR #2223, opened 2022-08-05
fix: down function for estimated seeds migration

fsayyid:


Cali0707 (approved):
Looks good to me!

jagDhillon077 (approved):
This works even after rolling back a migration when creating a crop plan that contains mt or t.

PR #2228, opened 2022-08-08
Lf 2502/addition of projected temp rect

anant00729:
To test:

1. go to the http://localhost:3000/sensor/<location_id>/readings.
2. add some sensor reading using the following script.

```
CREATE OR REPLACE FUNCTION generate_sensor_readings(
  readingType TEXT,
  unit TEXT,
  valid BOOLEAN,
  startDate DATE,
  endDate DATE,
  tMax REAL,
  tMin REAL 
  ) RETURNS INTEGER AS
$func$
DECLARE 
 currentDate DATE := NOW();
 valueData REAL := 0;
 r RECORD;
 s RECORD;
 intervalDuration TEXT := '1 hour';
 returnedID TEXT := '';
 counter INTEGER := 0;
BEGIN
  EXECUTE $$ TRUNCATE TABLE sensor_reading $$;  
  FOR s IN SELECT * FROM sensor LOOP 
    FOR r IN SELECT * FROM generate_series(
        startDate,endDate, 
        intervalDuration::INTERVAL) AS s(i) LOOP
        
        valueData := ROUND((RANDOM()*(tMax-tMin)+tMin)::NUMERIC, 1);
        EXECUTE $$
        INSERT INTO public.sensor_reading(
        location_id, 
        read_time, 
        created_at, 
        reading_type, 
        value, 
        unit, 
        valid)
        VALUES ($1, $2, $3, $4, $5, $6, $7)
        RETURNING reading_id $$     
        USING s.location_id, r.i ,currentDate, readingType, 
        valueData, unit ,valid
        INTO returnedID;
        counter := counter + 1;
    END LOOP;
  END LOOP;
  RETURN counter;
END
$func$ LANGUAGE plpgsql;
SELECT 
generate_sensor_readings AS g_asd 
FROM generate_sensor_readings(
    'temperature', 
    'C', 
    true, 
    '2022-08-04', 
    '2022-08-15',
    17,
    23.5
);
```

3. check if you can see the following line chart with predicted temp values: 

Screen Shot 2022-08-08 at 12 38 38 PM

For more details: 
https://lite-farm.atlassian.net/browse/LF-2502

Code ref: 
https://codesandbox.io/s/ib4z5?file=/src/App.tsx:1340-2152

### DIFF
--- a/packages/webapp/src/components/SensorReadingsLineChart/PredictedRect.jsx
+++ b/packages/webapp/src/components/SensorReadingsLineChart/PredictedRect.jsx
@@ -1,0 +1,19 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+function PredictedRect(props) {
+  const { x: oX, y: oY, width: oWidth, height: oHeight = 0, fill = '' } = props;
+
+  let x = oX;
+  let y = oHeight < 0 ? oY + oHeight : oY;
+  let width = oWidth;
+  let height = Math.abs(oHeight);
+
+  return <rect fill={fill} mask="url(#mask-stripe)" x={x} y={y} width={width} height={height} />;
+}
+
+PredictedRect.propTypes = {
+  fill: PropTypes.string,
+};
+
+export default PredictedRect;
--- a/packages/webapp/src/components/SensorReadingsLineChart/index.jsx
+++ b/packages/webapp/src/components/SensorReadingsLineChart/index.jsx
@@ -11,10 +11,12 @@ import {
   ResponsiveContainer,
   Surface,
   Symbols,
+  ReferenceArea,
 } from 'recharts';
 import { Label, Semibold } from '../Typography';
 import PropTypes from 'prop-types';
 import AxisLabel from './AxisLabel';
+import PredictedRect from './PredictedRect';
 import moment from 'moment';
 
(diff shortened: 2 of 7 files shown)


Cali0707 (approved):
This looks good

fsayyid:
- For the projected temperature, is it showing the projected temperature for the next day? I think it should the date for the projected temperature. 

- Also the when I convert to using imperial units, the sensor readings are not converted to Celcius. I think you mentioned some issue with handling this in stand up, does converting the values take too much time? I think we can talk about adding a loading screen if that's the case.

anant00729:
1. For the first point do you want to say that we should include 09 Aug, if the current day is 8th Aug?
2. Yes, I have created a ticket for that, Please check: 

https://lite-farm.atlassian.net/browse/LF-2720

fsayyid:
- I think that it would makes to include the next day to show that its a projection, but maybe talk to Kevin about
- Thanks for letting me know

Maybe I can merge this PR and you can add tomorrow's date in another PR after talking to Kevin?

anant00729:
As per the discussion, I have added exactly 20 points for the ambient temperature readings. the shaded region will shift as the day progresses.

Screen Shot 2022-08-09 at 9 49 51 AM

fsayyid (approved):

PR #2229, opened 2022-08-08
Fixed invite flow to have inline error when user with same email already exists

Cali0707:
To test:
1. Create a new user on a farm
2. Try and invite a second user who has the same email as either your current user or another user on the farm
3. There should be an inline error which prevents you from inviting the user until you have provided a valid email


anant00729:
Tested: 

Screen Shot 2022-08-08 at 1 05 22 PM

This looks good to me.

anant00729 (approved):

fsayyid (approved):
This looks good 👍 .

PR #2230, opened 2022-08-08
Fixed various bugs on the map

Cali0707:
This PR fixes a few bugs caused by updating the clustering library for the map
 
To test:
1. Create an area type location
2. Reload the map
3. It should still render
4. Try filtering out some point locations
5. The marker numbers should update accordingly


jagDhillon077 (approved):
This fixed my issues with the map and tests passed

fsayyid (approved):
This seems to be working. I tried making a few different types of locations and played around with the filters.

PR #2232, opened 2022-08-09
Fixed bug where it was impossible to invite a pseudo user

Cali0707:
This PR resolves the issue described in the comments of https://lite-farm.atlassian.net/browse/LF-2672

To test:
1. Try and invite a pseudo user
2. This should work
3. Try and invite a non pseudo user using an email of an existing user on the farm
4. You should be shown an inline error

### DIFF
--- a/packages/webapp/src/components/InviteUser/index.jsx
+++ b/packages/webapp/src/components/InviteUser/index.jsx
@@ -109,7 +109,7 @@ export default function PureInviteUser({ onInvite, onGoBack, userFarmEmails, rol
           },
           validate: {
             existing: (value) => {
-              if (role.value === 3) {
+              if (role.value === 3 && !value) {
                 return true;
               } else {
                 return (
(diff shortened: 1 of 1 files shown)


seinnlee:
@Cali0707 Why do we not get an inline error for inviting a worker with an email that's already associated with another account on the farm?

seinnlee:
@Cali0707 It's not the cleanest but something like this would solve the issue Mwaya found but also show inline error if the email already exists when inviting a farm worker.

`return ((((role.value === 3) ? true : value) && !userFarmEmails.includes(value.toLowerCase())) ||
                t('INVITE_USER.ALREADY_EXISTING_EMAIL_ERROR'));`

Cali0707:
@seinnlee good catch! I've fixed this now

seinnlee (approved):
LGTM!

jagDhillon077 (approved):
This works for me 👍🏽

PR #2233, opened 2022-08-09
Fixed regex and validation of empty case

Cali0707:
To test:
1. Upload a csv file where the `Reading_types` column is empty for a sensor through your favourite http client to `http://localhost:5001/sensor/`. You will need to set the Bearer token to be the one used by your currently authenticated user, and also set the farm_id header appropriately for this to work. The upload is a POST request with a multipart form data field where the key for the csv file is `sensors`
2. You should get a validation error


fsayyid (approved):
This looks good 👍

AbdusShaikh (approved):
Works as expected

PR #2234, opened 2022-08-09
Lf 2502/hotfix last updated time

anant00729:
To test: 

1. go to http://localhost:3000/sensor/<location_id>/readings
2. check the last updated text format: 

it should look like this: 
Screen Shot 2022-08-09 at 2 38 21 PM


fsayyid (approved):
Looks good and logic makes sense.

Cali0707 (approved):
Looks good!

PR #2235, opened 2022-08-09
[ngrok required to test] LF-2722: Long press for sensor reading preview on map is not working for clustered sensors

fsayyid:
Ticket: [LF-2722](https://lite-farm.atlassian.net/browse/LF-2722)

To test:
- Upload sensors with the same latitude and longitude.
- Run ngrok and access the farm map on a mobile device.
- Click the clustered sensors to see the list of sensors in one location.
- Long press a sensor to see the preview

### DIFF
--- a/packages/webapp/src/components/Map/SelectionHandler/index.jsx
+++ b/packages/webapp/src/components/Map/SelectionHandler/index.jsx
@@ -83,7 +83,7 @@ export default function PureSelectionHandler({ locations, history, sensorReading
     }
   };
 
-  const removeTextHighlight = isTouchDevice()
+  const removeSelect = isTouchDevice()
     ? {
         userSelect: 'none',
         WebkitUserSelect: 'none',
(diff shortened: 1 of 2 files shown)


Cali0707 (commented):
@fsayyid this is working, but it is very finicky to get the long press to work correctly. If I press for slightly the wrong amount of time, I get this behaviour:
IMG_B342A595BAF8-1

Up to you whether you think we should approve this now and open a bug ticket for that, or if you'd prefer to work on that now

fsayyid:
@Cali0707 I'm pretty sure that was just because of the icon. I think now that I made the icon "un-selectable", it should solve this issue. But if it doesn't we can create a separate bug ticket for this.

Cali0707 (approved):
@fsayyid looks like that fixed it

seinnlee (approved):
Tested with dev tools and it looks good to me.

PR #2239, opened 2022-08-12
LF-2626 - fix: email not appearing correctly on create account

fsayyid:
Ticket: [LF-2626](https://lite-farm.atlassian.net/browse/LF-2626)

Unfortunately this issue happens intermittently so it is hard to test. My hunch is that this issue had to do with the component not re-rendering on email change since it wasn't state. And I am no longer able to replicate the issue after making it state.

To test: 
- Start the flow to create a new account
- Go back and forth between entering your email and creating the account 
- Try this while using the Slow 3G option in the network tab
- Complete create account flow to see if that still works


seinnlee (approved):
I haven't been able to reproduce the original bug even on integration but this seems to fix one similar issue I saw. 

1. Enter an email and go to "create user account" page
2. Click "go back" and refresh the page
3. Edit the email address and click "continue", the email address is not updated.

This issue seems to be resolved with this PR and I assume the original issue came from something similar.

anant00729 (approved):
Looks good to me

PR #2240, opened 2022-08-12
LF-2730 augment sensor preview

seinnlee:
Ticket [LF-2730](https://lite-farm.atlassian.net/browse/LF-2730)

**To Test:**
Case 1:
1. Have at least 2 sensors uploaded to the map (apart from each other so they're not clustered up).
2. Long press a sensor for about 1 second (0.8 seconds to be exact) without lifting the mouse up.
3. The map should pan to the long pressed sensor location (smoother transition than before).
4. Repeat step 2 for the other sensor (same result expected).

Case 2:
1. Now have at least 2 sensors clustered up (update the lat lng to be the same in the details page).
2. Click the cluster to see the list of sensors/other locations.
3. Long press a sensor in the list without lifting the mouse up.
4. The map should show the sensor preview popup.

Mobile:
1. Repeat the above two cases on mobile devices (using ngrok).

*may not work on android but there's a [separate ticket](https://lite-farm.atlassian.net/browse/LF-2722) for that so this PR won't cover android specific long press*

### DIFF
--- a/packages/webapp/src/components/Map/SelectionHandler/index.jsx
+++ b/packages/webapp/src/components/Map/SelectionHandler/index.jsx
@@ -60,7 +60,7 @@ export default function PureSelectionHandler({ locations, history, sensorReading
         setIsSensor(true);
         setSensorIdx(idx);
       }
-    }, 1000);
+    }, 800);
   };
 
   const handleMouseUp = (location) => {
@@ -84,16 +84,6 @@ export default function PureSelectionHandler({ locations, history, sensorReading
       }
     : {};
 
-  const longPressHandlers = isTouchDevice()
-    ? {
-        onTouchStart: handleMouseDown,
-        onTouchEnd: handleMouseUp,
-      }
-    : {
-        onMouseDown: handleMouseDown,
-        onMouseUp: handleMouseUp,
-      };
-
   return locations.map((location, idx) => {
     const { type, asset, name } = { ...location };
     let icon = imgMapping(asset, type);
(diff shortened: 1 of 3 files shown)


packages/webapp/src/components/Map/SelectionHandler/index.jsx
```diff
@@ -50,22 +50,23 @@ export default function PureSelectionHandler({ locations, history, sensorReading
...
+  const handleMouseDown = (location, idx) => {
+    longPressed.current = false;
+    setTimeout(function () {
+      longPressed.current = true;
+      if (location.type === SENSOR && longPressed.current) {
+        setIsSensor(true);
+        setSensorIdx(idx);
       }
-    }, 200);
+    }, 800);
```
fsayyid, line 63: I think it could be nice to create a constant for the time so you don't have to change in 2 places.

fsayyid (approved):
This looks good overall, I tested on iPhone and that also seems to be working. 

I made one suggestion above, but it's not a big change so this can still move forward. 

@seinnlee You can handle this in another PR if you want its up to you:
- When I click on the compact temperature component from a cluster, it takes me to details rather than the readings page. Also from what I remember from planning, I believe when we click on a sensor without long press, it should still go to readings tab. With long press, it should go to the readings tab with the temperature readings graph scrolled into view.

packages/webapp/src/components/Map/SelectionHandler/index.jsx
seinnlee: That's a good idea!

seinnlee:
@fsayyid I fixed the cluster temperature component clicking issue

Cali0707 (changes requested):
@seinnlee things are only showing up for me once I release my mouse, not during the long press.

seinnlee:
@Cali0707 What device did you use to test it? It still works for me on my laptop (chrome). Did it not work for individual sensors or clustered sensors or both?

Cali0707 (approved):
@seinnlee my bad - looks like I misunderstood how this was supposed to work. This seems to be working on my end.

PR #2243, opened 2022-08-15
Other reason not showing up on read only task view

jagDhillon077:
Create a task and abandon the task and set the reason for task abandoning to other and fill in the explanation. 
You should be able to click on the task read only card and be able to see the reason set to "other" and explanation there as well


fsayyid (approved):
Looks good 👍

anant00729 (approved):
Looks good to me.

PR #2244, opened 2022-08-15
fix: sensor reading types for sensors without brand

fsayyid:
To test:
- Create a non ESCI sensor if you don't have one already
- Remove and re-add a sensor reading type


packages/webapp/src/containers/LocationDetails/PointDetails/SensorDetail/saga.js
```diff
@@ -83,6 +84,7 @@ export function* getSensorBrandSaga({ payload: { location_id, partner_id } }) {
     const brand_name = brand_name_response.data;
     yield put(onSensorBrandSuccess({ location_id, brand_name }));
   } catch (error) {
+    console.log(error);
```
Cali0707, line 87:
```suggestion

```

Cali0707 (approved):
Looks great!

jagDhillon077 (approved):
This works for me

PR #2245, opened 2022-08-15
LF-2712 improve error message format for reading types

seinnlee:
Ticket [LF-2712](https://lite-farm.atlassian.net/browse/LF-2712)

To Test:
1. Create a "add sensors to litefarm" CSV file with invalid reading types.
2. Upload the file and download the error message.
3. The error message format should be changed to as described by the ticket.

### DIFF
--- a/packages/api/src/controllers/sensorController.js
+++ b/packages/api/src/controllers/sensorController.js
@@ -658,7 +658,6 @@ const parseCsvString = (csvString, mapping, delimiter = ',') => {
       const parsedRow = headers.reduce((previousObj, current, index) => {
         if (allowedHeaders.includes(current)) {
           const val = mapping[current].parseFunction(values[index].replace(/^(["'])(.*)\1$/, '$2')); // removes any surrounding quotation marks
-          console.log(val, mapping[current].validator(val));
           if (mapping[current].validator(val)) {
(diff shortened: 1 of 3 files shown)


packages/api/src/controllers/sensorController.js
```diff
@@ -70,6 +70,7 @@ const sensorController = {
       const { access_token } = await IntegratingPartnersModel.getAccessAndRefreshTokens(
         'Ensemble Scientific',
       );
+      const invalidReadingTypes = [];
```
Cali0707, line 73: @seinnlee putting this here seems to lead to issues if there are more than one sensors with the same error in a single upload. Can you look at how the errors are constructed on line 664 and add them in there, rather than in this top-level call to the parseing function? I believe it should be fairly straightforward to add variables for the errors in there. You would also need to add them everywhere sensors are added to `errorSensors` (probably by spreading the variables from the error into the variables object)

Cali0707 (changes requested):
This seems to be working on the frontend, but not on the backend currently. I added a comment describing what I think is going on there

Cali0707 (approved):
Looks good to me 👍

anant00729:
This works for me!

anant00729 (approved):

PR #2247, opened 2022-08-16
Added the x-axix ticks and x axis labels

anant00729:
To test: 

1. go to http://localhost:3000/sensor/<location_id>/readings
2. check if you can find the following changes on the x-axis in the line chart:

Screen Shot 2022-08-15 at 5 41 44 PM

```
CREATE OR REPLACE FUNCTION generate_sensor_readings(
  readingType TEXT,
  unit TEXT,
  valid BOOLEAN,
  startDate DATE,
  endDate DATE,
  tMax REAL,
  tMin REAL 
  ) RETURNS INTEGER AS
$func$
DECLARE 
 currentDate DATE := NOW();
 valueData REAL := 0;
 r RECORD;
 s RECORD;
 intervalDuration TEXT := '1 hour';
 returnedID TEXT := '';
 counter INTEGER := 0;
BEGIN
  EXECUTE $$ TRUNCATE TABLE sensor_reading $$;  
  FOR s IN SELECT * FROM sensor LOOP 
    FOR r IN SELECT * FROM generate_series(
        startDate,endDate, 
        intervalDuration::INTERVAL) AS s(i) LOOP
        
        valueData := ROUND((RANDOM()*(tMax-tMin)+tMin)::NUMERIC, 1);
        EXECUTE $$
        INSERT INTO public.sensor_reading(
        location_id, 
        read_time, 
        created_at, 
        reading_type, 
        value, 
        unit, 
        valid)
        VALUES ($1, $2, $3, $4, $5, $6, $7)
        RETURNING reading_id $$     
        USING s.location_id, r.i ,currentDate, readingType, 
        valueData, unit ,valid
        INTO returnedID;
        counter := counter + 1;
    END LOOP;
  END LOOP;
  RETURN counter;
END
$func$ LANGUAGE plpgsql;
SELECT 
generate_sensor_readings AS g_asd 
FROM generate_sensor_readings(
    'temperature', 
    'C', 
    true, 
    '2022-08-05', 
    '2022-08-20',
    17,
    23.5
);
```

### DIFF
--- a/packages/webapp/src/components/SensorReadingsLineChart/index.jsx
+++ b/packages/webapp/src/components/SensorReadingsLineChart/index.jsx
@@ -97,8 +97,9 @@ const PureSensorReadingsLineChart = ({
   const renderDateOnXAxis = (tickProps) => {
     const { x, y, payload, index } = tickProps;
     const { value, offset } = payload;
-    const displayDate = dateFormat.format(new Date(value));
-    const displayDay = dayFormat.format(new Date(value));
+    const tickDate = new Date(value);
+    const displayDate = dateFormat.format(tickDate);
+    const displayDay = dayFormat.format(tickDate);
 
     if (index % 4 == 2) {
       return (
--- a/packages/webapp/src/components/SensorReadingsLineChart/index.jsx
+++ b/packages/webapp/src/components/SensorReadingsLineChart/index.jsx
@@ -93,12 +93,18 @@ const PureSensorReadingsLineChart = ({
   const renderDateOnXAxis = (tickProps) => {
     const { x, y, payload, index } = tickProps;
     const { value, offset } = payload;
-    const displayDate = moment(value).format('ddd, MMM DD');
+    const displayDate = moment(value).format('DD');
+    const displayDay = moment(value).format('ddd');
     if (index % 4 == 2) {
       return (
-        <text x={x} y={y - 16} textAnchor="middle">
-          {displayDate}
-        </text>
+        <>
+          <text x={x} y={y - 16} textAnchor="middle">
+            {displayDate}
+          </text>
+          <text x={x} y={y} textAnchor="middle">
+            {displayDay}
+          </text>
+        </>
       );
     }
     if (index % 4 === 0 && index !== 0) {
@@ -127,14 +133,14 @@ const PureSensorReadingsLineChart = ({
       </div>
       <Label className={styles.subTitle}>{subTitle}</Label>
       <Label className={styles.subTitle}>{weatherStationName}</Label>
-      <ResponsiveContainer width="100%" height="50%">
+      <ResponsiveContainer width="100%" height="60%">
         <LineChart
           data={chartData}
           margin={{
             top: 20,
             right: 30,
-            left: 5,
-            bottom: 20,
+            left: 20,
+            bottom: 60,
           }}
         >
           <pattern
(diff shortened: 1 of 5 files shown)


packages/webapp/src/components/SensorReadingsLineChart/index.jsx
```diff
@@ -93,12 +93,18 @@ const PureSensorReadingsLineChart = ({
   const renderDateOnXAxis = (tickProps) => {
     const { x, y, payload, index } = tickProps;
     const { value, offset } = payload;
-    const displayDate = moment(value).format('ddd, MMM DD');
+    const displayDate = moment(value).format('DD');
+    const displayDay = moment(value).format('ddd');
```
Cali0707, line 97: In general, we are trying to move away from using moment in LiteFarm as it is now deprecated (See `deprecate-momentjs` [here](https://lite-farm.atlassian.net/wiki/spaces/LITEFARM/pages/1191575576/Pre-integration+peer+review)). Can you refactor this to not use moment?

packages/webapp/src/containers/SensorReadings/saga.js
```diff
@@ -169,6 +169,14 @@ export function* getSensorsReadingsSaga({ payload }) {
     );
     const lastUpdatedReadingsTime = moment(lastUpdated).startOf('day').fromNow();
 
+    let xAxisLabel = '';
+    if (Object.keys(ambientDataWithSensorsReadings).length) {
+      const startDateObj = new Date(+Object.keys(ambientDataWithSensorsReadings)[0] * 1000);
+      const endDateObj = new Date(+Object.keys(ambientDataWithSensorsReadings).at(-1) * 1000);
+      xAxisLabel = `${moment(startDateObj).format('MMM DD')} - ${moment(endDateObj).format(
```
Cali0707, line 176: Can you get rid of moment here too?

packages/webapp/src/containers/SensorReadings/saga.js
```diff
@@ -169,6 +169,14 @@ export function* getSensorsReadingsSaga({ payload }) {
     );
     const lastUpdatedReadingsTime = moment(lastUpdated).startOf('day').fromNow();
 
+    let xAxisLabel = '';
+    if (Object.keys(ambientDataWithSensorsReadings).length) {
+      const startDateObj = new Date(+Object.keys(ambientDataWithSensorsReadings)[0] * 1000);
+      const endDateObj = new Date(+Object.keys(ambientDataWithSensorsReadings).at(-1) * 1000);
```
Cali0707, line 175: It would probably be best to put the `Object.keys()` result here into a variable rather than calling the `Object.keys()` function three times.

Cali0707 (changes requested):
This looks great and works well, just some small changes to the code and then this should be good to go!

packages/webapp/src/components/SensorReadingsLineChart/index.jsx
anant00729: Done

packages/webapp/src/containers/SensorReadings/saga.js
anant00729: resolved

packages/webapp/src/containers/SensorReadings/saga.js
anant00729: resolved

Cali0707 (approved):
This looks great!

jagDhillon077 (approved):
this worked for me

PR #2248, opened 2022-08-16
Various sensor fixes

Cali0707:
This PR resolves https://lite-farm.atlassian.net/browse/LF-2667, https://lite-farm.atlassian.net/browse/LF-2713, https://lite-farm.atlassian.net/browse/LF-2678, and https://lite-farm.atlassian.net/browse/LF-2725

To test:
1. Either use ngrok or comment out line 209 in `sensorController.js`
2. Take a valid sensor upload file
3. Duplicate one of the sensors (it should have the same brand and external_id as another sensor in the file)
4. Add an empty row after the existing rows (this works better in a raw text editor than in excel) - you can also put the empty rows between existing rows, but the frontend validation doesn't always allow this.
5. Upload this file - this should succeed! Only the first of the two duplicate sensors should be created on the farm, and it shouldn't complain about validation for the empty row (in the response from the backend - this PR doesn't change how it works on the frontend).
6. Look at a newly created sensor - it's depth should be the value from the upload file in cm.
7. Modify a sensor in the csv to have an External_Id longer than 40 characters and upload the file. You should get a validation error response from the API.

### DIFF
--- a/packages/api/src/controllers/sensorController.js
+++ b/packages/api/src/controllers/sensorController.js
@@ -81,7 +81,7 @@ const sensorController = {
         External_ID: {
           key: 'external_id',
           parseFunction: (val) => val.trim(),
-          validator: (val) => val.length <= 20,
+          validator: (val) => val.length <= 40,
           required: false,
           errorTranslationKey: sensorErrors.EXTERNAL_ID,
         },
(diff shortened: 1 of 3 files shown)


jagDhillon077 (changes requested):
This works correctly for me, just need to make some minor changes:

Update the translation in `packages/webapp/public/locales/en/translation.json` EXTERNAL_ID to 40 characters

Remove console log in line 122 of packages/webapp/src/components/Map/Modals/BulkSensorUploadModal/useValidateBulkSensorData.js

I noticed you commented out a console.table on line 115 of the same file. Maybe remove this if it is going to stay commented out.

Cali0707:
> This works correctly for me, just need to make some minor changes:
> 
> Update the translation in `packages/webapp/public/locales/en/translation.json` EXTERNAL_ID to 40 characters
> 
> Remove console log in line 122 of packages/webapp/src/components/Map/Modals/BulkSensorUploadModal/useValidateBulkSensorData.js
> 
> I noticed you commented out a console.table on line 115 of the same file. Maybe remove this if it is going to stay commented out.

@jagDhillon077 these should now be resolved

jagDhillon077 (approved):
👍🏽

fsayyid (approved):

PR #2252, opened 2022-08-17
LF-2528 compact temperature component type bug fix

seinnlee:
Read the latest comment by Mwaya from [LF-2528](https://lite-farm.atlassian.net/browse/LF-2528)

The issue was that sensor reading preview popup was being rendered for all sensors regardless of their reading types. If a sensor's reading type is "soil water content", the preview popup should not render compact temperature view. Currently, no matter what a sensor's reading types are, compact temperature view is visible.

Currently, reading types are added to the redux store (sensor individually) when a sensor detail's page is rendered. I tried to dispatch that action when a sensor's preview is activated, but that triggers re-rendering of the map component, which messes up the zoom level of the map. So I had to find a way to store reading types of all sensors of that farm when the map is rendered.

To Test:
1. Have multiple sensors on the map. (one sensor needs to have "temperature" as its reading types, other sensors need to exclude "temperature" as their reading types).
2. Long press the temperature sensor on the map. It should popup the preview component with "current soil temp".
3. Long press the other sensors on the map. It should not popup anything since we do not have any other compact reading components (i.e. soil water content, etc.)
4. You can try step 2 & 3 again after editing the sensors' reading types in their details page.

### DIFF
--- a/packages/api/src/routes/sensorRoute.js
+++ b/packages/api/src/routes/sensorRoute.js
@@ -45,6 +45,7 @@ router.get('/reading/farm/:farm_id', SensorController.getReadingsByFarmId);
 router.post('/reading/invalidate', SensorController.invalidateReadings);
 router.post('/unclaim', SensorController.retireSensor);
 router.get('/:location_id/reading_type', SensorController.getSensorReadingTypes);
+router.get('/:farm_id/reading_types', SensorController.getAllSensorReadingTypes);
 router.get('/partner/:partner_id/brand_name', SensorController.getBrandName);
 router.post('/reading/visualization', SensorController.getAllSensorReadingsByLocationIds);
 module.exports = router;
--- a/packages/api/src/routes/sensorRoute.js
+++ b/packages/api/src/routes/sensorRoute.js
@@ -45,7 +45,7 @@ router.get('/reading/farm/:farm_id', SensorController.getReadingsByFarmId);
 router.post('/reading/invalidate', SensorController.invalidateReadings);
 router.post('/unclaim', SensorController.retireSensor);
 router.get('/:location_id/reading_type', SensorController.getSensorReadingTypes);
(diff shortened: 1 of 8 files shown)


jagDhillon077 (approved):
This worked for me

fsayyid (approved):
This is working for me 👍

packages/api/src/routes/sensorRoute.js
```diff
@@ -45,6 +45,7 @@ router.get('/reading/farm/:farm_id', SensorController.getReadingsByFarmId);
 router.post('/reading/invalidate', SensorController.invalidateReadings);
 router.post('/unclaim', SensorController.retireSensor);
 router.get('/:location_id/reading_type', SensorController.getSensorReadingTypes);
+router.get('/:farm_id/reading_types', SensorController.getAllSensorReadingTypes);
```
fsayyid, line 48: @seinnlee one quick fix, can you make the route `/:farm_id/reading_type` singular to be consistent?

fsayyid (changes requested):
See above comment

packages/webapp/src/containers/Map/saga.js
```diff
@@ -207,11 +212,31 @@ export function* getSensorReadingsSaga() {
...
 }
 
+export const getAllSensorReadingTypes = createAction('getAllSensorReadingTypesSaga');
+
+export function* getAllSensorReadingTypesSaga() {
+  const { user_id, farm_id } = yield select(userFarmSelector);
+  const header = getHeader(user_id, farm_id);
+  try {
+    yield put(onLoadingSensorReadingTypesStart());
+    const result = yield call(axios.get, `${sensorUrl}/${farm_id}/reading_types`, header);
```
fsayyid, line 222: @seinnlee Should be updated here too

seinnlee: Thanks for catching that!

fsayyid (approved):

PR #2256, opened 2022-08-18
Rollback a transaction after failed add reading call

jagDhillon077:


Cali0707 (changes requested):
@jagDhillon077 can you also rollback the transaction for the delete sensor route while you're at it?

Cali0707 (approved):
Looks great!

fsayyid (approved):

PR #2257, opened 2022-08-19
Lf 2681 language support for sensor csv format back end

fsayyid:
Ticket: [LF-2681](https://lite-farm.atlassian.net/browse/LF-2681)

**Note:** the files with actual changes are `sensorController.js`, `sensorCSV.js`, and `csv.js`

To test: 
- Upload sensors in different languages (based on users current language preference) using a tool like Insomnia or Postman
- Try using a csv with reading types that would fail


Cali0707 (changes requested):
@fsayyid it seems that when I upload a sensor with an english reading_type when it is in spanish, the sensor still passes validation.

packages/api/src/util/sensorCSV.js
```diff
@@ -0,0 +1,278 @@
...
+  SENSOR_DEPTH: 'FARM_MAP.BULK_UPLOAD_SENSORS.VALIDATION.SENSOR_DEPTH',
+  SENSOR_BRAND: 'FARM_MAP.BULK_UPLOAD_SENSORS.VALIDATION.SENSOR_BRAND',
+  SENSOR_MODEL: 'FARM_MAP.BULK_UPLOAD_SENSORS.VALIDATION.SENSOR_MODEL',
+  SENSOR_HARDWARE_VERSION: 'FARM_MAP.BULK_UPLOAD_SENSORS.VALIDATION.SENSOR_HARDWARE_VERSION',
+  SENSOR_ALREADY_OCCUPIED: 'FARM_MAP.BULK_UPLOAD_SENSORS.SENSOR_CLAIM_ERROR.ALREADY_OCCUPIED',
+  SENSOR_DOES_NOT_EXIST: 'FARM_MAP.BULK_UPLOAD_SENSORS.SENSOR_CLAIM_ERROR.DOES_NOT_EXIST',
+  INTERNAL_ERROR: 'FARM_MAP.BULK_UPLOAD_SENSORS.SENSOR_CLAIM_ERROR.INTERNAL_ERROR',
+};
+
+const sensorCsvValidators = [
```
Cali0707, line 34: I still think that this should probably be passed into the function which parses the csv, instead of being the always called. There may be a scenario where LiteFarm uses a csv upload for something else now that it is already being done in one place, so I think there is a lot of value to making this function more reusable.

Cali0707: Maybe if we have a parseCsv function, and then a parseSensorCsv function which acts as a wrapper to the original function and uses this set of validators.

fsayyid: Sure

Cali0707 (approved):
Looks good to me!

Cali0707:
@fsayyid if this is merged before #2263 can you give me a heads up? This will be a breaking change for that PR

jagDhillon077 (approved):

PR #2258, opened 2022-08-19
Added notification resource refresh with docs

Cali0707:
This PR makes it so that the resources associated with a notification get refreshed when the user goes to the notifications page. To test:
1. Create at least one task on your farm
2. Do an action that would generate a task notification (e.g. assign the task to yourself)
3. Open the network tab in devtools and navigate to the notifications page
4. There should be a request made once you are on the notifications page to get all the tasks for the farm


fsayyid (approved):
This looks good, it's working on my end.

PR #2259, opened 2022-08-19
LF-2562: weather board multi language date support

fsayyid:
Ticket: [LF-2562](https://lite-farm.atlassian.net/browse/LF-2562)

To test: 
- Go to the home page with the weather board
- Switch languages
- Go back to the home page and see that the date is in that language


seinnlee (approved):
It looks good to me! @Cali0707 should review this PR since you can verify at least if the French date is correctly displayed.

Cali0707 (changes requested):
This works great! Could you just remove all the imports from `moment` and `moment/locale/xx` as they don't seem to be necessary anymore?

Cali0707 (approved):
LGTM

PR #2260, opened 2022-08-19
LF-2642 crop plan based pop up for adding a location

seinnlee:
Ticket [LF-2642](https://lite-farm.atlassian.net/browse/LF-2642)

To Test:
1. Create a new farm without adding any locations on the map.
2. Create a new crop plan.
3. When entering select location page, there should be a popup.
4. Clicking "go back" will return the user to crop plan creation flow.
5. Clicking "create location" will redirect the user to the farm map with "+" tab expanded.

### DIFF
--- a/packages/webapp/public/locales/en/translation.json
+++ b/packages/webapp/public/locales/en/translation.json
@@ -1105,6 +1105,11 @@
     "WHERE_TRANSPLANT_LOCATION": "Where will you transplant to?",
     "WILD_CROP": "Are you harvesting a wild crop?"
   },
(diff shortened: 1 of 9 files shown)


fsayyid (changes requested):
After I create a location, the map drawer is still showing. I believe we should see the map after creating a location.

seinnlee:
@fsayyid I fixed the issue you described!

fsayyid (approved):
This is looks good 👍

Cali0707 (approved):
Looks good to me, everything seems to be working as expected

PR #2261, opened 2022-08-19
LF-2592: hyphenate long crop names on crop catalogue

fsayyid:
Ticket: [LF-2592](https://lite-farm.atlassian.net/browse/LF-2592)

To test: 
- Go to the crop catalogue
- Add a crop with a long name
- See that the crop name breaks and hyphenates on the crop tile

### DIFF
--- a/packages/webapp/src/components/CropTile/styles.module.scss
+++ b/packages/webapp/src/components/CropTile/styles.module.scss
@@ -63,6 +63,10 @@
   font-weight: 600;
   font-size: 12px;
   line-height: 16px;
+  -webkit-hyphens: auto;
+  -moz-hyphens: auto;
+  hyphens: auto;
+  overflow-wrap: break-word;
 }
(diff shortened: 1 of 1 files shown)


seinnlee (changes requested):
It looks like on mobile, the long crop name on edit page is overflowing out of the screen. Can you also hyphenate in the edit crop plan component?

Also, shouldn't we hyphenate long crop names or variety names in the crop management page? Currently, it is using ellipsis (...) when viewed on mobile.

fsayyid:
@seinnlee the ellipses requirement for long crop names on the crop management page is from a different ticket, and I think it makes more sense in that case. https://lite-farm.atlassian.net/browse/F21-599

seinnlee (approved):
it lgtm!

Cali0707 (approved):
LGTM!

PR #2262, opened 2022-08-19
LF-2567: various issues on farm info page

fsayyid:
Ticket: [LF-2567](https://lite-farm.atlassian.net/browse/LF-2567)

To test: 
- Go to farm info page
- Enter a farm name longer than 255 characters
- Enter a phone number longer than 20 characters
- Should get inline error for both


Cali0707 (approved):
As the requirements for the phone number characters are out of scope, this looks good to me!

jagDhillon077 (approved):

PR #2264, opened 2022-08-19
Hotfix: sensor reading type route failing

seinnlee:
This PR fixes the issue where `sensor/:farm_id/reading_type` endpoint was not returning anything.

### DIFF
--- a/packages/api/src/routes/sensorRoute.js
+++ b/packages/api/src/routes/sensorRoute.js
@@ -45,7 +45,7 @@ router.get('/reading/farm/:farm_id', SensorController.getReadingsByFarmId);
 router.post('/reading/invalidate', SensorController.invalidateReadings);
 router.post('/unclaim', SensorController.retireSensor);
 router.get('/:location_id/reading_type', SensorController.getSensorReadingTypes);
-router.get('/:farm_id/reading_type', SensorController.getAllSensorReadingTypes);
+router.get('/farm/:farm_id/reading_type', SensorController.getAllSensorReadingTypes);
 router.get('/partner/:partner_id/brand_name', SensorController.getBrandName);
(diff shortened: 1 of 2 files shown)


packages/api/src/routes/sensorRoute.js
```diff
@@ -44,8 +44,8 @@ router.get('/:location_id/reading', SensorController.getAllReadingsByLocationId)
 router.get('/reading/farm/:farm_id', SensorController.getReadingsByFarmId);
 router.post('/reading/invalidate', SensorController.invalidateReadings);
 router.post('/unclaim', SensorController.retireSensor);
-router.get('/:location_id/reading_type', SensorController.getSensorReadingTypes);
 router.get('/:farm_id/reading_type', SensorController.getAllSensorReadingTypes);
+router.get('/:location_id/reading_type', SensorController.getSensorReadingTypes);
```
fsayyid, line 48: The issue is that we can't have two endpoints with the format `:id/reading_type` express isn't able to tell the difference. Please update to `/farm/:farm_id/reading_type` on both frontend and backend.

fsayyid (changes requested):

Cali0707:
@seinnlee if I'm understanding this properly, this PR just changes which endpoint will fail, correct?

packages/api/src/routes/sensorRoute.js
seinnlee: That makes sense. I updated the endpoint both front and backend.

seinnlee:
@Cali0707 Per @fsayyid's comment, I changed the endpoint so it's different from the other reading_type endpoint.

Cali0707 (approved):
LGTM!

fsayyid (approved):

PR #2265, opened 2022-08-22
Lf 2689 write cypress test for user invitation

MwayaB:


anant00729:
@Cali0707 checked the user invite flow. It looks good to me.

anant00729 (approved):

Cali0707 (commented):
@MwayaB I'm getting the following error when running the cypress tests. Any ideas?
```
Can't run because no spec files were found.

We searched for specs inside of this folder:

  > LiteFarm/packages/webapp/cypress/integration
```

Cali0707:
@MwayaB when I ran `pnpm install` I was able to get the correct version of cypress and the tests ran, however the `inviteUserFlow` spec seems to be failing. Is it supposed to fail currently?

PR #2267, opened 2022-08-23
LF-2658 task based pop up for adding a location

seinnlee:
Ticket [LF-2658](https://lite-farm.atlassian.net/browse/LF-2658)
Ticket [LF-2642](https://lite-farm.atlassian.net/browse/LF-2642)

To Test (for tasks):
1. Have a farm with no locations created.
2. Create a new task and there should be a popup.
4. Clicking "go back" will return the user to task creation flow.
5. Clicking "create location" will redirect the user to the farm map with "+" tab expanded.

To Test (for crops):
1. Create a new farm without adding any locations on the map.
2. Create a new crop plan and there should be a popup.
4. Clicking "go back" will return the user to crop plan creation flow.
5. Clicking "create location" will redirect the user to the farm map with "+" tab expanded.

Also, it is now refactored to use its own redux store instead of adding another value to the existing map filter store.


anant00729 (approved):
Worked for me.

jagDhillon077 (approved):
This works for me

PR #2268, opened 2022-08-23
Lf 2695 reconfigure nodemon for api package to support shared code

Cali0707:
This PR resolves https://lite-farm.atlassian.net/browse/LF-2695

There are two parts to this PR:
1. Is the developer tooling working?
2. Is the refactor of the sensor validation working?

To test part 1 of this PR:
1. Run both the webapp and the api
2. Make a change in the /shared package. The api should reload automatically, and if you are on a page which uses code from /shared in the webapp, that page should reload automatically as well (easiest way to verify this is to open the upload sensors modal and then modify something in the `parseSensorCsv` function in `shared/validation/sensorCSV.js`)

To test part 2 of this PR:
1. Try to upload an invalid csv file in english - you should get reasonable errors
2. Try again in another language - you should get reasonable errors
3. Try to upload a valid csv file in english - this should work (if the sensors are unclaimed)
4. Try to upload a valid csv file in another language - this should work (if the sensors are unclaimed)


fsayyid (approved):
This seems to be working on my end

anant00729 (changes requested):
Tested. it's not working on my end. Please check the screenshot
Screen Shot 2022-08-24 at 9 10 05 AM


Please use the following zip to test:
[test-sensors-multilanguage.zip](https://github.com/LiteFarmOrg/LiteFarm/files/9417753/test-sensors-multilanguage.zip)

fsayyid (changes requested):
When I upload an invalid csv, I'm not getting the text file on the frontend.

fsayyid:
Actually, I have only been able to create that issue once, and I think it was just because I manually messed with the database on my end earlier to change the user language. It seems to be working now that I set it back.

fsayyid (approved):

anant00729:
Test all the following scenarios. It looks good.

1. Empty file
2. Missing columns in different languages
3. Incorrect values
4. Too long of a file

anant00729 (approved):

PR #2269, opened 2022-08-24
LF-2566: various issues on account info page

fsayyid:
Ticket: [LF-2566](https://lite-farm.atlassian.net/browse/LF-2566)

To test:
- Input a first name and last name with a leading and trailing spaces. Save the name and refresh. Leading and trailing spaces should be removed.
- There should be an (optional) label next to the fields "Phone Number" and "User Address"
- Input a phone number greater than 20 characters in length. This should result in an inline error.

### DIFF
--- a/packages/webapp/public/locales/en/translation.json
+++ b/packages/webapp/public/locales/en/translation.json
@@ -1248,6 +1248,12 @@
       "USER_ADDRESS": "User Address"
     },
     "ACCOUNT_TAB": "Account",
+    "ERROR": {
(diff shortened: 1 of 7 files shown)


seinnlee (approved):
It looks good to me. It's not in scope of this ticket, but can we also have an in-line error for invite a user page? ("Full name" input does not show any in-line error when over 255 characters)

fsayyid:
@seinnlee I can do this in another PR since it's out of scope of the ticket

jagDhillon077 (approved):

PR #2270, opened 2022-08-24
Lf 2500/reading tab impl

anant00729:
To Test: 

1. update the sensor reading types.
2. check if the reading types are to be displayed on the "Readings" Tab page.

To Test:
1. Click on the sensor in the farm maps. check if it redirects to the "Readings" Tab page.

For more details:
https://lite-farm.atlassian.net/browse/LF-2500


fsayyid (approved):
This is working and looks good to me. There currently is no design criteria for these boxes, so this will work until there is a design.

Cali0707 (approved):
This is working well for me too

jagDhillon077 (changes requested):
Add react fragment instead of returning null

jagDhillon077 (approved):
LGTM

PR #2273, opened 2022-08-24
fix: order of validation for empty file error vs other errors

fsayyid:
Just a switch to the order of validation for header vs empty file validation

To test:
- Upload a file in a language different to the user's preference using Postman, Insomnia etc
- You should get header errors rather then empty file error


Cali0707 (approved):
LGTM!

anant00729:
Looks good to me

anant00729 (approved):

PR #2274, opened 2022-08-24
fix: invite user form name length validation

fsayyid:
To test:
- Go to invite a new user to your farm
- Enter a name with a first name and/or last name greater than 255 characters
- You should get an appropriate inline error


seinnlee (approved):
lgtm!

jagDhillon077 (approved):

PR #2277, opened 2022-08-25
fix: account info page user address char limit

fsayyid:
To test:
- Go to my info page
- Enter user address longer than 255 characters
- You should get an inline error


seinnlee (approved):
LGTM!

anant00729:
Looks good to me.

anant00729 (approved):

PR #2279, opened 2022-08-25
fix: change longpress to 0.5 seconds

seinnlee:
David wanted the long press to be shorter than what it currently is (0.8 seconds).

To Test:
1. Have sensors on the farm map.
2. Long press a sensor and the preview should pop up after holding the mouse down for 0.5 seconds.


Cali0707 (approved):
LGTM!

fsayyid (approved):
looks good

PR #2281, opened 2022-08-25
Retriveing the correct coordinates of a sensors using sensor query

anant00729:
To test,

1. create 3 - 4 sensors at different locations. 
2. check the station name in the readings tab and verify if they belong to the same location.

### DIFF
--- a/packages/api/src/models/sensorModel.js
+++ b/packages/api/src/models/sensorModel.js
@@ -129,23 +129,40 @@ class Sensor extends Model {
   static async getSensorLocationByLocationIds(locationIds = []) {
     return await knex.raw(
       `WITH filtered_sensors_info as (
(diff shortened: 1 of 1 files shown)


seinnlee (approved):
It worked for me. When I changed the sensor location, it updated the ambient temperature station correctly.

jagDhillon077 (commented):
What would you do in the case that there is no weather station nearby, shows no weather station if I input coordinates of a remote location.
Screen Shot 2022-08-25 at 3 40 18 PM


Coordinated I used: 75, -112

jagDhillon077 (approved):

anant00729:
I have created a separate bug ticket. https://lite-farm.atlassian.net/browse/LF-2754

PR #2283, opened 2022-08-26
Added some FR translations

Cali0707:
This PR adds translations for some of the missing french language tags.


jagDhillon077 (approved):
According to my 0 experience with french these translations are correct

seinnlee (approved):
also approving with no knowledge of french

PR #2300, opened 2022-09-16
singleValue for units should be black not grey

Duncan-Brain:
I returned the font color to the value set in `ReactSelect` referred to as default styles in `Unit`. Seems to work in views mentioned in Jira for: area creation, task creation, and crop plan creation.


anant00729:
Looks good to me!

anant00729 (approved):

kikemarquez8 (approved):

PR #2307, opened 2022-10-06
Fixed the null and undefined handling of array.length inside the render function.

anant00729:
To test: 

1. go to http://localhost:3000/Finances from the sidebar.
2. select the "Add New Expense" option.
3. this will take you to http://localhost:3000/expense_categories. 
4. check if you can find the expense types list.
5. logout from the app.
6. login again.
7. from the home page, try a direct redirect to http://localhost:3000/expense_categories

check if you can find any errors in the browser logs.

for more details: 
https://lite-farm.atlassian.net/browse/LF-2767

### DIFF
--- a/packages/webapp/src/containers/Finances/NewExpense/ExpenseCategories/index.jsx
+++ b/packages/webapp/src/containers/Finances/NewExpense/ExpenseCategories/index.jsx
@@ -90,50 +90,49 @@ class ExpenseCategories extends Component {
             width: '100%',
           }}
         >
-          {expenseTypes?.length > 0 &&
-            expenseTypes
-              .sort((firstExpenseType, secondExpenseType) => {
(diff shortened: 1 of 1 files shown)


Duncan-Brain:
From what I can tell this is not observable on the front end and is mostly housekeeping? My only comment is that I notice that PropTypes and prototype is not defined here like other components which could help with debugging in the future. If the direct mapping with redux doesn't interfere PropTypes it could be good to add it?

anant00729:
@Duncan-Brain Thanks for pointing that out. As you suggested, I have added the Prop types validation to ExpenseCategories page. I found that expense types were supplied as the array from the redux store, and that's the only data container and prop types coming in from the redux store.

anant00729:
updated the PR description to test out the scenario.

Duncan-Brain (approved):
Looks good to me!

wasswarichard (approved):

PR #2309, opened 2022-10-18
Lf 2796 lf 2795 irrigation task icon display

wasswarichard:
Display the irrigation task during task creation


Duncan-Brain:
@wasswarichard 

1. I am getting an error on the frontend regarding the taskComponents constant in <PureTaskDetails />. I think this is because the Irrigation Task Component is part of another job for this sprint. Without that component in place it doesn't make sense to me to push this ticket as it would be a broken component on it's own.

2. The package-lock changes.. I am guessing you are having the same issue as me where you have to change it to get the dev entrypoint to work. I do think that should be an Jira issue, but separate from this PR as it would need to be checked on other operating systems/setups.

3. Since we are changing the /en/ locale do the other locales also need to be changed?

wasswarichard:
> @wasswarichard
> 
> 1. I am getting an error on the frontend regarding the taskComponents constant in `<PureTaskDetails />`. I think this is because the Irrigation Task Component is part of another job for this sprint. Without that component in place it doesn't make sense to me to merge this ticket as it would be a broken component on its own.
> 2. The package-lock changes.. I am guessing you are having the same issue as me where you have to change it to get the dev entrypoint to work. I do think that should be an Jira issue, but separate from this PR as it would need to be checked on other operating systems/setups.
> 3. Since we are changing the /en/ locale do the other locales also need to be changed?

wasswarichard:
@Duncan-Brain for issue 1,  the <PureTaskDetails /> is not yet worked on,  for testing purposes the error has been fixed with a string.
For issue 2, the changes have been removed. I think this is an issue @kcussen has to give priority since running the project fails on some machines until  the run script is updated to "dev": "set CYPRESS_COVERAGE=TRUE && vite",
For issue 3 we are only considering /en/ locale for now.

Duncan-Brain:
1. My personal opinion is that I think I would ask @kcussen if it makes sense to pause this PR and finish `<PureIrrigationTask />` first before merging this in. The string does fix that error but it is still a dead-end task until that is complete because it will fail to create the task. If for whatever reason or accident/etc it gets merged into master it would be a big bug.

2. Great! Yeah I agree it is mildly annoying haha.  I figured it was just because I am on Windows but it works fine on Mac. If you make a Jira ticket for it I will test out your solution for that too!

3. Looks good now to me!

wasswarichard:
@Duncan-Brain we are using a trunk-based development approach with tasks and stories, tasks and stories can be merged in the trunk (integration branch) in case they are testable. The feature can't be shipped to production until the epic is completed.  The aim of the task or story is not to finish the epic but to have something testable.

Duncan-Brain:
@wasswarichard Okay thanks for clarifying that!

Duncan-Brain (approved):

kcussen:
Regarding #1 it's preferable to merge it in early even if it introduces a "loose end". From a functional perspective, seeing the icon on the task screen is the definition of done for this story. The owner of the "Select a date" view (also @wasswarichard) would then be responsible for connecting that screen based on the click of the irrigation icon. Let me know if that makes sense. 

I don't follow #2

anant00729:
This looks good to me.

anant00729 (approved):

PR #2312, opened 2022-10-21
Lf 2454 - revert code for sensor readings

anant00729:
This PR is made to revert the code related to the add Readings API of sensors. 

Please review the code and match it with the previous version. 

@Duncan-Brain and @wasswarichard


wasswarichard (approved):

kcussen (approved):
Approving based on the fact that we're reverting to a previous version to stabilize beta.

PR #2315, opened 2022-10-26
LF-2800 Irrigation Tasks

wasswarichard:
LF-2800 Irrigation Tasks


anant00729:
@wasswarichard your PR is 25 commits behind integration. Please take a sync

packages/webapp/src/components/Modals/WaterUseCalculatorModal/index.tsx
```diff
@@ -0,0 +1,35 @@
+import React, { FC } from 'react';
+import { useTranslation } from 'react-i18next';
+import ModalComponent from '../ModalComponent/v2';
+import { ReactComponent as Person } from '../../../assets/images/task/Person.svg';
+import styles from '../QuickAssignModal/styles.module.scss';
+import Button from '../../Form/Button';
+
+export interface IWaterUseCalculatorModal {
+  dismissModal: () => void;
+  // setTotalWaterUsage: () => number,
```
anant00729, line 10: you can remove this

packages/webapp/src/components/Task/PureIrrigationTask/index.tsx
```diff
@@ -0,0 +1,275 @@
...
+  };
+}
+export interface IPureIrrigationTask {
+  handleGoBack: () => void;
+  handleContinue: () => void;
+  system: ISystem;
+}
+
+const PureIrrigationTask: FC<IPureIrrigationTask> = ({
+  handleGoBack,
```
anant00729, line 38: Please add propTypes for validation

packages/webapp/src/components/Modals/WaterUseCalculatorModal/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+}
+
+const WaterUseCalculatorModal: FC<IWaterUseCalculatorModal> = ({ dismissModal }) => {
```
anant00729, line 13: Add propTypes

anant00729:
After, testing the irrigation flow. the end result in the persisted form state is as follows: 

Screen Shot 2022-11-02 at 5 32 22 PM

anant00729 (changes requested):

packages/webapp/src/components/Modals/WaterUseCalculatorModal/index.tsx
wasswarichard: @anant00729 with typescript, propTypes aren't required. Both typescript types/interfaces work the same as propTypes.

packages/webapp/src/components/Task/PureIrrigationTask/index.tsx
wasswarichard: @anant00729 with typescript, propTypes aren't required. Both typescript types/interfaces work the same as propTypes.  Typescript does validation for type checking.

packages/webapp/src/components/Modals/WaterUseCalculatorModal/index.tsx
wasswarichard: @anant00729 this has been removed.

packages/webapp/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "3.2.5",
   "description": "LiteFarm Web application",
   "scripts": {
-    "dev": "CYPRESS_COVERAGE=TRUE  vite",
+    "dev": "set CYPRESS_COVERAGE=TRUE && vite",
```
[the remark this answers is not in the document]
wasswarichard: @anant00729 this has been reverted.

wasswarichard:
@anant00729 Do you suggest any change to the persisted data, The branch is now in sync with the integration branch

packages/webapp/src/components/Modals/WaterUseCalculatorModal/index.tsx
anant00729: got it. thanks!

packages/webapp/src/components/Task/PureIrrigationTask/index.tsx
anant00729: got it

packages/webapp/src/components/Modals/WaterUseCalculatorModal/index.tsx
anant00729: great

packages/webapp/package.json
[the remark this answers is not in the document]
anant00729: nice

anant00729:
@wasswarichard to me it looks good. the structure that we discussed over the call looks the same. @Duncan-Brain Please review this PR.

anant00729 (approved):

packages/webapp/src/components/Modals/WaterUseCalculatorModal/index.tsx
```diff
@@ -0,0 +1,34 @@
...
+import Button from '../../Form/Button';
+
+export interface IWaterUseCalculatorModal {
+  dismissModal: () => void;
+}
+
+const WaterUseCalculatorModal: FC<IWaterUseCalculatorModal> = ({ dismissModal }) => {
+  const { t } = useTranslation();
+  return (
+    <ModalComponent>
```
Duncan-Brain, line 15: @wasswarichard I think this tag is meant to be self-closing, the attributes are showing up as text in the modal currently.

Duncan-Brain:
Hey @wasswarichard Big PR! The UI seems correct and working for the most part. My questions below:

1. The `Irrigate.svg`... is this part of the other [PR-2309](https://github.com/LiteFarmOrg/LiteFarm/pull/2309)? or is `Irrigation.svg` also a different image? I am seeing a few references to these in the existing code and wondering if there should be just one.  

2. In `<PureIrrigationTask />`  you are defining the system interface for metric and imperial, does this belong here or in the `<Unit />` component?

3. In `<PureIrrigationTask />` I am seeing 11 non-breaking style errors in VSCODE about several imports, and also the event(e) on line 162 as. I think it is related to the tsconfig? They all look like the following:
    > Parameter 'e' implicitly has an 'any' type. 

4. In `<PureIrrigationTask />` Should line 107-112 be put into a constants.js file? I see Anant do this regularly.

5. I am still learning how things work and connect to each other here so I don't know the answer to this but: Should `container/Task/IrrigationTask` line up more with Methods like `container/Task/TaskTransplantMethod`? In components the Methods are then used by the Tasks.

6. The Modal - I tagged you inline of the `<WaterUseCalculatorModal />` file change

7. I am getting a browser console Warning about changing an uncontrolled to controlled component when I select the checkbox, I think the other base components don't use controlled components.

8. Is React FC the best way to make a functional component in Typescript? I have never used it before so when I googled React FC this comes up: https://stackoverflow.com/questions/59988667/typescript-react-fcprops-confusion 

9. TypeScript. I think you are right that TS is better. But none of the other code uses TS. I  think it will be hard to easily compare and test between similar Task components.  It seems to me like a decent sized architecture change and I would be happy work towards an epic of "convert all code to TS".  For me it is all or nothing on JS style haha. Has @kcussen made a decision on this yet after doing more research? I won't feel comfortable approving until that architecture decision is made.

wasswarichard:
@Duncan-Brain 
issue 1, a different ticket https://lite-farm.atlassian.net/browse/LF-2832 was created to handle that change 
Issue 2, the unit component is a jsx component, not a TSX component therefore interfaces or types can't be defined in a non-typescript component 
issue 3, Your editor doesn't have ts plugins otherwise everything should be okay.
Issue 4, No, these are component-specific constants, not project reusable constants.
issue 5, No, this is a different task component, not a transplant task
issue 6, Has been replied to and resolved
issue 7, Has been resolved.
Issue 8, React FC provides more type-checking and autocomplete 
Issue 9, we already have typescript components in the project, I reviewed the project and it has support for Typescript. example of components with typescript 
image
Testing for components remains the same and with typescript, it gets easier due to type checking, Typescript is good for a large project and solves development errors and bugs. 

Lastly please read more about typescript.

Duncan-Brain:
Looks good to me, almost ready to approve! 

Though I will say that Issue 2 - where we are defining another components properties is a good example of how things will get messy by mixing TypeScript with plain JavaScript. 

@wasswarichard -- If Kevin approves the TS. I will also try to code in it from now on and convert as many existing components as I can, because as you said it it is better than React PropTypes for type checking. I am just saying that the codebase should try to be consistent one way or another.

wasswarichard:
@Duncan-Brain I removed the typescript, please review.

Duncan-Brain (commented):
@wasswarichard - it is working for me.

@anant00729 mentioned type checking with PropTypes for at least the new files (WaterUseCalculator, IrrigationTask and PureIrrigationTask) as we slowly transition the codebase to include type checking.

wasswarichard:
@Duncan-Brain PropTypes have been included

Duncan-Brain (approved):

Duncan-Brain:
Looks good to me!

PR #2316, opened 2022-10-26
Fix : on modal outside click the user is not taken to the next step

anant00729:
To Test:

1. Click outside of the modal and check if the user can able to move to the next step or not.
2. If the modal is dismissed and the user stays on the very same page, this means the bug is fixed.


Duncan-Brain (approved):
works for me

wasswarichard (approved):

PR #2319, opened 2022-10-30
Lf 2688 cypress tests for tasks

MwayaB:


Duncan-Brain:
Sorry for the lots of comments but its a big PR! haha

1. Should this be .gitignored?  `packages/webapp/.pnpm-debug.log`
2. Is `cypress.json` in the a) right place and b) necessary?
3. Should spherical geometry be a dev dependency? .. in `packages/webapp/package.json` 
4. Did eslint autoformat? It could be good in the future to do that as a separate PR
5. It could be useful to add a `package.json` command like` "test":"cypress run --config video=false [options]"` or `"cy:run":"cypress run --config video=false [options]"`
6. @wasswarichard mentioned it elsewhere but as it is written currently the dev command does not work for some environments. Maybe we can change it here if it works for others : `package.json` command "dev": "set CYPRESS_COVERAGE=TRUE && vite"
7. Tests failed for me because the API was hardcoded at localhost:5000, can we change this to get the process.env.VITE_API_URL var ?
8. Local testing errors as follows:
- Crops flow tests: `Expected to find content: 'What is your role on the farm' but never did.`
- LiteFarm end to end tests: `Expected to find content: 'What is your role on the farm' but never did.`
- Happy Path test: ``cy.type()` failed because this element is not visible `<input class="_input_ugk5b_1" aria-invalid="false" type="text" name="name" placeholder="e.g. Juan Perez" size="1" data-cy="createUser-fullName" style="padding-right: 4px;">`This element `<input._input_ugk5b_1>` is not visible because its parent `<div>` has CSS property: `display: none` Fix this problem, or use `{force: true}` to disable error checking.`
- Insights Flow test:` `cy.type()` failed because this element is not visible:`<input class="_input_ugk5b_1" aria-invalid="false" type="password" name="password" size="1" data-cy="enterPassword-password" id="password_input_to_focus" style="padding-right: 4px;">`This element `<input#password_input_to_focus._input_ugk5b_1>` is not visible because its parent `<div>` has CSS property: `display: none` Fix this problem, or use `{force: true}` to disable error checking.'
- invite users test: `Expected to find content: 'What is your role on the farm' but never did.`
- (Possibly I dont have this .env token?) Login flow tests : ` SSO Google Account:  CypressError: `cy.request()` failed on: https://www.googleapis.com/oauth2/v4/token The response we received from your web server was:  > 400: Bad Request This was considered a failure because the status code was not `2xx` or `3xx`. If you do not want status codes to cause failures pass the option: `failOnStatusCode: false``
- Notification flow: `cy.type()` failed because this element is not visible: `<input class="_input_ugk5b_1" aria-invalid="false" type="password" name="password" size="1" data-cy="enterPassword-password" id="password_input_to_focus" style="padding-right: 4px;">`
- Tasks flow tests: `cy.type()` failed because this element is not visible: `<input class="_input_ugk5b_1" aria-invalid="false" type="text" name="name" placeholder="e.g. Juan Perez" size="1" data-cy="createUser-fullName" style="padding-right: 4px;">`

MwayaB:
@Duncan-Brain I'll look into changes 1 through 7 and push a commit later today. As far as errors in the other test scripts alot of those are work in progress and need to be refactored to run correctly in CI, which isn't in the scope of the current PR

Duncan-Brain:
> As far as errors in the other test scripts alot of those are work in progress and need to be refactored to run correctly in CI, which isn't in the scope of the current PR

@maway It was local testing that the errors showed up on. I was not able to test these in CI I will just have to take the success based on the Actions succeed result and your cypress video. If the tasks tests are the only thing in the PR I would remove the other WIP for now then if it were me.

Duncan-Brain:
@MwayaB so what I am getting with your new commits is:
1,2,3 - taken care of
4,7 - maybe next time
5,6,8 - This is about happy path on CI -- not local testing, local testing may or may not work -- so maybe next time

So the only thing I have for this latest commit is that there is a new file added package-lock.json that I am not sure should be there.

Duncan-Brain (approved):
Looks like it is working on ci!

wasswarichard:
Looks working fine after review with @MwayaB

anant00729:
The cypress test seems to be working fine. Also, the cypress API links the test cases with github actions. the cypress dashboard shows me the happy path running which matches with the github actions. Here are the screenshots:

Screen Shot 2022-11-03 at 9 55 29 AM
Screen Shot 2022-11-03 at 9 55 54 AM

anant00729 (approved):

PR #2323, opened 2022-10-31
Fixing issues where units were not being displayed in the disabled units

kikemarquez8:
Fixes for 
https://lite-farm.atlassian.net/browse/LF-2792
and 
https://lite-farm.atlassian.net/browse/LF-2791

image
image


Duncan-Brain:
@kikemarquez8 

1. In LF-2791 it looks like @dtrapplitefarm wanted the Length added to the Buffer Zone creation flow and read only view. I don't know if this was discussed off of Jira, but to me it makes sense to keep it out of creation flow and only have it in the read only view (copying the watercourse).

2. I think something deeper is going on with the CSS, I have noticed a slight background-color difference between the units you fixed (light grey) and the ones that were working fine (white). I can't quite find the right place to fix it but I am not confident that the zIndex is the problem there.

anant00729:
@kikemarquez8 could you please discard the package-lock.json file

wasswarichard:
@kikemarquez8  this looks fine to me

wasswarichard (approved):

wasswarichard (approved):

Duncan-Brain:
@kikemarquez8 

Since some of the units are working correctly while others are not -- to me that's a clue it is not a z-index problem. Tough problem to find though! I've narrowed it down to line 375 `isSelectDisabled && disabled && styles.disableBackground` of `<Unit />`, basically the style .disableBackground is being applied to the units that are missing. And it is not being applied to the ones showing correctly. Something wrong with the logical expression there.

And also @anant00729 mentioned the package-lock changes should be removed.

kikemarquez8:
> @kikemarquez8
> 
> Since some of the units are working correctly while others are not -- to me that's a clue it is not a z-index problem. Tough problem to find though! I've narrowed it down to line 375 `isSelectDisabled && disabled && styles.disableBackground` of `<Unit />`, basically the style .disableBackground is being applied to the units that are missing. And it is not being applied to the ones showing correctly. Something wrong with the logical expression there.
> 
> And also @anant00729 mentioned the package-lock changes should be removed.

Hey Duncan! Thanks for taking the time, I'll re-review then, thanks for the input

@anant00729 Why do you want the package-lock.json removed? It seems to me that we should have it, it holds an exact reference of the dependencies the project should rely upon. I've read everywhere committing it is important. I'll share an example article
https://dev.to/adamklein/package-lock-json-in-git-or-not-50l5

Duncan-Brain:
@kikemarquez8 

My understanding about the `package-lock.json` is that it should only be included in a PR if a dependency has changed. And since there are no dependency changes that any addition of it would belong in a separate issue.

I think there is actually a lock file `pnpm-lock.yaml` in the project folder already. 

But generally speaking yeah my understanding is there should be a lockfile in the project. I might have also read something about having two different lockfiles may cause conflicts of some kind... but I don't know the details of that.

kikemarquez8:
> @kikemarquez8
> 
> My understanding about the `package-lock.json` is that it should only be included in a PR if a dependency has changed. And since there are no dependency changes that any addition of it would belong in a separate issue.
> 
> I think there is actually a lock file `pnpm-lock.yaml` in the project folder already.
> 
> But generally speaking yeah my understanding is there should be a lockfile in the project. I might have also read something about having two different lockfiles may cause conflicts of some kind... but I don't know the details of that.

@Duncan-Brain I just noticed the pnpm lock, I removed the package-lock now. 
Also I reviewed what you said, as you said there was a weird background change that was caused by those 2conditions (isSelectDisabled && disabled) being required to apply the "disabledBackground" stlyes. I removed the isSelectDisabled condition so now all disabled inputs look right, if they are of type select or not. I still found the issue to be that "div" stepping over the input's div.

Duncan-Brain:
@kikemarquez8 

Your solution is visibly working. But I am still noticing a few non-visible styling issues that this component has relating to this problem. Overall I am happy to approve this if you decide that's best! 

Since I got addicted to the problem I have also found a solution without z-index that fixes a few non-visible issues to check out if you're interested. 

I've tagged you on a [PR on this branch](https://github.com/LiteFarmOrg/LiteFarm/pull/2335) for you to view it.

These are the non-visible issues:
- input container blocking the units multi-select
image
- mismatch between pseudo-container sizes and combined input and units size.
image
image
- future z-index problems

anant00729 (approved):

PR #2328, opened 2022-11-09
Lf 2805 migrate irrigation table

wasswarichard:
The pull request consists of https://lite-farm.atlassian.net/browse/LF-2800 https://lite-farm.atlassian.net/browse/LF-2801 https://lite-farm.atlassian.net/browse/LF-2805 @Duncan-Brain  @anant00729  please review.


packages/api/db/migration/20221118081755_alter_irrigation_task_table.js
```diff
@@ -0,0 +1,23 @@
...
+      table.string('created_by_user_id').references('user_id').inTable('users').defaultTo('1');
+      table.string('updated_by_user_id').references('user_id').inTable('users').defaultTo('1');
+      table.dateTime('created_at').defaultTo(new Date('2000/1/1').toISOString()).notNullable();
+      table.dateTime('updated_at').defaultTo(new Date('2000/1/1').toISOString()).notNullable();
+    }),
+  ]);
+};
+
+export const down = function (knex) {
+  return Promise.all([knex.schema.dropTable('irrigation_task')]);
```
anant00729, line 22: In case we want to roll back the migration on beta or on production. we will lose all the data from the table. could you please drop only the columns in the down function?

packages/webapp/src/components/Modals/WaterUseCalculatorModal/index.jsx
```diff
@@ -1,32 +1,315 @@
...
+const TotalWaterUsage = ({ totalWaterUsage }) => {
   const { t } = useTranslation();
+  const labelStyle = {
+    fontSize: '16px',
+    lineHeight: '20px',
+    color: 'var(--teal900)',
+  };
+  return (
+    <div
+      style={{
```
anant00729, line 31: can create a 'styles.scss' file and add these styling

packages/webapp/src/components/Modals/WaterUseCalculatorModal/index.jsx
```diff
@@ -1,32 +1,315 @@
...
+
+const WaterUseDepthCalculator = ({ system, setTotalWaterUsage, totalWaterUsage, formState }) => {
+  const { t } = useTranslation();
+  const { register, getValues, watch, control, setValue } = formState();
+  const { irrigated_area, application_depth } = getValues();
+  const modalState = useForm({ mode: 'onChange', shouldUnregister: false });
+  const { percentage_location_irrigated } = modalState.getValues();
+
+  const locationSize = useSelector(cropLocationsSelector).filter(
+    (location) => location.location_id === getValues().locations[0].location_id,
```
anant00729, line 124: can use optional for null safety e.g. location?.location_id === getValues()?.locations[0]?.location_id

packages/webapp/src/components/Modals/WaterUseCalculatorModal/index.jsx
```diff
@@ -1,32 +1,315 @@
...
+  )[0].total_area;
```
anant00729, line 125: can use optional for null safety e.g )[0]?.total_area

packages/webapp/src/components/Modals/ModalComponent/v2/index.jsx
```diff
@@ -38,6 +40,15 @@ export default function ModalComponent({
             {error && <FiSlash style={{ marginTop: '1px' }} />}
             {icon && icon}
             {title}
+            {tooltipContent && (
+              <>
+                &nbsp;&nbsp;
```
anant00729, line 45: instead of using "& nbsp;& nbsp;" use "{'    '}"

packages/webapp/src/components/Modals/WaterUseCalculatorModal/index.jsx
```diff
@@ -1,32 +1,315 @@
...
+          displayUnitName={LOCATION_SIZE_UNIT}
+          unitType={location_area}
+          system={system}
+          hookFormSetValue={setValue}
+          hookFormGetValue={getValues}
+          hookFromWatch={watch}
+          control={control}
+          required
+          value={locationSize}
+          disabled={true}
```
anant00729, line 212: just write disabled no need to write disabled={true}. as you can see 'required' is also written the same way

packages/webapp/src/components/Task/PureIrrigationTask/index.jsx
```diff
@@ -168,7 +179,10 @@ export default function PureIrrigationTask({
         style={{ marginTop: '10px', marginBottom: '25px' }}
         hookFormRegister={register(DEFAULT_IRRIGATION_TASK_LOCATION)}
       />
-      <Label className={styles.label} style={{ marginBottom: '12px', marginTop: '10px' }}>
+      <Label
+        className={styles.label}
+        style={{ marginBottom: '12px', marginTop: '10px', fontSize: '16px' }}
```
anant00729, line 184: avoid in line style

packages/webapp/src/components/Task/PureIrrigationTask/index.jsx
```diff
@@ -197,24 +211,23 @@ export default function PureIrrigationTask({
       <Checkbox
         label={t('ADD_TASK.IRRIGATION_VIEW.SET_AS_DEFAULT_MEASUREMENT_FOR_THIS_IRRIGATION_TYPE')}
         sm
-        style={{ marginTop: '2px', marginBottom: '20px' }}
+        style={{ marginBottom: '24px', marginTop: '-5px' }}
```
anant00729, line 214: avoid in line style

packages/webapp/src/components/Task/PureIrrigationTask/index.jsx
```diff
@@ -197,24 +211,23 @@ export default function PureIrrigationTask({
...
         hookFromWatch={watch}
-        name={DEPTH}
+        name={ESTIMATED_WATER_USAGE}
         unitType={waterUsage}
         max={999999999}
         system={system}
         control={control}
-        style={{ paddingBottom: '32px' }}
-        disabled={false}
+        style={{ marginTop: '10px' }}
```
anant00729, line 230: avoid in line style

packages/webapp/src/containers/Task/saga.js
```diff
@@ -364,13 +364,84 @@ const getTransplantTaskBody = (data, endpoint, managementPlanWithCurrentLocation
...
+  const flow_rate_unit = data.estimated_flow_rate_unit;
+  const default_application_depth = data.default_location_application_depth;
+  const application_depth = data.application_depth;
+  const application_depth_unit = data.application_depth_unit;
+  return produce(
+    getPostTaskBody(data, 'irrigation_task', managementPlanWithCurrentLocationEntities),
+    (data) => {
+      data.managementPlans = managementPlans?.map((management_plan_id) => ({
+        planting_management_plan_id:
+          managementPlanWithCurrentLocationEntities[management_plan_id].planting_management_plan
```
anant00729, line 389: use optional

packages/webapp/src/containers/Task/saga.js
```diff
@@ -364,13 +364,84 @@ const getTransplantTaskBody = (data, endpoint, managementPlanWithCurrentLocation
...
+        flow_rate,
+        flow_rate_unit,
+        application_depth,
+        application_depth_unit,
+        default_application_depth,
+        default_flow_rate,
+      };
+      set_default_irrigation_task_location &&
+        (data.location.irrigation_task_type = irrigation_task_type);
+      delete data.irrigation_task_type_other;
```
anant00729, line 413: can take out the useful values in a separate object and override it with the data.

packages/webapp/src/util/convert-units/unit.js
```diff
@@ -113,6 +114,61 @@ export const container_planting_depth = {
   databaseUnit: databaseUnit.length,
 };
 
+export const estimated_duration = {
+  metric: {
+    units: ['h', 'm'],
+    defaultUnit: 'h',
+    breakpoints: [9999999999],
```
Duncan-Brain, line 121: Based off of the pattern of the others should this be [60] ?

anant00729 (changes requested):

packages/webapp/src/util/convert-units/unit.js
```diff
@@ -113,6 +114,61 @@ export const container_planting_depth = {
...
+  },
+  databaseUnit: databaseUnit.hour,
+};
+
+export const irrigation_depth = {
+  metric: {
+    units: ['mm', 'm'],
```
Duncan-Brain, line 133: Kevin mentioned that irrigation depth should never be m, you can remove m. This also relates to my comments on `<Unit />`

packages/webapp/src/components/Form/Unit/index.jsx
```diff
@@ -135,6 +138,7 @@ const useReactSelectStyles = (disabled, { reactSelectWidth = DEFAULT_REACT_SELEC
 };
 const Unit = ({
   disabled = false,
+  selectDisabled = false,
```
Duncan-Brain, line 141: Remove, see comment below.

packages/webapp/src/components/Form/Unit/index.jsx
```diff
@@ -188,7 +193,7 @@ const Unit = ({
     const options = getOptions(unitType, system);
     const hookFormValue = hookFormGetValue(name);
     const value = hookFormValue || (hookFormValue === 0 ? 0 : defaultValue);
-    const isSelectDisabled = options.length <= 1;
+    const isSelectDisabled = options.length <= 1 || selectDisabled;
```
Duncan-Brain, line 196: The previous expression should be sufficient. Once the irrigation_depth has only one option in util/convert-units this should work as expected.

packages/webapp/src/components/Modals/WaterUseCalculatorModal/index.jsx
```diff
@@ -1,32 +1,315 @@
...
+        label={t('ADD_TASK.IRRIGATION_VIEW.ESTIMATED_FLOW_RATE')}
+        hookFormSetValue={setValue}
+        hookFormGetValue={getValues}
+        hookFromWatch={watch}
+        name={FLOW_RATE}
+        unitType={water_valve_flow_rate}
+        max={999999.99}
+        system={system}
+        control={control}
+        onChangeUnitOption={(e) => {
```
Duncan-Brain, line 79: This is not working correctly. If you select the same units you are already on it will keep dividing. (eg current value: 60 l/h, "change" units to l/h (even though already this value), new value: 1 l/h).

Duncan-Brain:
@wasswarichard @anant00729

1. Default settings pattern.

Since LiteFarm does not yet seem to have a pattern for this type of user setting. I like how you add the 'location_defaults' table as the easiest to maintain solution for now. But to Anant's point in our meeting: putting the defaults in the irrigation task would reduce data duplication and sort of align with existing patterns. But I also agree with what I think you were saying that it is much more difficult to put it into the irrigation task table without adding unnecessary checks/overwrites for existing defaults around something like: `t.boolean('is_this_the_default_flow_rate_for_this_location_id').defaultTo(false);`. So therefore location_defaults is the best option for now. 

We may want to get feedback from @kcussen about if he has any procedures about db schema design. A bit of research here about what we could be doing for user default settings: https://stackoverflow.com/questions/10204902/database-design-for-user-settings.

2. Extra columns in 'irrigation_task' (notes, due_date, created_by, created_at etc. etc.)

So what I gathered from this discussion was that maybe there is a miscommunication here. Two different reasons to add this data:

The Jira story asks specifically for the base attributes on the irrigation task, but it also says "(please treat this as guidance rather than gospel)". Currently if we add these to 'irrigation_task' table this would unnecessarily duplicate data that already exists with the same PK in the 'task' table. The 'task' table is the base object here and already implements the base model properties.

But if this situation is like @anant00729 PR for [LF-2818](https://github.com/LiteFarmOrg/LiteFarm/compare/LF-2818/refactor-custom-field-work-tasks-to-mirror-pattern-for-harvest_use_task) then maybe these columns should be added to 'irrigation_task_type' table. If we add to the 'irrigation_task_type' table it will allow for custom irrigation task types to be selectable and conform to base model. So now 'irrigation_task_type' would be a base object here. 

3. I think there might me some missing migrations for the 'irrigation_task' table to deal with renaming columns 'flow_rate_l/min' and 'hours'. Also, if existing data in the irrigation task table for task_id does not match up to a task in the 'task' table. Then the created_at = '2000/1/1' would need to be added to 'task' table... but I think it should already exist so hopefully no action needed there.

4. `irrigationTaskModel.js` -- See # 2, if true revert changes to back to Model. See # 3, renaming columns.

5. Inline comments: 
- Total water usage event is not working when units changed to existing value.
- Convert units remove m as option for depth
- On `<Unit />`  new attribute isSelected not necessary

6. General comments 
- For edits to components like` <Unit />` and `<ReactSelect />` a quick description of the need for the change in this PR desrciption would help me review faster and help if someone looks back at changes in the future.
- If we can keep LF issues on separate PR's I think that would be best. This contains code pertaining to Lf-2800, LF-2801 and LF2805

wasswarichard:
@Duncan-Brain  changes to ReactSelect Component fixes issue 
image
on https://lite-farm.atlassian.net/browse/LF-2800 . The rest of the refactors have been effected.

anant00729 (approved):
This worked for me.

Duncan-Brain:
@wasswarichard @anant00729 

1. On` <Unit />` you need to handle the case where onChangeUnit does not exist. You can test by changing units on Estimated Water Usage
image

2. Is the Total Estimated water usage supposed to persist between two different depth and volume water calculators? To me it only makes sense to persist if the same calculator is used. Consider persisting both depth_total_water & volume_total_water separately.

3. irrigationTaskModel.js & irrigat_task migration - add estimated_duration_unit and populate with value 'hours' according to https://lite-farm.atlassian.net/l/cp/kW7o92mJ

wasswarichard:
@Duncan-Brain 
issue 1, has been resolved.
issue 2, has been resolved.
issue 3, depending on the requirements from Kevin 
image
 that column is renamed.

Duncan-Brain:
for 3. duration unit is in requirements, since duration is staying in the table so should the units:
Screen Shot 2022-11-22 at 8 43 50 AM

Added an inline comment about keeping the mock functional. If you are changing it, change it all.

packages/api/db/migration/20221118081755_alter_irrigation_task_table.js
```diff
@@ -0,0 +1,50 @@
...
+    table.string('estimated_water_usage_unit');
+    table.float('application_depth');
+    table.string('application_depth_unit');
+  });
+
+  for (const row of irrigation_task_rows) {
+    await knex.raw(
+      `UPDATE "task" task SET duration = ${row.hours * 60} WHERE task.task_id = row.task_id`,
+    );
+    await knex('irrigation_task').insert({ estimated_flow_rate_unit: 'l/min' });
```
Duncan-Brain, line 24: @wasswarichard something like this one I think

Duncan-Brain (approved):
Seems good to me, just blocked on the automated tests by the POST ticket.

PR #2331, opened 2022-11-10
Lf 2798 select the task location

Duncan-Brain:
[LF-2798](https://lite-farm.atlassian.net/browse/LF-2798?atlOrigin=eyJpIjoiNzczNzUzZWIwN2Y2NDBlYzgyYTRkMTkxODcyM2YxOGYiLCJwIjoiaiJ9) - 

My previous understanding was that crop enabled locations meant crop areas with current crops. The actual goal was to use any crop capable area regardless of current crops.

**Changes since previous merge:**
- Made a separate functional `<TaskIrrigationLocations />`  for the switch statement.
- Added the `readOnlyPinCoordinates()` so that wild crops are included as a location
- Handled the case where multi select is not enabled but also wild crops are selected
- Included title translations
- Refactored the goBack function to reuse for all


packages/webapp/src/components/Task/TaskLocations/index.jsx
```diff
@@ -55,10 +55,13 @@ export default function PureTaskLocations({
   );
 
   const onSelectLocation = (location_id) => {
-    setValue(
-      LOCATIONS,
-      isMulti ? getSelectedLocations(location_id, selectedLocations) : [{ location_id }],
-    );
+    if(!isMulti && getValues('show_wild_crop')){
```
anant00729, line 58: use early return 
```
const onSelectLocation = (location_id) => {
  if(!isMulti && getValues('show_wild_crop')) return;
  setValue(
    LOCATIONS,
    isMulti ? getSelectedLocations(location_id, selectedLocations) : [{ location_id }],
  );
};
```

anant00729 (changes requested):

packages/webapp/src/components/Task/TaskLocations/index.jsx
Duncan-Brain: @anant00729 Good catch! Can you check this again. 

I instead handled the !multi condition for the wild crop here. Now when a location is selected and the checkbox is also selected it will now deselect the checkbox and select the area to fit in with how the map multiselect works.

Duncan-Brain:
> @Duncan-Brain Please check the PR comments. Also, I would like to know the reason to remove isMulti as a prop from the component

So I added the isMulti prop to the component for my first PR and I reverted it back to what it was before.

anant00729 (approved):
This looks good to me. @wasswarichard Please review.

wasswarichard (approved):

wasswarichard:
@Duncan-Brain it looks good.

PR #2333, opened 2022-11-21
update the status code to 200 as per the ensemble's api requirements for add readings

anant00729:
The add readings status code is been passed by the ensemble API in the registration response of sensors. This was causing us the issue to register sensors. 

The ensemble calls the webhook API to confirm if it working or not with the empty data object. 

Previously, I was sending 400 status code when the req.body was {}. After discussing this with Morten, it's now changed to 200.

the add readings API's response is revived by the ensemble.

### DIFF
--- a/packages/api/src/controllers/sensorController.js
+++ b/packages/api/src/controllers/sensorController.js
@@ -401,7 +401,7 @@ const sensorController = {
 
   async addReading(req, res) {
     if (!Object.keys(req.body).length) {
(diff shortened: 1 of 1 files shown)


wasswarichard (approved):
This is okay for me depending on the discussion I have had with @anant00729

Duncan-Brain (approved):
Ok sounds reasonable.

PR #2334, opened 2022-11-21
Lf 2765 incorporate happy path tests into GitHub actions based build process

MwayaB:


Duncan-Brain:
@MwayaB 

It looks like this is supposed to be git-ignored:  `packages/webapp/.nyc_output/out.json`

I believe this should be deleted since we already have a pnpm-lockfile: `packages/webapp/package-lock.json`

Duncan-Brain (approved):
Looks like those were removed and the test is succeeding in actions.

.github/workflows/cypress_tests.yml
```diff
@@ -7,7 +7,6 @@ on:
 
 jobs:
   build:
-    timeout-minutes: 10
```
mika-robots, line 10: @MwayaB I think we should keep the timeout, but make it 20 or 30 min

mika-robots (approved):

PR #2336, opened 2022-11-23
Lf 2818 : refactor custom field work tasks to mirror pattern for harvest task

anant00729:
To Test: 

1. run the migration script 
2. go to the field task page
3. select other as the task type and add the task
4. go again to the field task page and check the custom task getting added to the drop-down.


packages/api/db/migration/20201124123610_crop_farm_field_traceability.js
```diff
@@ -10,6 +10,7 @@ const tables = [
   'pesticide',
   'shift',
   'taskType',
+  'fieldWorkLog',
```
Duncan-Brain, line 13: @anant00729 

This looks like a migration from 2020. Is it okay to edit old migrations? I think this could have negative consequences.

packages/api/db/migration/20221114230005_alter_field_work_task_table.js
```diff
@@ -0,0 +1,97 @@
...
+    {
+      field_work_type_translation_key: 'TILLAGE',
+      field_work_name: 'Tillage',
+    },
+    {
+      field_work_type_translation_key: 'WEEDING',
+      field_work_name: 'Weeding',
+    },
+  ];
+  await knex.schema.createTable('field_work', function (t) {
```
Duncan-Brain, line 38: @anant00729 

Consider naming table 'field_work_type'

packages/api/db/migration/20221114230005_alter_field_work_task_table.js
```diff
@@ -0,0 +1,97 @@
...
+
+  await knex('field_work').insert(fieldWorkData);
+
+  const field_work_rows = await knex('field_work').select(
+    'field_work_id',
+    'field_work_type_translation_key',
+  );
+
+  await knex.schema.alterTable('field_work_task', (t) => {
+    t.integer('field_work_id').references('field_work_id').inTable('field_work');
```
Duncan-Brain, line 58: @anant00729 

Similar to other comments consider naming column 'field_work_type_id'

packages/api/db/migration/20221114230005_alter_field_work_task_table.js
```diff
@@ -0,0 +1,97 @@
...
+  });
+
+  await knex('field_work').insert(fieldWorkData);
+
+  const field_work_rows = await knex('field_work').select(
+    'field_work_id',
+    'field_work_type_translation_key',
+  );
+
+  await knex.schema.alterTable('field_work_task', (t) => {
```
Duncan-Brain, line 57: @anant00729 

Is it ok to alter two tables in one migration? What is best practice?

packages/api/db/migration/20221114230005_alter_field_work_task_table.js
anant00729: the refactored tables will look like this

Screen Shot 2022-11-29 at 10 01 36 AM

packages/api/db/migration/20221114230005_alter_field_work_task_table.js
anant00729: the refactored tables will look like this

Screen Shot 2022-11-29 at 10 01 36 AM

packages/api/db/migration/20221114230005_alter_field_work_task_table.js
anant00729: in the alter table query I am just adding the columns in the up function. so it's a safe change.

packages/api/db/migration/20201124123610_crop_farm_field_traceability.js
Duncan-Brain: Yes I can see how it would work fine to rebuild database and work as expected. But why in the other migration file (alter_filed_work_task_table) did you use `alterTable()`?  Could you have gone back to the original table creation migration too and make the table correctly in an older file?  If we did this we could probably condense down all the migrations into one or two files.

> The process of adding a column is similar to creating a table. We never want to edit a migration file after it has been run because when we run knex migrate:latest knex will not make the change. Migrations will only run once.
 
> [From: http://perkframework.com/v1/guides/database-migrations-knex.html](
http://perkframework.com/v1/guides/database-migrations-knex.html)

packages/api/db/migration/20221114230005_alter_field_work_task_table.js
Duncan-Brain: Are you changing the current table naming convention?

image

image

kcussen:
Hey guys, I might need a quick call to make sure I understand correctly. I assumed it would be:
- task: Defines the generic attributes of a task
- field_work_task: Defines the field work specific attributes of the task (similar to harvest_task)
- field_work_type: Defines the global and farm specific types of field work (similar to harvest_use_type or task_type)

packages/api/src/routes/taskRoute.js
```diff
@@ -250,4 +250,6 @@ router.get(
   taskController.getHarvestUsesByFarmId,
 );
 
+router.get('/get_field_work_types/:farm_id', taskController.getFieldWorkTypes);
```
Duncan-Brain, line 253: @anant00729 

Should this not check farm access, or scope or anything?

packages/webapp/src/main.jsx
```diff
@@ -53,6 +53,7 @@ import consentSaga from './containers/Consent/saga';
 import callbackSaga from './containers/Callback/saga';
 import inviteUserSaga from './containers/InviteUser/saga';
 import exportSaga from './containers/ExportDownload/saga';
+import fieldWorkTaskSaga from './components/Task/FieldWorkTask/saga';
```
Duncan-Brain, line 56: I think this saga file should be somewhere in containers -- there are no other saga files in components that I can see.

packages/api/db/migration/20201124123610_crop_farm_field_traceability.js
anant00729: I tested it using the latest dump of the beta database. it looks like to migration only ran for the newly added files under the migration folder. Therefore, I had to remove this. In the migration file 20221114230005_alter_field_work_task_table.js

I have added these Base Model related columns

anant00729: ie. 
created_by_user_id
updated_by_user_id
created_at
updated_at

anant00729:
I have renamed the table name and primary key. It looks like this: 

Screen Shot 2022-11-30 at 2 14 50 PM

packages/api/src/routes/taskRoute.js
anant00729: added the scope 
```
router.get(
  '/get_field_work_types/:farm_id',
  hasFarmAccess({ params: 'farm_id' }),
  taskController.getFieldWorkTypes,
);

```

packages/webapp/src/main.jsx
anant00729: Changed the location and shifted to the containers folder

anant00729:
@Duncan-Brain I have worked on the changes that you asked for. Please take a look.

Duncan-Brain:
@anant00729 We have already talked about most of the following and you have completed/working on:

1. Table migrations : 
- [Completed] The simple naming conventions mentioned in some lines.
- [Completed] Editing past migrations seems bad
- Migrating/dropping 'other_type' column on up and migrating back/adding column 'other_type' on down

2. Routes to Controllers:
- [Completed] Task route doesnt check farm access? Any other checks needed here?
- taskController - make the function generic something like -> checkAndAddCustomType
- [Not this PR] Thanks for discussion on controller architecture - I am thinking some of these files could get big as we add more custom x_task types. Right now the endpoint route is taskRoute -> taskController -> fieldWorkTaskTypeModel -> field_work_task[db] I am wondering if we should be OOP-ing things towards taskRoute -> fieldWorkTaskRoute -> fieldWorkTask -> fieldWorkTaskController -> fieldWorkTaskModel -> field_work_task[db]. Resource for nesting routers like so in the future: https://stackoverflow.com/a/25305272

3. Components:
-  [Completed] Inline comment about moving file location of saga.js

4. containers/Task/saga - this kind of mirrors the task controller
- make the function generic something like -> getCompleteCustomTaskTypeBody

5. Frontend
- Consider making the label in the FieldWorkTask component, the name entered previously instead of lowercasing the translation key -- see inline comment with pictures

packages/webapp/src/components/Task/FieldWorkTask/index.jsx
```diff
@@ -41,6 +36,31 @@ const PureFieldWorkTask = ({
...
+  useEffect(() => {
+    dispatch(getFieldWorkTypes());
+  }, []);
+
+  const displayLabel = (value) => {
+    const translationKey = `ADD_TASK.FIELD_WORK_VIEW.TYPE.${value}`;
+    if (i18next.exists(translationKey)) {
+      return t(translationKey);
+    } else {
+      let customValue = value.toLocaleLowerCase().replaceAll('_', ' ');
```
Duncan-Brain, line 59: You can consider using the name supplied instead of lowercasing the translation key. Maybe the capitalization is important?:
image

image

packages/webapp/src/components/Task/FieldWorkTask/index.jsx
```diff
@@ -41,6 +36,31 @@ const PureFieldWorkTask = ({
   useEffect(() => {
     if (fieldWorkTypeExposedValue?.value !== 'OTHER') setValue(FIELD_WORK_OTHER_TYPE, null);
   }, [fieldWorkTypeExposedValue]);
+
+  useEffect(() => {
```
JimmyRowland, line 40: use useMemo and remove useState setState

JimmyRowland: const fieldWorkTypeOptions = useMemo(()=>{
let options = ....
return options
}, [fieldWorkTypes])

packages/webapp/src/components/Task/FieldWorkTask/index.jsx
```diff
@@ -13,23 +17,14 @@ const PureFieldWorkTask = ({
   disabled = false,
 }) => {
   const { t } = useTranslation();
+
+  const { fieldWorkTypes = [] } = useSelector(fieldWorkSliceSliceSelector);
```
JimmyRowland, line 21: The purpose of component/container folder is separate testable/untestable components.
In component folder, all components do not use selectors or dispatch/network requests so that components can be unit tested without mocking redux store/network requests.

packages/webapp/src/components/Task/FieldWorkTask/index.jsx
```diff
@@ -41,6 +36,31 @@ const PureFieldWorkTask = ({
...
+      return options;
+    });
+  }, [fieldWorkTypes]);
+
+  useEffect(() => {
+    dispatch(getFieldWorkTypes());
+  }, []);
+
+  const displayLabel = (value) => {
+    const translationKey = `ADD_TASK.FIELD_WORK_VIEW.TYPE.${value}`;
```
JimmyRowland, line 55: types are usually in their only translation files like crop, then this can be t(translationKey)

packages/webapp/src/containers/fieldWorkSlice.js
```diff
@@ -0,0 +1,43 @@
+import { createSlice } from '@reduxjs/toolkit';
+
+const initialState = {
+  loading: false,
+  fieldWorkTypes: [],
```
JimmyRowland, line 5: Shouldn't this be aggregated by farm_id? or there is no custom fieldWorkType

packages/api/src/models/fieldWorkTaskModel.js
```diff
@@ -33,21 +35,10 @@ class FieldWorkTaskModel extends Model {
...
-            'PREPARING_BEDS_OR_ROWS',
-            'PRUNING',
-            'SHADE_CLOTH',
-            'TERMINATION',
-            'TILLAGE',
-            'WEEDING',
-            'OTHER',
-          ],
-        },
+        type: { type: 'string' },
```
JimmyRowland, line 38: Why it exist along with field work type id?

packages/webapp/src/containers/fieldWorkSlice.js
JimmyRowland: Sorry. type_id

anant00729:
@Duncan-Brain I have covered all the points that you have mentioned. Please review it again. Thanks.

packages/webapp/src/components/Task/FieldWorkTask/index.jsx
```diff
@@ -61,11 +81,7 @@ const PureFieldWorkTask = ({
             value={
               // TODO: refactor value reading here and in pest control
               // this solution keeps placeholder while accommodating the read-only view
-              !value
-                ? value
-                : value?.value
-                ? value
-                : { value, label: t(`ADD_TASK.FIELD_WORK_VIEW.TYPE.${value}`) }
+              !value ? value : value?.value ? value : { value, label: displayLabel(value) }
```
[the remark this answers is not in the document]
anant00729: this was added previous the ternary operation .

packages/webapp/src/components/Task/FieldWorkTask/index.jsx
anant00729: Didn’t get you

Duncan-Brain:
@anant00729 Everything is working for me mostly.

Just on the migration from pre-existing data I expected both user_id 's to be correct -- similar to how you inserted the farm_id.

image

anant00729:
@Duncan-Brain Added the correct user ids to the columns of the field_task_type table. Please check 

Screen Shot 2022-12-01 at 5 07 36 PM

Duncan-Brain (approved):
Looks good to me

wasswarichard (approved):
This looks fine

PR #2337, opened 2022-11-23
water usage model refactor

wasswarichard:
Fixes migrations issues and refactors water usage calculator component


packages/api/db/migration/20221118081755_alter_irrigation_task_table.js
```diff
@@ -45,9 +45,9 @@ export const down = async function (knex) {
 
   for (const row of task_rows) {
     await knex.raw(
-      `UPDATE 'irrigation_task' task SET hours = ${
+      `UPDATE "task" t SET duration = ${
         row.duration / 60
```
Duncan-Brain, line 49: @wasswarichard 

Is the previous state of task.duration = irrigation_task.duration/60 or is the previous state NULL? 
image

Duncan-Brain:
@wasswarichard 

I think it mostly looks ok to me. I added one inline comment on the alter table.

Just one question -- since this migration has already been merged. And is probably running on beta. Is there a process to rollback the previous db update and re-run the migration?

wasswarichard:
@Duncan-Brain the previous state is not deployed to beta, I rolled back.

Duncan-Brain (approved):
I think that's good.

anant00729:
The only change I can see is you are setting the duration = null in the update query and there is a modal name change in the files.
This looks good to me.

anant00729 (approved):

PR #2339, opened 2022-11-23
Server reload change in base url response message

anant00729:


kcussen (approved):
Seems much more welcoming now!

PR #2346, opened 2022-11-28
LF-2861 Post Irrigation data 

wasswarichard:
https://lite-farm.atlassian.net/browse/LF-2861 https://lite-farm.atlassian.net/browse/LF-2806 irrigation task post


anant00729:
This worked for me. @Duncan-Brain Please review.

anant00729 (approved):

Duncan-Brain:
@anant00729 @wasswarichard  

I probably will not have time to check this this week. 

All I can say is that migrations are pretty serious actions in my opinion. If you have time to double check this @anant00729. Specifically that the migration down is the opposite of up and works well (even on a non beta database) than it is probably ok. It seems strange to me that some of the enums are capitalized while others are not --- but that is probably part of the fix.

PR #2359, opened 2022-12-02
Lf 2865 solidify unit styling and color styling in modals

Duncan-Brain:
For ticket : [LF-2865](https://lite-farm.atlassian.net/browse/LF-2865?atlOrigin=eyJpIjoiOTljMDg5Mzc1ZTc4NDk0MGI3OTY5ZjRjMDkwMzhlMGQiLCJwIjoiaiJ9)

1. Fixes how unit inputs are displayed based on disabled or isSelectDisabled.
2. Fixes Modal background color to not be equal to --inputDisabled color


anant00729:
@Duncan-Brain this looks good to me.

anant00729 (approved):

packages/webapp/src/components/Form/Unit/index.jsx
```diff
@@ -115,12 +115,12 @@ const useReactSelectStyles = (disabled, { reactSelectWidth = DEFAULT_REACT_SELEC
         width: `${reactSelectWidth - 19}px`,
         display: 'flex',
         justifyContent: 'center',
-        background: state.isDisabled ? 'var(--inputDisabled)' : 'inherit',
+        background: disabled ? 'var(--inputDisabled)' : 'inherit',
```
JimmyRowland, line 118: Do not use disabled outside of reactSelect here. 
disabled is passed to the react select component, and then state.isDisabled is synced internally in reactSelect library.
state.isDisabled should always be up to date.

Duncan-Brain: @JimmyRowland 

The reason I changed this is because I need the Unit based disabled distinction to render this css. Essentially: disabled != isDisabled. For Select we set the attribute as: 
`isDisabled = {isSelectDisabled || disabled} ` for units. 

Since disabled is about Unit and not about Select I think the pattern here is fine until the css pattern is updated as a whole, since it is passed to Select through Unit and the useMemo handles the update.

Let me know if you have another pattern in mind for this distinction.

JimmyRowland: Thanks. Then isSelectDisabled || disabled is definitely wrong.  || disabled should go into useMemo

JimmyRowland: The component has completely changed since the last time I looked it. Ignore my comments if they do not make sense.

Duncan-Brain: @JimmyRowland 

Thanks. It was a good thing to double check its a weird component I second guessed it too. It made sense. I don't ignore though haha we always need more approvers/reviewers!

Duncan-Brain: I just tried adding disable to useMemo and it caused problems. Double checking the react docs I am seeing a note that says [The array of dependencies is not passed as arguments to the function.](https://reactjs.org/docs/hooks-reference.html#usememo) .. just the dependency array like useEffect.

JimmyRowland: https://beta.reactjs.org/apis/react/useMemo#examples-recalculation

JimmyRowland: Something might be seriously wrong with unit component

JimmyRowland (approved):

PR #2363, opened 2022-12-03
LF-2788 

cfwm:
For password requirement indications, add some examples of special characters


kcussen:
Thanks so much @cfwm!  This looks great. Two quick notes:

1. Per @anant00729 's note it looks like you're about 122 commits behind integration at this point. Given the trivial changes in your PR, I'm not too worried. We'll shortly be enforcing that PRs have to be up to date with integration via Github in which case this situation will no longer be a problem =)
2. Typically with language tags, we just mass update translations right before a release so that it's one consistent voice / conjugation, etc. Obviously with this ticket that isn't an issue - but for future notice. If you want to add your own translations make sure to chat with David so he can give you a quick training on how we approach that.

kcussen (changes requested):
Can you bring your branch up to date with integration? Then I can merge. Thanks again @cfwm !

cfwm:
@kcussen @anant00729 branch merged with integration!

PR #2366, opened 2022-12-05
Added error handling of soil water potential unit in the preview tab

anant00729:
To Test,

1. Add a sensor to the farm
2. on the farm map page, long click on the sensor point.
3. check if the error is handled 

Before: 

Screen Shot 2022-12-05 at 1 55 06 PM

After: 

Screen Shot 2022-12-05 at 1 55 19 PM


wasswarichard (approved):
This looks fine to me

PR #2374, opened 2022-12-07
Addition of the deleted column to the field_work_type table

anant00729:
Based on the following comment: 

Screen Shot 2022-12-07 at 10 04 30 AM

Here is the new migration change I have made.

To Test,

1. run the latest migration 
2. go to the field_work_type table and check deleted column is present with the default value as false.


wasswarichard (approved):
looks fine to me

PR #2375, opened 2022-12-08
Conversion from kPa to PSI

anant00729:
To Test,

1. add some sensors to the farm and make sure you have the sensor readings
2. long click on the sensor to see the temperature and soil water potential readings 
3. go to http://localhost:3000/profile and change metric to imperial 
4. come back to the farm page 
5. long press and check if you can see the value of soil water potential is in PSI

### DIFF
--- a/packages/webapp/src/components/Map/PreviewPopup/index.jsx
+++ b/packages/webapp/src/components/Map/PreviewPopup/index.jsx
@@ -5,13 +5,19 @@ import { makeStyles } from '@material-ui/core/styles';
 import { colors } from '../../../assets/theme';
 import CompactPreview from './CompactPreview';
 import { TEMPERATURE, SOIL_WATER_POTENTIAL } from '../../../containers/SensorReadings/constants';
-import { getTemperatureUnit, getTemperatureValue } from './utils';
+import {
+  getSoilWaterPotentialUnit,
+  getSoilWaterPotentialValue,
+  getTemperatureUnit,
+  getTemperatureValue,
+} from './utils';
 import { userFarmSelector } from '../../../containers/userFarmSlice';
 import { useSelector } from 'react-redux';
 import {
   sensorReadingTypesByLocationSelector,
   sensorReadingTypesSelector,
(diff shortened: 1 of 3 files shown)


packages/webapp/src/components/Map/PreviewPopup/utils.js
```diff
@@ -1,5 +1,5 @@
-const metric = ['metric', 'celsius'];
-const imperial = ['imperial', 'fahrenheit'];
+const metric = ['metric', 'celsius', 'kpa'];
+const imperial = ['imperial', 'fahrenheit', 'psi'];
```
JimmyRowland, line 3: temperature is already in convert-unit library
https://www.npmjs.com/package/convert-units#temperature
Also, the library support custom units
https://github.com/convert-units/convert-units

JimmyRowland (changes requested):

packages/webapp/src/components/Map/PreviewPopup/index.jsx
```diff
@@ -55,7 +66,7 @@ export default function PurePreviewPopup({ location, history, sensorReadings, st
   const classes = useStyles();
   const { t } = useTranslation();
   const { units } = useSelector(userFarmSelector);
-  const { reading_types } = useSelector(sensorReadingTypesByLocationSelector(location.id));
+  const { reading_types = [] } = useSelector(sensorReadingTypesByLocationSelector(location.id));
```
JimmyRowland, line 69: why it's location.id not location.location_id

packages/webapp/src/components/Map/PreviewPopup/utils.js
anant00729: The implementation of temperature conversion was added by sein. I have just worked on the soil water potential

packages/webapp/src/components/Map/PreviewPopup/index.jsx
anant00729: This was added by sein. I don't know why the variable name is kept as id

wasswarichard (approved):
looks fine

PR #2376, opened 2022-12-08
Lf 2804 irrigation read only view

wasswarichard:
https://lite-farm.atlassian.net/browse/LF-2804


wasswarichard:
@JimmyRowland Please review. let's stick to the scope of the ticket.

anant00729 (approved):
I have already reviewed this PR. The read only view of irrigation task works on my end

PR #2379, opened 2022-12-08
Lf 2530/conversion from kpa to psi

anant00729:
To Test,

1. add some sensors to the farm and make sure you have the sensor readings
2. long click on the sensor to see the temperature and soil water potential readings
3. go to http://localhost:3000/profile and change the metric to imperial
4. come back to the farm page
5. long press and check if you can see the value of soil water potential in PSI


wasswarichard (approved):
Looks fine

PR #2381, opened 2022-12-09
Lf 2691 placeholder add a spotlight to anants component to highlight that labels are clickable

MwayaB:


wasswarichard (approved):
This looks fine

PR #2384, opened 2022-12-13
PLACEHOLDER: Refresh store with newly POSTed irrigation task

anant00729:
To Test,

1. create an irrigation task. make sure to checkbox everything that can be persisted
2. create another irrigation task by selecting the exact farm location.
3. check if the default values are persisted in the create irrigation task form


wasswarichard:
@anant00729 I think this ticket is being blocked by https://lite-farm.atlassian.net/browse/LF-2875 cc @kcussen . There are a number of table migrations.

anant00729:
@wasswarichard as discussed we can proceed with the review. Please check

Duncan-Brain:
@anant00729 

Two things I am noticing:

1. I am seeing that the "set as default measurement type for this location" checkbox is not persisting, but the radio value it is referencing (depth/volume) seems correct.
2. In the volume water use calculator the "set as default flow rate for this location" I think the checkbox was falsely checked. The first time I opened this calculator the value was showing a checkmark in the box. Possible step to reproduce previously I did save the defaults for just the irrigation task portion on this location.

Duncan-Brain:
@wasswarichard This is not working for me.

Things I am noticing:
- When I select zero defaults: The next time I try there is a default.  And it looks like a location_default is saved
- When I go back and switch locations before saving. If that location has no default type it will take on the other location_default from previously selected area.

Duncan-Brain (approved):
Looks like the checkbox issue is fixed. The location_defaults were being created with no defaults set that now seems fixed.  Existing issues that still remain for another ticket are : always unchecked "set as" and then data persisting when switching location.

PR #2387, opened 2022-12-13
LF-2875_Irrigation_task_refactor

wasswarichard:
LF-2875_Irrigation_task_refactor
https://lite-farm.atlassian.net/browse/LF-2875


Duncan-Brain:
I am concerned that we are snowballing LF2800 and LF2801 week over week because these are large PRs. 

I would suggest breaking this into several smaller pieces so that we can consider each piece individually -- this should reduce the need to refactor in the future. 

- For example in the subtasks - adding support for and migrating just one table at a time

anant00729 (approved):
@wasswarichard this looks good to me. the changes I asked you to add are in place.

PR #2391, opened 2022-12-16
Fix: the issue of data overflow in the redux in the farm map page.

anant00729:
To Test:

1. add sensors to the farm / take a dump of the beta database in a local database.
2. check the store output under 'persistedStateReducer' > 'mapSensorReducer' in beta 
3. now check the store output under 'persistedStateReducer' > 'mapSensorReducer' in the local database which will be less than 10. 
4. you will find the count on the beta will be around 24k which is causing the data overflow.

Before:

Screen Shot 2022-12-15 at 3 04 59 PM

After: 

Screen Shot 2022-12-15 at 4 50 50 PM


wasswarichard (approved):
This looks okay. 
image
The response data provide a few important sensor readings.

PR #2392, opened 2022-12-16
Lf 2887 add an image in the add new crop workflow

Duncan-Brain:
**PLEASE LET ME KNOW WHEN MERGING THIS COULD CAUSE BREAKING CHANGES** 
[Issue LF-2887](https://lite-farm.atlassian.net/browse/LF-2887)
[Parent Issue LF-2072](https://lite-farm.atlassian.net/browse/LF-2072)

**Description**
We want to be able to add a custom crop image when creating a custom crop. Similar to how custom crop_variety image works.

**Testing**
This is not very testable without pushing to beta. But one thing you can do that illustrates some details:
1. Enter debug mode in integration and put a breakpoint on line 77 (data definition) in the postCropAndVarietal method of AddCropVariety/saga.js. 
2. Then edit crop_photo_url, and note the crop_variety_photo_url
3. Continue in debug, then check the crop table in the database to see that crop_variety_photo_url was actually persisted
4. Redo the process after deleting line 80 of the same file.
5. You should now see crop_photo_url persisted in the crop table 

**Notes**
- made ImagePickerWrapper more generic
- updated all components that use generic ImagePickerWrapper to stop using unused attribute (except in stories/storybook)
- updated main to use generic saga from ImagePickerWrapper
- Copied AddCropVariety for: Controller, Route, Component, CSS, Container
- Removed overwriting of crop_photo_url by crop_variety_photo_url in saga


wasswarichard (approved):
The code looks fine to me. I couldn't test it on local machine because of s3 bucket env file

anant00729:
@Duncan-Brain its 52 commits behind.

PR #2399, opened 2022-12-22
Fixed the issue of last updated text in the visualization 

anant00729:
To Test,

1. go to the sensor readings page 
2. check the text of the "last updated" sensor readings


wasswarichard (approved):
Looks fine to me

PR #2408, opened 2023-01-06
Lf 2804 manual irrigation task

wasswarichard:
Add default columns to store boolean values and include them in the post request.
https://lite-farm.atlassian.net/browse/LF-2804?focusedCommentId=14497

### DIFF
--- a/packages/api/db/migration/20230106112659_add_extra_columns_irrigation_table.js
+++ b/packages/api/db/migration/20230106112659_add_extra_columns_irrigation_table.js
@@ -10,7 +10,7 @@ export const up = async function (knex) {
     'irrigation_type_name',
   );
   await knex.schema.alterTable('irrigation_type', (table) => {
-    table.string('irrigation_type_translation_key').notNullable();
+    table.string('irrigation_type_translation_key').nullable();
   });
 
   for (const row of irrigation_task_type) {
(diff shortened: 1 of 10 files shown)


packages/api/db/migration/20230106112659_add_extra_columns_irrigation_table.js
```diff
@@ -0,0 +1,43 @@
...
+    const irrigation_type_translation_key = row.irrigation_type_name
+      .toUpperCase()
+      .split(' ')
+      .join('_');
+    await knex('irrigation_type')
+      .update({ irrigation_type_translation_key })
+      .where({ irrigation_type_id: row.irrigation_type_id });
+  }
+
+  await knex.schema.alterTable('location_defaults', (table) => {
```
Duncan-Brain, line 26: @wasswarichard 

The opposite of this should be in 'down' function.

Duncan-Brain:
@wasswarichard 

Before I continue the down function should be fixed so I do not corrupt my database in testing.

Also For your tables, please check and add as many constraints to as many columns as make sense like .notNullable(). This would also be reflected in required sections of Models etc.

packages/api/db/migration/20230106112659_add_extra_columns_irrigation_table.js
wasswarichard: @Duncan-Brain this has been resolved

wasswarichard:
@Duncan-Brain we accept null  irrigation_type_id values for new locations since values have not been set

PR #2412, opened 2023-01-11
Lf 2892 user nominated data ie crops species and varietals

Duncan-Brain:
Jira Ticket: [LF-2892](https://lite-farm.atlassian.net/browse/LF-2892?atlOrigin=eyJpIjoiNWZkYThmNjQzY2M1NGE1MmI2ZTdlZTEyNjEzOWEwNjIiLCJwIjoiaiJ9) 

Description:
Adds nomination db architecture. Including:
- Database architecture creation
- Models for each table
- A controller for the overall nomination
- A route that includes create, update, delete methods
- A new farm access verification
- Updated server to include route

To test: 
Use Postman or similar api request option.
1. Create 
- POST to /nomination
- { auth: id_token, header: user_id, farm_id, body: user_id, farm_id, nomination_type="CROP_NOMINATION"}
- view in table if nomination and status log is present
2. Delete
- DELETE to /nomination/[nomination_id from previous step eg. 1]
- { auth: id_token, header: user_id, farm_id }
- check the nomination table and see that deleted=true in base object property

[LF-2892]: https://lite-farm.atlassian.net/browse/LF-2892?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/api/db/migration/20230111164146_create_nomination_tables.js
```diff
@@ -0,0 +1,117 @@
+export const up = async function (knex) {
+  const now = knex.fn.now();
+  const litefarmUserId = 1;
+
+  await knex.schema.createTable('nomination_type', (table) => {
+    table.string('nomination_type').primary().notNullable();
```
wasswarichard, line 6: @Duncan-Brain strings cannot be a primary key

packages/api/db/migration/20230111164146_create_nomination_tables.js
```diff
@@ -0,0 +1,117 @@
...
+    table.string('created_by_user_id').references('user_id').inTable('users').notNullable();
+    table.dateTime('updated_at').notNullable();
+    table.string('updated_by_user_id').references('user_id').inTable('users').notNullable();
+    table.boolean('deleted').notNullable().defaultTo(false);
+  });
+  await knex.schema.createTable('nomination', (table) => {
+    table.increments('nomination_id').primary();
+    table
+      .string('nomination_type')
+      .references('nomination_type')
```
wasswarichard, line 17: @Duncan-Brain you have to reference an integer not a string value

packages/api/db/migration/20230111164146_create_nomination_tables.js
```diff
@@ -0,0 +1,117 @@
...
+    table.dateTime('updated_at').notNullable();
+    table.string('updated_by_user_id').references('user_id').inTable('users').notNullable();
+    table.boolean('deleted').notNullable().defaultTo(false);
+  });
+  await knex.schema.createTable('nomination_workflow', (table) => {
+    table.increments('workflow_id').primary();
+    table.string('status').notNullable();
+    table
+      .string('type_group')
+      .references('nomination_type')
```
wasswarichard, line 32: @Duncan-Brain you have to reference an integer not a string value

wasswarichard (changes requested):

packages/api/db/migration/20230111164146_create_nomination_tables.js
wasswarichard: we can not have duplicate strings. same string value should only appear once in a database. A single change in string value reference by an id indicates an updated to all foreign keys.

Duncan-Brain: Its a primary key so any attempt to create a duplicate will fail. Do you have any articles I can read saying this is a bad idea? I have personally read the opposite.

Duncan-Brain: I double checked because I am not 100% confident on this but trying to change it in PG results in this: 
image

JimmyRowland:
After reading the requirement, only nomination table is needed.
I don't get why nomination type needs to be a separate table. nomination type can just be an enum representing table names in nomination table. Nonimation workflow/status can also be merged into nomination table.

Agree with Richard, user input/string shouldn't be primary key. I usually just use UUID. serial/integer PK behaves differently in different versions of pg.

BaseController is deprecated. trx is not needed if only one table is modified. 

Don't add relationMappings if you do not intend to use fetchGraph API.

Use recommended transaction syntax when possible.
https://vincit.github.io/objection.js/guide/transactions.html#creating-a-transaction

missing await trx.commit() / await trx.rollback() caused too many bugs

packages/api/db/migration/20230111164146_create_nomination_tables.js
wasswarichard: @Duncan-Brain you review this https://online.umich.edu/courses/database-design-postgresql/

JimmyRowland:
Hi @Duncan-Brain 

> a robust model to follow for future new features

It's nice to look ahead. I just don't see how the enum constraint will change very often or how the nomination type will be referenced by other tables. Litefarm is still in feature building stage. Simplicity/maintainability beats everything else. It wouldn't be hard to split/normalize a single table in the future. On the other hand, 3 tables are definitely more difficult to work with. More tables == more mental overhead. When you design the database schema, you also need to consider ORM API (in this case ObjectionJS) and understand what kind of query is easy using a given ORM. Since we are not optimizing column indexing. changing enum constraint is straightforward. 

> For UUID vs integer I have this reading that basically mentions how Jira would choose at the end there-

In this case, serial might be better. I personally had a bad experience with serial. When you want to share your local db dumb with coworkers, restore will fail if your coworker uses a different PG version since serial works differently. 

Writing crop migration for both beta db and production db is a pain. It's easy to make mistakes when some crops have identical primary keys.

> model methods 

Server is only responsible basic curd operations + validation. Queries are almost never reusable. Adding another layer of abstraction on top of objectionJS does not make sense. Occasionally duplication is way better than unnecessary abstraction.

Add methods in objection model is a bad idea.
1: this structure will not work with transactions.
2: All the util function does is to hide what the query returns and the util functions are almost never reusable. In this case, the util function only select a few fields and the detail would be hidden away from reader.
3: Code navigation with JS class is always broken. Use go to definition on method name will give you all the methods/member variables/ string with same name in different classes.
4. You lose all IntelliSense when you wrap a query. Since typescript is not available, you wouldn't be able to type function argument correctly.

> baseController
It's easier to read plain ObjectionJS queries. BaseController tries to accommodate too many use cases it will eventually end up with a mess if people keep modifying each function to accommodate new use cases. On the other hand, ObjectionJS has great documentation. It's just easier to read the documentation. Reading abstracted code is painful.

> get the model well defined

upsertGraph also uses relationalMappings. If all tables are connected and one post endpoint uses upsertGraph without proper validation, a malicious user can send a request in a particular shape to mess up the database. It's better to add relationalMappings when you need to use upsertGraph/fetchGraph with proper validations

JimmyRowland:
I don't see anything wrong with using string as PK if only a handful of entries are added. When there are lots of entries, indexing might be messed up. When there are very few entries, enum is probably better.

If you guys have a bunch of nomination features planned just ignore my comments.

JimmyRowland (dismissed):

Duncan-Brain:
So I have removed:
- all relation mappings
- the model method

I have added 
- a previously missed explicit nullable() constraint to the notes column.

JimmyRowland (approved):

JimmyRowland (approved):

JimmyRowland (dismissed):

packages/api/src/models/nominationWorkflowModel.js
```diff
@@ -0,0 +1,43 @@
...
+
+// Describes the nomination table
+// Base model extends objection.js
+class NominationWorkflow extends baseModel {
+  static get tableName() {
+    return 'nomination_workflow';
+  }
+
+  static get idColumn() {
+    return 'id';
```
SayakaOno, line 25: Should this be `workflow_id`?

Duncan-Brain: Good catch! I will wait to see if you have more suggestions but I will change that once you are finished reviewing.

Duncan-Brain:
@SayakaOno 

Change has been made on id column.

packages/api/src/controllers/nominationController.js
```diff
@@ -0,0 +1,163 @@
...
+  deleteNomination() {
+    return async (req, res) => {
+      const trx = await transaction.start(Model.knex());
+      try {
+        const params = req.params;
+        const success = await baseController.delete(NominationModel, params.nomination_id, req, {
+          trx,
+        });
+        await trx.commit();
+        res.status(201).send({ success });
```
SayakaOno, line 138: I found that other delete APIs I referred were returning a 200 status code like below:
`res.sendStatus(200);`

Can we confirm what the response should be?
Other than that, everything worked great! Great job!!

JimmyRowland: 200 Kike had a confluence doc on status code :)

Duncan-Brain: Cool found it in "The api package", he references REST standards. Thanks for finding I complete forgot about this part. I also fixed the update response codes in update as well as delete based on the other controllers.

Duncan-Brain:
@SayakaOno @JimmyRowland 

Status codes for both update and delete fixed.

SayakaOno (approved):

PR #2414, opened 2023-01-16
fix water use calculator

wasswarichard:
persist application depth and % of location to be irrigated values for the depth measurement type when user clicks out of the model


Duncan-Brain (approved):
This works for me

PR #2415, opened 2023-01-16
display set default custom irrigation type

wasswarichard:
Display set as a default custom irrigation type during irrigation task creation

### DIFF
--- a/packages/webapp/src/components/Task/PureIrrigationTask/index.jsx
+++ b/packages/webapp/src/components/Task/PureIrrigationTask/index.jsx
@@ -186,30 +186,33 @@ export default function PureIrrigationTask({
 
   return (
     <>
-      <Controller
-        control={control}
-        name={IRRIGATION_TYPE}
-        rules={{ required: true }}
-        render={({ field: { onChange, onBlur, onFocus } }) => {
-          return (
-            <ReactSelect
-              onFocus={onFocus}
-              label={t('ADD_TASK.IRRIGATION_VIEW.TYPE_OF_IRRIGATION')}
-              options={IrrigationTypeOptions}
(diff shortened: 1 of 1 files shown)


packages/webapp/src/components/Task/PureIrrigationTask/index.jsx
```diff
@@ -186,30 +186,33 @@ export default function PureIrrigationTask({
...
-                setValue(IRRIGATION_TYPE_ID, e.irrigation_type_id);
-              }}
-              isDisabled={disabled}
-              value={IrrigationTypeOptions.find((options) => options.value === irrigation_type)}
-              defaultValue={getDefaultIrrigationTypeOptions}
-            />
-          );
-        }}
-      />
+      {IrrigationTypeOptions.length >= defaultIrrigationTaskTypes.length && (
```
Duncan-Brain, line 189: @wasswarichard 

Would this always be true? And I don't think we want to conditionally render this component do we?

wasswarichard: @Duncan-Brain we need to render the component after loading the irrigation types for the defaults to be displayed and it will always be true.

Duncan-Brain: I think the core problem is that you should be passing the data you want loaded from an existing container to pass in the data to the "pure" component. 

@JimmyRowland  I think tried to discuss this with Anant here: https://github.com/LiteFarmOrg/LiteFarm/pull/2336#discussion_r1036613780

For example: PurePestControl has data passed down through three other containers: TaskDetails, StepOne, and TaskReadOnly.

PR #2421, opened 2023-01-24
Chore/setup script

baweinbe:
This PR updates the setup process for Mac users to add a bash script for running the various install commands. It also changes the instructions for cloning to use SSH and adds commands for copying over the default environment files. Open to feedback!

Problems I encountered while getting setup that are not addressed in this PR:
1. Using `psql` didn't work for me. I had to login to `template1` database specifically to then be able to create databases from there.
2. I couldn't create a database called `pg-litefarm`. It had to be called `pg_litefarm`

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,23 +22,24 @@ LiteFarm is comprised of three applications which all reside in this monorepo.
 ## Preliminaries
 
 1. Clone the repository from Github to your computer. In a Terminal window navigate to the directory you want to put the files in. Then use the command `git clone https://github.com/LiteFarmOrg/LiteFarm.git`. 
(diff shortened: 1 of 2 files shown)


craigyu (approved):
Tested and it's working on my mac, good job!

Duncan-Brain:
@baweinbe 

Thanks for doing this, I like anything that helps make onboarding easier! 

My only concern/question is that I can't really think of any comparable projects that do this sort of thing. It seems unusual to create a script that installs common tools like node, homebrew and nvm as part of the project itself. Just wondering if you can think of any comparable project examples for me to look at?

PR #2423, opened 2023-01-24
Lf 2930 form element checkbox does not have the option for a tooltip

Duncan-Brain:
[Jira Ticket LF-2930](https://lite-farm.atlassian.net/browse/LF-2930)

**Description**
Adding tooltipContent functionality to Checkbox Form element

**What was done**
Pushed the tooltipContent attribute through Checkbox and into the existing Main component in /Typography.
Removed excess styling causing alignment issues when adding a tooltip.
Another component being used by checkbox: `<Main />`. It uses a toolTipContent as a bool so the 'undefined' was necessary to prevent always being rendered. Used `<PurePlantingDate />` as similar use of Main's tooltip.

**How it was tested**
1. Added _tooltipContent="your tooltip content here"_ to one Checkbox on a form and compared to one without tooltip. (Irrigation task is a good candidate for this).
2. Made sure alignment is good in multiple existing views that use `<Checkbox />` and `<Checkbox sm />` and also those same views with a tooltip.

### DIFF
--- a/packages/webapp/src/components/Form/Checkbox/index.jsx
+++ b/packages/webapp/src/components/Form/Checkbox/index.jsx
@@ -15,7 +15,7 @@ const Checkbox = ({
   hookFormRegister,
   errors,
   sm,
-  tooltipContent = '',
+  tooltipContent = null,
   ...props
 }) => {
   const name = hookFormRegister?.name ?? props?.name;
@@ -42,7 +42,7 @@ const Checkbox = ({
       <Main
         className={clsx(styles.label, sm && styles.smallLabel)}
         style={classes.label}
-        tooltipContent={tooltipContent ?? undefined}
+        tooltipContent={tooltipContent}
       >
         {label}
       </Main>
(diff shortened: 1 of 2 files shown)


packages/webapp/src/components/Form/Checkbox/index.jsx
```diff
@@ -15,6 +15,7 @@ const Checkbox = ({
   hookFormRegister,
   errors,
   sm,
+  tooltipContent = '',
```
SayakaOno, line 18: Can this be `tooltipContent = undefined` so that the line 45 can simply be `tooltipContent={tooltipContent}`?
I'm just wondering.

Duncan-Brain: Good question. I think it is more succinct to do it the way you are suggesting. 

To be honest I was mostly copying the PurePlantingDate component way of doing it. 

For some reason setting anything as undefined makes me uncomfortable for some undefinable reason.

Are you good if also switch it to null?

SayakaOno: Yes :)

SayakaOno (approved):

PR #2424, opened 2023-01-25
submission

baweinbe:


packages/api/db/migration/20230124222902_update_support_type_enum.js
```diff
@@ -0,0 +1,15 @@
+export const up = async function (knex) {
+  await knex.schema.alterTable('supportTicket', () => {
+    knex.raw(`ALTER TYPE support_type ADD VALUE 'Delete my farm'`);
```
baweinbe, line 3: According to the knex documentation (https://knexjs.org/guide/schema-builder.html#enum-enu) "Knex does not provide any way to alter enumerations after creation. To change an enumeration later on you must use Knex.raw, and the appropriate command for your database.". I looked up how to do so and found this syntax. 

NOTE: this did not work. It did not throw an error when the migration ran, but when I went and looked at the constraint check, it had not been updated. I ran this directly against the database and got an error that support type does not exist. 

Truthfully, I have never used Postgres (I've only used MySQL) and have always been told by those more experienced than me to stay away from the enum type when using MySQL.

packages/api/db/migration/20230124222902_update_support_type_enum.js
```diff
@@ -0,0 +1,15 @@
...
+  });
+};
+
+export const down = async function (knex) {
+  await knex.schema.alterTable('supportTicket', () => {
+    knex.raw(`
+            CREATE TYPE support_type1 AS ENUM ('Request information', 'Report a bug', 'Request a feature', 'Other');
```
baweinbe, line 10: I read that you cannot remove a value from an enum and found this approach. Not tested

packages/api/package.json
```diff
@@ -18,7 +18,7 @@
     "integration": "NODE_ENV=integration node src/server.js",
     "migrate:testing:db": "NODE_ENV=test knex migrate:latest --env=test --knexfile=./.knex/knexfile.js",
     "migrate:dev:db": "knex migrate:latest --knexfile=./.knex/knexfile.js",
-    "migrate:make:dev": "knex migrate:latest --knexfile=./.knex/knexfile.js",
+    "migrate:make:dev": "knex migrate:make --knexfile=./.knex/knexfile.js",
```
baweinbe, line 21: `migrate:make:dev` was the exact same thing as `migrate:dev:db` and I assumed this was a mistake. Looked up in knex documentation how to create a new migration file

packages/api/src/models/supportTicketModel.js
```diff
@@ -32,7 +32,13 @@ class supportTicketModel extends BaseModel {
         support_ticket_id: { type: 'string' },
         support_type: {
           type: 'string',
-          enum: ['Request information', 'Report a bug', 'Request a feature', 'Other'],
+          enum: [
```
baweinbe, line 35: Part of my attempt to add a value to the enum

packages/webapp/src/components/Help/index.jsx
```diff
@@ -11,15 +11,21 @@ import TextArea from '../Form/TextArea';
...
 import Radio from '../Form/Radio';
 import { Label } from '../Typography/index';
+import { useSelector } from 'react-redux';
+import { userFarmSelector } from '../../containers/userFarmSlice';
+import ConfirmModal from '../Modals/Confirm';
 
-export default function PureHelpRequestPage({ onSubmit, goBack, email, phone_number, isLoading }) {
+export default function PureHelpRequestPage({ onSubmit, goBack, isLoading }) {
   const [file, setFile] = useState(null);
+  const [showConfirmModal, setShowConfirmModal] = useState(false);
```
baweinbe, line 20: I'm not sure what the policy is in this code base for when to use `useState` and when to use `redux`. I saw another form that was doing it this and so I copied the approach because I thought it was simpler

packages/webapp/src/components/Help/index.jsx
```diff
@@ -11,15 +11,21 @@ import TextArea from '../Form/TextArea';
...
   const validEmailRegex = RegExp(/^$|^[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,4}$/i);
   const { register, handleSubmit, watch, control, setValue, formState } = useForm({
     mode: 'onTouched',
-    defaultValues: { 'contact_method': 'email' }
+    defaultValues: { contact_method: 'email' },
```
baweinbe, line 24: Auto format change

packages/webapp/src/components/Help/index.jsx
```diff
@@ -43,110 +60,124 @@ export default function PureHelpRequestPage({ onSubmit, goBack, email, phone_num
     const contactInformation = contactMethodSelection === 'email' ? email : phone_number;
     setValue(CONTACT_INFO, contactInformation);
   }, [contactMethodSelection]);
-  const submit = (data) => {
+
+  const confirmedSubmit = (data) => {
     data.support_type = data.support_type.value;
     data[data[CONTACT_METHOD]] = data.contactInfo;
     data.attachments = {};
+    data.farm_id = farm_id;
```
baweinbe, line 68: I noticed in the database that there is a column for this, but it wasn't getting sent. I thought it could be helpful. Also, my original approach was going to delete the farm as part of the create ticket operation on the back end and this is left over from that

packages/webapp/src/containers/Help/index.jsx
```diff
@@ -16,15 +15,7 @@ export default function HelpRequest() {
   const handleBack = () => {
     history.push('/');
   };
-  const { email, phone_number } = useSelector(userFarmSelector);
```
baweinbe, line 19: This got moved lower down because I needed more things from it

packages/webapp/src/containers/Help/saga.js
```diff
@@ -31,12 +32,26 @@ export function* supportFileUploadSaga({ payload: { file, form } }) {
     const formData = new FormData();
     formData.append('_file_', file);
     formData.append('data', JSON.stringify(form));
-    const result = yield call(axios.post, supportUrl(), formData, {
-      headers: {
-        'Content-Type': 'multipart/form-data',
-        Authorization: 'Bearer ' + localStorage.getItem('id_token'),
-      },
-    });
+    let result;
```
baweinbe, line 35: Because I could not get the enum to update in the database, I decided to make this an either or situation

packages/webapp/src/containers/Help/saga.js
```diff
@@ -31,12 +32,26 @@ export function* supportFileUploadSaga({ payload: { file, form } }) {
...
+    if (form.support_type === 'Delete my farm') {
```
baweinbe, line 36: I'm not a big fan of this being a string that needs to coincidentally match with the value defined in the form, but I didn't see an already established pattern for handling this situation

packages/webapp/src/containers/Help/saga.js
```diff
@@ -31,12 +32,26 @@ export function* supportFileUploadSaga({ payload: { file, form } }) {
...
+      const { farmUrl } = apiConfig;
+      let { user_id, farm_id } = yield select(userFarmSelector);
+      result = yield call(axios.delete, farmUrl + '/' + farm_id, {
```
baweinbe, line 39: This feels like it should be its own isolated function, but I did it inline here for convenience to show something working. I don't have very much experience with generator functions

setup.sh
```diff
@@ -0,0 +1,50 @@
+BREW_VERSION=$(brew -v)
```
baweinbe, line 1: Left over from other pull request

PR #2425, opened 2023-01-25
Lf 2889 allow farmers to nominate a custom crop to be added to core crop catalogue

Duncan-Brain:
[Jira Ticket LF-2889](https://lite-farm.atlassian.net/browse/LF-2889)
[Parent Jira Ticket LF-2072](https://lite-farm.atlassian.net/browse/LF-2072)

**Description**
This is the last piece for LF-2072. This adds the nomination_id column to the crop table as well as the frontend Checkbox in the add new crop workflow.

**What's Been done**
- Migrated crop table to add new column
- Updated cropController to retrieve generated nomination_id
- Added new function to nominationController for calls that do not come from the /nomination route
- Updated Model and Slice to include nomination_id
- Added Frontend UI
- Added translations

**To test**
1. Go to crop catalogue and start the add new crop workflow
2. Fill in the new crop details and check(or don't check) the box 
3. Complete the form flow through the varietal creation
4. Check the database in crop table, nomination and nomination_status. There should(or should not) be entries for the crop you entered.


JimmyRowland:
Nomination id shouldn't be in crop table. Add a connecting nomination-crop table
Addnominatio controller will send an additional response in crop endpoint.
I haven't looked into pr in detail. Will take another look when I am free (probably weekend ) if pr still needs review

Duncan-Brain:
Okay so my problems with automated tests seem to be based around the fact that objection validation was preventing null values from being valid values for nomination_id during updateCrop. 

So no test change was necessary, I allowed for setting null values which I think makes sense as an option.

Sorry I don't have local tests working hopefully you didn't get too many emails!

Duncan-Brain:
@JimmyRowland 

Will wait to push through until you have had a chance to review on the weekend but for your first two comments: 
- Junction table - this isn't a many-to-many relationship so I don't think crop-nominations will be necessary [https://stackoverflow.com/questions/7113295/sql-should-i-use-a-junction-table-or-not](https://stackoverflow.com/questions/7113295/sql-should-i-use-a-junction-table-or-not)
- Additional response from nomination controller - a new controller function was made to specifically avoid this and use a parent trx object. Just in case you thought this too: the baseController.postWithResponse() does not send a response it just returns data from the database.

JimmyRowland:
Relationship is not 1 to 1 either. 
Nomination can reference many tables. The best way is to create junction tables for each type of nomination.

Suppose a nomination for a crop is rejected, I think it makes more sense to create new nomination on the same crop if user decided to nominate the rejected crop. If a crop is not nominated, it doesn't make sense to have empty values in nomination id column. There might be nomination type specific values such as supplier, seed price, or any fields that are not already available in crop table.

Location table has same relationship with each location type.

Duncan-Brain:
@JimmyRowland 

Yeah thanks for the suggestion I think I see what you are saying now. The many null values is the big point for me. 

We were just going to let the nominations become orphans if they were deleted but this makes sense too. Let me try that out.

packages/api/src/controllers/nominationController.js
```diff
@@ -75,6 +75,31 @@ const nominationController = {
...
+    //TODO: Hopefully this gets changed/removed with workflow ranking
+    const { workflow_id } = await NominationWorkflowModel.query(trx)
+      .select('workflow_id')
+      .where('status', initialStatus)
+      .andWhere('type_group', nominationType)
+      .first();
+    data.workflow_id = workflow_id;
+    const nomination = await baseController.postWithResponse(NominationModel, data, req, {
+      trx,
+    });
```
SayakaOno, line 96: You might go in a different direction, but `data`(line 97) needs to have `nomination_id`.
UI looks good!

Duncan-Brain: Ah thanks!, Looks like I accidentally dropped it in commit 4293d61

JimmyRowland:
> We were just going to let the nominations become orphans

If a nomination is soft deleted, and crop nomination id is then overwritten, the soft deleted nomination is practically broken. It's better to move nomination to conjunction table or remove soft delete for all nomination-related tables to reduce maintenance work.

The series of insert for nomination tables can be a single upsertGraph if nomination_id is in a conjunction table. 

Also, I don't think you should worry about the eager loading warning on ObjectJS doc. 
1. The warning is not applicable to insert. 
2. It's very unlikely that multiple farm admin would edit the same object at the same time.
3. If upsertGraph implication fails, location/task endpoints would fail first.
4. Prisma has the same API and `relation query` does not have the same limitations. The backend needs to migrate to something else at some point when obejctionJS is no longer maintained.

Avoid variable rename
type_group -> nomination_type
Should rename the column to nomination_type.

workflow table is unnecessary complexity.

The Redux store is set up in a way to support offline use. The relationalDB tables need to be implemented as key-value store on the frontend. While designing DB schema, you also need to consider how redux store/selectors need to be structured.

In the current UI, if a user wants to nominate an existing crop. UI needs to fetch all nomination-related tables and calculate the nomination status of each crop. Having nomination_id  on crop table will not work. If crop is soft deleted, crop nomination selectors will break. If nomination is deleted when nomination id on crop table is not removed, crop nomination selectors will also break. I think it's better to create pages dedicated to nominations instead of having checkboxes in edit/create crop flow. The flow is too long and it's not good to have nomination inserts in crop endpoints.

I am not sure what's the current workflow. I think devs should add backend use case unit testings (both happy/unhappy paths), maybe in this pr or another pr. The process will reveal all the design flaws/pain points.

For example, previous nomination endpoints do not work with the current design. 
Add nomination endpoint will only add broken nominations. 
If a user creates a crop along with the nomination and then farm member uses put endpoint to change crop.nomination_id to a nomination.nomination_id in another farm. Then the crop entry and all crop-related selectors will be broken.
In this case, there should be a validation middleware to guard the put crop endpoint.
All duplicate farm_id in connected tables need extra care. 

Be prepared to redo the schema multiple times 😉. I lost count of how many times management plan had been redone

packages/api/src/controllers/nominationController.js
Duncan-Brain: See line 104 - linting likely moved it down a bit

Duncan-Brain:
@JimmyRowland 
- I have added the NominationCrop junction table
- I will consider eager loading, and other renaming suggestions for a future upgrade
- Redux store - Currently there is zero way to view on the frontend -- that may be upcoming and will create selectors slice refreshes as needed
-I agree with your nomination flow argument and it is creating a bad pattern -- I am going to try to push that with Kevin and have created [LF-2945](https://lite-farm.atlassian.net/browse/LF-2945) for you to view.
- I agree we need tests -- I think we are going to prioritize that soon but there has been no one to pass that process knowledge on to the previous devs and so no one passed it on to me. I have created a PR here to force my familiarity with it and test these PRs  see [LF-2946](https://lite-farm.atlassian.net/browse/LF-2946)

@SayakaOno 
- I fixed that missing line thanks for catching that!

Anything else needed on this one?

JimmyRowland (approved):

packages/api/src/models/nominationCropModel.js
```diff
@@ -0,0 +1,40 @@
...
+ *  LiteFarm is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  LiteFarm is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ */
```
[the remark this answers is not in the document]
Duncan-Brain: Lol thanks, how did that sneak in!

Duncan-Brain:
Correct copyrights are now added to all touched files minus json.

SayakaOno (approved):
Approved!

PR #2426, opened 2023-01-26
LF-2643: When a task is assigned to a user without a wage prompt to provide a wage

SayakaOno:
## Description
When a user without a wage is selected as task's assignee, the additional section shows up so that the user can provide a wage.
Screenshot 2023-01-25 at 11 38 29 PM

Jira link: https://lite-farm.atlassian.net/browse/LF-2643

- This section should be shown if these condition below are met:
   - the user is admin
   - the assignee does not have a wage
   - the task does not have `wage_at_moment`
- The hourly wage input should be displayed only if "Yes, set hourly wage" or "Yes, just for this task" option is selected.

## What does this PR do?
- add `wage_do_not_ask_again` column to `userFarm` table
- create APIs
   - `/user_farm/wage_do_not_ask_again/farm/:farm_id/user/:user_id`
   - `/task/patch_wage/:task_id`
- modify styles of the info icon in `Tooltip` component (it was being affected by a className of its parent component and affecting another component)
- add the hourly wage section

## How Has This Been Tested?
Login as admin, go to `/tasks` and:
1.
- click assignee of a task that has `wage_at_moment`. The wage section is hidden regardless of the selected worker
- click assignee of a task that does not have `wage_at_moment`
  - assignee input
    - select "Unassigned" -> the wage section is hidden
    - select a worker with wage -> the wage section is hidden
    - select a worker without wage
       - the wage section is shown.
         - select Yes option. The wage input is shown. An error message is shown if `number < 0 || 999,999,999 < number`
            - "yes, set hourly wage" and click "Update" -> `wage` in `userFarm` table(DB) is updated with the wage inputted
            - "yes, just for this task" and click "Update" -> `wage_at_moment` in `task` table(DB) is updated with the wage inputted
            - "no" and click "Update" -> nothing happens except for task assigning
            - "no, do not ask again" -> `wage_do_not_ask_again` in `userFarm` table(DB) is updated to `True`
    - select a worker that has had “Don’t ask again” set is chosen -> the wage section is hidden

2.
- click a task that has `wage_at_moment`. Click edit icon for the assignee input. The wage section is hidden regardless of the selected worker
- click a task that does not have `wage_at_moment`. The same behaviours are above is seen.

**When logged in as farm worker, the wage section should not be shown.** 

## Note
Task assigning should work as before. (functionality was not updated)
There are two existing bugs that will not be addressed in this PR.
- [LF-2932](https://lite-farm.atlassian.net/browse/LF-2932) - Multiple tasks cannot be unassigned at once
- [LF-2935](https://lite-farm.atlassian.net/browse/LF-2935) - Completed task's assignee is clickable although it cannot be updated

[LF-2932]: https://lite-farm.atlassian.net/browse/LF-2932?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/api/db/migration/20230124_add_wage_do_not_ask_again_column_to_userFarm.js
```diff
@@ -0,0 +1,30 @@
...
+ *
+ *  LiteFarm is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ */
+
+export const up = function (knex) {
+  return Promise.all([
+    knex.schema.table('userFarm', (table) => {
```
Duncan-Brain, line 18: In both up and down functions consider using the alterTable() knex function instead of table() for readability.

packages/api/src/controllers/taskController.js
```diff
@@ -167,6 +167,27 @@ const taskController = {
     }
   },
 
+  async patchWage(req, res) {
+    try {
+      const { task_id } = req.params;
+      const { wage_at_moment } = req.body;
+
+      //Ensure only adminRoles can modify task wage
+      if (!adminRoles.includes(req.role)) {
```
Duncan-Brain, line 176: I think this same thing be accomplished by adding another checkScope('edit:user_wage?') to the taskRoute.

I see there is another patch in this controller above this one that may handling permissions dangerously.

packages/api/src/routes/userFarmRoute.js
```diff
@@ -110,6 +110,13 @@ router.patch(
   userFarmController.updateWage(),
 );
 
+router.patch(
+  '/wage_do_not_ask_again/farm/:farm_id/user/:user_id',
+  hasFarmAccess({ params: 'farm_id' }),
+  checkScope(['edit:user_wage']),
```
Duncan-Brain, line 116: I don't know the answer to this one --- Is there better or different scope that maybe should be used like a "user_setting" or something that would be more appropriate since no user wage is being set?

packages/api/tests/task.test.js
```diff
@@ -2339,4 +2349,53 @@ describe('Task tests', () => {
...
+
+    test('EO must be able to patch task wage', async (done) => {
+      testWithAdminRoleAndValidInput(5, 27, done);
+    });
+
+    test('Managers must be able to patch task wage', async (done) => {
+      testWithAdminRoleAndValidInput(2, 37, done);
+    });
+
+    test('Farm worker must not be able to patch task due date', async (done) => {
```
Duncan-Brain, line 2384: Test name should be wage instead of task due date. And just wondering if you could reuse the same function testAdminRoleWithValidInput() here by splitting the expect statement by roleId.

packages/api/tests/userFarm.test.js
```diff
@@ -1013,6 +1022,55 @@ describe('User Farm Tests', () => {
...
+
+      test('Manager should be able to set wage_do_not_ask_again', async (done) => {
+        testWithAdminRole(2, 3, done);
+      });
+
+      test('EO should be able to set wage_do_not_ask_again', async (done) => {
+        testWithAdminRole(5, 3, done);
+      });
+
+      test('Farm worker should not be able to set wage_do_not_ask_again', async (done) => {
```
Duncan-Brain, line 1059: Same as other test - since this is so similar to testWithAdminRole() could it be simply testWithRole() and split expect statements ?

packages/webapp/src/components/Modals/QuickAssignModal/index.jsx
```diff
@@ -60,24 +113,134 @@ export default function TaskQuickAssignModal({
...
+        </Label>
+        <Main style={{ marginBottom: 10 }}>
+          {t('ADD_TASK.HOURLY_WAGE.WANT_TO_SET_HOURLY_WAGE')}
+        </Main>
+        <RadioGroup
+          hookFormControl={control}
+          name={HOURLY_WAGE_ACTION}
+          radios={radioOptions}
+          data-cy="quickAssign-hourlyWageAction"
+          style={{ marginBottom: 10 }}
```
Duncan-Brain, line 220: I am personally open to changing this standard but most of the app explicitly states the units 'px'

packages/api/src/routes/userFarmRoute.js
SayakaOno: I don't think there's an option other than creating an appropriate one. What do you think about adding a permission for `edit:user_wage_do_not_ask_again`?

packages/api/src/controllers/taskController.js
SayakaOno: Thank you, that's much better.

Do you mean we should add a permission for `patchTaskDate` so we can use `checkScope`?

packages/webapp/src/components/Modals/QuickAssignModal/index.jsx
```diff
@@ -103,21 +266,30 @@ export default function TaskQuickAssignModal({
...
+            defaultValue={selectedWorker}
+            label={t('ADD_TASK.ASSIGNEE')}
+            options={options}
+            onChange={onChange}
+            style={{ marginBottom: 10 }}
+            isSearchable
+          />
+        )}
       />
+      {renderHourlyWageSection()}
```
Duncan-Brain, line 284: Please tell me if this is a bad idea or you do not want to do it. But I am thinking that this could get reused in the create task workflow since this same problem likely happens so it could maybe be its own separate component?

packages/api/tests/task.test.js
SayakaOno: Good catch, thank you! I agree, I think reusing the function is better.

packages/webapp/src/components/Modals/QuickAssignModal/styles.module.scss
```diff
@@ -29,3 +29,6 @@
   width: 96px;
 }
 
+.warning {
+  color: var(--brown700);
```
Duncan-Brain, line 33: Couldn't find the UI specs on the Jira ticket -- we do have a colors.scss file that has a Litefarm standard warning color. Ignore this if this color was specifically asked for.

Duncan-Brain:
@SayakaOno 

Everything is seems to be working well for me as far as expected functionality.

Other things I noticed:
1.  I have added some inline comments to the code
2.  For the migration naming convention you can use:  `knex migrate:make migration_name` to get the format output correctly
3.  UI/UX - I didn't see a Figma file attached to the Jira to see pixel spacing between elements but it looks a little squished to me. Might be worth checking with Carolina on that.
4. For the tooltip -- I get it looks goofy so big haha but I think if we put that on a separate PR that would be best that way if the display change affects other instances we can check that. And my understanding about !important is that it should be avoided in most scenarios. A new Jira ticket for this issue I think would just help robustly improve tooltip.

Testing: 
I love that you added tests but I don't have mine set up just yet, I think I am almost there. But if it works for you I would approve the way it is currently looks good to me.

packages/api/src/controllers/taskController.js
Duncan-Brain: I meant that checkScope() checks and verifies the role right. Since this is editing the user_wage that the existing edit:user_wage permission can be added to the patchwage task route you made.

packages/api/src/routes/userFarmRoute.js
Duncan-Brain: I think an existing scope should be good. I am seeing the onboarding flags also in the userFarm table are set without permission in userFarmRoute -- which seems weird to me. Maybe one of the others used in userFarmRoute?

SayakaOno: We only have these scopes, and I can't reuse any of them here. The first one includes FW, which should not, the rest are dedicated to each column. I'll create a new one.
Screenshot 2023-02-02 at 4 32 46 PM

packages/webapp/src/components/Modals/QuickAssignModal/index.jsx
SayakaOno: I prefer to take advantage of not having to type `''` and units, but I'll follow what's been done for now.

packages/webapp/src/components/Modals/QuickAssignModal/index.jsx
SayakaOno: I'll extract this and create a reusable component in my next ticket.

SayakaOno:
@Duncan-Brain 
Thank you for reviewing this big PR!

 1. 🙏 
> 2. For the migration naming convention you can use: knex migrate:make migration_name to get the format output correctly

I got this error: `command not found: knex`. I'll try it next time and ask you for help if I don't figure out how to make it work.

3.  👍 

> 4. For the tooltip

Yeah, the style is not great. I'll create an enhancement ticket. `!important` is necessary here unless we improve the "Assign task" modal to specify which svg should be `font-size: 20px;`.

I don't think we need to work on it just to remove `!important`.
Screenshot 2023-02-02 at 5 03 45 PM

packages/api/src/routes/userFarmRoute.js
Duncan-Brain: Yeah a FW would never see the prompt in the UI but really it is just a userFarm default setting which I think even a FW should be able to control on their own account? This is a weird one

Duncan-Brain:
> I got this error:` command not found: knex`
You'll just need to install knex globally. For now you can just change the name manually with some random time during the day. If the name stays short it wont list chronologically in folders, VBCode etc.

Tooltip:
My concern is that by editing Tooltip directly it may be affecting the UI/UX of Tooltips across the whole app. If there is some example where the specification for style was 14px and was being correctly ordered in the hierarchy before -- this may affect instances like that. Is there any way you can pass style for this specific instance -- maybe just for now.

packages/api/src/routes/userFarmRoute.js
SayakaOno: Okay, you must be right :) I'll go with edit:users

SayakaOno:
@Duncan-Brain 
I understand your concern. I'll try to override it using a scss file.

packages/webapp/src/components/Modals/QuickAssignModal/styles.module.scss
SayakaOno: I checked figma, and this color was required.

packages/webapp/src/components/Modals/QuickAssignModal/index.jsx
```diff
@@ -36,13 +68,34 @@ export default function TaskQuickAssignModal({
     } else return [selfOption, unAssignedOption];
   }, []);
 
-  const [selectedWorker, setWorker] = useState(isAssigned ? unAssignedOption : selfOption);
-  const [assignAll, setAssignAll] = useState(false);
+  const [showHourlyWageSection, setShowHourlyWageSection] = useState(false);
+  const [showHourlyWageInput, setShowHourlyWageInput] = useState(false);
+
```
JimmyRowland, line 73: These two useState can just be calculated

packages/webapp/src/components/Modals/QuickAssignModal/index.jsx
```diff
@@ -36,13 +68,34 @@ export default function TaskQuickAssignModal({
...
+    control,
+    register,
+    resetField,
+    watch,
+    formState: { isValid, errors },
+  } = useForm({
+    mode: 'onTouched',
+    defaultValues: {
+      [ASSIGNEE]: isAssigned ? unAssignedOption : selfOption,
+      [HOURLY_WAGE_ACTION]: '',
```
JimmyRowland, line 84: Shouldn't HOURLY_WAGE_ACTION have a default value?

packages/webapp/src/components/Modals/QuickAssignModal/index.jsx
```diff
@@ -36,13 +68,34 @@ export default function TaskQuickAssignModal({
...
+    },
+  });
+
+  const selectedWorker = watch(ASSIGNEE);
+  const selectedHourlyWageAction = watch(HOURLY_WAGE_ACTION);
+  const hourlyWage = watch(HOURLY_WAGE);
+  const assignAll = watch(ASSIGN_ALL);
 
   const tasks = useSelector(tasksSelector);
+  const currencySymbol = grabCurrencySymbol(getCurrencyFromStore());
```
JimmyRowland, line 96: getCurrencyFromStore is deprecated.
It was used as walk around in some old class component.

packages/webapp/src/components/Modals/QuickAssignModal/index.jsx
```diff
@@ -60,24 +113,134 @@ export default function TaskQuickAssignModal({
...
+        />
+        {showHourlyWageInput && (
+          <Input
+            unit={currencySymbol + t('ADD_TASK.HR')}
+            data-cy="quickAssign-hourlyWage"
+            label={t('WAGE.HOURLY_WAGE')}
+            step="0.01"
+            type="number"
+            onKeyPress={numberOnKeyDown}
+            hookFormRegister={register(HOURLY_WAGE, {
```
JimmyRowland, line 230: {
    valueAsNumber: true,
  }
  https://react-hook-form.com/api/useform/register/

packages/webapp/src/components/Modals/QuickAssignModal/index.jsx
```diff
@@ -60,24 +113,134 @@ export default function TaskQuickAssignModal({
...
+      label: t('common:NO'),
+      value: hourlyWageActions.NO,
+    },
+    {
+      label: t('ADD_TASK.HOURLY_WAGE.DONT_ASK'),
+      value: hourlyWageActions.DO_NOT_ASK_AGAIN,
+    },
+  ];
+
+  const renderHourlyWageSection = () => {
```
JimmyRowland, line 202: Should create a new component or useMemo/useCallback

component overhead is bigger than the function. component will be recreated on each render

packages/webapp/src/components/Modals/QuickAssignModal/index.jsx
```diff
@@ -60,24 +113,134 @@ export default function TaskQuickAssignModal({
...
+  }, [user.is_admin, selectedWorker, wageAtMoment]);
+
+  useEffect(() => {
+    if (!user.is_admin) {
+      return;
+    }
+
+    resetField(HOURLY_WAGE_ACTION);
+    resetField(HOURLY_WAGE);
+  }, [user.is_admin, selectedWorker]);
```
JimmyRowland, line 168: Inputs usually don't reset

packages/webapp/src/components/Modals/QuickAssignModal/index.jsx
```diff
@@ -60,24 +113,134 @@ export default function TaskQuickAssignModal({
...
+            onKeyPress={numberOnKeyDown}
+            hookFormRegister={register(HOURLY_WAGE, {
+              valueAsNumber: true,
+              min: { value: 0, message: t('WAGE.HOURLY_WAGE_RANGE_ERROR') },
+              max: { value: 999999999, message: t('WAGE.HOURLY_WAGE_RANGE_ERROR') },
+            })}
+            style={{ marginBottom: '24px' }}
+            errors={errors[HOURLY_WAGE] && (errors[HOURLY_WAGE].message || t('WAGE.ERROR'))}
+            toolTipContent={t('WAGE.HOURLY_WAGE_TOOLTIP')}
+            optional
```
JimmyRowland, line 238: It's odd that wage is optional when user selects input wage radio button

packages/webapp/src/components/Modals/QuickAssignModal/index.jsx
```diff
@@ -103,21 +266,30 @@ export default function TaskQuickAssignModal({
...
-        onChange={setWorker}
-        style={{ marginBottom: '24px' }}
-        isSearchable
+      <Controller
+        control={control}
+        name={ASSIGNEE}
+        render={({ field: { onChange } }) => (
+          <ReactSelect
+            data-cy="quickAssign-assignee"
+            defaultValue={selectedWorker}
```
JimmyRowland, line 275: Shouldn't ReactSelect automatically update the default option when selectedWorker is set in react hook form?

packages/webapp/src/containers/Task/taskCardContentSelector.js
```diff
@@ -23,6 +23,7 @@ const getTaskContents = (tasks, userFarmEntities, { farm_id }) => {
       happiness: task.happiness,
       abandon_date: task.abandon_date,
       date: task.abandon_date || task.complete_date || task.due_date,
+      wageAtMoment: task.wage_at_moment,
```
JimmyRowland, line 26: all database fields are in snake_case

JimmyRowland:
Don't forget to squash merge

JimmyRowland (dismissed):

packages/webapp/src/components/Modals/QuickAssignModal/index.jsx
SayakaOno: I found a way to do it, thank you!

SayakaOno:
@JimmyRowland
Thank you so much for reviewing! I am going to raise another PR against this branch to extract some part of `QuickAssignModal` as a new component. I asked some questions related to the new component in my comments, it would be great if you could answer those. I would also like you to review my PR for the new component later. (I will address the rest of your comments after creating the PR) Thank you again!

packages/webapp/src/components/Modals/QuickAssignModal/index.jsx
```diff
@@ -60,24 +113,134 @@ export default function TaskQuickAssignModal({
...
+    if (selectedHourlyWageAction) {
+      shouldShowWageInput = isYesOptionSelected(selectedHourlyWageAction);
+    }
+    if (!shouldShowWageInput) {
+      resetField(HOURLY_WAGE);
+    }
+
+    setShowHourlyWageInput(shouldShowWageInput);
+  }, [selectedHourlyWageAction]);
+
```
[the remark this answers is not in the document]
JimmyRowland: I think checking selected radio option before calling apis is enough. I am not sure whether reset is desired behavior
In other forms, inputs are not reset when they are hidden.
For example, in management plan flow, a user first chooses broadcast.
image
fill out the form
image
If the user goes back to choose row planting method and then goes back to broadcast, form values are still there. The user doesn't have to refill the form.

Whether input should be reset depends on requirements. In the past, inputs don't reset when inputs are hidden. Maybe  this ticket is different?

packages/webapp/src/components/Modals/QuickAssignModal/index.jsx
JimmyRowland: {
valueAsNumber: true,
} props can be used so that react-hook-form will treat HOURLY_WAGE as a number and the value won't have be to parsed to number manually

packages/webapp/src/components/Modals/QuickAssignModal/index.jsx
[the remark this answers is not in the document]
SayakaOno: Thank you for the example. It is not specified if we should keep user inputs or not in the ticket, but I don't think inputs need to be reset as long as the check for selected radio option before calling apis is reliable. I'll follow the existing components' behaviour.

packages/api/src/controllers/userFarmController.js
```diff
@@ -485,6 +485,31 @@ const userFarmController = {
     };
   },
 
+  setWageDoNotAskAgain() {
+    return async (req, res) => {
+      const trx = await transaction.start(Model.knex());
```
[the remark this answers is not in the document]
SayakaOno: Thank you! I need to study this area...

packages/api/db/migration/20230203164549_add_wage_do_not_ask_again_column_to_userFarm.js
```diff
@@ -0,0 +1,30 @@
...
+ *  LiteFarm is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ */
+
+export const up = function (knex) {
+  return Promise.all([
+    knex.schema.alterTable('userFarm', (table) => {
+      table.boolean('wage_do_not_ask_again').nullable().defaultTo(false);
```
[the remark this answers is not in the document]
SayakaOno: Removed defaultTo.

SayakaOno:
@JimmyRowland 
> Don't forget to squash merge
It looks like squash is not enabled for this repo. I will ask Kevin about it.

I have addressed your comments for the backend. It would be great if you could review my changes.
Thank you so much again!

@Duncan-Brain 
I decided to create a new component in this PR and made changes. Could you please re-review?
[This](https://github.com/LiteFarmOrg/LiteFarm/pull/2449) is the main changes I made after you reviewed and it might make a review easier. Thank you so much!

JimmyRowland (approved):

packages/api/db/migration/20230203164549_add_wage_do_not_ask_again_column_to_userFarm.js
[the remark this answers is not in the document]
kcussen: Thank you!

packages/api/src/controllers/userFarmController.js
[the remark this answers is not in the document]
Duncan-Brain: @JimmyRowland we still need a transaction trx object though right?

Are you saying that you prefer the first mentioned syntax on objection:
```
try {
  const returnValue = await Person.transaction(async trx => {
    // Here you can use the transaction.

    // Whatever you return from the transaction callback gets returned
    // from the `transaction` function.
    return 'the return value of the transaction';
  });
  // Here the transaction has been committed.
} catch (err) {
  // Here the transaction has been rolled back.
}
```
Over the alternative one described right below also on objection that we use regularly in the code?:
```
const { transaction } = require('objection');

const trx = await Person.startTransaction();

try {
  // If you created the transaction using `Model.startTransaction`, you need
  // commit or rollback the transaction manually.
  await trx.commit();
} catch (err) {
  await trx.rollback();
  throw err;
}
```

Duncan-Brain:
@SayakaOno 

Its all working for me. My big two questions are about two of Jimmy's comments:

- (optional) tag on wage input (From other PR): I agree 100% but check in with Kevin and Mwaya about this or it may get sent back from validation
- The trx object is now gone from the controller: I think we still need a transaction here it's just that he prefers a different syntax. I might be wrong about that but have left a comment in the line itself.

JimmyRowland:
@Duncan-Brain 
TRX is only needed if there are multiple updates in an endpoint. 
If there are two updates in an endpoint and second fails, the first one would be rolled back on TRX.rollback.

If there is only one update, the db update wouldn't go through if there is any error.

SayakaOno:
@Duncan-Brain 
Thank you for reviewing!

>  tag on wage input
I'll confirm with them!

> trx
My understanding about `trx` is that we need it when we modify/add data in/to multiple tables. If any error happens during the transaction, we `rollback` not to add any piece of data to the DB for consistency, and if it is successful, we `commit` to save all data. The API I created does only one operation, so `trx` is not necessary.

Jimmy has already replied, but since I typed, I'll leave it :)

Duncan-Brain (approved):
Approved - If (optional) tag is needed will re-approve if necessary

SayakaOno:
I confirmed that the wage input should be required. It's ready for merging :)

PR #2432, opened 2023-01-31
Lf 2934 upon loading select the task date screen date field should be pre selected

MwayaB:
## Description

Upon loading "Select the task date" screen, "Date" field should be pre-selected

Jira link: 
https://lite-farm.atlassian.net/browse/LF-2934
## Type of change

Please delete options that are not relevant.

## How Has This Been Tested?
Test plan:
https://lite-farm.atlassian.net/l/cp/nVyfZeR2

- [ ] Passes test case

**Test Configuration**:
* Firmware version:
* Hardware:
* Toolchain:
* SDK:

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [ ] Any dependent changes have been merged and published in downstream modules
- [x] I have checked my code and corrected any misspellings


packages/webapp/src/components/Task/TaskDate/index.jsx
```diff
@@ -58,6 +58,7 @@ export default function PureTaskDate({
       <Main style={{ marginBottom: '16px' }}>{t('TASK.SELECT_DATE')}</Main>
 
       <Input
+        autoFocus
```
SayakaOno, line 61: Maybe this can be used in the Input component?

packages/webapp/src/components/Form/Input/index.jsx
```diff
@@ -106,6 +114,7 @@ const Input = ({
       {unit && <div className={styles.unit}>{unit}</div>}
       {currency && <div className={styles.currency}>{currency}</div>}
       <input
+        onFocus={inputType === 'date' ? showPickerHandler() : null}
```
SayakaOno, line 117: We don't always want a date input to be focused, so I think this component should take `autoFocus`(I used what you passed from `TaskDate` component) as prop.

I don't think those date inputs in the screenshot below need focus on them, and we should be able to focus on one of them when we want to.
Screenshot 2023-02-02 at 11 40 28 AM

SayakaOno (changes requested):

Duncan-Brain:
@MwayaB 
When testing manually I had an issue where it was staying open even after selecting a date. I can also confirm it also looks like this would apply to all views beyond just tasks.

It looks like there is a line where it could get stuck in the open position. 

I would consider finding an effect that can run only once based upon a configurable condition.

packages/webapp/src/components/Form/Input/index.jsx
```diff
@@ -58,8 +59,20 @@ const Input = ({
...
+      input.current.showPicker();
+    } catch (e) {
+      console.log({ e });
+    }
+  };
+
   const onKeyDown = ['number', 'decimal'].includes(type) ? numberOnKeyDown : undefined;
 
+  useEffect(() => {
+    name === 'due_date' ? showPickerHandler() : null;
```
Duncan-Brain, line 73: It looks like you are creating a new feature on Input with an attribute called autoFocus. I think if you carry forward that option as a boolean here, and scrap the try catch and console.log then you'll have a succinct and reusable feature.

packages/webapp/src/components/Form/Input/index.jsx
```diff
@@ -58,8 +59,20 @@ const Input = ({
     setShowError(false);
   };
 
+  const showPickerHandler = () => {
+    try {
+      input.current.showPicker();
+    } catch (e) {
+      console.log({ e });
```
SayakaOno, line 66: console.log(e);

packages/webapp/src/components/Form/Input/index.jsx
```diff
@@ -12,6 +12,7 @@ import { ReactComponent as Leaf } from '../../../assets/images/signUp/leaf.svg';
 import Infoi from '../../Tooltip/Infoi';
 import { get } from 'react-hook-form';
 import i18n from '../../../locales/i18n';
+import { use } from 'i18next';
```
SayakaOno, line 15: This can be removed?

packages/webapp/src/components/Form/Input/index.jsx
SayakaOno: I think what Duncan means is:
```
 useEffect(() => {
    if (autoFocus) {
        try {
           input.current.showPicker();
         } catch (e) {
           console.log(e);
         }
  }, []);
```

and I know someone wants to use this feature for a different ticket.

SayakaOno (changes requested):

packages/webapp/src/components/Form/Input/index.jsx
gursimran-singh: I agree with @SayakaOno addressing this requirement in one of the tickets i am working on. There we also need to focus on the input showing the date picker. So we can add this line of code "input.current.focus();":
`useEffect(() => {

    if (autoFocus) {
        try {
           input.current.showPicker();
           input.current.focus();
         } catch (e) {
           console.log(e);
         }
  }, [autoFocus]);`

packages/webapp/src/components/Form/Input/index.jsx
```diff
@@ -58,8 +58,20 @@ const Input = ({
     setShowError(false);
   };
 
+  const showPickerHandler = () => {
+    try {
+      input.current.showPicker();
```
gursimran-singh, line 63:
```suggestion
      input.current.showPicker();
      input.current.focus();
```

packages/webapp/src/components/Form/Input/index.jsx
```diff
@@ -60,6 +61,17 @@ const Input = ({
 
   const onKeyDown = ['number', 'decimal'].includes(type) ? numberOnKeyDown : undefined;
 
+  useEffect(() => {
+    if (autoFocus) {
+      try {
+        input.current.focus();
```
SayakaOno, line 67: Even if the focus is on the input, you cannot start typing while the calendar is open, which I don't think is a good user experience.(the focus is on the calendar, and you can select a date using your keyboard) I would like to discuss the UX, so I will bring this up at standup.

SayakaOno: Never mind, it looks like that is what it is. I will approve!

SayakaOno (approved):

Duncan-Brain:
@MwayaB @SayakaOno @gursimran-singh 

Looking really good. I can approve, but I would say that I think Mwaya was right to use the autoFocus input prop ([see Mozilla](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autofocus)) instead of `input.current.focus(); `. 

Since autoFocus is a globalAttribute for `<input/>` we should avoid using that name and choose something like 'openCalendar'. And so what that would look like would be: 

in TaskDate/index
`<Input
        autoFocus openCalendar ........ />`
        
in Input/index
```
const Input = ({
  openCalendar,
  .....
  ...props // <--- autoFocus pushed through here

useEffect(() => {
    if (openCalendar) {
        try {
           input.current.showPicker();
         } catch (e) {
           console.log(e);
         }
  }, [openCalendar]);
```

SayakaOno:
@MwayaB @Duncan-Brain 
I didn't know that `autoFocus` was global attribute, sorry!
We need to add `autoFocus` to `<input />`(line 120) then:

```
<input
    autoFocus={autoFocus}
    disabled={disabled}
    ...
/>
```

MwayaB:
@Duncan-Brain that makes sense making that change now

packages/webapp/src/components/Form/Input/index.jsx
```diff
@@ -60,6 +62,17 @@ const Input = ({
...
+  useEffect(() => {
+    if (openCalendar) {
+      try {
+        input.current.focus();
+        input.current.showPicker();
+      } catch (e) {
+        console.log(e);
+      }
+    }
+  }, [autoFocus]);
```
Duncan-Brain, line 74: @MwayaB this would need to be openCalendar

gursimran-singh:
`input.current.focus();` on line no. 68 can be removed as it will be taken care by when `autoFocus` is passed through `props`. 
 
Also `autoFocus` on line no. 18 can be removed as it is automatically applied to input using `{...props}` on line no. 162

packages/webapp/src/components/Form/Input/index.jsx
```diff
@@ -14,6 +14,8 @@ import { get } from 'react-hook-form';
 import i18n from '../../../locales/i18n';
 
 const Input = ({
+  openCalendar,
+  autoFocus,
```
Duncan-Brain, line 18: @MwayaB 
This shouldn't be necessary anymore either if you remember the way you had it originally it was getting passed down through {...props} the way you had it before.

packages/webapp/src/components/Form/Input/index.jsx
```diff
@@ -60,6 +61,16 @@ const Input = ({
...
 
+  useEffect(() => {
+    if (openCalendar) {
+      try {
+        input.current.showPicker();
+      } catch (e) {
+        console.log(e);
+      }
+    }
+  }, []);
```
Duncan-Brain, line 72: @MwayaB 

Sorry if I wasn't clear in my last comment! The useEffect() hook just needs to have the dependent variables added to its dependencies so [openCalendar] would go where you had autoFocus previously in this line. The linting rule React made for it is called exhaustive-deps in case your interested... one of these days we should declare it to our linter.

SayakaOno (approved):

Duncan-Brain (approved):
Looks great! Approved!

If it hasn't already been done can we create a JIRA ticket that references this to do research to find an always open date picker? Just since it is a non recommended pattern from Mozilla.

gursimran-singh (approved):

PR #2449, opened 2023-02-05
LF-2643(2): Create assign task component

SayakaOno:
## Description
This is part of [LF2643](https://lite-farm.atlassian.net/browse/LF-2643)
(parent PR: https://github.com/LiteFarmOrg/LiteFarm/pull/2426)
- Create `AssignTask` component and `useTaskAssignForm` hook and use them in `QuickAssignModdal`
  (they are going to be used in another component later on)

### DIFF
--- a/packages/webapp/src/components/Modals/QuickAssignModal/index.jsx
+++ b/packages/webapp/src/components/Modals/QuickAssignModal/index.jsx
@@ -23,7 +23,7 @@ export default function TaskQuickAssignModal({
   onSetUserFarmWageDoNotAskAgain,
   users,
   user,
-  wageAtMoment,
+  wage_at_moment,
 }) {
   const { t } = useTranslation();
 
@@ -45,7 +45,7 @@ export default function TaskQuickAssignModal({
     users,
     isAssigned,
     additionalFields: { [ASSIGN_ALL]: false },
-    wageAtMoment,
+    wage_at_moment,
   });
 
   const assignAll = watch(ASSIGN_ALL);
--- a/packages/webapp/src/components/Modals/QuickAssignModal/index.jsx
+++ b/packages/webapp/src/components/Modals/QuickAssignModal/index.jsx
@@ -1,35 +1,15 @@
-import React, { useMemo, useState, useEffect } from 'react';
+import React from 'react';
 import { useTranslation } from 'react-i18next';
-import { useForm, Controller } from 'react-hook-form';
 import ModalComponent from '../ModalComponent/v2';
 import styles from './styles.module.scss';
+import { useSelector } from 'react-redux';
 import Button from '../../Form/Button';
-import ReactSelect from '../../Form/ReactSelect';
 import Checkbox from '../../Form/Checkbox';
-import Input, { numberOnKeyDown } from '../../Form/Input';
-import RadioGroup from '../../Form/RadioGroup';
 import { ReactComponent as Person } from '../../../assets/images/task/Person.svg';
 import { tasksSelector } from '../../../containers/taskSlice';
-import { useSelector } from 'react-redux';
-import { Label, Main } from '../../Typography';
-import { getCurrencyFromStore } from '../../../store/getFromReduxStore';
-import grabCurrencySymbol from '../../../util/grabCurrencySymbol';
-
-const ASSIGNEE = 'assignee';
-const HOURLY_WAGE = 'hourly_wage';
-const HOURLY_WAGE_ACTION = 'hourly_wage_action';
-const ASSIGN_ALL = 'assign_all';
-const hourlyWageActions = {
-  SET_HOURLY_WAGE: 'set_hourly_wage',
-  FOR_THIS_TASK: 'for_this_task',
-  NO: 'no',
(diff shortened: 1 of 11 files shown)


packages/webapp/src/components/Modals/QuickAssignModal/index.jsx
```diff
@@ -126,7 +83,7 @@ export default function TaskQuickAssignModal({
           assignee_user_id: assigneeUserId,
         });
 
-    if (isYesOptionSelected(selectedHourlyWageAction)) {
+    if (shouldSetWage) {
       const wage = +parseFloat(hourlyWage).toFixed(2);
```
JimmyRowland, line 87: if {valueAsNumber: true} is passed as option in hourlywage register, HOURLY_WAGE will be treated as a number by  react-hook-form and the value won't have be to parsed to number manually

packages/webapp/src/components/Modals/QuickAssignModal/index.jsx
```diff
@@ -141,106 +98,19 @@ export default function TaskQuickAssignModal({
...
-          />
-        )}
-      </div>
-    );
-  };
+  {
+    /*TODO: properly fix checkbox label overflow ST-272*/
+    /*TODO: LF-2932 - need to be able to unassign mutiple tasks at once */
+  }
+  const assignAllCheckbox = (
```
JimmyRowland, line 105: useMemo(()=>{}, [selectedWorker.label])

JimmyRowland: https://twitter.com/0xca0a
This guy always preaches not to declare components in the component body since components will always be recreated on each render.

packages/webapp/src/components/Task/AssignTask/HourlyWageInputs.jsx
```diff
@@ -0,0 +1,81 @@
...
+          unit={currencySymbol + t('ADD_TASK.HR')}
+          data-cy="hourlyWageInputs-wage"
+          label={t('WAGE.HOURLY_WAGE')}
+          step="0.01"
+          type="number"
+          onKeyPress={numberOnKeyDown}
+          hookFormRegister={register(HOURLY_WAGE, {
+            valueAsNumber: true,
+            min: { value: 0, message: t('WAGE.HOURLY_WAGE_RANGE_ERROR') },
+            max: { value: 999999999, message: t('WAGE.HOURLY_WAGE_RANGE_ERROR') },
```
JimmyRowland, line 70: if {valueAsNumber: true}

packages/webapp/src/components/Task/AssignTask/index.jsx
```diff
@@ -0,0 +1,86 @@
...
+  control,
+  selectedWorker,
+  errors,
+  showHourlyWageInputs,
+  shouldSetWage,
+  currency,
+}) => {
+  const { t } = useTranslation(['translation']);
+
+  const AssigneeSelect = (
```
JimmyRowland, line 39: useMemo(()=> {},[assigneeOptions, optional, showHourlyWageInputs, selectedWorker.label])

packages/webapp/src/components/Task/AssignTask/useTaskAssignForm.jsx
```diff
@@ -0,0 +1,171 @@
...
+  }, [user.is_admin, selectedWorker]);
+
+  const shouldSetWage = useMemo(() => {
+    let shouldSet = false;
+
+    if (selectedHourlyWageAction) {
+      shouldSet = isYesOptionSelected(selectedHourlyWageAction);
+    }
+    if (!shouldSet) {
+      resetField(HOURLY_WAGE);
```
JimmyRowland, line 149: If reset is necessary, it's probably better to reset in another effect hook.

JimmyRowland (approved):

JimmyRowland:
I honestly don't know the difference between declaring a component in the component function body and spreading the component directly in JSX expression after the return. 
I just blindly follow rule to move all component declaration out of function body or wrap component declaration in memo

packages/webapp/src/components/Task/AssignTask/HourlyWageInputs.jsx
SayakaOno: The way I'm passing {valueAsNumber: true}(line 68) is correct, right?

SayakaOno:
Thank you so much, Jimmy.
I wasn't also sure about what you mentioned plus when to use `useMemo`(I used to use it only when a component uses lots of dynamic props), but I'll follow the rule from now on.
Thank you for the twitter link too!

packages/webapp/src/components/Task/AssignTask/HourlyWageInputs.jsx
JimmyRowland: I missed that 🤡

SayakaOno: I was assuming so but wanted to make sure. Thank you!

PR #2450, opened 2023-02-07
Lf 2656 not entering a date on document upload archives the document

ebertmota:
<h2>Description</h2>
Set "This file does not expire" to true if user does not provided an "Valid until" date in add document page.

Jira link: [LF-2656](https://lite-farm.atlassian.net/browse/LF-2656)

<h2>Type of change</h2>

<h2>How has this been tested?</h2>

1. Go to documents page and click in "+Add a new document" link.
2. In put a name, but don't add a "Valid until" date nor select "This file does not expire".
3. Click in "Save" button.

[LF-2656]: https://lite-farm.atlassian.net/browse/LF-2656?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/webapp/src/components/Documents/Add/index.jsx
+++ b/packages/webapp/src/components/Documents/Add/index.jsx
@@ -89,8 +89,8 @@ function PureDocumentDetailView({
       }
       return undefined;
     };
-    let validUntil = !!data.valid_until ? data.valid_until : null;
-    data.type = !!data.type ? data.type.value : data.type;
+    let validUntil = data.valid_until ? data.valid_until : null;
+    data.type = data.type ? data.type.value : data.type;
     validUntil = data.no_expiration ? null : validUntil;
     submit({
       ...data,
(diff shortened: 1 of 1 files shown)


Duncan-Brain (approved):
@ebertmota 

This looks to me like it will work. But with no way to test locally we will have to check immediately after merge.

I am not super confident on the removal of the double negation casting. I think I agree they are uneccessary because they are just used as the conditional in the ternary. But wondering if they need an existence check like your line:  
```
!!data.valid_until --> !data?.valid_until
```
We will need a second review on this one because it could break things. I will request one from Jimmy and Sayaka.

JimmyRowland (approved):

SayakaOno (commented):
It works, but I made some suggestions.

PR #2455, opened 2023-02-07
Lf 2599 Part One - Action needed to react to google sign in being discontinued

kathyavini:
## Description

This PR updates the Google Login Button component on the primary login page, migrating from the deprecated [react-google-login](https://github.com/anthonyjgrove/react-google-login) package to [@react-oauth/google](https://github.com/MomenSherif/react-oauth).

The react-googe-login component used the old Google Sign-In, which will be discontinued on March 31, 2023. @react-oauth/google uses the new Google identity Services and the new [Sign in with Google](https://developers.google.com/identity/gsi/web/guides/overview) button.

A separate draft PR will be created for the Invite Sign Up view, which also uses react-google-login, but implements it differently.

Jira link: https://lite-farm.atlassian.net/browse/LF-2599

## Type of change

## How Has This Been Tested?

- Tested locally using the frontend (Note that this PR allows Google Authentication to work on localhost, which it previously did not if your Google Client ID was generated after July 29, 2022)
- Migration off the deprecated API was also verified with a debug cookie as described in the [Google Identity Services Migration Guide](https://developers.google.com/identity/oauth2/web/guides/migration-to-gis#identifying_affected_code_and_testing)

## Checklist:

- [x] My code follows the style guidelines of this project 
- [x] I have performed a self-review of my own code
- [x] Any dependent changes have been merged and published in downstream modules
- [x] I have checked my code and corrected any misspellings


SayakaOno (approved):
It worked great!

kathyavini:
Sorry for making your review stale @SayakaOno! I was doing more testing around this and saw the app crash when attempting to log in using email to an account previously created with Google SSO. 

That flow tried to select and click the Sign in with Google button in a way that I don't think is supported anymore.

packages/webapp/src/containers/GoogleLoginButton/index.jsx
```diff
@@ -1,39 +1,48 @@
...
+  credential: <equivalent to previous id_token below>,
+  clientId: <same as the one imported from .env file above>,
+  select_by: "btn"
+}
+
+*/
+
 /**
- Google res shape
+ Original Google response shape (Google Sign In, discontinued March 31, 2023, via react-google-login)
```
Duncan-Brain, line 45: @kathyavini 

I am thinking we should just replace the text. No need to store old info since its all in the commit history.

Duncan-Brain:
@kathyavini 

Can you confirm your api/token settings with me on slack? My personal tokens are not working.

I think my client_id is configured wrong: before merge can we confirm with Kevin this setting on beta and prod:

image

packages/webapp/src/containers/GoogleLoginButton/index.jsx
kathyavini: Good point :)

Duncan-Brain:
@kathyavini 
This is looking good to me and works well.

Thanks for the help with the keys! So in the google api cloud console my Oauth key needed to be a web application, I with missing plain localhost as an authorized JS origins. And then that key needed to be in both webapp and api .env files.

Things I noticed:
1. Shape comments -- see inline
2. Google authenticator - This may not need to be addressed since it works well currently but in the @react-oauth/google docs I noticed it said to wrap your "app" in the Google Authenticator portion. In the past I have seen this to mean add it to the App.jsx file similar to how Snackbar Provider is used. I don't suppose you saw any examples either way?
image

kathyavini:
@Duncan-Brain

Yes, I believe the package author intended for App itself to be wrapped 😄 I did that in [this commit](https://github.com/LiteFarmOrg/LiteFarm/pull/2461/commits/af75f46832b7e5118df191254af41b22b3415699) as it's necessary for that PR. For this one (login just from one view with one component), I tested it both ways, found that both worked,  and went with this to keep the changes local to the file. I wasn't following any code examples; I just went rogue 😁  I believe the provider sets up and inserts the GIS script and that can happen anywhere if it only happens once.

However, I'll copy that other commit to this branch (if I can figure out the git for that!) and wrap App right from the start, as I agree that's where we should go eventually.

SayakaOno (approved):

MwayaB (approved):
The changes to the cypress test look good to me

Duncan-Brain (approved):
Working great with the changes!

PR #2461, opened 2023-02-10
LF-2967 Implement new Google SSO requirements when accepting an invite to a farm

kathyavini:
## Description

This PR outlines a possible way to update the email invite sign-up flow from the deprecated [react-google-login](https://github.com/anthonyjgrove/react-google-login) package to [@react-oauth/google](https://github.com/MomenSherif/react-oauth)

[This PR](https://github.com/LiteFarmOrg/LiteFarm/pull/2455) updated the main LiteFarm login view to the new [Sign in with Google](https://developers.google.com/identity/gsi/web/guides/overview) button using the @react-oauth/google package.

What the new Google Identity Services no longer supports is receiving an identification token from a custom button, as had been done in this email invite view. This PR outlines the easiest way of moving over to the new API (before the old one is discontinued on March 31, 2023) that doesn't change the UI/UX of signing up after an email invite.

Ultimately, the Google Identity Services encourages using the new Sign in with Google button when only authenticating for login (as we do). Therefore we might later consider refactoring this user flow to incorporate that new button.

Jira link: https://lite-farm.atlassian.net/browse/LF-2967

## Previous implementation
The old Google Sign-In (implemented via react-google-login) returned both an id_token and readable user info immediately upon calling the API. Our email invite view used the returned user info to give immediate feedback about whether the correct Google account (the one matching the email invite) was used, then sent the identification token as an authorization header to the backend's `checkGoogleJwt()` where it was verified using [google-auth-library](https://github.com/googleapis/google-auth-library-nodejs).

## Issue
When using a custom button, it is impossible to get this authentication token (id_token) back in this way. Instead, the @react-oauth/google package provides a `useGoogleLogin()` hook to interact with the Google Identity Services authorization (not authentication) API, and the kind of data returned is different.

The GitHub issues for the @react-oauth/google package include several threads of people dealing with the same migration and challenge:

- [How to get id_token using useGoogleLogin Hook](https://github.com/MomenSherif/react-oauth/issues/144)
- [Not returning id_token or credential with Hook](https://github.com/MomenSherif/react-oauth/issues/151)
- [GoogleLogin and useGoogleLogin are not returning the same response](https://github.com/MomenSherif/react-oauth/issues/12)

From my reading of those issues and the Google [Migration Guide](https://developers.google.com/identity/gsi/web/guides/gis-migration), this is not a shortcoming of the package but a design of Google Identity Services; see especially [separated authentication and authorization moments](https://developers.google.com/identity/gsi/web/guides/overview#separated_authentication_and_authorization_moments).

## Approach taken
The @react-oauth/google package maintainer outlines the two data flows that can be accessed from the `useGoogleLogin()` hook in [this GitHub issues comment](https://github.com/MomenSherif/react-oauth/issues/12#issuecomment-1131408898).

This PR implements the implicit flow, which returns an access token which can be used immediately to call the Google API and verify that the Google email matches the mail invite email. The access token is then sent to the backend and verified with google-auth-library's `tokenInfo()` in place of `verifyIdToken()`.

### Benefits
- We maintain the UI/UX of immediate feedback as to whether the Google email selected matches the invite
- Very little downstream code is changed other than handling the access_token case alongside the id_token in the backend middleware
- Nothing coming from the frontend is used directly in the database

### Possible drawbacks
- I couldn't find any information in the [Google Auth Library Client API reference](https://cloud.google.com/nodejs/docs/reference/google-auth-library/latest/google-auth-library/oauth2client#google_auth_library_OAuth2Client_getTokenInfo_member_1_) specifically about the robustness/security of using `tokenInfo()` in place of `verifyIdToken()` (or, alternatively, calling the same Google API used in the frontend again in the middleware). Generally speaking [an access token is not meant for authentication](https://auth0.com/blog/id-token-access-token-what-is-the-difference/), but we need very, very little data from the Google SSO (just the email).
- We make an extra API call in the frontend component (which probably should be refactored into a saga; not sure how difficult that will be)
- There are enough GitHub issues still being opened about the hook and the component returning different data that it's not inconceivable the package author will end up changing the API

## Alternatives considered
The package maintainer recommends the `auth-flow` route (getting code that can be exchanged for an id_token on the backend) over the `implicit-flow` / access token route implemented here. I considered implementing this in two ways:
1. Send it to the backend in place of the id_token and decode it in the `checkGoogleJwt()` middleware. I implemented this initially but found it unacceptable UI/UX because we don't receive any user information (including email) until too late in the flow, after the user already submits their new account information.
2. Send it to the backend immediately (at the same moment the code in this PR calls the Google API with the access token) to decode and return to the frontend. The advantage would be that we could return the data in the exact same shape as previously (with id_token and user info). The disadvantage is that we would need a new API endpoint to do this processing, and it would be fully redundant with the verification we already do (and would want to keep doing) before saving the email to the database in the current route.

An entirely separate option would be writing all the API calls ourselves and not using the @react-oauth/google package. However, as mentioned above, I don't think this solves the token type problem, and to write this code ourselves would also bring in a fair amount of new code that was previously handled by a third-party package.

## Testing
Edit: I've tested this now with invite emails. Please see https://github.com/LiteFarmOrg/LiteFarm/pull/2467 for sending emails from a source other than system@litefarm.org, or use the beta/prod keys to send from the official Gmail. I was previously incorrect in that emails do send from localhost :)

### DIFF
--- a/packages/api/src/middleware/acl/checkGoogleAccessToken.js
+++ b/packages/api/src/middleware/acl/checkGoogleAccessToken.js
@@ -31,6 +31,7 @@ async function checkGoogleAccessToken(req, res, next) {
     return next();
   } catch (error) {
     console.error(error);
+    return res.status(401).send('Not authenticated');
   }
 }
 
--- a/packages/api/src/middleware/acl/checkGoogleAccessToken.js
+++ b/packages/api/src/middleware/acl/checkGoogleAccessToken.js
@@ -1,3 +1,18 @@
+/*
+ *  Copyright 2019, 2020, 2021, 2022, 2023 LiteFarm.org
+ *  This file is part of LiteFarm.
+ *
+ *  LiteFarm is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  LiteFarm is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ */
+
 import { OAuth2Client } from 'google-auth-library';
 const client_id = process.env.GOOGLE_OAUTH_CLIENT_ID;
 const client = new OAuth2Client(client_id);
--- a/packages/api/src/middleware/acl/checkGoogleAccessToken.js
+++ b/packages/api/src/middleware/acl/checkGoogleAccessToken.js
@@ -1,0 +1,22 @@
+import { OAuth2Client } from 'google-auth-library';
+const client_id = process.env.GOOGLE_OAUTH_CLIENT_ID;
+const client = new OAuth2Client(client_id);
+
+// Token returned from the @react-oauth useGoogleLogin() hook in the email invite view
+async function checkGoogleAccessToken(req, res, next) {
+  const authorization = req.headers['authorization'];
+  if (!authorization) {
+    return res.status(401).send('Not authenticated');
+  }
+  const token = authorization.replace('Bearer ', '');
+
+  try {
+    const tokenInfo = await client.getTokenInfo(token);
+    req.user = { email: tokenInfo.email, sub: tokenInfo.sub };
+    return next();
+  } catch (error) {
+    console.error(error);
+  }
+}
+
+export default checkGoogleAccessToken;
--- a/packages/api/src/middleware/acl/checkGoogleJwt.js
+++ b/packages/api/src/middleware/acl/checkGoogleJwt.js
@@ -9,6 +9,18 @@ async function checkGoogleJwt(req, res, next) {
   }
   const token = authorization.replace('Bearer ', '');
 
+  // For implicit flow from @react-oauth useGoogleLogin() hook from the email invite view
+  if (req.body.isAccessToken) {
+    try {
+      const tokenInfo = await client.getTokenInfo(token);
+      req.user = { email: tokenInfo.email, sub: tokenInfo.sub };
+      return next();
+    } catch (error) {
+      console.error(error);
+    }
+  }
+
+  // Original flow & Sign in with Google Button on main login page
   try {
     const ticket = await client.verifyIdToken({
       idToken: token,
--- a/packages/api/src/middleware/acl/checkGoogleJwt.js
+++ b/packages/api/src/middleware/acl/checkGoogleJwt.js
@@ -2,6 +2,7 @@ import { OAuth2Client } from 'google-auth-library';
 const client_id = process.env.GOOGLE_OAUTH_CLIENT_ID;
 const client = new OAuth2Client(client_id);
 
+// Token returned from the Sign in with Google button on main login page
 async function checkGoogleJwt(req, res, next) {
   const authorization = req.headers['authorization'];
   if (!authorization) {
@@ -9,18 +10,6 @@ async function checkGoogleJwt(req, res, next) {
   }
   const token = authorization.replace('Bearer ', '');
 
-  // For implicit flow from @react-oauth useGoogleLogin() hook from the email invite view
-  if (req.body.isAccessToken) {
-    try {
-      const tokenInfo = await client.getTokenInfo(token);
-      req.user = { email: tokenInfo.email, sub: tokenInfo.sub };
-      return next();
-    } catch (error) {
-      console.error(error);
-    }
-  }
-
-  // Sign in with Google Button on main login page
   try {
     const ticket = await client.verifyIdToken({
       idToken: token,
--- a/packages/api/src/middleware/acl/checkGoogleJwt.js
+++ b/packages/api/src/middleware/acl/checkGoogleJwt.js
@@ -20,7 +20,7 @@ async function checkGoogleJwt(req, res, next) {
     }
   }
 
-  // Original flow & Sign in with Google Button on main login page
+  // Sign in with Google Button on main login page
   try {
     const ticket = await client.verifyIdToken({
       idToken: token,
--- a/packages/webapp/src/containers/InviteSignUp/index.jsx
+++ b/packages/webapp/src/containers/InviteSignUp/index.jsx
@@ -11,7 +11,6 @@ import { isChrome } from '../../util';
 function InviteSignUp({ history }) {
   const invite_token = history.location.state;
   const GOOGLE = 1;
-  const LITEFARM = 2;
   const [selectedKey, setSelectedKey] = useState(0);
   const { i18n, t } = useTranslation(['translation', 'common']);
   const [email, setEmail] = useState();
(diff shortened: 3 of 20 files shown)


packages/api/src/middleware/acl/checkGoogleJwt.js
```diff
@@ -9,6 +9,18 @@ async function checkGoogleJwt(req, res, next) {
...
+    try {
+      const tokenInfo = await client.getTokenInfo(token);
+      req.user = { email: tokenInfo.email, sub: tokenInfo.sub };
+      return next();
+    } catch (error) {
+      console.error(error);
+    }
+  }
+
+  // Original flow & Sign in with Google Button on main login page
```
Duncan-Brain, line 23: Are we using the original flow anymore? (Can we delete it?)

packages/webapp/src/containers/InviteSignUp/index.jsx
```diff
@@ -68,37 +73,76 @@ function InviteSignUp({ history }) {
...
+        given_name: ...
+        locale: ...,
+        name: ...
+        picture: <link>
+        sub: ...
+      }
+
+  */
+
+  /* ProceedButton is a wrapper for the renderProps that handles the onClick similar to how the previous package's GoogleLogin component did: https://github.com/anthonyjgrove/react-google-login/blob/master/src/google-login.js I wrote it like this in order to keep the surrounding code unchanged. I don't really use/understand renderProps myself so I will defer to someone else if maybe they can be refactored out, now that our new package doesn't require them. */
```
Duncan-Brain, line 119: @kathyavini 
I too am not familiar with this syntax. How it is right now it doesn't look to me like it fits the pattern I have seen in the React docs. 

From what I can tell the previous GoogleLogin returned a component, whereas the new google login is a hook. I am not confident keeping this syntax is appropriate.

packages/webapp/src/containers/InviteSignUp/index.jsx
```diff
@@ -68,37 +73,76 @@ function InviteSignUp({ history }) {
       });
     }
   };
-  const clientId = import.meta.env.VITE_GOOGLE_OAUTH_CLIENT_ID;
+
+  const login = useGoogleLogin({
+    onSuccess: async (tokenResponse) => {
+      try {
+        // I have not previously used redux or redux-saga, but I imagine a query to an API like this should be in a saga.js file?
```
Duncan-Brain, line 80: @kathyavini 

I might not have noticed this otherwise haha. But I think your instinct is correct.

packages/api/src/middleware/acl/checkGoogleJwt.js
kathyavini: Yes, this is used by the main login page (the one that now has the new personalized Sign in with Google button). I will take "Original flow" out of that comment!

packages/webapp/src/containers/InviteSignUp/index.jsx
kathyavini: Haha, alas. Ok, I'll do some reading to figure out how to structure that.

packages/webapp/src/containers/InviteSignUp/index.jsx
kathyavini: Yeah, I have also never seen or used a pattern like this before. I kept the render props because I think the next component up the tree (that imports this one) expects them. But I'm definitely open to a refactor; I was hoping that someone more familiar with this structure might chime in on the simplest way to do that!

packages/webapp/src/containers/InviteSignUp/index.jsx
```diff
@@ -68,37 +73,76 @@ function InviteSignUp({ history }) {
...
+              data-cy="invitedUser-proceed"
+              onClick={onClickProceed(renderProps)}
+              disabled={(selectedKey === GOOGLE && renderProps.disabled) || !selectedKey}
+              fullLength
+            >
+              {t('common:PROCEED')}
+            </Button>
+          )}
+        />
+      }
```
SayakaOno, line 139: If we don't need to wait for someone to refactor this...

I believe this `renderProps` was previously passed from the `GoogleLogin` component and is no longer necessary. I also believe that `renderProps`(line 128) is the `onClick` that you are passing in line 121, so the `googleButton` can be:
```
  googleButton={
    <Button
      data-cy="invitedUser-proceed"
      onClick={onClickProceed}
      disabled={!selectedKey}
      fullLength
    >
      {t('common:PROCEED')}
    </Button>
  }
/>
```

`onClickProceed`(line 61) does not need to return a function anymore, and `renderProps.onClick();` can be removed because what actually is happening is `setSelectedKey();`

packages/webapp/src/containers/InviteSignUp/index.jsx
SayakaOno: This does not update any Redux state, so I don't think it needs to be done as a Redux action. I might be wrong though...

packages/webapp/src/containers/InviteSignUp/index.jsx
kathyavini: THANK YOU!!! My head was spinning with all the functions being passed around and this is the simplification that I hoped existed 😄 So much better!!!

SayakaOno (approved):
Looks good to me!

packages/webapp/src/containers/InviteSignUp/index.jsx
Duncan-Brain: I think asynchronous requests are usually handled in sagas. For example this would be the first axios request in any non-saga.js file in /containers .

packages/api/src/middleware/acl/checkGoogleJwt.js
```diff
@@ -9,6 +9,18 @@ async function checkGoogleJwt(req, res, next) {
   }
   const token = authorization.replace('Bearer ', '');
 
+  // For implicit flow from @react-oauth useGoogleLogin() hook from the email invite view
+  if (req.body.isAccessToken) {
```
Duncan-Brain, line 13: Looking at this again I am wondering if this should just be separated into two functions -- one for login and one for access token. Then the isAccess token would not have to be created at all on the frontend and the routes would just point to their respective functions. What do you think?

Duncan-Brain:
Added two comments:

1. Axios request - see previous discussion added a comment, I'm getting more confident that needs to be in a saga
2. isAccessToken - take it or leave it

kathyavini:
Here is the follow-up Jira ticket for evaluating where that axios request should ideally go!

https://lite-farm.atlassian.net/browse/LF-3043

SayakaOno (commented):
Sorry Joyce, I just found one thing missing!

packages/api/src/middleware/acl/checkGoogleAccessToken.js
```diff
@@ -0,0 +1,37 @@
...
+    return res.status(401).send('Not authenticated');
+  }
+  const token = authorization.replace('Bearer ', '');
+
+  try {
+    const tokenInfo = await client.getTokenInfo(token);
+    req.user = { email: tokenInfo.email, sub: tokenInfo.sub };
+    return next();
+  } catch (error) {
+    console.error(error);
```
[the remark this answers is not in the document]
kathyavini: Thank you!

Duncan-Brain (approved):
Looks good to me!

SayakaOno (approved):

PR #2465, opened 2023-02-15
LF-2951-Retired-sensors-become-active-again-when-attempting-to-upload-new-sensors

javiscript13:
fix: sensors from no integrating partners are assumed that are always new since there is no external_id to match if it has been previously deleted

### DIFF
--- a/packages/api/src/models/sensorModel.js
+++ b/packages/api/src/models/sensorModel.js
@@ -67,19 +67,21 @@ class Sensor extends Model {
     const trx = await transaction.start(Model.knex());
 
     try {
-      const existingSensor = await LocationModel.getSensorLocation(
-        farm_id,
-        partner_id,
-        sensor.external_id,
-        trx,
-      );
-      if (existingSensor) {
-        if (existingSensor.deleted) {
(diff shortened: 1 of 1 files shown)


SayakaOno (approved):

packages/api/src/models/sensorModel.js
```diff
@@ -67,19 +67,21 @@ class Sensor extends Model {
...
-        farm_id,
-        partner_id,
-        sensor.external_id,
-        trx,
-      );
-      if (existingSensor) {
-        if (existingSensor.deleted) {
-          await LocationModel.unDeleteLocation(user_id, existingSensor.location_id, trx);
-          existingSensor.deleted = false;
+      if (partner_id !== 0) {
```
Duncan-Brain, line 70: @javiscript13 

I think at some point we will be registering more partners. So my question is if we create a partner with partner_id = 1 will this still work? 

So would it make sense to make condition more specific -- maybe something: 
`if( (!partner_id || partner_id !==0) || (partner_id && (!sensor.external_id || sensor.external_id !==0))`

I might not have all the conditions there ^^ and its kind of messy haha

Duncan-Brain (approved):
Works for me great work!

PR #2466, opened 2023-02-15
LF-2957-Bulk-sensor-upload-ignores-sensors-beyond-the-first

javiscript13:
fix: sensor CSV now checks row uniqueness by brand:external_id if available, else it uses sensor name


SayakaOno (approved):

Duncan-Brain (approved):
This works for me unless the issue described in LF-2951( Retired sensors becoming active again) is also present. This should work perfect in combination with that one.

PR #2467, opened 2023-02-16
LF-2976 (Feat) Allow sending LiteFarm emails from personal accounts in development environment

kathyavini:
Currently, LiteFarm emails (sent when, e.g., joining a farm or inviting users) are only sent from system@litefarm.org and require the Beta/Production gmail API keys and refresh token to work.

This PR adds the option of sending, in the development environment only, emails from another account with either 1) the correct OAuth2 credentials or 2) an app password set up through Google. Instructions for generating those credentials were added to `.env.default` along with more precise instructions about what the beta/production keys are for. To set up OAuth2, the currently unused `LITEFARM_SERVICE_EMAIL` environment variable was re-introduced.

This change is entirely opt-in; emails should continue to work as before whether the developer's `.env` file contains the official beta/production keys or not. That is, if the developer possessed the keys and was previously getting development emails sent for localhost actions via system@litefarm.org, they will continue to receive them. Contributors not using the LiteFarm gmail API keys will not start getting emails unless they set up their own account credentials following the `.env.default` instructions.

There are no beta or production changes; this only concerns development.

Jira link: https://lite-farm.atlassian.net/browse/LF-2976

**Type of change**

**How Has This Been Tested?**

In my local environment:
- Tested using both OAuth2 and App Password `.env` variables using my personal Gmail account. Emails are then sent upon localhost actions with my own gmail as the sender
- Tested with the beta keys in `.env` and with no opt-in variables; emails are successfully sent from system@litefarm.org

**Checklist:**

### DIFF
--- a/packages/api/.env.default
+++ b/packages/api/.env.default
@@ -30,11 +30,21 @@ GOOGLE_API_KEY=?
 OPEN_WEATHER_APP_ID=?
 
 # Create your own Google API key at https://developers.google.com/identity/oauth2/web/guides/get-google-api-clientid
-GMAIL_API_CLIENT_ID=?
-GMAIL_API_CLIENT_SECRET=?
-GMAIL_ACCESS_TOKEN=?
-GMAIL_REFRESH_TOKEN=?
 GOOGLE_OAUTH_CLIENT_ID=?
+
+# For sending automatic LiteFarm emails on certain events (e.g. joining a farm)
+# To send from system@litefarm.org you will need the corresponding beta/production values for the Client ID, Secret, and Refresh Token
+LITEFARM_SERVICE_EMAIL=
+GMAIL_API_CLIENT_ID=
+GMAIL_API_CLIENT_SECRET=
+GMAIL_REFRESH_TOKEN=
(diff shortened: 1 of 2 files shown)


JimmyRowland (approved):

Duncan-Brain:
@kathyavini 

Looks good to me, my only worry is that we may need to do the old .env check again to see if node_env is set there. The Dockerfile uses start:prod which doesn't explicitly set node_env. So there is a small chance that the safety net to keep system@litefarm.org the default could fail at the environment check. 

Do you remember from your meeting with Kevin the other day?

kathyavini:
@Duncan-Brain I don't know if we looked at the prod `.env` file the other day, and definitely not for the NODE_ENV (just keys). That would be a quick check for @kcussen.

I agree with double-checking! But I took the environment code from the `homeUrl()` function in the same file:

```
const environment = process.env.NODE_ENV || 'development';
...
  else if (environment === 'production') {
    homeUrl = 'https://app.litefarm.org';
  }
```

so if email links are working from production, the NODE_ENV must be set right! If not, our emails are linking to the wrong domain too 😄

Duncan-Brain (approved):
@kathyavini 

Good call I missed that. NODE_ENV must be set to integration then. There is a few things that use dev scripts on beta but that alleviates the concern. Cool!

SayakaOno (approved):

PR #2471, opened 2023-02-21
Lf 2848 augment the existing varietal creation flow

MwayaB:
**Description**
This story augments the crop varietal creation flow, by adding two new inputs and a custom image upload feature

Jira link: https://lite-farm.atlassian.net/browse/LF-2848

**Type of change**

**How Has This Been Tested?**

tested as per https://lite-farm.atlassian.net/wiki/spaces/LITEFARM/pages/1273135133/LF+-2848+Test+plan

- [x] Passes test case
- [ ] Other (please explain)

**Checklist:**


Duncan-Brain:
@MwayaB 

So I think you mentioned you were looking for high level feedback like a draft PR so I wont go into some of the small changes. It mostly was working... but here is some of the high level things I noticed on this one:

1. All instances that use the addCropVariety components need to be tested.  It is used in both the addNewCrop workflow and the addNewVariety creation flow (at least). It currently only works in the add new variety flow.
2. I really like that crop variety name is being changed to "variety_common_name", changes like this I would be interested in carrying through the new naming convention all the way to backend--  I would normally ask Kevin if he is okay with that first -- but that way crop_variety_name and crop_varietal don't get mixed up and its pretty nice looking that way
3. For things like the following I don't personally know the best way to do it off the top of my head but it makes my spidey senses think there is a more succinct way to do it.
```
const specieForLabel = () => {
    if (crop.crop_specie) {
      if (crop.crop_specie.length > 22) {
        return crop.crop_specie.slice(0, 22) + '...';
      }

      return crop.crop_specie;
    }

    return '';
```
4. The new header with crop photo and genus and species:  It seems like a lot of divs -- spidey sense says its not good to have that many divs with all that inline styles but I don't know the best way to fix that either. But I am thinking it should at least be outside of the form element, maybe even its own component. 
5. I think your instinct about the subtext was right haha. I like how it is formed but for me all instances of Info from Typography would need to be tested with that change to Info. I would normally move this out of scope to a new ticket because it could have a fair amount of checking other elements. The purpose of Info I think was to be a `<p>` so some other creative solution may or may not be preferred. If you are seeing errors about child elements in `<p>` I found this list of allowed elements so maybe it has to be base elements?

packages/webapp/src/components/AddCropVariety/index.jsx
```diff
@@ -52,18 +51,67 @@ export default function PureAddCropVariety({
...
+  const imageUrlRegister = register(CROP_VARIETY_PHOTO_URL, { required: false });
   const crop_variety_photo_url = watch(CROP_VARIETY_PHOTO_URL);
   const cropTranslationKey = crop.crop_translation_key;
   const cropNameLabel = cropTranslationKey
     ? t(`crop:${cropTranslationKey}`)
     : crop.crop_common_name;
+  const genusForLabel = () => {
+    if (crop.crop_genus) {
+      if (crop.crop_genus.length > 22) {
+        return crop.crop_genus.slice(0, 22) + '...';
```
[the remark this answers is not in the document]
MwayaB: @Duncan-Brain I thought about changing the the crop variety name all the way through to the back end, but realized that may be a breaking change for some functionality that was not necessarily within the scope of this ticket.

packages/webapp/public/locales/en/translation.json
```diff
@@ -299,6 +299,16 @@
...
     "ADD_IMAGE": "Add Custom Image",
+    "VARIETAL_SUBTITLE": "Tell us about the specific varietal or cultivar you intend to grow",
+    "VARIETAL_IMAGE": "Customize image for varietal or cultivar",
+    "VARIETAL_IMAGE_INFO": "If you would like this varietal or cultivar to have a different image from the default crop, you can customize the image here.",
+    "VARIETY_COMMON_NAME": "Common Name",
+    "VARIETY_VARIETAL": "Varietal",
+    "VARIETY_CULTIVAR": "Cultivar",
+    "CULTIVAR_SUBTEXT": "Learn more about cultivars",
+    "VARIETAL_SUBTEXT": "Learn more about varietals",
+    "CULTIVAR_PLACEHOLDER": "Ex. Cabernet sauvignon",
```
Duncan-Brain, line 310: @MwayaB looks like these are backwards: varietal => cab sauv, cultivar => red delicious.

packages/api/src/models/cropVarietyModel.js
```diff
@@ -83,6 +83,8 @@ class CropVariety extends BaseModel {
         crop_id: { type: 'integer' },
         farm_id: { type: 'string' },
         crop_variety_name: { type: ['string', null] },
+        crop_varietal: { type: ['string', null] },
+        crop_cultivar: { type: ['string', null] },
```
Duncan-Brain, line 87: @MwayaB 

These two need the backend character limit validation.

packages/webapp/src/components/AddCropVariety/index.jsx
```diff
@@ -81,24 +90,99 @@ export default function PureAddCropVariety({
...
-      <img
-        src={crop_variety_photo_url}
-        alt={crop.crop_common_name}
-        className={styles.circleImg}
-        onError={(e) => {
-          e.target.onerror = null;
-          e.target.src = 'crop-images/default.jpg';
-        }}
-      />
+      <p style={{ fontSize: '16px', marginBottom: '20px' }}>
```
Duncan-Brain, line 93: @MwayaB there are a lot of inline styles can we move these to the stylesheet?

packages/webapp/src/components/AddCropVariety/index.jsx
```diff
@@ -81,24 +90,99 @@ export default function PureAddCropVariety({
...
+        label={t('CROP.VARIETY_CULTIVAR')}
+        type="text"
+        hookFormRegister={cultivarRegister}
+        errors={getInputErrors(errors, 'crop_cultivar')}
+        hasLeaf={false}
+        optional
+        subText={t('CROP.CULTIVAR_SUBTEXT')}
+        link={'https://www.litefarm.org/post/cultivars-and-varietals'}
+        placeholder={t('CROP.CULTIVAR_PLACEHOLDER')}
+        hideOnFocus
```
Duncan-Brain, line 166: @MwayaB 

Getting an error on this line: 
image

packages/webapp/src/components/AddCropVariety/index.jsx
```diff
@@ -107,19 +191,16 @@ export default function PureAddCropVariety({
       >
         {React.cloneElement(imageUploader, {
           hookFormRegister: imageUrlRegister,
-          targetRoute: 'crop_variety',
+          targetRoute: 'crop',
```
Duncan-Brain, line 194: @MwayaB 

Just wondering if you had issues with this -- crop_variety should be correct. The folder is where all the crop_vareity photos would be stored  on Digital Ocean.

packages/webapp/src/components/Typography/index.tsx
```diff
@@ -205,3 +207,53 @@ export const Text = ({ children = 'Text', className = '', style, ...props }: Typ
     </p>
   );
 };
+
+export const ExternalLink = ({ children = 'Link', className = '', style, link, ...props }) => {
+  return (
+    <>
+      <div
+        className={clsx(styles.text, className)}
+        style={{ width: 'fit-content', display: 'inline-block', ...style }} // I'm not sure about style...
```
Duncan-Brain, line 216: @MwayaB 

I'm not sure about styles either lol (probably should remove this comment). 

If anything base element styles like div's and p's etc can be put in the styles css file.

packages/webapp/src/util/index.js
```diff
@@ -131,3 +132,13 @@ export const getObjectInnerValues = (data) => {
...
 };
+
+export const truncateText = (text, length) => {
+  if (text) {
+    if (text.length > length) {
+      return text.slice(0, 22) + '...';
+    }
+    return text;
+  }
+  return '';
```
Duncan-Brain, line 143: @MwayaB 

Possible one liner but up to you: 

`return text?.length > length ? text.slice(0, 22) + '...' : text ?? '';`

Duncan-Brain:
@MwayaB 

I think it is looking pretty darn good just a few small things:

1. externalLink icon -- image looks the same as icon_launch.svg and use case the same. Wondering if it was asked about reusing icon_launch but would have to change the color on that one. Previous naming isn't super consistent but I am seeing lower case, icon in the name, and either dash or underscore. Something like icon_external_link to fit in with the others.

2. Some inline comments --  nothing major except the console error

3. Just addressing my previous review:
- 1. seems fixed
- 2. I'd still like to see this changed at some point -- but ya there is a lot of references to it so I think a future cleanup would be best
- 3. See inline comment 
- 4. I still think this could still be moved outside of the form element maybe that causes issues I'm unaware of.
- 5. New fix is looking like a good solution!

packages/webapp/src/components/Form/Input/index.jsx
```diff
@@ -165,6 +167,7 @@ const Input = ({
           {errors}
         </Error>
       ) : null}
+      {subText ? <TextWithExternalLink link={link}>{subText}</TextWithExternalLink> : null}
```
SayakaOno, line 170: I don't think subtext will always be `TextWithExternalLink`. Could we have this component to take and show `subText` to make it more flexible? (and pass `<TextWithExternalLink link={link}>{subText}</TextWithExternalLink>` from the parent)

packages/webapp/src/components/Typography/index.tsx
SayakaOno: Haha, that was in my code I shared with Mwaya.
Mwaya, please remove it lol

packages/webapp/src/components/AddCropVariety/index.jsx
MwayaB: Yes, that happens because of this bit of code `input:focus::placeholder {
  color: transparent;
}` . Forgot to delete that bit of code

packages/webapp/src/util/index.js
MwayaB: That's beautiful, but not very readable to someone new to javascript. I think i'll leave it as is

packages/webapp/src/components/AddCropVariety/index.jsx
MwayaB: Yes, the file upload fails when the target route is set to crop_variety

packages/webapp/src/components/Form/Input/index.jsx
MwayaB: i think there's `info` for other cases, but that didn't quite fit for this particular case

packages/webapp/src/components/AddCropVariety/index.jsx
Duncan-Brain: Would you be willing to show me the error you see for that? The crop_variety route for the image has been working on beta since it was made by someone however long ago.

I personally added the crop version of the route to AddNewCrop, and renamed the attribute, when making the AddNewCrop adjustments. 

They both don't work for me locally since I am not using beta .env vars.

packages/api/src/models/cropVarietyModel.js
```diff
@@ -83,6 +83,8 @@ class CropVariety extends BaseModel {
         crop_id: { type: 'integer' },
         farm_id: { type: 'string' },
         crop_variety_name: { type: ['string', null] },
+        crop_varietal: { type: 'string', minLength: 1, maxLength: 255 },
+        crop_cultivar: { type: 'string', minLength: 1, maxLength: 255 },
```
Duncan-Brain, line 87: It looks like in the frontend register you limit these to 200 chars?

MwayaB: Yes, this came up before with your Crop_type creation flow ticket, and it was agreed that the character limit on the back end for Genus and Species fields be kept as 255 characters even though the requirement on the ticket and enforced on the front end was 200 characters. I figured this also followed for these fields

Duncan-Brain: Yeah kinda, the agreement was 200 is arbitrary but it needs to match frontend. This can stay 255 but then the frontend validation needs to change. Either way is good enough -- I tagged those spots

packages/webapp/src/components/AddCropVariety/index.jsx
MwayaB: Just tried it again it again to recreate the error and it looks to be working now with the target route set to 'crop_variety'

Duncan-Brain: Perfect thanks for checking that!

packages/api/src/models/cropVarietyModel.js
MwayaB: Oh, I see

packages/api/jest-global-setup.cjs
```diff
@@ -4,4 +4,5 @@
  */
 module.exports = async () => {
   process.env.TZ = 'UTC';
+  testTimeout = 30000;
```
Duncan-Brain, line 7: @MwayaB 

Is this piece still necessary with the new mock additions?

Duncan-Brain (approved):
Looks good to me!

packages/webapp/src/util/index.js
```diff
@@ -131,3 +132,13 @@ export const getObjectInnerValues = (data) => {
     return { ...reduced, [k]: data[k] };
   }, {});
 };
+
+export const truncateText = (text, length) => {
+  if (text) {
+    if (text.length > length) {
+      return text.slice(0, 22) + '...';
```
SayakaOno, line 139: You forgot to use `length` here.

packages/webapp/src/components/Typography/index.tsx
```diff
@@ -205,3 +207,57 @@ export const Text = ({ children = 'Text', className = '', style, ...props }: Typ
     </p>
   );
 };
+
+export const ExternalLink = ({ children = 'Link', className = '', style, link, ...props }) => {
+  return (
+    <>
```
SayakaOno, line 213: Could you remove this fragment?

packages/webapp/src/components/Typography/index.tsx
```diff
@@ -205,3 +207,57 @@ export const Text = ({ children = 'Text', className = '', style, ...props }: Typ
...
+  children = 'SubText',
+  className = '',
+  style,
+  link,
+  ...props
+}) => {
+  const { t } = useTranslation(['translation', 'common', 'crop']);
+
+  return (
+    <>
```
SayakaOno, line 249: Could you remove this fragment?

packages/webapp/src/components/Typography/index.tsx
```diff
@@ -205,3 +207,57 @@ export const Text = ({ children = 'Text', className = '', style, ...props }: Typ
...
+      <div
+        className={styles.info}
+        style={{ width: 'fit-content', display: 'inline-block' }}
+        {...props}
+      >
+        {children}
+
+        <ExternalLink link={link} style={''} {...props}>
```
SayakaOno, line 257: `style={''}` can be removed.

SayakaOno (commented):
@MwayaB 
Sorry I couldn't come back sooner.
It would be great if you could address my comments.
Please also resolve conflicts. (sorry about that!)

packages/webapp/src/util/index.js
MwayaB: this works as intended, i'm not sure i understand what you mean

packages/webapp/src/components/Typography/index.tsx
MwayaB: this fragment is required for translation

SayakaOno (approved):

Duncan-Brain (approved):
reapproved

PR #2473, opened 2023-02-22
Lf 2955 spike investigate current irrigation task architecture

craigyu:
**Description**
This irrigation task PR includes the following:
- In the “Mark complete” flow, when a user select “Yes” to “Did you have to make any changes to this task?” Each of the default checkboxes should be visible but disabled
    -  When a user goes to complete an irrigation task that was created with default checkboxes checked, the task completion page will show which boxes the user checked and the checkboxes are disabled. However, if a user select "Yes" to make changes to this task, the check boxes uncheck themselves and are still disabled. In either cases, they won't set default values again once the task is complete. Default values are only set during creation.

- In the original file, there was a hard coded constant with all the default irrigation types for translation. The updated method in this PR has made it so that it generates a variable from the data provided by the back end server. The back end was not touched in this PR. 

Note:
I noticed this warning when I'm on the Tasks page:
`TypeError: Cannot read properties of undefined (reading 'irrigation_type')
    at getIrrigationTask (irrigationTaskSlice.js:43:1)
    at irrigationTaskSlice.js:57:5
    at Array.map (<anonymous>)
    at upsertManyIrrigationTask (irrigationTaskSlice.js:53:5)
    at createReducer.ts:280:20
    at produce (immerClass.ts:94:14)
    at createReducer.ts:279:18
    at Array.reduce (<anonymous>)
    at reducer (createReducer.ts:246:25)
    at reducer (createSlice.ts:325:14)`
I couldn't resolve this issue but it shouldn't be from changes from this PR, I think it's due to my local setup.

Jira link: [LF-2955](https://lite-farm.atlassian.net/browse/LF-2955?atlOrigin=eyJpIjoiMmRlYzk2NmJmZjlkNDljZjg2MTgzMDJjYzlkNDM0MjEiLCJwIjoiaiJ9)

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)
    - tested locally on my machine
    - Create an irrigation task with irrigation type "Channel", make that the default method
    - Create another irrigation task on the same field with irrigation type "Pivot", make it default as well
    - Complete the task with the "Channel" type
    - Try to create another irrigation task on the same field, the irrigation type should be "Pivot" instead of "Channel"
    - You shouldn't be able to set default values during the task completion process
    
**Checklist:**

[LF-2955]: https://lite-farm.atlassian.net/browse/LF-2955?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/webapp/src/components/Task/PureIrrigationTask/index.jsx
+++ b/packages/webapp/src/components/Task/PureIrrigationTask/index.jsx
@@ -77,8 +77,7 @@ export default function PureIrrigationTask({
         value: irrigationType.irrigation_type_name,
         label: defaultIrrigationTaskTypes.includes(irrigationType.irrigation_type_name)
           ? t(`ADD_TASK.IRRIGATION_VIEW.TYPE.${irrigationType.irrigation_type_name}`)
-          // is the false case here really needed?
-          : t(irrigationType.irrigation_type_name),
+          : irrigationType.irrigation_type_name,
         default_measuring_type: irrigationType.default_measuring_type,
         irrigation_type_id: irrigationType.irrigation_type_id,
       };
--- a/packages/webapp/src/components/Task/PureIrrigationTask/index.jsx
+++ b/packages/webapp/src/components/Task/PureIrrigationTask/index.jsx
@@ -17,16 +17,6 @@ import { useDispatch, useSelector } from 'react-redux';
 import { irrigationTaskTypesSliceSelector } from '../../../containers/irrigationTaskTypesSlice';
 import { cropLocationsSelector } from '../../../containers/locationSlice';
 
-const defaultIrrigationTaskTypes = [
-  'HAND_WATERING',
-  'CHANNEL',
-  'DRIP',
-  'FLOOD',
-  'PIVOT',
-  'SPRINKLER',
-  'SUB_SURFACE',
-  'OTHER',
-];
 export default function PureIrrigationTask({
   system,
   register,
@@ -44,7 +34,7 @@ export default function PureIrrigationTask({
   createTask = false,
 }) {
   const { t } = useTranslation();
-  const { errors, isValid } = formState;
+  const { errors } = formState;
   const [showWaterUseCalculatorModal, setShowWaterUseCalculatorModal] = useState(false);
   const { irrigationTaskTypes = [] } = useSelector(irrigationTaskTypesSliceSelector);
   const cropLocations = useSelector(cropLocationsSelector);
@@ -69,13 +59,25 @@ export default function PureIrrigationTask({
 
   const dispatch = useDispatch();
 
+  const getDefaultIrrigationTypes = () => {
+    let  defaultIrrigationTaskTypes = [];
+    for (let type of irrigationTaskTypes) {
+      if (type.farm_id === null) {
+        defaultIrrigationTaskTypes.push(type.irrigation_type_name);
+      }
+    }
+    return defaultIrrigationTaskTypes;
+  }
+
   const IrrigationTypeOptions = useMemo(() => {
     let options;
+    let defaultIrrigationTaskTypes = getDefaultIrrigationTypes()
     options = irrigationTaskTypes.map((irrigationType) => {
       return {
         value: irrigationType.irrigation_type_name,
         label: defaultIrrigationTaskTypes.includes(irrigationType.irrigation_type_name)
           ? t(`ADD_TASK.IRRIGATION_VIEW.TYPE.${irrigationType.irrigation_type_name}`)
+          // is the false case here really needed?
           : t(irrigationType.irrigation_type_name),
         default_measuring_type: irrigationType.default_measuring_type,
         irrigation_type_id: irrigationType.irrigation_type_id,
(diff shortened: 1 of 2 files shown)


packages/webapp/src/components/Task/PureIrrigationTask/index.jsx
```diff
@@ -68,13 +59,25 @@ export default function PureIrrigationTask({
...
+
   const IrrigationTypeOptions = useMemo(() => {
     let options;
+    let defaultIrrigationTaskTypes = getDefaultIrrigationTypes()
     options = irrigationTaskTypes.map((irrigationType) => {
       return {
         value: irrigationType.irrigation_type_name,
         label: defaultIrrigationTaskTypes.includes(irrigationType.irrigation_type_name)
           ? t(`ADD_TASK.IRRIGATION_VIEW.TYPE.${irrigationType.irrigation_type_name}`)
+          // is the false case here really needed?
```
Duncan-Brain, line 80: @craigyu 

I think yes, if the tasktype is included in the default then we need to translate to local language. If it is not a default type then the user will have named it appropriately to their own language.

craigyu: In the false case it's calling `t(irrigationType.irrigation_type_name)`, but we know there won't be any translation for user generated task name, so I'm not sure if calling the `t` function here is necessary since we could just display it as it is. It's working the way it is though.

Duncan-Brain: Oh I think I see what you are saying. Do you mean just the t() should be removed from the false case?  I am surprised that translation function would be working without a value for that haha Yeah I think you can remove that.

craigyu: haha yes it's basically relying on the `t()`'s fault tolerance ability

craigyu: Ok it's removed and tested working

packages/webapp/src/components/Task/TaskComplete/StepOne.jsx
```diff
@@ -88,6 +87,7 @@ export default function PureCompleteStepOne({
           farm,
           task: selectedTask,
           disabled: !changesRequired,
+          isModified: changesRequired,
```
Duncan-Brain, line 90: @craigyu 

It seems weird to send the same attribute down twice. But I don't immediately see a better option.
```
disabled: !changesRequired,
isModified: changesRequired,

// In PureIrrigationTask the following is always true when changesRequired is set. 
disabled={disabled || isModified} 
```

craigyu: unfortunately no, I didn't really want to add that either, but couldn't think of another way of doing it :/

craigyu:
> @craigyu
> 
> I am still seeing this behaviour:
> 
> > However, if a user select "Yes" to make changes to this task, the check boxes uncheck themselves and are still disabled.
> 
> Is that expected?
> 
> For the TypeError:
> 
> I see the scope was limited to make it more actionable. But I think to make it releasable we would like to fix that. Maybe could make high priority new tickets (if they don't already exist) for the remaining items you saw that are not part of the limited scope.

@Duncan-Brain 
After discussing it with Kevin, the behaviour is acceptable. 
As for the error, I think it's actually something to do with my local setup. I'm getting this error when doing `pnpm install` uner `/webapp`: 
image

Have you seen an error like this? and do you mind checkout this branch see if the error shows up on your side? Thanks

Duncan-Brain (dismissed):
So I am seeing the checkboxes are disabled and the task types selected from the backend.

Looks like you worked with Kevin on the scope so it definitely looks better!

craigyu:
> 

thanks, do you see the TypeError warning on your end?

Duncan-Brain:
@craigyu Sorry I totally forgot to address that -- I did not see that on my tasks page console or terminal

craigyu:
> @craigyu Sorry I totally forgot to address that -- I did not see that on my tasks page console or terminal

@Duncan-Brain Nice, must be my local setup then! thanks for checking

SayakaOno (commented):
@craigyu 
Thank you for the PR, the description made it very easy for me to review! I appreciate it.
I created a [PR](https://github.com/LiteFarmOrg/LiteFarm/pull/2488) to add a missing dependency array to the useEffect, please review and merge it if it is good. (Update: thank you for reviewing, I merged!)

Also, I noticed that a new `irrigation_type` is created when I:
1.  create a new irrigation task with one of the default types and check `Set as default type for this location`
2. create another irrigation task with the same type of irrigation
Screenshot 2023-03-10 at 5 28 37 PM

Do you happen to know if it is expected? (It looks like `irrigation_type_id` is not sent from the frontend when creating a task for the second time) If you don't, I will look into it, so never mind!

Duncan-Brain:
@SayakaOno 

What I know of this one is:
1. Setting any default is expected to make a new irrigation type.
2. Do you mean Custom task type using "Other" option and same spelling? If yes I believe this is expected but possibly could be improved.

craigyu:
@SayakaOno 
My apologies, I did not notice this issue(or pay enough attention to the DB) while working on it.
Upon testing locally I'm able to reproduce this behavior but it's inconsistent. This happens even when a default irrigation type is selected, but it does not happen every time.

SayakaOno:
@craigyu 
No worries!!! Thank you for confirming!

@Duncan-Brain 
> Do you mean Custom task type using "Other" option and same spelling?

No, it's automatically created without selecting the "Other" option! I will add the details to the ticket :)

SayakaOno:
I found a root cause and created a [ticket](https://lite-farm.atlassian.net/browse/LF-3057)!

SayakaOno (approved):

PR #2477, opened 2023-02-24
LF-2990 Ensure private SurveyStack submissions can be read

kathyavini:
**Description**

This PR allows viewing (export) of private survey responses by adding cookies to the SurveyStack data request.

Please see the complete discussion of this feature on Confluence: [Allow LiteFarm to access private SurveyStack data](https://lite-farm.atlassian.net/wiki/spaces/LITEFARM/pages/1271889935/Allow+LiteFarm+to+access+private+SurveyStack+data)

This functionality requires adding two variables to your `.env` file, obtained from inspecting either your cookies or local storage when visiting app.surveystack.io, logged in with an account with access right to the private survey in question. Access rights mean administrator of the group that owns the survey, or administrator in the parent group to a subgroup that owns the survey. 

```
SURVEY_USER=<user ID>
SURVEY_TOKEN=<token>
```

Jira link: https://lite-farm.atlassian.net/browse/LF-2990

**Type of change**

**How Has This Been Tested?**
- [x] Other (please explain)

That the SurveyStack API returns private question data only when the correct user + token cookies are supplied was verified with:

- An API testing tool (Postman) on these SurveyStack endpoints:

From the Confluence document:
```
https://app.surveystack.io/api/submissions?survey={{survey_id}}&match={"_id":{"$oid":"{{submission_id}}"}}
```

From the existing document export code:
```
https://app.surveystack.io/api/submissions/{{submission_id}}
```

Small aside/note: the top endpoint appears preferable to me (it's much faster to return a response), but to adapt our code to use it would take some refactoring of document export that isn't directly related to auth and is hard to test without local document export; therefore I didn't consider it within the scope of this ticket.

- Setting cookies using `request-promise` (as in this PR's code) and testing that within a backend controller

The survey ID and submission ID can be found using the SurveyStack webapp. They can also be found from the Network tab when submitting the survey via the LiteFarm frontend. To access the survey from the LiteFarm frontend, you must first manually add the survey ID to your certifiers table.

What I have not been able to test is the full document export flow, because I have not gotten document export working from localhost.

**Checklist:**

### DIFF
--- a/packages/api/src/jobs/certification/pdf.js
+++ b/packages/api/src/jobs/certification/pdf.js
@@ -3,6 +3,14 @@ import fs from 'fs';
 import rp from 'request-promise';
 const surveyStackURL = 'https://app.surveystack.io/api/';
 
+// Add cookies to the SurveyStack requests
+rp.defaults({ jar: true });
+const cookiejar = rp.jar();
+cookiejar.setCookie(
+  `user=${process.env.SURVEY_USER}; token=${process.env.SURVEY_TOKEN};`,
+  'https://app.surveystack.io',
+);
+
 export default (nextQueue, emailQueue) => async (job) => {
   console.log('STEP 3 > PDF');
   const {
--- a/packages/api/.env.default
+++ b/packages/api/.env.default
@@ -61,6 +61,10 @@ REDIS_HOST=localhost
 REDIS_PORT=6379
 REDIS_PASSWORD=test
 
+# For document export of private SurveyStack responses
+SURVEY_USER=<userId>
+SURVEY_TOKEN=<token>
+
 # The following vars are only needed if working with Ensemble Scientific sensors and actuators
 #ENSEMBLE_USERNAME=?
 #ENSEMBLE_PASSWORD=?
--- a/packages/api/src/jobs/certification/survey_record.js
+++ b/packages/api/src/jobs/certification/survey_record.js
@@ -2,6 +2,14 @@ import XlsxPopulate from 'xlsx-populate';
 import rp from 'request-promise';
 const surveyStackURL = 'https://app.surveystack.io/api/';
 
+// Add cookies to the SurveyStack requests
+rp.defaults({ jar: true });
+const cookiejar = rp.jar();
+cookiejar.setCookie(
+  `user=${process.env.SURVEY_USER}; token=${process.env.SURVEY_TOKEN};`,
+  'https://app.surveystack.io',
+);
+
 export default async (emailQueue, submission, exportId, organicCertifierSurvey) => {
   if (!submission) {
     emailQueue.add({ fail: true });


packages/api/src/jobs/certification/pdf.js
```diff
@@ -16,18 +16,28 @@ export default (nextQueue, emailQueue) => async (job) => {
...
+    headers: { Authorization: `${process.env.SURVEY_USER} ${process.env.SURVEY_TOKEN}` },
   });
+
   if (!submission || !survey) {
     emailQueue.add({ fail: true });
     return Promise.resolve();
   }
+
+  if (!Object.keys(submission.data).length) {
+    console.error(`No data was returned from submission ${job.data.submission}`);
```
Duncan-Brain, line 38: @kathyavini 

So it is ok to continue on with the export if no survey submission data is returned?

Duncan-Brain:
@kathyavini 

So I think everything seems to be working for me. Questions I had were:

1. Couple inline questions
2. The survey stack token in .env -- does this ever refresh or change? And if so will we be needing to update .env regularly to keep this functioning?
3. The excel export in documents is that supposed to have the data,Or is it private so not retrievable? If it should be visible maybe it is related to the inline variable question.

Screenshot 2023-03-20 at 1 45 33 PM

packages/api/src/jobs/certification/pdf.js
```diff
@@ -16,18 +16,28 @@ export default (nextQueue, emailQueue) => async (job) => {
...
+
+  const [submission] = await rp({
+    uri: `${surveyStackURL}/submissions?survey=${certifier.survey_id}&match={"_id":{"$oid":"${job.data.submission}"}}`,
     json: true,
+    headers: { Authorization: `${process.env.SURVEY_USER} ${process.env.SURVEY_TOKEN}` },
   });
+
   const survey = await rp({
-    uri: `${surveyStackURL}/surveys/${submission.meta.survey.id}`,
+    uri: `${surveyStackURL}/surveys/${certifier.survey_id}`,
```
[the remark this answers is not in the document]
kathyavini: It should not, because of the `await`!

packages/api/src/jobs/certification/pdf.js
kathyavini: Definitely. There are several documents that get exported; the survey is only one of them.

This error is only to help us troubleshoot (e.g. by looking at the logs of the export server) when the issue was accessing survey data in particular.

kathyavini:
@Duncan-Brain:

1. Replied inline
2. I discussed that in the Confluence document linked above, but in short, yes, it would change on account password reset or accepting an email invite invitation to a new group. Neither of those should happen spontaneously though, and the second is not necessary if following a group-subgroup structure.
3. The answers would be one column over from the screenshot you posted (not the most intuitive format, I admit). Are they missing?

Duncan-Brain:
1. Woops ahah
2. Ah yes the comments of that page. I see now. Hmm that is sketchy for the potential to change on us -- especially if person running our surveystack account changes and forgets to pass on the knowledge. Especially since SSO isn't an option. Not much we can do I guess with the limited options.
3. So weird... the B column had a zero width. And no, no values in there. It shows for you?

Screenshot 2023-03-20 at 3 13 39 PM

Duncan-Brain (approved):
Dunno what happened before but my response now looks like yours so seems good to go for me.

Still concerned about the token changing though. This risk needs to be really well communicated at LiteFarm with Kevin and to whomever uses SurveyStack.

kathyavini:
@Duncan-Brain agreed. I'll add something to Jira.

SayakaOno (approved):
Do you know how we add `process.env.SURVEY_USER` and `process.env.SURVEY_TOKEN`?
And how are we going to update survey_ids in the beta database?

Also, I happened to know `request-promise` used for this feature has been deprecated. Do you think we should find an alternative at some point? (It seems that the library is still widely used...)
https://www.npmjs.com/package/request-promise

kathyavini:
@SayakaOno Oh YES thank you! I left a tiny little comment about the request-promise deprecation in the Confluence document and then completely didn't follow up with the team about it! I think it should go on our tech debt list, yes!

Right now Kevin handles all the `.env` files for beta + prod and he is also (manually) adding the survey ids for the certifier table. I don't know much more about that, but we could discuss it with Kevin.

SayakaOno:
Wow, how thorough you are! I'll create a ticket :)

Okay, so I'll leave this PR to Kevin so that he can merge when he has added those environment variables.

PR #2478, opened 2023-03-01
LF-3003: Extract logic from unit component and create a hook

SayakaOno:
**Description**

I am working on refactoring the Unit component, and I plan to do it in 3+ PRs. The PRs I am creating are going to be merged to the "LF-2995+3002/Unit_displays" branch, and when everything is ready, the branch is going to be merged to the integration branch.
- [X] **LF-3003 Extract logic from Unit component and create a hook**
- [ ] LF-3004 Update Unit component to make it work as a presentation component
- [ ] LF-3005 Fix useUnit hook

In this PR, I just extracted logic from the unit component as `useUnit` hook and the logic is going to be fixed later on. No changes have been made to the code except that I removed `reactSelectWidth` from `useMemo` and left it in the Unit component. I plan to make a file for the `useReactSelectStyles` hook, pass `measure` and calculate `reactSelectWidth` in the hook in the next PR. (@JimmyRowland I know you worked on this component, please let me know if you see any issues with this approach! Basically, what I'm trying to do is to add necessary changes on top of your code which was working as expected)
- extract logic from the unit component and create a `useUnit` hook
- extract `getUnitOptionMap` and create a new file for it. "import" is updated in 5 files
- remove `reactSelectWidth` from `useMemo` and keep it in the Unit component

Jira link: https://lite-farm.atlassian.net/browse/LF-3003

**Checklist:**

- [X] I have added [the GNU General Public License](https://lite-farm.atlassian.net/l/cp/BT0Dd7WW) to all new files


JimmyRowland:
I think extracting custom hooks is great

packages/webapp/src/components/Form/Unit/useUnit.jsx
```diff
@@ -0,0 +1,212 @@
...
+            displayUnit: to,
+            displayValue:
+              defaultValue && roundToTwoDecimal(convert(value).from(databaseUnit).to(to)),
+            options,
+            databaseUnit,
+            isSelectDisabled,
+            measure,
+          }
+        : {
+            ...getDefaultUnit(unitType, value, system, databaseUnit),
```
JimmyRowland, line 81: This conditional branch is only used in location pages. Not sure if it's better to extract this to somewhere else.

packages/webapp/src/components/Form/Unit/useUnit.jsx
```diff
@@ -0,0 +1,212 @@
...
+            options,
+            databaseUnit,
+            isSelectDisabled,
+            measure,
+          };
+    }, []);
+
+  const hookFormUnitOption = hookFromWatch(displayUnitName);
+  const hookFormUnit = databaseUnit;
```
JimmyRowland, line 90: This is probably wrong

packages/webapp/src/components/Form/Unit/useUnit.jsx
```diff
@@ -0,0 +1,212 @@
...
+  useEffect(() => {
+    if (hookFormUnit && convert().describe(hookFormUnit)?.system !== system && measure !== 'time') {
+      hookFormSetValue(displayUnitName, getUnitOptionMap()[displayUnit]);
+    }
+  }, [hookFormUnit]);
+
+  useEffect(() => {
+    !hookFormGetValue(displayUnitName) &&
+      hookFormSetValue(displayUnitName, getUnitOptionMap()[displayUnit]);
+  }, []);
```
JimmyRowland, line 105: These three useEffect are probably wrong.

packages/webapp/src/components/Form/Unit/useUnit.jsx
```diff
@@ -0,0 +1,212 @@
...
+          };
+    }, []);
+
+  const hookFormUnitOption = hookFromWatch(displayUnitName);
+  const hookFormUnit = databaseUnit;
+  useEffect(() => {
+    if (typeof hookFormUnitOption === 'string' && getUnitOptionMap()[hookFormUnitOption]) {
+      hookFormSetValue(displayUnitName, getUnitOptionMap()[hookFormUnitOption]);
+    }
+  }, []);
```
JimmyRowland, line 95: When default values are put into react hook form, displayUnit will be 'm' instead of {lable: 'm', value: 'm'}.
This hook convert unit to unitOption

packages/webapp/src/components/Form/Unit/useUnit.jsx
```diff
@@ -0,0 +1,212 @@
...
+  useEffect(() => {
+    if (hookFormUnit && convert().describe(hookFormUnit)?.system !== system && measure !== 'time') {
+      hookFormSetValue(displayUnitName, getUnitOptionMap()[displayUnit]);
+    }
+  }, [hookFormUnit]);
```
JimmyRowland, line 100: When database display unit is in metric but farm is using an imperial unit, set unit to recommended imperial unit

packages/webapp/src/components/Form/Unit/useUnit.jsx
JimmyRowland: These three useEffect are probably wrong.
When the first effect hook set the initial unit option, second effect hook won't get proper hookFormUnit if block. hookFormUnit will update on next render.

packages/webapp/src/components/Form/Unit/useUnit.jsx
```diff
@@ -0,0 +1,212 @@
...
+      hookFormSetValue(displayUnitName, getUnitOptionMap()[displayUnit]);
+  }, []);
+
+  const [visibleInputValue, setVisibleInputValue] = useState(displayValue);
+  const hookFormValue = hookFromWatch(name, defaultValue);
+
+  useEffect(() => {
+    hookFormSetHiddenValue(hookFormValue, { shouldValidate: true, shouldDirty: false });
+  }, []);
+
```
JimmyRowland, line 113: This is probably related to some react hook form racing condition bug.
Maybe react hook form should be updated

packages/webapp/src/components/Form/Unit/useUnit.jsx
```diff
@@ -0,0 +1,212 @@
...
+  useEffect(() => {
+    if (hookFormUnit && hookFormValue !== undefined) {
+      setVisibleInputValue(
+        roundToTwoDecimal(convert(hookFormValue).from(databaseUnit).to(hookFormUnit)),
+      );
+      //Trigger validation
+      (hookFormValue === 0 || hookFormValue > 0) && hookFormSetHiddenValue(hookFormValue);
+    }
+  }, [hookFormUnit]);
+
```
JimmyRowland, line 123: This should depend on hookFormUnit and hookFormValue. Then if 
roundToTwoDecimal(convert(hookFormValue).from(databaseUnit).to(hookFormUnit)) is different from visible value, setVisibleValue.

packages/webapp/src/components/Form/Unit/useUnit.jsx
JimmyRowland: It's cleaner to write effect hooks + set values in terms of state sync conditions.
for example, 
convert(hookFormValue).from(databaseUnit).to(hookFormUnit) === displayValue should always hold.
Then there should be 3 effect hooks that listen to hookFormValue/hookFormUnit/displayValue, then setHookFormValue/displayValue so that the relationship holds.

I just don't know how to make it clean.

SayakaOno:
@JimmyRowland
Thank you so much for all the helpful comments!!! I will refer those when I update the logic!

I was trying writing tests using the testing-library last night, but I had to install everything from `testing-libary/react` to transformers... I knew `setUpTests.js` was already included and assumed that the environment was set up already, but it is not, right? Do you know if there is the right way to set it up?

JimmyRowland:
I also had trouble installing testing stuff after migrating to vite.

SayakaOno:
Okay...
I'm trying to mock `i18n` now, I didn't imagine running a test for the unit component would need such a lot of work!

Duncan-Brain:
Things I am thinking about... but I would still approve as it is now since no logic has changed:

1. Should `useUnit` hook be in `/containers` ? With the exception of just one other custom hook (which may be in the wrong place based on how recent it was made) all other custom hooks are in /containers. 

If yes ==> If useUnit should be in /containers are there any pieces that should be extracted from useUnit to a Unit container?

2. If you are passing `measure` next pr will that include `getOnKeyDown` as well?

JimmyRowland:
useUnit should be in the component folder as Unit component should be a pure UI component without side effect/global states.

All hooks/components in the container folder have side effects (network calls) or rely on global states (redux), If someone puts selectors in useUnit hook, all components that use Unit component need to be moved into the container folder as those components will not be reusable. 

Maybe there is a way to remove implicit coupling between the unit component and react hook form without moving unit component logic to the parent.

SayakaOno:
@Duncan-Brain 
> 1. Should `useUnit` hook be in `/containers` ? With the exception of just one other custom hook (which may be in the wrong place based on how recent it was made) all other custom hooks are in /containers.

I prefer to make a Unit container and call `useUnit` there. (the unit component will be almost like a presentation component) 
@JimmyRowland Do you have any thoughts on this?

And sorry for the hook in /components!

> If yes ==> If useUnit should be in /containers are there any pieces that should be extracted from useUnit to a Unit container?
 
I don't think so, but I would also call `useReactSelectStyles` that needs `reactSelectWidth`. So I would extract `reactSelectWidth` from `useUnit`. (my initial plan was to pass `measure` to `useReactSelectStyles` and calculate `reactSelectWidth`, but now I plan to let the hook calculate `reactSelectWidth`)


> 2. If you are passing `measure` next pr will that include `getOnKeyDown` as well?

I am not so sure what you are referring to, but I plan to clean up the Unit component (which means to move more things to the hook). I am going to return `onKeyDown` function from the hook.

Cleanup 🔽 
Screenshot 2023-03-03 at 1 55 17 PM

Duncan-Brain (approved):
@SayakaOno

So I am trying to get some LiteFarm architecture intuition out of this -- thanks for the guidance @JimmyRowland . The following is mostly for me but I will share anyways.

At a very basic level what I am getting from this is that: 

Despite our general practice that logic is put into a container --- Unit should not have a container because it is used by many components. It is generally bad practice to import a container from a component (though there may be special cases). All global states and side effects should be happening at a higher level.

JimmyRowland:
Apps with normalized redux stores are different from apps using react query/rtk-query.
With react query, you always know that the data returned from react query hook exists.
With a normalized redux store, when selectors are used, devs need to read through the component tree/user flow and make sure that the required data is fetched somewhere. It's easier to put all get requests/selectors at 'root' level.

It would be very difficult to build PWA with react query. On the task read-only page, to display the page offline, useTaskById for every task need to be cached. If there are 100 tasks, useTaskById needs to be called 100 times with 100 task ids when the app is online. 

I couldn't find good offline apps on github to learn from. Top-down prop drilling is the default/old react architecture. Hope you guys can find better pwa solutions

SayakaOno:
I need to learn about PWA first...!
Thank you for sharing your knowledge Jimmy, I really appreciate it.

By the way, I tried setting up vitest and react-testing-library. I got some errors in the integration branch, but I didn't in Jimmy's pwa1 branch!

JimmyRowland:
@SayakaOno must be vite version

kathyavini (approved):
The extractions seem good to me. I'm looking forward to going through the updated logic!

PR #2488, opened 2023-03-11
Add dependency to useEffect

SayakaOno:
**Description**

This PR is to add a missing dependency array to useEffect.


craigyu (approved):
Oops! thanks for fixing the lint

PR #2493, opened 2023-03-15
LF-3005(1): Add JSDoc and test ids to Unit and useUnit

SayakaOno:
**Description**

- add JSDoc
- add test ids for component testing

I found that "data-testid" cannot be added to `react-select`.
This is the [doc](https://testing-library.com/docs/ecosystem-react-select-event/) by testing-library for how to test the select, but it seems to need a label...

Once any of us finds a solution, I will add whatever necessary to the component.


packages/webapp/src/components/Form/Unit/index.jsx
```diff
@@ -137,6 +138,7 @@ const Unit = ({
           {showError && <Cross onClick={onClear} className={styles.crossWrapper} />}
         </div>
 
+        {/* TODO: add id to react-select */}
```
kathyavini, line 141: There is an example on GitHub of react-select-event being used without label here:
- https://github.com/romgain/react-select-event/issues/51

Copying some of the code from that example, we could just wrap the whole Controller + Select in a div with `data-testId`:

```
<div data-testId={`${testId}-select`}>
  <Controller
	control={control}
	name={displayUnitName}
   
	// ...
    
</div>
```

And then in @storybook/testing-library with react-select-event:

```js
import selectEvent from 'react-select-event';

//...

// Changing the unit of the select
await selectEvent.select(
  within(canvas.getByTestId('plant-spacing-select')).
  getByRole('combobox'),
  'm',
);

// Querying the unit of the select
const unit = await within(
  canvas.getByTestId('length-of-bed-select')).
  findByText('m');

expect(unit).toBeInTheDocument();
```

I think this might be the most reliable way to control the Select; I also tried `.getByRole('combobox')`, clicking, and searching for the unit text, but that was extremely flaky for me.

I haven't tried it with the fully updated Unit yet, though! Just on this current branch.

kathyavini:
Thank you so much for finding react-select-event! It worked very well for me, and I left a comment above with one possibility for where to implement the data-testId for the Select. The other two testIds (for the inputs) worked great!

I'm not familiar with JSDoc, so I can't really comment on it, but it looks useful.

SayakaOno:
@kathyavini 
Thank you so much for the example, it worked perfect for me!! I wrapped the select with a div as you did in your example and added some more test ids.

kathyavini:
That's a good idea to test the errors 🙂 I had forgotten about that!

I think we could remove the divider testid, if it was only used for testing the disabled status, just because it might make others think (as I did) that there is something functionally significant to test on the divider itself.

But of course, if you see it being useful for something else (or it's already a big part of your unit tests), it should stay!

SayakaOno:
I removed the id for the divider, thank you Joyce!

kathyavini (approved):

Duncan-Brain (approved):
Thats some nice JSDoc'ing !

PR #2499, opened 2023-03-17
LF-2986-Sensor-position-changes-when-there-are-multiple-sensors-at-the-same-location-and-tooltip-is-opened

javiscript13:
**Description**

This pull request addresses the new design requirements for the popup components when more than one items are on the same location. In order to achieve the new design, some rework on the 3 components that formed the popups has to be done.

Previously, the CompactPreview component was only used on PurePreviewPopup. However, with the new design, the CompactPreview component is used on both PurePreviewPopup and PureSelectionHandler, so CompactPreview now has full responsibility to manage the data to be able to display sensors information. As a result, PurePreviewPopup and PureSelectionHandler now only pass that information to the CompactPreview component so it can display it.

PureSelectionHandler has also a new design and functions to be able to display the CompactPreview when necessary.

I'm also adding a new selector so PureSelectionHandler can know which entities are sensors so it can add the arrow and CompactPreview for each legible sensor.

Jira link: https://lite-farm.atlassian.net/browse/LF-2986

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/webapp/src/components/Map/SelectionHandler/index.jsx
```diff
@@ -1,24 +1,102 @@
...
+    paddingBottom: '8px',
+  },
+  sensorArrow: {
+    display: 'flex',
+    flexDirection: 'row',
+    alginSelf: 'center',
+    transition: 'transform 1s',
+  },
+  sensorArrowUp: {
+    transform: 'rotate(-180deg)',
```
SayakaOno, line 99: Nice animation :)
(I'm not sure what Carolina will say, but personally I prefer faster 😆)

packages/webapp/src/components/Map/SelectionHandler/index.jsx
```diff
@@ -48,89 +126,117 @@ export default function PureSelectionHandler({ locations, history, sensorReading
...
+                  })}
+            >
+              <div className={classes.title}>
+                <div
+                  className={clsx(classes.itemIcon, showDetails(location.id) && classes.sensorIcon)}
+                >
+                  {' '}
+                  {icon}{' '}
+                </div>
+                <div className={classes.itemName}>{name}</div>
```
SayakaOno, line 196: It looks like some style adjustments are needed for a soil water content sensor.
Also, when a soil water content sensor is clicked, I think we need to open the edit view. Otherwise, we won't have access to that view.
Screenshot 2023-03-21 at 10 23 10 PM

SayakaOno (changes requested):
Thank you for working on this complicated UI!!
I left some comments :)

packages/webapp/src/components/Map/PreviewPopup/CompactPreview.jsx
```diff
@@ -27,9 +37,46 @@ const useStyles = makeStyles((theme) => ({
...
+  const loadReadingView = () => {
+    history.push(`/${location.type}/${location.id}/readings`);
+  };
+
+  let sensorData = [];
+  let sixHoursBefore = new Date();
+  sixHoursBefore.setHours(sixHoursBefore.getHours() - 6);
+
+  if (readings.length) {
+    sensorData = sensorReadings.filter(
```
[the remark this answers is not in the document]
javiscript13: fixed to the right parameter name.

packages/webapp/src/components/Map/SelectionHandler/index.jsx
```diff
@@ -48,89 +126,117 @@ export default function PureSelectionHandler({ locations, history, sensorReading
...
+                {showDetails(location.id) && (
+                  <div>
+                    {' '}
+                    {
+                      <ShowMore
+                        className={clsx(
+                          classes.sensorArrow,
+                          sensorIdx === idx && classes.sensorArrowUp,
+                        )}
+                      />
```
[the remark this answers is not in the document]
javiscript13: The spaces were already passed to css but I did left behind the {' '} that were usless. It is now fixed.

packages/webapp/src/components/Map/SelectionHandler/index.jsx
javiscript13: I did sent a video to Caro, she approved it as it was, so I guess I will leave it that way.

javiscript13:
@Duncan-Brain 

About 1:  I did fix the 4 elements limit, and ordered by type (as it was) and then by name.
About 2: I really didn't do anything to fix it, but tried several times and wasn't able to replicate the issue. Please check again and let me know if it continues to fail to investigate further.
About 3: I will quote "would be cool but no obligation" on this one :D

Thanks for the past and for the coming review!!

packages/webapp/src/components/Map/SelectionHandler/index.jsx
```diff
@@ -139,6 +139,11 @@ export default function PureSelectionHandler({ locations, history, sensorReading
     );
   };
 
+  const isSensor = (id) => {
+    let sensor = readingTypes.find((sensor) => sensor.location_id === id);
+    return sensor?.reading_types;
```
SayakaOno, line 144: `let` -> `const`
I think it is better to return boolean: `return !!sensor?.reading_types`

javiscript13:
@mickenorlen the branch for this PR was too many commits behind, I did pull all integration changes on it and then tried your csv and it worked (I didn't check before pulling integration changes, but it might be that), would be great if you could try again (the long click was for the previous component, in the new one you only have to click the number in order to appear).

image

SayakaOno (approved):
Looks good to me, thank you!!

mickenorlen:
@javiscript13 The problem remains. I was trying to find some clues to why the popup isn't showing. The code was quite complex and I would need more time to find any clear answers. I'm not sure what event is supposed to be triggered by clicking on the numbers? At least I can say that `useSelectionHandler.js#handleSelection` isn't triggered at all :P

Would be nice if you could point me in the right direction for debugging. Or we could do this later and let Duncan review it for now.

Duncan-Brain (approved):
Thanks for fixing! It seems to be working with new changes on Farm map.

For some reason the task views have returned to the old view as you mentioned. Maybe #2 was a caching thing. @mickenorlen you could possibly try in incognito or clear your browser cache. We can debug at some point if you like.

I am thinking it might be good to open two new Jira tickets:
- Update multiple map locations in same place across the whole app (example Field work task)
- Also in the (rare) case that there are a lot of sensors this component may need a way to handle overflow with a scroll option.

PR #2460, opened 2023-02-10
update react, vite, storybook, chromatic, and replace vite-mdx with r…

JimmyRowland:
update react, vite, storybook, chromatic, and replace vite-mdx with rollup mdx plugin.
Add vite pwa.
Fix cypress imports.
Fix stories.
Add offline crop grey box to pwa.
Move crop selector out of PlantingLocation component Fix crop variety photo url prop typo
Add OfflineDetector and Selector
Lazy load compressorjs
deprecate getMeasurementFromStore/getCurrencyFromStore/grabCurrencySymbol and add useCurrencySymbol hook deprecate cloneObject
Fix onReqSuccessSaga so that history stack will reset to a proper state after multi page form submission.
Fix some cypress tests, still can't figure out a reliable react select selector

PR #2514, opened 2023-03-24
LF-3078 Allow mobile device testing over local network by replacing @tsndr/cloudflare-worker-jwt with react-jwt

kathyavini:
**Description**

Currently, it is not possible to test LiteFarm locally using our mobile devices because our frontend jwt package [tsndr/cloudflare-worker-jwt](https://github.com/tsndr/cloudflare-worker-jwt) gives an app-crashing error about unsupported crypto over http addresses other than localhost.

This PR replaces this package with the equally lightweight [react-jwt](https://github.com/gustavo0197/react-jwt) to solve this issue and allow localhost testing on mobile devices. The only necessary change to start using LiteFarm on your phone is changing the `VITE_API_URL` in `webapp/.env` to your computer's network name or local IP address.

**What will work out of the box:**
- Once you update your `webapp/.env` file with your computer's hostname or IP, you can access LiteFarm over your local area network on any device (NOTE: if you try to log in and nothing happens when you click the "Continue" button, that's a good sign that you haven't set the correct network address for the API in the `.env` file!)
- Most features of LiteFarm will work automatically over the local network

**What will not work:**
- Google SSO (without some proxying or networking workarounds that I have not tried). Google does not allow you to add a local domain or IP to the authorized JavaScript origins of your Google OAuth Client ID. You will have to use password-based accounts when testing on your phone

**What can work, but is probably not worth the effort:**
- It's fairly easy to change the hardcoded instance of "localhost" in the API `sendEmailTemplate.js` to an environment variable to get password resets and invites that you can open on your phone, document export and retrieval on your phone (if you have the export server running locally already and update your MinIO endpoint in the frontend `.env`), and probably even more! But I decided to keep that type of flexibility out of this PR, as it becomes a bit of a pain when you switch from localhost to device and back again
- I recommend changing those hardcoded instances only when you specifically need to test one of those services on your phone for an extended period of debugging 

Jira link: https://lite-farm.atlassian.net/jira/software/c/projects/LF/issues/LF-3078

**Type of change**

**How Has This Been Tested?**

- [x] Other (please explain)

Both @javiscript13 and I have tested this on our phones over the local area network. @javiscript13's local development environment (unlike mine) is set up in Docker so it should be compatible with that as well (thanks for the testing and feedback, Javier!)

**Checklist:**


Duncan-Brain:
Everything works as advertised! Bot reset password and invite linked worked for me.

Things I am thinking about:

1. @JimmyRowland mentioned large package size last we discussed it. I think I saw on npm this is ~19kB compared to ~17kB from cloudflare. I have no concept of what is too big but would this be a opportunity to do some codesplitting like Jimmy did in the ImagePickerWrapper from the big [vite update pr](https://github.com/LiteFarmOrg/LiteFarm/pull/2460/files#diff-f895604b31b5a42475911bf9d351ab01714304a452c8aca7ec3a635471c55758)? Like:

`const Compressor = await import('compressorjs').then((Compressor) => Compressor.default);`

2. Is the VITE_API_URL a candidate for the old `development ? (VITE_API_URL ?? localhost:5001) : VITE_API_URL`

3. An update to the .env.default help text would be nice.

SayakaOno (commented):
I think you have changes related to storybook in `pnpm-lock.yaml`...

kathyavini:
@SayakaOno Very strange, because I installed from a frozen lockfile! Even stranger: I just now added react-jwt from a fresh `--frozen-lockfile` (integration branch) pnpm install again, and it has updated yet again:

Screenshot 2023-03-24 at 6 04 06 PM

So this is the first I've learned that the pnpm default is to [update all packages on install](https://stackoverflow.com/questions/74305317/pnpm-add-package-name-updating-all-packages). A strange default!

Not sure if the best approach with this is to 1) `add --patch` on the lockfile, which makes me a little nervous, 2) be more strict with our Storybook-related versioning in package.json, or 3) just commit the Storybook upgrades as they happen. Having done a normal install once (and presumably cached the result) neither `--offline` or `--prefer-offline` tags do anything -- Storybook is ALWAYS upgraded!

JimmyRowland:
https://bundlephobia.com/package/react-jwt@1.1.8
https://bundlephobia.com/package/@tsndr/cloudflare-worker-jwt@2.2.1
react jwt is smaller. bundle size should be smaller than 582B with tree shaking and gzip. compressor.js is quite large. I should have checked bundle size. Sorry for the confusion 

npm updates minor versions. 
Storybook 7 is still in beta. Not sure whether what's the best solution

SayakaOno:
@kathyavini 
Thank you for looking into it!
Should we be concerned about having the latest version of storybook...? If not, I think we can go with the default. (Sorry but I'm not sure about what to do...)

kathyavini:
@Duncan-Brain:

1. Jimmy answered this one
2. See below!
3. & 4. Done! But.... check out below. I'm actually now thinking this should be approached differently.

Re: question #2... this is actually really interesting!! I think `VITE_API_URL` it was originally meant to be used that way, as an optional variable. Check out the code in `webapp/src/apiConfig.js`:

```
  } else if (import.meta.env.VITE_API_URL?.length) {
    URI = import.meta.env.VITE_API_URL;
  } else {
    if (VITE_ENV === 'development') {
      URI = window.location.href.replace(/3000.*/, '5000');
    } else if (VITE_ENV === 'production') { //...
```

Currently, we have to set this variable to get the webapp to work, but that's only because that port above _should_ be 5001, not 5000. If we set the port correctly in that line and then _delete_ the `VITE_API_URL` entirely, you can switch between localhost or device without having to change any `.env` variable at all!

What do you think of removing `VITE_API_URL` and just instructing to make sure it's removed in the readme for those who set up their `.env` previously? I would have to change the readme and the `.env.default` again but I think that automatic URL switching is a way better solution.

Duncan-Brain (approved):
As discussed, the automatic port switching would be nice but it seems like we may need to iron out some conflicts with Cypress. Both ports 5000 and 5001 are used in the codebase and it is probably outside of the scope of this PR.

Everything else looks good to me!

SayakaOno (approved):

PR #2517, opened 2023-03-24
LF-3008 Augment additional survey questions xlsx to handle nested null responses without failing

kathyavini:
**Description**

Certification document export will fail (the whole export process will crash and no documents will be sent) if an exported certification survey has a structure with nested null values.

Nested null values means that a skipped question/group also contains one or more child questions/groups, e.g. when a whole section of a survey is not shown due to [Relevance Expressions](https://our-sci.gitlab.io/software/surveystack_tutorials/relevance_expressions/) (conditional scripts) set up in the survey. Note that skipped questions in themselves are not a problem if they don't contain child questions/sections.

This PR adds a check for undefined parent groups so that this kind of nested null value does not lead to document export crashing.

Jira Link: https://lite-farm.atlassian.net/browse/LF-3008

**Type of change**

**How Has This Been Tested?**

- [x] Other (please explain)

To test this, you will need to have the [export server running locally](https://lite-farm.atlassian.net/wiki/spaces/LITEFARM/pages/1190101039/The+export+jobs+pseudo-package#Running-the-export-server-locally). Add the [IAF Survey Portugues](https://app.surveystack.io/surveys/63fd38cf9c98ce0001ace151) to your certifiers table and access it on the frontend.

To generate the nested null values that previously crashed the export server, answer the survey in such a way that you skip at least one section. I recommend:

- For Question 5.1 "Você tem relação direta (sem a existência de intermediários) com os consumidores?" select "Nao"
- For Question 6.1 "Após a pandemia, mudou a importância dos mercados locais (circuitos curtos de comercialização) para escoamento da produção agroecológica?" select "Nao"
- For Question 7 "Os membros da propriedade trabalham exclusivamente com agricultura?" select "Sim"

The expected behaviour is that the survey (and other documents) will export successfully. Note that the Excel columns generated are a bit funky in width... you might have to scroll significantly to the right to see the answer "x" marks.

**Checklist:**

### DIFF
--- a/packages/api/src/jobs/certification/survey_record.js
+++ b/packages/api/src/jobs/certification/survey_record.js
@@ -85,6 +85,11 @@ export default async (emailQueue, submission, exportId, organicCertifierSurvey, 
     for (const group of parentGroups) {
       if (group != null) {
         answer = answer[group];
+
+        if (answer == undefined) {
+          // i.e. parent group had no answer
+          return undefined;
+        }
       }
     }
     return answer[name]?.value;


packages/api/src/jobs/certification/survey_record.js
```diff
@@ -85,6 +85,11 @@ export default async (emailQueue, submission, exportId, organicCertifierSurvey,
     for (const group of parentGroups) {
       if (group != null) {
         answer = answer[group];
+
+        if (answer == undefined) {
+          // i.e. parent group had no answer
+          return undefined;
+        }
```
SayakaOno, line 92: I am confused...
If the first `group` has `answer[group]` and the second `group` also has `answer[group]`, is the `answer` overridden? Or `parentGroups` length is less than 2?

I am not sure about the details, but it seems more natural to place the if block right before the return. or do `return answer?.[name]?.value;`
But It looks like the length is always 1? I'm completely confused...

kathyavini: Ok, it's very possible that my sense of this function could be quite wrong 😜 

But my understanding is that each iteration through parentGroups is replacing the submission data object with the subset of data that applies to that level of nesting (narrowing it each time).

My thinking with placing the check inside the loop was to prevent a TypeError in the following iterations when accessing `answer[group]` once we've gotten to the first undefined parent question. But clearly that wasn't right because `return answer?.[name]?.value` works as well -- at least for all the IAF submissions! 😆 

I thought both the "name" and the "parentGroups" are originally coming from the survey questions (and not the submission data) so I'll have to look at that again and report back.

But I don't think that answers your concern about length... does it?

SayakaOno: That is something that I was not thinking about!
This function deals with something like these, right?
```
// parentGroups = ["group1", "group2", ...];
// submissionData.data ⬇️ 
// { 
//   "group1": {
//       "group2: {
//          ...
//        }
//    }
// }
```
I don't think length matters, and you don't need to change your code. Sorry!

kathyavini: Yes, that's exactly right. Would it be good to include the shape of the data in a code comment here (just like your example above)? I don't think this is the easiest function to read...!

SayakaOno: I guess I should have figured it out from the JSDoc, so I don't think comment is necessary!
I'll approve, thank you Joyce!

SayakaOno (approved):

packages/api/src/jobs/certification/survey_record.js
Duncan-Brain: @kathyavini 

(I think) my question is related so I will hop on this thread haha. 

Maybe you can provide some info for me on this. Looking at the function it seems like it can only go one nest deep. I was half expecting a recursive function here in case it is possible to deeply nest questions. But maybe I don't understand surveystacks and the parent groups: Does Parent groups return all questions with a child? And is it even possible to have multiple nesting of questions in surveystack? (I tried but it was hard to navigate and figure out).

Duncan-Brain:
@kathyavini 

The survey exports successfully for me. Let me know if I am missing something on my inline comment -- otherwise I can approve.

packages/api/src/jobs/certification/survey_record.js
kathyavini: @Duncan-Brain oops I left this PR for a while between being sick + Units but I should really wrap it up before the end of sprint! 🙂 

> Looking at the function it seems like it can only go one nest deep.

The `getNestedAnswer` function can go any level of nesting deep. How nested the question is determines the length of `parentGroups` -- I think the clearest illustration is in Sayaka's code example above.

> And is it even possible to have multiple nesting of questions in surveystack?

Definitely. To use a real question from the [IAF survey](https://app.surveystack.io/surveys/63fd38cf9c98ce0001ace151/edit), the `ocorre_relacao` item is three levels of nesting deep (5 group_relacoes > 5.2 producao_consumo > 5.2.1 ocorre_relacao), so for that item, this function is called with `parentGroups = [null, 'group_relacoes', 'producao_consumo']`.  The `for.. of` loop runs from the outer group to the inner, returning a more narrow subset of the submission data each time as described in the thread above.

> Does Parent groups return all questions with a child?

Not sure I 100% understand, because the function is only ever called on one question/item (the `name` argument) at a time and with that particular question's parent groups. But the `answer[group]` part of the loop would return all the child nodes for that group that are present in the submission data.

> I was half expecting a recursive function here in case it is possible to deeply nest questions

I guess It is recursive-ish because each iteration through the `for... of` loop is acting on a further subset of the same data in a similar way, traversing down the tree. I'm pretty sure that could be written recursively, although I don't think there is any actual recursion right now.

kathyavini:
@Duncan-Brain Ok,I responded to your inline comment! Hopefully in a way that was helpful and not just complicating the issue unneessarily 😆

Duncan-Brain (approved):

PR #2521, opened 2023-03-27
Improve api logging in development

mickenorlen:
## Description
- Improve/fix api development logs using the already included Winston log package.
- Add new `console.dev` that prints heavily highlighted output for dev debugging (see second screenshot below). This is useful also to easily find and clean out debugging log messages before pushing your work - they aren't meant to be committed to the app.
- Timestamp respects `TZ` in `.env`. 
- Also added a simple route logging middleware (blue lines in new log).

This PR intends to target only api development logs for now.

I think the node knex db transaction log is very useful, activated by `DEBUG=knex:query`. Once we merge the docker setup, it will be easy to set the DEBUG  variable directly in `.env`. But for now, you have to integrate it in your own commands, see examples in api `package.json`.

Old api log
image

New api log
image

## Testing

1. Add `console.dev('Hello')` at beginning of action `notificationUserController#subscribeToAlerts`
2. Go to http://localhost:3000, refresh
3. See the api logs and the highlighted hello output
4. Check that the code logic doesn't change non-development environments

### DIFF
--- a/packages/api/src/common/logger.js
+++ b/packages/api/src/common/logger.js
@@ -1,31 +1,80 @@
 import winston from 'winston';
 import DailyRotateFile from 'winston-daily-rotate-file';
+import moment from 'moment';
+import express from 'express';
+const router = express.Router();
 
-const logger = winston.createLogger({
-  level: 'info',
-  format: winston.format.json(),
-  defaultMeta: { service: 'user-service' },
-  transports: [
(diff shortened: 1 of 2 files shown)


iperdomo:
Thanks for your contribution. We'll take as inspiration for future improvements. As of today, we'll not merge the PR as is.

PR #2522, opened 2023-03-27
LF-3077 Mismatch between read-only and modify to complete task views

liao02x:
**Description**

A follow-up on https://github.com/LiteFarmOrg/LiteFarm/commit/603e8f7faa0d008835f630df3a09e75f5911e866. According to the commit, the checkbox values should be set to false when sending the data through the API call. What this commit did is change the form value from the form component packages/webapp/src/components/Task/PureIrrigationTask/index.jsx, which cause the visual mismatch and confusion for the user. What I did here is change those values only when it is sending the data through the API, so it maintains the functionality and resolves the visual issue.

Jira link: https://lite-farm.atlassian.net/browse/LF-3077

**Type of change**

**How Has This Been Tested?**

I populated my local database with beta server backup and tested it on my local. The visual components show up correctly, and the API call is verified to send the correct data.

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/containers/Task/saga.js
+++ b/packages/webapp/src/containers/Task/saga.js
@@ -616,6 +616,12 @@ const getCompleteIrrigationTaskBody = (task_translation_key) => (data) => {
         ? data.taskData[taskType]?.irrigation_task_type_other
         : data.taskData[taskType]?.irrigation_type_name;
       data.taskData[taskType].location_id = data.location_id;
+      if (data.taskData[taskType]?.is_modified_on_completion) {
+        if (data.taskData[taskType]?.default_irrigation_task_type_location)
+          data.taskData[taskType].default_irrigation_task_type_location = false;
+        if (data.taskData[taskType]?.default_irrigation_task_type_measurement)
+          data.taskData[taskType].default_irrigation_task_type_measurement = false;
+      }
       data.taskData.location_defaults = [
         {
           location_id: data.location_id,
@@ -643,6 +649,7 @@ const getCompleteIrrigationTaskBody = (task_translation_key) => (data) => {
           'location_size_unit',
           'irrigation_type',
           'irrigation_type_translation_key',
+          'is_modified_on_completion',
         ].includes(element) && delete data.taskData[taskType][element];
       }
     }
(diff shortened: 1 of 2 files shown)


packages/webapp/src/containers/Task/saga.js
```diff
@@ -616,6 +616,12 @@ const getCompleteIrrigationTaskBody = (task_translation_key) => (data) => {
         ? data.taskData[taskType]?.irrigation_task_type_other
         : data.taskData[taskType]?.irrigation_type_name;
       data.taskData[taskType].location_id = data.location_id;
+      if (data.taskData[taskType]?.is_modified_on_completion) {
+        if (data.taskData[taskType]?.default_irrigation_task_type_location)
+          data.taskData[taskType].default_irrigation_task_type_location = false;
+        if (data.taskData[taskType]?.default_irrigation_task_type_measurement)
+          data.taskData[taskType].default_irrigation_task_type_measurement = false;
```
SayakaOno, line 623: @Duncan-Brain I think this was what Craig did in his last PR, but do we need to update these when completing a task?

packages/webapp/src/containers/Task/saga.js
```diff
@@ -643,6 +649,7 @@ const getCompleteIrrigationTaskBody = (task_translation_key) => (data) => {
           'location_size_unit',
           'irrigation_type',
           'irrigation_type_translation_key',
+          'is_modified_on_completion',
```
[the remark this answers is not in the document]
liao02x: done

packages/webapp/src/containers/Task/saga.js
Duncan-Brain: @SayakaOno It is similar and continues the work(his file changes: https://github.com/LiteFarmOrg/LiteFarm/pull/2473/files). 

These fields show the user what they did on that task during task creation. It should not be editable on completion. How defaults work in our architecture was not well considered on creation. Hopefully we can refactor user and farm level defaults in the future.

Duncan-Brain:
@liao02x 

Thank you for looking into this! So the expected behaviour for this is really confusing. Some of what I noticed may or may not be related to your fix. I have asked for clarification on the Jira ticket about the expectations so that I am not giving too much bad advice haha

In short: The problem I am seeing (possibly also on integration) is that the default boolean value on irrigation_task table should not be updated when the read-only form is submitted. Once the task is created these values should not be editable/updateable.

PR #2525, opened 2023-03-28
Lf 2996 update radio component

gursimran-singh:
**Description**

We wanted to use the `<Pill />` in Radio, but it was not possible because Radio (`<label />`) cannot have Pill that has  `<div />` inside it. So, performed the following changes to accept `<Pill />` component inside `<Radio />` component. 

- updated the Radio component(change `<p />` to `<span />` )
- updated the Pill component (change `<div />` to `<span />`) 

Jira link: https://lite-farm.atlassian.net/browse/LF-2996

**Type of change**

**How Has This Been Tested?**

- got to `/tasks`
- The `<Pill />` component having logged in person name, should be in `<span />`. Use Inspect element to verify that.
- Click on any task and then click on "Abandon this task" link.
- The label for date options for "When was the task abandoned?" should be in `<span />`

- [x] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/Filter/Pill/index.jsx
+++ b/packages/webapp/src/components/Filter/Pill/index.jsx
@@ -6,7 +6,7 @@ import { BsX } from 'react-icons/bs';
 
 const Pill = ({ className, label, selected, removable, onRemovePill }) => {
   return (
-    <div className={clsx(className, styles.pill, selected ? styles.selected : styles.deselected)}>
+    <span className={clsx(className, styles.pill, selected ? styles.selected : styles.deselected)}>
       {label}
       {removable && (
         <BsX
@@ -18,7 +18,7 @@ const Pill = ({ className, label, selected, removable, onRemovePill }) => {
           onClick={onRemovePill}
         />
       )}
-    </div>
+    </span>
   );
 };
 
--- a/packages/webapp/src/components/Form/Radio/index.jsx
+++ b/packages/webapp/src/components/Form/Radio/index.jsx
@@ -38,9 +38,9 @@ const Radio = ({
         {...props}
         disabled={disabled}
       />
-      <p className={clsx(styles.label)} style={classes.label}>
+      <span className={clsx(styles.label)} style={classes.label}>
         {label}
-      </p>
+      </span>
       {toolTipContent && <Infoi content={toolTipContent} />}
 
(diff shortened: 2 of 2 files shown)


SayakaOno (approved):
Looks good, thank you for taking care of this!

mickenorlen (changes requested):
Hey!

I just worked with Pills and radios on another [issue](https://github.com/LiteFarmOrg/LiteFarm/pull/2536) where I had to add dates in a pill next to a radio-box. I found that there were already 3 different pill components in the code:

- `Filter/Pill`
- `Filter/checkboxPill`
- `Filter/filterPillSelect`

I felt that all of them were intended/created for a particular context around filters, and I wasn't working on filters. I also started to feel like it probably isn't necessary to have that many Pill components - instead other components could reuse a variant of a base `/Pill`.

So instead of for example creating a `radioPill` or reusing the filter Pills, I created the "base" `/Pill` - a slightly modified `Filter/Pill`. Then I added `pill` as a property (similarly to how we use eg: `label`) to the `radio` and `radio-group` components, which in turn render the new base Pill component and manage its active/inactive color states.

In the future I thought we could even consider removing the filter pill components, and instead having relevant checkboxes or filter-select components take `pill` as a property - which in turn renders some version of the base `/Pill`. 

- Do you think this approach makes sense, would it be something that you could integrate in this issue as well?
- I'm not sure how far ahead you are in your work? How do we best move forward together?

gursimran-singh:
Hi Mikael, yes you are right. we do not need to have a different component, like in your case `radioPill`, for every use case instead we can use the previously build components. This is what we are intending to do in this ticket. Modifying `<Pill />` so that it can be used inside `Label` of other HTML elements. I'm working on same kind of [issue](https://lite-farm.atlassian.net/browse/LF-2997). 

I think it will be out of scope of this ticket to modify other filter components and integrate base `<Pill />` into them. we can create another ticket to address this issue after discussion with team.

gursimran-singh:
I think @SayakaOno would be the right person to decide as what we should be doing as both have same issue and require a common solution.

SayakaOno:
I don't think we need to create a new component if the existing pill component can be used for the "date".
If I am not mistaken, what's confusing is that the location of the pill component. I think we can create a ticket to consider relocating the file later to `/components`.

PR #2533, opened 2023-03-30
Lf 3050 fix estimated price per unit input in update estimated crop revenue

Duncan-Brain:
**Description**
Since very little change was needed in the actual UpdateEstimatedCropRevenue component... 
 
This PR also makes changes necessary to base component in order to fix the bug: 
- augments the unit component to allow for inverse units conversions
- refactors code to use this new convert (except `/util/index.js` I think these functions are deprecated)
- augments the unit component to display the currency symbol and removes deprecated currency grabbing from this file

(DRAFT) Yet to do:
- Add tests for new features - COMPLETE
- Comments explaining a few things - COMPLETE
- Decide whether breaking WaterUseCalculator is ok - OK BUT UNDONE (reverted all updates to other component files)
- memoize convert function - NOT DEEMED NECESSARY

Jira link: [LF-3050](https://lite-farm.atlassian.net/browse/LF-3050?atlOrigin=eyJpIjoiYjA1OTU4YzJhMTgzNDRiNzhiNWRmN2U2MDkzYTYyZGEiLCJwIjoiaiJ9)

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain): Using storybook run `pnpm test-storybook src/stories/Form/Unit/` , view currency changes in storybook itself

**Checklist:**

[LF-3050]: https://lite-farm.atlassian.net/browse/LF-3050?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/webapp/src/util/convert-units/unit.js
```diff
@@ -253,6 +253,21 @@ export const seedYield = {
...
+    breakpoints: [1000],
+  },
+  imperial: {
+    units: ['lb', 't'],
+    defaultUnit: 'lb',
+    breakpoints: [2000],
+  },
+  databaseUnit: databaseUnit.mass,
+  invertedUnit: true,
+};
```
SayakaOno, line 269: Maybe we can remove the breakpoints for now?

My concern:
1. User's unit system is "imperial", and user inputted "460$/lb" (= 1014$/kg)
2. The user changed unit system to "metric"
3. The value (460$/lb) is now displayed as "1014000$/mt" (it's okay if what the user wants to see is 1014000$/mt rather than 1014$/kg)

packages/webapp/src/components/Form/Unit/index.jsx
```diff
@@ -120,13 +121,20 @@ const Unit = ({
           )}
         </div>
       )}
-
+      {currency && (
+        <div className={styles.currency} data-testid={`${testId}-currency`}>
+          {currency}
+        </div>
+      )}
```
SayakaOno, line 128: Is there a reason you chose to add the currency here? I would add it to line 131, so I'm curious.

SayakaOno (commented):
Thank you for adding the test and stories! Everything seems working well!!
I left a few comments :)

packages/webapp/src/components/Form/Unit/index.jsx
Duncan-Brain: Yeah, I agree line 131 is better. But essentially I copied `/Form/Input` and on that component all the elements are flat with not as much nesting. The css seems to be based upon the entire element including label.  

So that  allowed me to simply copy and paste, but I can adjust the css so that it works in that position if you would prefer? ...  Using `UpdateEstimatedCropRevenue` as a starting point it looks like ..~1/3 of a pixel out of alignment either way haha (As it is currency symbol looks slightly higher than Input, if I change the css translation from (8,34) to (8,13) it looks slightly lower than Input).

I am happy to change if you prefer.

SayakaOno: Since you agree, it would be great if you could change.
Thank you Duncan!

Duncan-Brain: Changed

SayakaOno (approved):

packages/webapp/src/components/Form/Unit/unit.module.scss
```diff
@@ -149,3 +149,15 @@
...
+.currency {
+  font-size: 16px;
+  line-height: 20px;
+  color: var(--labels);
+  font-style: normal;
+  font-weight: normal;
+  font-family: 'Open Sans', 'SansSerif', serif;
+  position: absolute;
+  left: 0;
+  transform: translate(8px, 13px);
```
kathyavini, line 162: To fix the slight vertical offset issue you could use 13.5px here. Alternatively (and I think preferably), since the currency and the input are in a flex container, you could add `align-self: center` here and make this `transform: translateX(8px)` only; that would also line up the symbol and the number and I think it would be a bit more robust (but a bit less similar to Input).

kathyavini (approved):
CSS can be fixed in another quick PR so it shouldn't hold up merging!

PR #2519, opened 2023-03-24
LF-3055: Replace Unit component for % with Input component in WaterUsageCalculatorModal

SayakaOno:
**Description**

- Replace `<Unit />` used for `% with Input component` with `<Input />` in WaterUsageCalculatorModal. (The unit component is not designed to be used for percentage)
- pass `errors` to `WaterUsageCalculatorModal` from its parent component.
- delete `percentage_location` object which should not be used.

Note: You will see a warning in the console when you open the water use calculator modal, and it is going to be addressed in this [ticket](https://lite-farm.atlassian.net/browse/LF-3080). (it is not related to the input component)

Jira link: https://lite-farm.atlassian.net/browse/LF-3055

**Type of change**

- [x] Bug fix (non-breaking change which fixes an issue)

**How Has This Been Tested?**

- [x] UI components visually reviewed on desktop view

1. go to /tasks
2. "+ Create a task"
3. Select “Irrigate”
4. "Continue" until you see "Tell us about the Irrigation work task"
5. Select "Depth" for “How do you measure water use for this irrigation type?”
6. Click “Calculate Water Usage”
7. the calculator should look and work as before

**Checklist:**

- [ ] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [x] The precommit and linting ran successfully
- [ ] I have added or updated language tags for text that's part of the UI
- [ ] I have added "MISSING" for all new language tags to languages I don't speak
- [ ] I have added [the GNU General Public License](https://lite-farm.atlassian.net/l/cp/BT0Dd7WW) to all new files

PR #2550, opened 2023-04-06
LF-3048 Replace Location size Input with Unit component in PureBroadcastForm

kathyavini:
**Description**

This PR updates the following in the Broadcast Method planting method view (PureBroadcastForm component):
- Fixes Location Size so that it is a `<Unit />` rather than an `<Input />` component
- Since two new values are being registered with React Hook Form, cleans the broadcast_method data of these values before sending to the API
- Removes the dependency between the units of Area planted and the value of Location Size. Previously Location Size would be re-calculated (converted) according to the unit of Area planted; now only Area planted will change its unit according to the normal breakpoint for area when the % of Location is adjusted. See Jira ticket for details.
- Fixes the validation of Location Size (min to 0 instead of 1) and matches its format to the corresponding % Input component in #2519

Not touched in this PR:
- This PR hasn't changed any calculations (other than removing the conversion of Location Size as described above) as they appear to already be functioning correctly

Jira link: https://lite-farm.atlassian.net/browse/LF-3048

**Type of change**

**How Has This Been Tested?**

- [x] Passes test case
- [ ] Other (please explain)

**Checklist:**


SayakaOno (dismissed):
Thank you so much for taking care of this, and thank you for deleting `location_size` and `location_size_unit` properly!!

Duncan-Brain (approved):
Works for me!

Interesting that you have to delete extraneous form data. I actually haven't had to use this yet I seem to have mostly worked on components that use the baseController in the api that automatically removes extra properties.

It does seem like there is an issue pushing data down from the map selection. There should be some combination of setting default values on useForm() and defaultValues in Unit component.

SayakaOno (approved):

PR #2552, opened 2023-04-06
LF-3054: Fix calculation for total water usage in water usage calculation modal

SayakaOno:
**Description**

- Fix total water usage calculation (Please check if the formulas are correct!)
- Use `defaultValue` and `to` props for `Unit` component to set default values (flow rate, location size, and application depth). `setValues` are removed.
- Update irrigated area unit depending on “% of location to be irrigated”
- Pass `location` from its parent component rather than doing `useSelector` in the child component (this is to be able to write storybook tests)
- Delete `location_size` before sending an API request for irrigation task creation (this should have been done in [LF-3055](https://lite-farm.atlassian.net/browse/LF-3055))
- Fixed `location_area` constant (now it’s the same as `area_total_area`)
- Update `UnitTest` class so that it can be used for modals. 
`canvas` cannot be used for this reason:
  > If you do have elements that get projected into document.body, then you should, for that particular case, use screen.getBy instead of canvas.getBy, and that should suffice.
  
   (https://github.com/storybookjs/testing-library/issues/25#issuecomment-1184255825)

Jira link: https://lite-farm.atlassian.net/browse/LF-3054

storybook UI: http://127.0.0.1:6006/?path=/docs/components-modals-waterusagecalculatormodal--docs
test: `pnpm test-storybook /src/stories/Modal/WaterUsageCalculatorModal.stories.jsx`

**Type of change**

**How Has This Been Tested?**

- [x] Passes test case
- [ ] Other (please explain)

**Checklist:**

[LF-3055]: https://lite-farm.atlassian.net/browse/LF-3055?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/webapp/src/components/Modals/WaterUsageCalculatorModal/index.jsx
+++ b/packages/webapp/src/components/Modals/WaterUsageCalculatorModal/index.jsx
@@ -13,6 +13,7 @@ import {
   location_area,
   roundToTwoDecimal,
   water_valve_flow_rate,
+  getDefaultUnit,
 } from '../../../util/convert-units/unit';
 import Checkbox from '../../Form/Checkbox';
 import { Label } from '../../Typography';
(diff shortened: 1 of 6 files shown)


kathyavini (commented):
The calculations look 👌 to me!

Just two small bugs...

1. I know you didn't touch this part of the code, but the `<Unit />` component that stores the result of the calculation has a buggy `onChangeUnitOption` handler (PureIrrigationTask/index.jsx at line 294). Could that be replaced with `autoConversion={true}` since the autoconversion prop already works perfectly?
2. Just a small thing, and not sure why this happens,  but the visible Input of the Estimated flow rate doesn't re-populate its value when you select "Set as default flow rate for its location" -- only the hidden Input!

Screenshot 2023-04-07 at 10 56 23 AM

Duncan-Brain:
I agree with @kathyavini ... calculations look good.

But Water Use Calculator is so buggy in general. I do not think it should be using any conversion functions and just letting Unit component handle everything.

Happy to approve since calculations look good

SayakaOno:
@kathyavini 
1. Thank you so much! I wrote tests for estimated water usage, but I didn't catch that bug! What are the tests for!!!
I will just remove `setValue` in `onChangeUnitOption`. I honestly can't wait to use `autoConversion` somewhere, but it feels strange that only that input is auto converted...
2. This was basically the same thing as the total area inputs. `defaultValue` prop is not easy to work with when react hook form is used in its parent component...
  I couldn't use `setValue` for the flow rate input because the value is set after useMemo (which determines default value and unit) in the useUnit hook is called for some reason, so I tried to resolve the issue by sending correct default values. It would be great if you could re-test, thank you!!

kathyavini:
@SayakaOno I also (clearly) can't wait to stick the `autoConversion` on everything! It's so good!!

Lovely fix with the useMemo. Works perfectly for me now.

kathyavini (approved):

Duncan-Brain (approved):
@SayakaOno 

Seems to still work for me, thanks for getting rid of most of those extra conversions ! haha

Is anyone else having the first number they put into the calculator disappear? Its weird I type 10 into the input and then its only 0 -- only when there is no stored value.

SayakaOno:
@Duncan-Brain 
😭 Reproduction rate is 100%! and I have spent some time looking into it, but I couldn't figured it out.
Since it's happening in beta, I will create a bug ticket.
Thank you for reviewing!

PR #2575, opened 2023-04-12
LF-3113  Fixing issues with the field work complete endpoint 

kikemarquez8:
**Description**

Fixed an issue where the complete endpoint for "field_work" tasks was attempting to create a new type of field work when changing the field work type upon completing the task.

When you selected a value on the "Type Of Field Work" dropdown it appended an object property "field_work_task_type" to the payload, which is used by the create task endpoint if you are attempting to create a new (custom) type of field work. 

Since the user is not attempting to create a new type of field work the appened data is incomplete as one would expect.
The frontend however does have to inform somehow that the user is attempting to **change** the field work type. I achieved this by sending only the work_type_id. 

I modified the backend to ignore the current functionality if the sent attribute for field_work_task_type is a number. And to continue using the current funcionality when creating. 

Jira link:

**Type of change**

**How Has This Been Tested?**
Steps on [LF-3113](https://lite-farm.atlassian.net/browse/LF-3113) should now pass. 
- [X] Passes test case
- [ ] Other (please explain)

**Checklist:**

[LF-3113]: https://lite-farm.atlassian.net/browse/LF-3113?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/api/src/controllers/taskController.js
+++ b/packages/api/src/controllers/taskController.js
@@ -361,7 +361,7 @@ const taskController = {
       const row = {
         farm_id,
         field_work_name: field_work_task_type.field_work_name,
-        field_work_type_translation_key: field_work_task_type.field_work_type_translation_key,
+        field_work_type_translation_key: field_work_task_type.field_work_name.toUpperCase().trim(),
         created_by_user_id: data.owner_user_id,
         updated_by_user_id: data.owner_user_id,
       };
--- a/packages/api/src/controllers/taskController.js
+++ b/packages/api/src/controllers/taskController.js
@@ -356,7 +356,7 @@ const taskController = {
       data.field_work_task,
       'field_work_task_type',
     );
-    if (containsFieldWorkTask) {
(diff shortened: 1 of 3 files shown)


Duncan-Brain:
Hey @kikemarquez8 thanks for this!

So there is a selection "other" where the user is actually able to create a new type of field work. 

This fix seems to be working for all but the "other" when updating/switching to other on task completion.

SayakaOno (commented):
Thank you for cleaning my code!

There is `completeTaskSaga` in `/packages/webapp/src/containers/Task/saga.js` where a task sent to an API is formatted. It's weird that `getCompleteFieldWorkTaskBody` is used in `createTaskSaga`, but I wonder formatting should be done on the client side...

kikemarquez8:
I think the latest changes should've solved the other type, the solution was very hacky though, I think it's worth noticing we need to refactor this to be consistent accross task types.

PR #2536, opened 2023-03-31
LF-2776 New completion date component for task completion flow

mickenorlen:
**Description**

Jira link: https://lite-farm.atlassian.net/browse/LF-2776

**Type of change**
- [x] New feature (non-breaking change which adds functionality)

**How Has This Been Tested?**

- [ ] Passes test case
- [x] UI components visually reviewed on desktop view
- [x] UI components visually reviewed on mobile view
- [x] Other (please explain)

1. Create a task and select it
2. Mark as Completed and step forward to the new Completed Date
3. See Jira for how new completed date component should work

**OBS**

I had to add some additional fixes for edge cases.

1. `Due date >= Today`: Disable "Due date" and use "Today" as default date choice as the completion date cannot be in future and if `Due date == Today` there isn't any point to have both active.
2. When completing an irrigation task that targets a wild crop. There is no location available so completion logic fails. Separate fix PR here: https://github.com/LiteFarmOrg/LiteFarm/pull/2535

Also see this [discussion](https://github.com/LiteFarmOrg/LiteFarm/pull/2525#issuecomment-1493881136) about the structure of the Pill components

**Checklist:**

- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] The precommit and linting ran successfully
- [x] I have added or updated language tags for text that's part of the UI
- [ ] I have added "MISSING" for all new language tags to languages I don't speak
- Wasn't needed as the only one I updated already had MISSING values present
- [ ] I have added [the GNU General Public License](https://lite-farm.atlassian.net/l/cp/BT0Dd7WW) to all new files
- This isn't present in any of the files I'm working in. Should I still add it to eg: `Pill/index.jsx`?

PR #2578, opened 2023-04-12
LF-2776: New completion date component for task completion flow resolve conflict

SayakaOno:
**Description**

This PR is to complete @mickenorlen's PR: https://github.com/LiteFarmOrg/LiteFarm/pull/2536
I created the branch from his branch, merged integration branch and resolved conflicts.

Jira link: https://lite-farm.atlassian.net/browse/LF-2776


Pazzolo (dismissed):

packages/webapp/src/components/Task/TaskComplete/index.jsx
```diff
@@ -26,23 +27,38 @@ export default function PureTaskComplete({
...
-  const COMPLETE_DATE = 'complete_date';
+  const DATE_CHOICE = 'date_choice';
+  const ANOTHER_DATE = 'date_another';
 
   const { t } = useTranslation();
 
+  // Prepare dates
+  const date_due = getDateInputFormat(persistedFormData.due_date);
+  const date_today = getDateInputFormat();
+  const dueDateDisabled = date_due >= date_today;
```
Duncan-Brain, line 38: @SayakaOno This is disabling the ability to select the first option which I don't think is desired.

packages/webapp/public/locales/en/translation.json
```diff
@@ -1670,7 +1670,7 @@
       "WHICH_DATE": "Which date?",
       "DATE_ORIGINAL": "The original due date",
       "DATE_TODAY": "Today",
-      "DATE_ANOTHER": "Another",
+      "DATE_ANOTHER": "Another date",
```
Duncan-Brain, line 1673: The other translations should be reset to MISSING

packages/webapp/src/components/Form/Radio/index.jsx
```diff
@@ -18,6 +19,9 @@ const Radio = ({
   ...props
 }) => {
   const name = hookFormRegister?.name ?? props?.name;
+  const pill = props?.pill;
+  const checked = props?.checked;
```
Duncan-Brain, line 23: same I think we could add this to explicit props?

Duncan-Brain:
@SayakaOno and @mickenorlen 

Looks good to me, only dealbreaker is the original due date logic -- I don't think we want any:
Screenshot 2023-04-12 at 7 04 39 PM

Also the translations: any updates to english means we need to invalidate the others by replacing with MISSING.

SayakaOno:
@Duncan-Brain Thank you for the comments!
1. another date default value - removed
2. translations - fixed

- Radio component - please check if I understand what you meant correctly
- Several changes were made (including making the radio options' "value" static) to enable all options
- The request for this task was to make the behaviour consistent with `AbandonTask`, so I removed validations although they were good.

packages/webapp/src/components/Task/TaskComplete/index.jsx
SayakaOno: Removed!

packages/webapp/src/components/Form/Radio/index.jsx
```diff
@@ -18,6 +19,9 @@ const Radio = ({
   ...props
 }) => {
   const name = hookFormRegister?.name ?? props?.name;
+  const pill = props?.pill;
```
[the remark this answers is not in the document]
SayakaOno: Now it's destructured.

packages/webapp/src/components/Form/Radio/index.jsx
SayakaOno: Now it's destructured too!

packages/webapp/src/components/Task/TaskComplete/index.jsx
Duncan-Brain: @SayakaOno 

I think I was wrong to suggest removing this even though abandon task does it... If you think it is correct I would leave it in.

Duncan-Brain:
@SayakaOno @mickenorlen 

I think I was wrong to suggest removing the "original due date" logic. Even though it is meant to match the abandon task I think the validation is good. I was thrown off by the unselectable radio button -- I think the best solution would be to not show the option at all.

packages/webapp/src/components/Task/TaskComplete/index.jsx
SayakaOno: I don't think the option is disabled in AbandonTask. Could you double check?
`/packages/webapp/src/components/Task/AbandonTask/index.jsx`

SayakaOno:
I added validation to both `TaskComplete` and `AbandonTask`.

packages/webapp/src/components/Task/TaskComplete/index.jsx
Duncan-Brain: @SayakaOno -- no it doesn't disable it you are right. But it seems like they both _should_ prevent abandoning or completing on a future date. Even though it is not currently in place it seems like that is desired functionality to disable(or hide) the option if the original due date is at some point in the future.

SayakaOno:
We decided to disable the original due date option when it is a future date.

Pazzolo (approved):

Duncan-Brain (approved):
Thanks for this and approved. 

Possible future question: If todays date = original due date maybe we should not disable 😂 ? Not necessary for this pr tho.

SayakaOno:
@Duncan-Brain Thank you Duncan! That's what I was wondering! Users can use the second option for now...😂

PR #2583, opened 2023-04-13
LF-3122 Cannot login with institutional domain hosted on Google via SSO and don't receive email

kathyavini:
**Description**

Farm invitations are re-sent every time someone tries to log into an invited (but not yet created) account without accepting the invitation email. Currently, these emails cap out at three, at which point no more emails are sent. If all three emails had been sent sometime in the past, this can cause issues like the one documented in this Jira ticket because the error message will reference an invite email but no new emails will have been sent.

This PR adds a 24-hour reset to the sent email count, so that after 24 hours three more farm invitations can be sent. This mirrors the behaviour of the password reset email, which also sends up to three times and resets after 24 hours.

To implement this functionality the `emailToken` model also had to be fixed so that it correctly tracks `updated_at` (previously `updated_at` was always the same value as `created_at`). This PR also updates the error message to something more helpful.

**To test this ticket:**
1. Send yourself an invitation to an email that has not yet been associated with a LiteFarm account (or that has never accepted an invitation and so has user status of 2 "Invited")
2. Try to log in to the main login page with this email. You should see the updated error message "A farm invitation has been emailed to you, please check your inbox and spam folder."
3. Do this several more times and see that only three emails are sent. Also confirm that the `times_sent` value in the `emailToken` table is at `3`.
4. Reset the `updated_at` value in the `emailToken` table to a date-time at least 24 hours in the past
5.  Attempt to log in again. Check that a new email is sent and the `times_sent` value has been reset to `1`

Jira link: https://lite-farm.atlassian.net/browse/LF-3122

**Type of change**

**How Has This Been Te

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/api/src/models/emailTokenModel.js
+++ b/packages/api/src/models/emailTokenModel.js
@@ -46,9 +46,21 @@ class emailTokenModel extends Model {
         farm_id: userFarm.farm_id,
       })
       .first();
+
+    // Reset times_sent if the last invitation was sent more than 24 hours ago
+    let hasBeenReset;
+    const now = new Date();
+    const diffDays = Math.abs(now - emailSent?.updated_at) / (1000 * 60 * 60 * 24);
+    if (diffDays > 1) {
+      emailSent.times_sent = 0;
+      hasBeenReset = true;
+    }
+
     if (!emailSent || emailSent.times_sent < 3) {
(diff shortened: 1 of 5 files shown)


Duncan-Brain (approved):
This looks good to me

PR #2594, opened 2023-04-14
Lf 3091 the user is able to create a duplicate of the crop that already exists in the farm if he lefts fields genus species blanc

Duncan-Brain:
**Description**
User is able to add duplicate crops when no genus or species is added.

The database stores crops as unique values for (common name, species, genus, farm id). Because species and genus are optional we regularly send null values. 

Since null != null (Ginger, null, null, this_is_my_farm id) != (Ginger, null, null, this_is_my_farm id)

By requiring a value and using form validation to set the value as a single space ' ' we fix this condition and make it harder to add bad data. (Ginger, ' ', ' ', this_is_my_farm id) == (Ginger, ' ', ' ', this_is_my_farm id)

An alternate solution could be:
Exploring setting the defaultTo('') in the database

Jira link: [LF-3091](https://lite-farm.atlassian.net/browse/LF-3091)

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain):
- tested with empty values for each and entered values for each.

**Checklist:**

[LF-3091]: https://lite-farm.atlassian.net/browse/LF-3091?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/api/src/models/cropModel.js
+++ b/packages/api/src/models/cropModel.js
@@ -81,8 +81,8 @@ class Crop extends BaseModel {
   static get jsonSchema() {
     return {
       type: 'object',
-      required: ['crop_common_name', 'farm_id'],
-
+      // genus and species are required for database uniqueness
+      required: ['crop_common_name', 'farm_id', 'crop_genus', 'crop_specie'],
       properties: {
         crop_id: { type: 'integer' },
         farm_id: { type: 'string' },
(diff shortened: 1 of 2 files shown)


Pazzolo (approved):

kathyavini:
It _does_ work -- definitely can't create a duplicate crop now, but I'm really hesitant to start storing spaces in the database when we mean a null value -- that sounds like it could get very messy later! If it's done at the level of the controller, wouldn't a query comparison for `null` be ok?

Duncan-Brain:
@kathyavini As on the call.. Agreed this solution isn't great. I am onboard with the query on backend and new selector on frontend. I still think it would be good to not have a unique constraint with optional values -- but that is a larger fix.

PR #2611, opened 2023-04-15
LF-3092 The user is able to create a duplicate variety of the same crop

kathyavini:
**Description**

This PR adds a uniqueness constraint to adding a variety in the frontend, so that the Add Variety form cannot be submitted if the given Variety Common Name already exists for that given crop on this farm.

The same constraint has also been added to the API for both the Add Variety and Edit Variety screens. _The frontend validation for Edit Variety should be added at some point in the future._

**This ticket has a language component for the validation error displayed when a Variety is not unique.**

Jira link: https://lite-farm.atlassian.net/browse/LF-3092

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/cropVarietyController.js
+++ b/packages/api/src/controllers/cropVarietyController.js
@@ -72,7 +72,7 @@ const cropVarietyController = {
         const [relatedCrop] = await CropModel.query()
           .context({ showHidden: true })
           .where({ crop_id });
-        if ((farm_id, crop_id, crop_variety_name)) {
+        if (farm_id && crop_id && crop_variety_name) {
           const duplicateVariety = await CropVarietyModel.query().findOne({
             farm_id,
             crop_id,
@@ -98,7 +98,7 @@ const cropVarietyController = {
       const { crop_variety_id } = req.params;
       const { farm_id, crop_id, crop_variety_name } = req.body;
       try {
-        if ((farm_id, crop_id, crop_variety_name, crop_variety_id)) {
+        if (farm_id && crop_id && crop_variety_name && crop_variety_id) {
           const duplicateVariety = await CropVarietyModel.query()
             .whereNot('crop_variety_id', crop_variety_id)
             .findOne({
--- a/packages/api/src/controllers/cropVarietyController.js
+++ b/packages/api/src/controllers/cropVarietyController.js
@@ -72,16 +72,18 @@ const cropVarietyController = {
         const [relatedCrop] = await CropModel.query()
           .context({ showHidden: true })
           .where({ crop_id });
-        const duplicateVariety = await CropVarietyModel.query().findOne({
-          farm_id,
-          crop_id,
-          crop_variety_name,
-          deleted: false,
-        });
-        if (duplicateVariety) {
-          return res.status(400).json({
-            error: 'This crop variety already exists, please choose a different variety name',
+        if ((farm_id, crop_id, crop_variety_name)) {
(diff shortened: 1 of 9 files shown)


Duncan-Brain:
@kathyavini  FYI will want to add the same to EditCropVariety as well. Crop doesnt have edit capability just variety

Working on backend now.

Pazzolo (approved):

packages/webapp/src/components/AddCropVariety/index.jsx
```diff
@@ -73,14 +75,41 @@ export default function PureAddCropVariety({
     truncateText(crop.crop_genus, 22) + ' ' + truncateText(crop.crop_specie, 22);
   const progress = 33;
 
+  const uniqueVarietyName = () => {
+    const formVarietyName = getValues('crop_variety_name');
```
SayakaOno, line 79: `crop_variety_name` could be replaced with `COMMON_NAME`, but I will approve after doing some more testing!

kathyavini: Ah yes, it definitely could. I used the hook form registered name in the Add Crop version of this component as no naming variables had been set up in that file. I would slightly prefer to keep the parallelism, but either way!

SayakaOno (commented):
I know it's not ready for a review yet, but I will just bring this to attention.
I guess it's out of scope, but the same check needs to be added for EditCropVariety on the frontend. Only the API will be updated.

kathyavini:
@SayakaOno Yes, this was intentional as I was having trouble getting the same code to work in the EditVariety component! I was going to open placeholder ticket for adding frontend validation at some later point.

However, a big difference between this screen and the other (and why I'm not so concerned about getting that in right away) is that the API call happens on the same page. Therefore the user is altered at the exact same place in the flow, it's just that the error message appears in a snackbar rather than in an inline validation message.

Regarding review readiness, the functionality should be ready to review -- it's only the changes to the API tests that still have to be added!

packages/api/src/controllers/cropVarietyController.js
```diff
@@ -68,10 +68,23 @@ const cropVarietyController = {
   createCropVariety() {
     return async (req, res, next) => {
       try {
-        const { crop_id } = req.body;
+        const { crop_id, crop_variety_name, farm_id } = req.body;
         const [relatedCrop] = await CropModel.query()
           .context({ showHidden: true })
           .where({ crop_id });
+        if ((farm_id, crop_id, crop_variety_name)) {
```
SayakaOno, line 75: @Duncan-Brain `||` ? `&&`?

Duncan-Brain: @SayakaOno @kathyavini 

yah I messed up my eyes arent working and im rushing

kathyavini: Obviously mine too because I didn't see that when I reviewed for merge!

Thank you @SayakaOno!!

Duncan-Brain: lol that would have broke my half of it -- @SayakaOno @kathyavini please check again

SayakaOno: LOL!!! ❤️

SayakaOno (approved):
Approved!

Pazzolo (approved):

PR #2615, opened 2023-04-16
sending welcome email only if second and third steps are not completed.

kikemarquez8:
**Description**

sending welcome email only if second and third steps are not completed.

Jira link:

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/userFarmController.js
+++ b/packages/api/src/controllers/userFarmController.js
@@ -210,8 +210,8 @@ const userFarmController = {
           .where({ user_id, farm_id })
           .patch({ has_consent, consent_version });
         res.sendStatus(200);
-        const { step_two_end, step_three_end, step_five_end } = userFarm;
-        const isWelcomeEmailSent = !!step_two_end && !!step_three_end && !step_five_end;
+        const { step_two_end, step_three_end } = userFarm;
+        const isWelcomeEmailSent = !!step_two_end && !!step_three_end;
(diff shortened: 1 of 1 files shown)


packages/api/src/controllers/userFarmController.js
```diff
@@ -210,8 +210,8 @@ const userFarmController = {
           .where({ user_id, farm_id })
           .patch({ has_consent, consent_version });
         res.sendStatus(200);
-        const { step_two_end, step_three_end, step_five_end } = userFarm;
-        const isWelcomeEmailSent = !!step_two_end && !!step_three_end && !step_five_end;
+        const { step_two_end, step_three_end } = userFarm;
+        const isWelcomeEmailSent = !!step_two_end && !!step_three_end;
```
Pazzolo, line 214: shouldnt it be an "or" operator? `!!step_two_end || !!step_three_end;`

kikemarquez8: I think not, this endpoint triggers in the onboarding flow after you select a role (which is step 2) right at the point you accept the policy  step 2 is complete but step 3 isn't. So the email should be sent exactly then. 

When this condition is met again (when you accept consent because we introduced a new policy) you will have both steps finished.

Pazzolo (approved):

PR #2633, opened 2023-04-17
fix the order of hooks issue

javiscript13:
**Description**

A hook was modified in a way that the component didn't render the same way each time, which is a hook rule and it was being broken.

It caused issues on delete tasks.

Jira link:

**Type of change**

**How Has This Been Tested?**

- [x] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/containers/Task/TaskReadOnly/useReadonlyTask.js
+++ b/packages/webapp/src/containers/Task/TaskReadOnly/useReadonlyTask.js
@@ -29,10 +29,10 @@ const getManagementPlanTile = (managementPlan, tasksByManagementPlanId) =>
 
 export const useReadonlyTask = (task_id) => {
   const task = useSelector(taskWithProductSelector(task_id));
-  if (task === undefined) return undefined;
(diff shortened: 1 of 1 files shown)


packages/webapp/src/containers/Task/TaskReadOnly/useReadonlyTask.js
```diff
@@ -29,10 +29,10 @@ const getManagementPlanTile = (managementPlan, tasksByManagementPlanId) =>
 
 export const useReadonlyTask = (task_id) => {
   const task = useSelector(taskWithProductSelector(task_id));
-  if (task === undefined) return undefined;
   const tasksByManagementPlanId = useSelector(taskEntitiesByManagementPlanIdSelector);
 
   const getTransplantTask = (task) => {
+    if (task === undefined) return undefined;
```
kathyavini, line 35: I don't think it's possible to reach here with task `undefined` because of your optional chaining and the ternary expression -- right?

I think maybe I _do_ prefer putting `if (task === undefined) return undefined;` right into the first line of the useMemo 😆 It is a bit clearer!

SayakaOno (approved):

PR #2634, opened 2023-04-17
Add check to API to prevent crash on partially deleted tasks

kathyavini:
**Description**

This PR adds a check for an undefined task in assignTask.js that is capable of crashing the API with a TypeError.

A React Hooks-related error in the frontend was making it easier to create partially deleted tasks that were deleted in the database but not the Redux store. Reassigning these tasks would cause a TypeError that crashes the API.

With the React Hooks error addressed (see PR #2633) these partially-deleted tasks will be much harder to generate. But there is no downside to including this check 🙂 

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

I crashed the API on my local reliably and verified that this file was at fault.

**Checklist:**


SayakaOno (approved):
Looks good!

PR #2636, opened 2023-04-24
LF-3218 An error is displayed: "LOGIN.ERROR.LOGIN_FAIL" once the invited user tries to accept the invitation using the Google SSO method

kathyavini:
**Description**

When an invite is accepted using GoogleSSO, a new `user_id` is generated at the point of account creation in the format specific to GoogleSSO accounts. (For a password-based account, the `user_id` generated at the moment of invite is retained).

If tasks are assigned to an invited user before they have accepted their invitation, they will not be able to create their account using Google SSO because their original `user_id` will exist in the `task` and `notifcation_user` tables and the database will give a Foreign Key Violation error when the original record is deleted in the Objection.js transaction.

This PR fixes this issue by updating the `user_id` referenced in the `tasks` and  `notification_user` tables as is already done for several other tables within the transaction.

Jira link: https://lite-farm.atlassian.net/browse/LF-3218

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/userController.js
+++ b/packages/api/src/controllers/userController.js
@@ -20,6 +20,8 @@ import UserFarmModel from '../models/userFarmModel.js';
 import PasswordModel from '../models/passwordModel.js';
 import EmailTokenModel from '../models/emailTokenModel.js';
 import ShiftModel from '../models/shiftModel.js';
+import TaskModel from '../models/taskModel.js';
+import NotificationUserModel from '../models/notificationUserModel.js';
 import FarmModel from '../models/farmModel.js';
 import { transaction, Model } from 'objection';
 import bcrypt from 'bcryptjs';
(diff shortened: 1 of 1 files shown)


Duncan-Brain (approved):
@kathyavini 

This solves the issue for me nice fix!

One question: I noticed you also added more than just the Task model. Did you look into any other tables we should be doing this for? Would it be fair to say this can only happen to records where there is an assignee?

kathyavini:
@Duncan-Brain Correct; this should only occur when there is an assignee. At the moment, I think all the affected tables are updated.

It's probably a good thing that this transaction fails with a nice clear error though, in case we ever end up adding more functionality that can be performed on invited (but not yet active) users.

SayakaOno (approved):
I didn't know how it was working! Great job!!

PR #2640, opened 2023-04-25
LF-3187: No results are shown when the user searches for a crop using its Scientific Name when the Scientific name Consists of more than 1 word

SayakaOno:
**Description**

Scientific name consists of "crop_genus"+ "crop_specie". This PR is to update `useStringFilteredCrops` to allow crops to be searched by scientific name.

Jira link: https://lite-farm.atlassian.net/browse/LF-3187

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/containers/CropCatalogue/useStringFilteredCrops.ts
+++ b/packages/webapp/src/containers/CropCatalogue/useStringFilteredCrops.ts
@@ -49,8 +49,17 @@ export default function useStringFilteredCrops(
         'crop_cultivar', // crop_variety.crop_cultivar
       ];
       for (let i = 0; i < keys.length; i++) {
-        if (keys[i] in crop && check(crop[keys[i] as keyof typeof crop])) {
-          return true;
+        if (keys[i] in crop) {
+          if (!i) {
+            const scientificName =
+              crop[keys[0] as keyof typeof crop] + (crop[keys[1] as keyof typeof crop] || '');
+            if (scientificName && check(scientificName)) {
+              return true;
+            }
+            i++;
+          } else if (check(crop[keys[i] as keyof typeof crop])) {
+            return true;
+          }
         }
       }
       return false;


kathyavini (dismissed):
It works well!!

I kind of wish there was a way to write it without the type redundancies, but after puzzling at it for a bit, I think your way is the best 🙂

packages/webapp/src/containers/CropCatalogue/useStringFilteredCrops.ts
```diff
@@ -49,8 +49,17 @@ export default function useStringFilteredCrops(
         'crop_cultivar', // crop_variety.crop_cultivar
       ];
       for (let i = 0; i < keys.length; i++) {
-        if (keys[i] in crop && check(crop[keys[i] as keyof typeof crop])) {
-          return true;
+        if (keys[i] in crop) {
+          if (!i) {
```
Duncan-Brain, line 53: @SayakaOno This works but I am not sure I fully understand.

Is this when i=1 or when keys[i] == 'crop_specie'?

packages/webapp/src/containers/CropCatalogue/useStringFilteredCrops.ts
```diff
@@ -49,8 +49,17 @@ export default function useStringFilteredCrops(
...
+            const scientificName =
+              crop[keys[0] as keyof typeof crop] + (crop[keys[1] as keyof typeof crop] || '');
+            if (scientificName && check(scientificName)) {
+              return true;
+            }
+            i++;
```
Duncan-Brain, line 59: Is this an efficiency to avoid an extra check during the loop since we know the next key is 'crop_variety_name'?

I am not sure I have seen for loop indexes bypassed this way before

Duncan-Brain:
@SayakaOno this works for me, but I am not sure I understand some of the logic. I have added a few inline questions.

SayakaOno:
@kathyavini I tried to find a better way, but I couldn't either... I hope to improve it in the future!! Thank you for reviewing :)

packages/webapp/src/containers/CropCatalogue/useStringFilteredCrops.ts
Duncan-Brain: I prefer the use of the explicit if (keys[i] === 'crop_genus') since there is no way to misunderstand that. Right now there is a chance some developer will come along and want to search by "crop_notes" or something and then add that in as the first element in the keys array.

packages/webapp/src/containers/CropCatalogue/useStringFilteredCrops.ts
Duncan-Brain: Thanks yah. The search function overall could be improved, we could probably use a library for this even to tokenize and allow for cross searching like "apple malus" working for apple. -- Future us problem haha

packages/webapp/src/containers/CropCatalogue/useStringFilteredCrops.ts
SayakaOno: Thanks to your input, it should be more readable now. I appreciate it!!

packages/webapp/src/containers/CropCatalogue/useStringFilteredCrops.ts
SayakaOno: Yeah, it could be an independent function at least. Tech debt candidate :)

packages/webapp/src/containers/CropCatalogue/useStringFilteredCrops.ts
Duncan-Brain: ooh thats a pretty solution!

Duncan-Brain (approved):
very succint!

PR #2650, opened 2023-04-27
LF-3157b Reset file picker after filesize warning or upload

kathyavini:
**Description**

When ticket [LF-3157](https://lite-farm.atlassian.net/browse/LF-3157) was in verification, QA noticed that the filesize warning modal doesn't fire a second time if the exact same file is selected again. This is because the default behaviour of the HTML input element (type file) is to not fire an `onChange` if the user selects the same file twice in a row because the value of the input hasn't changed.

In our case, we would like the file picker `onChange` to fire even when the same file is selected twice, so this PR adds a value reset after both the filesize warning modal and a successful upload with "Add more pages." The file picker is an uncontrolled input not registered with React Hook Form so it is safe to do this 🙂 

Jira link: https://lite-farm.atlassian.net/browse/LF-3157

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

If you have MinIO working but not the Imaginary token, you can test this using "Add more pages" on a .csv or .txt file. The filesize warning should re-fire regardless of whether you have the token.

**Checklist:**


SayakaOno (approved):
Looks good!

Duncan-Brain (approved):

PR #2654, opened 2023-04-28
LF-3186 filter pill is wonky

Duncan-Brain:
**Description**

Well this is what happens when I try to take a quick ticket... This edits the vite configuration to allow for a proper compile of emotion css through @emotion/react. If you don't know how react-select works this is how I landed on this solution:

- I tried to find the css that was out of place in ReactSelect under multiValueRemove -- everything looked correct and I tried fidgeting around with it with no success
- I found that react-selects native MultiValueRemove was being overridden to add in a different cross (x) than the provided one and so maybe the css wasn't being passed down
- css was being passed down but not to styles and class element attributes instead to a css element attribute
- I tried passing the css attribute to element style attribute -- it mostly worked but not the hover modifiers
- I tried the @emotion/react css interpreter but got an error on the element text itself (weird)
- I tried adding a new package @emotion/css css interpreter and it works!
- But why doesn't the @emotion/react css interpreter work? I don't want to add new packages.
- Don't need to edit anything on React-select component, just edit the compiler

Answer found here: https://dev.to/mehdinajafi/how-to-use-emotion-css-prop-in-vite-39pj

This was only a problem for the components overriding the react-select base components. The only other component that did so was clearIndicator and it had no custom css. So no other changes should be visible.

Jira link: https://lite-farm.atlassian.net/browse/LF-3186

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain): 

**Checklist:**

### DIFF
--- a/packages/webapp/vite.config.ts
+++ b/packages/webapp/vite.config.ts
@@ -9,7 +9,12 @@ import { VitePWA } from 'vite-plugin-pwa';
 export default defineConfig({
   plugins: [
     { enforce: 'pre', ...mdx({ providerImportSource: '@mdx-js/react' }) },
-    react(),
+    react({
+      jsxImportSource: '@emotion/react',
+      babel: {
+        plugins: ['@emotion/babel-plugin'],
+      },
+    }),
     svgrPlugin({
       svgrOptions: {
         icon: false,


SayakaOno:
> this is what happens when I try to take a quick ticket

🤣🤣🤣 I'm glad you took care of it!

So, were they not pills but styled using CSS!?
I wonder if we should update `tsconfig.json` and `vite-env.d.ts` for the future as it's shown in the [article](https://dev.to/mehdinajafi/how-to-use-emotion-css-prop-in-vite-39pj) you shared...

Duncan-Brain:
@SayakaOno Yes, using css styles on the react select core components. I think the overridden component was an attempt to make the cross(x) thinner looking like the other pills.

I thought about updating the ts config but this file isn't in .ts. So I thought I might leave it -- if at some point we convert to Typescript I thought there might be many other configurations needed as well. So I couldn't really verify it works until it is converted to .ts.

SayakaOno:
@Duncan-Brain Ah! I understand.
I hope anyone does not need to go through everything you did to find the solution for TypeScript 😄
Anyways, nice fix 👏

SayakaOno (approved):

kathyavini (approved):
Oh wow, nice sleuthing!!

PR #2656, opened 2023-05-01
LF-2980 Add "loading" image when uploading a custom varietal image

kathyavini:
**Description**

This PR adds a loading spinner to image selection and replacement in AddCrop, AddCropVariety, and EditCropVariety. Logic to manage the loading spinner state has been added to ImagePickerWrapper, the image uploader component that all three of these views use.

Due to differences in page layout, only EditCropVariety also needed a wrapper div to contain the loading spinner.

Note: To test the full behaviour successfully, you will need the LiteFarm imaginary token (for thumbnail creation) + MinIO.

Jira link: https://lite-farm.atlassian.net/browse/LF-2980

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/EditCropVariety/index.jsx
+++ b/packages/webapp/src/components/EditCropVariety/index.jsx
@@ -1,5 +1,5 @@
 import Button from '../Form/Button';
-import React from 'react';
+import React, { useState } from 'react';
 import PropTypes from 'prop-types';
 import { useTranslation } from 'react-i18next';
 import { Label } from '../Typography';
@@ -12,6 +12,7 @@ import PageTitle from '../PageTitle/v2';
 import RadioGroup from '../Form/RadioGroup';
 import Infoi from '../Tooltip/Infoi';
 import Leaf from '../../assets/images/farmMapFilter/Leaf.svg';
+import Spinner from '../Spinner';
 
 export default function PureEditCropVariety({
   onSubmit,
@@ -85,6 +86,11 @@ export default function PureEditCropVariety({
     ? t(`crop:${cropTranslationKey}`)
     : cropVariety.crop_common_name;
 
+  const [showSpinner, setShowSpinner] = useState(false);
+  const handleSpinnerState = (state) => {
+    setShowSpinner(state);
+  };
+
   return (
     <Form
       buttonGroup={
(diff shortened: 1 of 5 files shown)


javiscript13 (approved):
I think it works perfect and the code is very clean.

If something fails with the bucket the error snackbar displays that there was an error, the app continues to work but the spinner keeps going without returning to the previous or default image.

As the probabilities on the bucket failing are very low, the user is notified and the app keeps working I think there is not need to change anything.

Great work!

kathyavini:
@javiscript13 actually I'm SUPER interested in protecting from that situation as I would have thought that it would have been covered within the error block! Could you tell me how to recreate the bucket error where you saw that?

packages/webapp/src/components/EditCropVariety/index.jsx
```diff
@@ -101,15 +107,22 @@ export default function PureEditCropVariety({
...
-        alt={cropVariety.crop_common_name}
-        className={styles.circleImg}
-        onError={(e) => {
-          e.target.onerror = null;
-          e.target.src = 'crop-images/default.jpg';
-        }}
-      />
+
+      {showSpinner ? (
+        <div style={{ height: 'fit-content' }}>
```
SayakaOno, line 109: This is probably not working as you intended...? Nothing changes when removing `fit-content`(or `div`) for me.

javiscript13: @kathyavini  You just have to shut down the MinIO and then try to upload a file. It will fail on the saga file, and the spinner will persist instead of any other previous image.

javiscript13:
> @javiscript13 actually I'm SUPER interested in protecting from that situation as I would have thought that it would have been covered within the error block! Could you tell me how to recreate the bucket error where you saw that?

@kathyavini  You just have to shut down the MinIO and then try to upload a file. It will fail on the saga file, and the spinner will persist instead of any other previous image.

packages/webapp/src/components/EditCropVariety/index.jsx
kathyavini: So this is so subtle maybe it's ridiculous to have been thinking about it, but if you remove the `<div>`, the Spinner takes up more space than the image, and the rest of the page moves down during loading:

https://user-images.githubusercontent.com/82857964/235545683-cf559c95-94ff-4d8e-8d42-fecbb5c76d31.mov



While with the `<div>` included, the Spinner is smaller than the image so the rest of the page moves up during loading:

https://user-images.githubusercontent.com/82857964/235545697-3c8fdda3-d76d-4914-b30c-79358bed9afa.mov


The only reason I adjusted was because the move-up behaviour was what was seen in the other two views and I wanted to be consistent; although I doubt anyone will notice at all, especially unless comparing them one after another!

kathyavini:
@javiscript13 Thank you, Javier!!! I added error handling to the saga to cover that case, so the spinner shouldn't persist now even on bucket connection issues.

@SayakaOno thanks for your help with the React! I think the weirdness is all cleaned up now with the props, if you could take another look?

SayakaOno (approved):

javiscript13 (approved):
Everything working great!!

packages/webapp/src/components/EditCropVariety/index.jsx
kathyavini: @SayakaOno Oh wow, that freaks me out a little 😬

PR #2658, opened 2023-05-02
LF-3114 When creating a varietal without adding a new image, the image should be inherited from the crop_type

kathyavini:
**Description**

When no custom varietal image is uploaded, the varietal should inherit the image of the parent crop.

Logic to handle this inheritance did exist, but it was in the data-transformation step which is only applied to non-certification farms. Therefore, certification-pursuing farms did not see crop image inheritance.

This PR moves that logic to Add Crop Variety by setting the parent crop image as the default value in the React Hook Form, so image inheritance will be applied regardless of certification status. The image preview was updated to only display if the crop variety image was changed off of this default.

Jira link: https://lite-farm.atlassian.net/browse/LF-3114

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/AddCropVariety/index.jsx
+++ b/packages/webapp/src/components/AddCropVariety/index.jsx
@@ -39,12 +39,12 @@ export default function PureAddCropVariety({
     getValues,
     watch,
     setError,
-    formState: { errors, isValid },
+    formState: { errors, isValid, dirtyFields },
   } = useForm({
     mode: 'onChange',
     shouldUnregister: true,
     defaultValues: {
-      crop_variety_photo_url: null,
+      crop_variety_photo_url: crop.crop_photo_url,
       [LIFE_CYCLE]: crop[LIFE_CYCLE],
       [HS_CODE_ID]: crop?.[HS_CODE_ID],
       ...persistedFormData,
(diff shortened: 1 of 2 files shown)


packages/webapp/src/components/AddCropVariety/index.jsx
```diff
@@ -181,7 +181,8 @@ export default function PureAddCropVariety({
       {showSpinner ? (
         <Spinner />
       ) : (
-        crop_variety_photo_url && (
+        crop_variety_photo_url &&
+        dirtyFields.crop_variety_photo_url && (
```
SayakaOno, line 185: I think we need another solution here.
This is the current behaviour:
https://user-images.githubusercontent.com/33141219/235772448-70231266-ad85-4a6c-8b6c-d9e27af1c22c.mov

With the new code, the image will not be shown when coming back from the next page.

kathyavini: Thank you for catching this!!

My initial thought was to do a straight comparison anyway (`crop_variety_photo != crop.crop_photo`), so I think I'll implement that again. Unless do you think there are other issues with using the default value of the form to hold the image value?

SayakaOno: I think using the default value is a great idea! and the code looks good :)

SayakaOno (approved):

PR #2659, opened 2023-05-02
LF-2975 Revert to react-redux-form style and adopt css suggestion

Duncan-Brain:
**Description**

Hi @bailey-mc Thank you for your work on LF-2975 sorry it is a weird release time and are pushing it through. Normally there would be more time for you to work on it at your own schedule. Looks like we were hoping to get out another release patch today. 

Basically the InputAutoSize works with react-hook-forms so it wasn't working with the model to register values. This component needs to be completely refactored as it uses the old react-redux-forms. Pretty much everything else I kept the same I think.

Jira link:

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/containers/Finances/NewExpense/AddExpense/styles.module.scss
+++ b/packages/webapp/src/containers/Finances/NewExpense/AddExpense/styles.module.scss
@@ -64,6 +64,7 @@
     border-radius: 4px;
     padding: 0 5%;
     line-height: 37px;
+    font-family: 'Open Sans', 'SansSerif', serif;
   }
   display: flex;
   flex-direction: row;
--- a/packages/webapp/src/containers/Finances/NewExpense/AddExpense/styles.module.scss
+++ b/packages/webapp/src/containers/Finances/NewExpense/AddExpense/styles.module.scss
@@ -57,23 +57,13 @@
     letter-spacing: 0.324px;
     color: #4d4d4d;
   }
-  input {
+  input, textarea {
     width: 60%;
-    height: 37px;
     border: 1px solid #efefef;
     box-sizing: border-box;
     border-radius: 4px;
     padding: 0 5%;
-  }
-  .textArea {
-    font-size: 16px;
-    line-height: 24px;
-    font-family: 'Open Sans', 'SansSerif', serif;
-    height: 37px;
-    border: 1px solid #efefef;
-    box-sizing: border-box;
-    border-radius: 4px;
-    padding: 0 5%;
+    line-height: 37px;
   }
   display: flex;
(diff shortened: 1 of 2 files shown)


packages/webapp/src/containers/Finances/NewExpense/AddExpense/styles.module.scss
```diff
@@ -57,23 +57,13 @@
...
-    height: 37px;
-    border: 1px solid #efefef;
-    box-sizing: border-box;
-    border-radius: 4px;
-    padding: 0 5%;
-  }
-  .textArea {
-    font-size: 16px;
-    line-height: 24px;
-    font-family: 'Open Sans', 'SansSerif', serif;
```
SayakaOno, line 71: I think we should keep font-size and font-family. Otherwise...
Screenshot 2023-05-02 at 1 44 46 PM

```
  textarea {
    font-size: 16px;
    font-family: 'Open Sans', 'SansSerif', serif;
  }
```

Duncan-Brain: Isn't this normally set at the top level component instead of individual inputs? and since this needs a refactor to use react-hook-forms will that fix it also?

kathyavini: @Duncan-Brain Yeah font-family is usually inherited, but for a textarea, most browsers apply monospace font 😁 So you will have to add it in manually to reverse that. I don't think the font size is 16px though... if it's going to be kept equivalent to Edit Expense I think the default font-size is ok. My browser is calculating that at 13.33px (??!) but maybe check with yours.

SayakaOno: I agree, they should not be written over and over again. But `InputAutoSize` also needed them, so I thought global styling was not working.
I'm not sure how the refactor will change the input, but I think we could change the font-family for the patch? The font is `Arial` in beta which is probably wrong, but `monospace` looks very different.

SayakaOno (approved):

packages/webapp/src/containers/Finances/NewExpense/AddExpense/styles.module.scss
Duncan-Brain: Font-size: both are inherited

Font-family: done

kathyavini (approved):
I'm all in support for merging this branch, but I think I have some reservations that belong on the main ticket 😄

PR #2667, opened 2023-05-08
LF-3193 Sensors fail to upload for cryptic reasons

kathyavini:
**Description**

Note: This is a change that is only going to be visible until a fix for [LF-3240](https://lite-farm.atlassian.net/browse/LF-3240) is implemented.

If sensors fail to save to the database (something that was happening regularly with bulk sensor upload due to concurrent knex transactions, see ticket LF-3240 above), the same error message is used as when an ensemble sensor fails to be registered. This error message prints the name of the ensemble external ID for the sensor, and hardcodes the problematic column as "External_ID", leading to cryptic and confusing error messages for sensors without external IDs.

This PR fixes that error message to show the sensor name when External ID is not available and removes the "External_ID" column value in the case of database save failures (which has nothing to do with any particular column). Please note that failures to save to the database are not going to happen once the changes in LF-3240 are implemented, but this ticket can be tested and validated until then.

The error message generation function, which creates a string from an array of IDs, was also refactored from a reduce to a join for improved readability.

Jira link: https://lite-farm.atlassian.net/browse/LF-3193

**Type of change**

**How Has This Been Tested?**

Please test this by uploading the CSV file linked in the Jira ticket (or any other bulk upload that passes validation) and verifying that the error message accurately reflects which sensors were and were not uploaded. It is not necessary to change any latitude or longitude values.

Please only attempt to upload once, unless you would also like to reproduce the API crash described in LF-3240 😄

- [ ] Passes test case
- [x] Other (please explain)

**Checklist:**

[LF-3240]: https://lite-farm.atlassian.net/browse/LF-3240?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/api/src/controllers/sensorController.js
+++ b/packages/api/src/controllers/sensorController.js
@@ -221,11 +221,13 @@ const sensorController = {
           if (curr.status === 'fulfilled') {
             prev.push(curr.value);
           } else {
+            // These are sensors that were not saved to the database as locations
             errorSensors.push({
               row: data.findIndex((elem) => elem === registeredSensors[idx]) + 2,
-              column: 'External_ID',
               translation_key: sensorErrors.INTERNAL_ERROR,
-              variables: { sensorId: registeredSensors[idx].external_id },
+              variables: {
+                sensorId: registeredSensors[idx].external_id || registeredSensors[idx].name,
+              },
             });
           }
           return prev;
@@ -249,7 +251,7 @@ const sensorController = {
                   error_download: {
                     errors: errorSensors,
                     file_name: 'sensor-upload-outcomes.txt',
-                    success: successSensors.map((s) => s.sensor?.external_id), // Notification download needs an array of only ESIDs
+                    success: successSensors.map((s) => s.sensor?.external_id || s.name), // Notification download needs an array of only ESIDs
                     error_type: 'claim',
                   },
                 },
--- a/packages/webapp/src/containers/Map/saga.js
+++ b/packages/webapp/src/containers/Map/saga.js
@@ -169,7 +169,7 @@ export function* bulkUploadSensorsInfoFileSaga({ payload: { file } }) {
             }
             yield put(
               bulkSensorsUploadFailure({
-                success: success.map((s) => s?.sensor?.external_id),
+                success: success.map((s) => s?.sensor?.external_id || s?.name),
                 errorSensors,
               }),
             );
--- a/packages/webapp/src/util/sensor.js
+++ b/packages/webapp/src/util/sensor.js
@@ -42,11 +42,8 @@ export const generateClaimSensorErrorFile = (errors, success) => {
   let errorText = '';
   if (success.length > 0) {
     errorText += i18n.t('FARM_MAP.BULK_UPLOAD_SENSORS.DOWNLOAD_FILE.PARTIAL_SUCCESS_TOP_TEXT');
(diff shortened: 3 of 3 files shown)


SayakaOno (approved):
Worked for me :)

packages/api/src/controllers/sensorController.js
```diff
@@ -221,11 +221,13 @@ const sensorController = {
           if (curr.status === 'fulfilled') {
             prev.push(curr.value);
           } else {
+            // These are sensors that were not saved to the database as locations
             errorSensors.push({
               row: data.findIndex((elem) => elem === registeredSensors[idx]) + 2,
-              column: 'External_ID',
```
Duncan-Brain, line 226: @kathyavini Did you mean to remove this?

It is now leaving the Column part blank in the errors text:

Screenshot 2023-05-10 at 4 37 10 PM

I am guessing it was removed because it is not the core problem -- but it is probably good to keep the data point if it will be because of External ID once 3142 is fixed?.

Duncan-Brain:
@kathyavini 

It seems to work for me. Just mentioned that column: 'External_id' field inline.

Is this worth fixing then if 3240 fixes it? (I would still want the join succinctness of course wither way 😁 )

packages/api/src/controllers/sensorController.js
kathyavini: Yup! Super intentional (see comments in the Jira ticket).

The `external_id` is not at fault if the record fails to save to the database -- as far as I can gather this code block was just copied verbatim from further up in the controller where the errors were related to sensor registration; it's not like that here and I didn't change how the error is generated if there is an actual registration issue. We could adjust the error text template to delete the Column part when one isn't present, but that's going to touch on the translation files. Given that we're not going to see this very much in the future, I strongly feel it's not worth the effort. The blank column indicates that no column is at fault, which is correct 🙂

kathyavini:
@Duncan-Brain I replied inline re: the Column part.

As to your larger question, YES, I absolutely think it's worth fixing even though we don't expect to see database errors once the other PR is merged! Why leave a misleading error message in the code if we know about it, even if we don't expect to see that error in the future? And who knows, we may very well do something to break knex or this database table again! And then whoever sees this is going to be confused by the cryptic errors... again! 😆

packages/api/src/controllers/sensorController.js
Duncan-Brain: To your other point people may continue being confused if we do see it in the future. I think a language translation or removal of the column field should not be off the table. 

Probably we have something like "not specified" or "unknown" already translated.

Duncan-Brain:
@kathyavini 

I get now.. it is an error that could occur either way. Of course we should always fix errors. It sounded like it was "not going to be visible" after 3240. 

So now my question is if external_id is not the problem why `s.sensor?.external_id || s.name` and not just `s.name` which is required and always present?

packages/api/src/controllers/sensorController.js
kathyavini: I asked Kevin to be the final judge on this one, and he has cast his vote for blank column!

Duncan-Brain: Ok cool made a note on the jira ticket about our discussion. And will approve as is.

PR #2668, opened 2023-05-08
LF-3240 The uploading the sensors group causes the crash of the beta

kathyavini:
**Description**

As demo-ed in standup this morning, May 8, 2023... but there is an update! 🙂

The transaction that saves newly created sensors to the database is called in the controller using `promise.allSettled()`, meaning all the transactions are initiated concurrently. In theory, as many concurrent transactions can be created as the size of the [connection pool](https://knexjs.org/guide/#pool) defined in the knex configuration, but in our case even relatively low numbers (below our defined pool size) caused knex timeouts and even an API crash (sometimes) if done multiple times in a row.

Update: actually, a closer read of our `knexfile.js` shows that the connection pool was only set on the testing databases and for migration, meaning that for all other environments, we are using the default value of 10. Therefore, there are a few possible approaches to fixing this error:

1. Add the configuration `pool: { min: 0, max: 101 }` to the development, integration, and production environments in the knexfile (100 is the max number of sensors that can be uploaded at once and +1 for the API connection)
2. Limit the number of sensors that can be uploaded at once to 9 (again +1 for the API connection)
3. What I recommend and have implemented: remove the concurrency

The reason I recommend (3) is that timeouts with the connection pool are a [common knex GitHub issue](https://github.com/knex/knex/issues?q=is%3Aissue+timeout+connection+pool), and sometimes other factors beyond connection pool size (such as the docker postgres configuration) can come into play. Creating new records in the database for the permitted number of sensors that can be uploaded in one CSV is a fast operation. It doesn't need to be done concurrently, particularly with the failsafe of our async (notification) mode.

Also, I would be hesitant to modify the knex configuration for only for this ticket, as it does not look like we use `promise.all()` or `promise.allSettled()` to initiate concurrent transactions elsewhere in the app. **However, I recommend we keep the pool maximum in mind for the future, and I have created a new placeholder Jira ticket for future connection pool/knex timeout issues here: [LF-3253](https://lite-farm.atlassian.net/browse/LF-3253)**

Jira link: https://lite-farm.atlassian.net/browse/LF-3240

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

This has been tested with the CSV in this Jira ticket and the CSV in [LF-3193](https://lite-farm.atlassian.net/browse/LF-3193). Please test with other sensor combinations as well. If you want to see the API crash, checkout integration and upload either of those CSVs several times in a row -- it doesn't happen every time, but it should happen eventually.

**Bonus**: If you would like to see the effect of the max connections pool in realtime, checkout integration and remove two sensors from the CSV so that only 9 concurrent transactions are created. You'll want to shut down pgAdmin4 / any other programs that keep open connections to the database so that there is only the API's one connection live. See that the sensors will upload successfully. Then open a database connection via pgAdmin4 and see how that causes a knex timeout failure on the upload of the same file 🙂 

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/sensorController.js
+++ b/packages/api/src/controllers/sensorController.js
@@ -206,16 +206,20 @@ const sensorController = {
         );
 
         // Save sensors in database
-        const sensorLocations = await Promise.allSettled(
-          registeredSensors.map(async (sensor) => {
-            return await SensorModel.createSensor(
+        const sensorLocations = [];
+        for (const sensor of registeredSensors) {
+          try {
+            const value = await SensorModel.createSensor(
               sensor,
               farm_id,
               user_id,
               esids.includes(sensor.external_id) ? 1 : 0,
             );
-          }),
-        );
+            sensorLocations.push({ status: 'fulfilled', value });
+          } catch (error) {
+            sensorLocations.push({ status: 'rejected', reason: error });
+          }
+        }
 
         const successSensors = sensorLocations.reduce((prev, curr, idx) => {
           if (curr.status === 'fulfilled') {


SayakaOno (approved):
Great job!

Duncan-Brain:
@kathyavini 

This seems all too similar to other issues we have seen. 

Pool size: I believe we considered pool size when Anant was working on inserting Sensor Readings  becasue there were so many of them. I believe we tried it but ultimately dropped it because there seems to be a special formula for pool size that is not as straight forward as increasing the amount : https://stackoverflow.com/questions/66067124/sequelize-postgres-how-to-calculate-connection-pool-size 

I am slightly concerned about the current implementation because of how 1 transaction per sensor is being made. Should the unlikely situation of many sensors being uploaded at once occur we could too many transactions.

Have you tried:
- a regular knex statement like [knex.insert() or knex.upsert() or knex.batchInsert()](https://knexjs.org/guide/query-builder.html#insert)
  - sensor readings uses batchInsert() but could probably just be done with insert/update/upsert or something
  - this could maybe mean just fixing the model to allow for multiple sensors at a time

- moving the transaction outside of the model and outside of the try/catch statement
  - one of the controllers I copied resulted in something like nomination controller where trx is passed to each db we just can't forget to add await trx.commit() / await trx.rollback() if that method is helpful.

kathyavini:
> Should the unlikely situation of many sensors being uploaded at once occur we could too many transactions.

@Duncan-Brain Do you mean many sensors uploaded by one user (in one CSV) or by many different users using the app at the same time?

If the first, then no; that is exactly the issue that was causing the crash and that is what this PR addressed. One transaction is still created per sensor but they are created sequentially, not in parallel (because of the `for ... of` loop with the `await`). So a `trx.commit()` or `trx.rollback()` is always called before the next sensor is processed and there is only ever one transaction active at a time.

If many different users at the same time, then I guess that's always true. We need a transaction and not an insert or a batch insert because both the location table and the sensor table are going to be updated (or not) within that transaction -- this is how the location table is set up. But do you see a difference between this case and all the other places we use transactions in the app?

Duncan-Brain (approved):
So this is a definite improvement in my books !

So suggestions brought up on the zoom call that don't affect my approval.

Transaction handling
- this controller is making multiple transactions which is not ideal -- but the inconsistent trx pattern exists consistently in our controllers (😛 ) and so this fix of the controllers is a step in the right direction.
- Tech debt Jira ticket made : https://lite-farm.atlassian.net/browse/LF-3276

Nested Try{} Catch{}
- This helps achieve the desired pattern for mixed success/fail for individual sensors
- I think we should push back more on this pattern because it leads to weird situations like multiple transactions and nested try catch but not necessary to address on this ticket.

PR #2669, opened 2023-05-08
LF-3160: Misleading temperature line for sensors without temperature data

SayakaOno:
**Description**

When farm's unit system is `imperial` and temperature data is "null", `convertCelsiusToFahrenheit` function returns `32` (see the old line 15), and the temperature chart shows 32F when there is no data. This PR updates `convertCelsiusToFahrenheit` and `getTemperatureValue` to return `NaN` when a non-number is passed to the function.

Jira link: https://lite-farm.atlassian.net/browse/LF-3160

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/Map/PreviewPopup/utils.js
+++ b/packages/webapp/src/components/Map/PreviewPopup/utils.js
@@ -1,3 +1,5 @@
+import { convert } from '../../../util/convert-units/convert';
+
 const metric = ['metric', 'celsius', 'kpa'];
 const imperial = ['imperial', 'fahrenheit', 'psi'];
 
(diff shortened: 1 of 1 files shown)


kathyavini (dismissed):
This bug kind of feels like an argument for using TypeScript 🙂 

Nice fix.

Duncan-Brain:
@SayakaOno 

Could we use the Units functionality and deprecate this utility? convert package should be able to handle temperature.

SayakaOno:
@kathyavini 
Thank you Joyce! TypeScript would have caught this!

@Duncan-Brain 
I was in two minds if I should use the library! `convertCelsiusToFahrenheit` has been removed, thank you!

kathyavini (approved):
Oops, sorry I didn't immediately see I needed to re-review! It still works 😄

Duncan-Brain (approved):
Very nice! Yes its tough but I think we should deprecate as many of our custom utilities as possible.

I think I heard Joyce say that she was going to fix the soil water potential version this maybe?

Also since we hardcoded 'from' to 'C' it made me realize that our database can never allow non-metric sensor data. To be fixed hopefully in the future.

PR #2670, opened 2023-05-09
LF-2993 On iOS sensor icon does not appear next to "Sensor" in the "+" Drawer

kathyavini:
**Description**

The sensor SVG used in these three views:

- map drawer
- notification list (if you have an async sensor upload)
- selection handler (grouped locations on the map)

used the `<feBlendMode />` filter component, which is not supported in Safari (iOS/macOS). Therefore that icon was missing from those views in that browser. (Note the sensor icon on the map itself is different and not an SVG.)

This PR replaces the sensor SVG with another that doesn't use `<feBlendMode />`. To avoid touching any code in those three views, I added a bit of transform and line weight to make the sensor appear as close as possible to the original (it's not exact, but it's _pretty_ close).

Jira link: https://lite-farm.atlassian.net/browse/LF-2993

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

I tested this on Safari 16.4 on my mac. I don't own an iOS device so I could not test it in iOS Safari.

**Checklist:**


Duncan-Brain (approved):
Wow thats a nice find!

PR #2674, opened 2023-05-11
Lf 3161 soil water content component should show no sensor readings found when no data is present

Duncan-Brain:
**Description**

So for this one I had to change a couple of constants and add a few translations to make the correct text display.

Subtitle was originally required and so I worked on fixing it up a bit and that included renaming a few things and fixing how Units work.

Once I fixed units I thought that other components should not be using Weatherboard utility for units either so I fixed those.

Then I thought the subtitle shouldn't be required so I removed all that but decided to keep the positive changes regarding the units.

Jira link: [LF-3161](https://lite-farm.atlassian.net/browse/LF-3161?atlOrigin=eyJpIjoiMDI2ZDYyYTA3ODM5NGI5ZjlkZWNjNmQxNjUzMmNjZmUiLCJwIjoiaiJ9)

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

[LF-3161]: https://lite-farm.atlassian.net/browse/LF-3161?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


SayakaOno (commented):
Nice cleanup!
I wonder we can make the constant names `ambientTemperature` and `soilWaterPotential` more generic. For example, I had to use `area_total_area` for partial area not total area, and I didn't like it... I hope we could avoid that kind of situation. (I'm guessing any temperature and pressure are going to have the same object... maybe I'm overthinking or missing something)

Duncan-Brain:
@SayakaOno 

I actually kind of like the specific nature of these. 

I am thinking about how the Units form component works and how sometimes we only want to show a subset of units in a unit selection dropdown. Instead of all mm, cm, m, dm, km, etc we just want m and km but the case where mm and cm only also exists. 

I am wondering if in those cases area_total_area we should just be making another more specific case like partial_area. But for that one I think what you mean is maybe a "field_area" could cover both since partial field and total field would likely have the same magnitude context.

So if I were to look at what I have done on this jira ticket maybe in terms of <context>_<unit type> it might be :
- context: human_scale & unit: temperature
- context: water & unit: pressure

SayakaOno:
@Duncan-Brain 
I agree with you, I like having constants that have different subsets of units when there are multiple units. (I was assuming there was just one unit for pressure and temperature, and there would be no variation in constants)

> I am wondering if in those cases area_total_area we should just be making another more specific case like partial_area. But for that one I think what you mean is maybe a "field_area" could cover both since partial field and total field would likely have the same magnitude context.

You are right, I wish the name were something like "field_area". I think we are supposed to make a new constant for something like partial area, but it would have just been a copy of area_total_area with a different name.

> context: human_scale & unit: temperature
context: water & unit: pressure

I'm not sure if I understand these... are you saying the constant names could be "human_scale" and "water"??

Duncan-Brain:
@SayakaOno 

Yeah I am not sure the best naming here. Temperature could be in metric Kelvin (Rankine imperial) or maybe even a milliKelvin etc. Pressure could be in mmHg, Torr, Atm. 

I was saying human_scale_temperature & water_pressure. I could maybe use code context too like sensor_chart_temperature, sensor_chart_pressure.

Do you have any suggestions?

SayakaOno (dismissed):

packages/webapp/public/locales/pt/translation.json
```diff
@@ -1791,19 +1792,18 @@
...
-      "LAST_UPDATED_TEMPERATURE_READINGS": "Última atualização {{latestReadingUpdate}}"
+      "AMBIENT_TEMPERATURE_FOR": "Temperatura ambiente para"
     },
     "SOIL_WATER_POTENTIAL_READINGS_OF_SENSOR": {
       "TITLE": "Potencial hídrico do solo",
-      "SUBTITLE": "MISSING",
-      "Y_AXIS_LABEL": "Potencial hídrico do solo em {{soilWaterPotentialUnit}}",
-      "NO_DATA": "(sem dados))",
-      "LAST_UPDATED_TEMPERATURE_READINGS": "Última atualização {{latestReadingUpdate}}"
+      "Y_AXIS_LABEL": "Potencial hídrico do solo em {{units}}"
```
kathyavini, line 1802: So this is one of the phrases that was translated in two different ways for the sensor reading types, and the version that Isabela verified is potencial_de_água_do_solo -- and there was even a third translation that was not grammatical and removed!

Screenshot 2023-05-12 at 2 37 37 PM

Do you think we should keep these consistent across the CSV and the chart view? ChatGTP says that hídrico is more technical so I wonder if the CSV should be updated to use this too.

And then if we are going to keep the phrasing consistent in these different places, I wonder if we should pre-populate here (rather than using the MISSING tags) with the validation translations? What do you think?

kathyavini (approved):
Looks like a great fix and refactor!

I'm a little worried about sending forth "MISSING" tags for some of these translations because we have already seen inconsistencies created by translating reading types on a case-by-case basis. I left an inline comment where we already have a different existing Portuguese translation for "soil water potential."

I approve the code though! But still not sure about the best way to handle establishing canonical translations across the charts + CSVs...

packages/webapp/public/locales/pt/translation.json
Duncan-Brain: @kathyavini 

I think it is a good point. My comments for this might be:
- Maybe where it is critical if used in multiple places for validation we the developers should likely add some sort of comment or warning or separate file highlighting the implications of changing translations
- where it is only "inconsistent" maybe a process suggestion for translation PRs could be proposed where a search for "like" translations is done and an attempt at consistency is made

kathyavini: Hmmm! Very good point that in non-validation contexts it's not critical.

I don't love the idea of putting more work on the translators by having them refer to a process doc if we could just enforce it by doing the translations ourselves (for devs I don't mind an external doc!)... but in any case, new issue, new ticket 😁 I'll bring it in Standup to see how consistent we want these non-critical contexts.

Duncan-Brain: Process doc is always good even for trivial things!.... but yah translation file is also in need of some de-duping and dev strategy too.

SayakaOno (approved):

kathyavini (approved):

PR #2682, opened 2023-05-16
LF-3278 Incoming Ensemble sensor data gives 413 (payload too large) errors

kathyavini:
**Description**

This PR sets the incoming request size limit to 1MB on the sensor webhook route, while maintaining a 100kB limit for all other routes.

### Technical Details
It turns out it was not possible to override the app-wide `express.json()` with route-specific settings (in other words what I had insisted to @Duncan-Brain  we had to do was totally backwards!); once the app-wide middleware is defined, the route-specific ones are never used.

Therefore it was necessary to remove the global middlware and apply `express.json({ limit: '100kB' })` to all routes except the Ensemble webhook routes, basically following [this Stack Overflow](https://stackoverflow.com/questions/12921658/use-specific-middleware-in-express-for-all-paths-except-a-specific-one). It has been scoped to the Ensemble webhooks specifically as other partners might have different requirements and data limits.

Part of the custom middleware we need to call before our routes checks for a `req.body` on GET and DELETE requests. I refactored that particular check out of the containing middleware function so that it could also be called on the route with the higher limit (it needs to be called after `express.json()` or `req.body` will always be undefined). With the refactor it should continue to work as before.

### body-parser and lockfile
I removed and uninstalled body-parser as importing it no longer needed since Express [v4.16.0](https://github.com/expressjs/express/releases/tag/4.16.0) and now deprecated. The package-lock.json was generated with 

```
nvm use 16.15.0
npm ci
npm uninstall body-parser
```

so it should be correct for our repo settings. Nonetheless it has some differences from the existing lockfile with the `"requires" : {}` syntax.

If you are able to create a `package-lock.json`  without these changes, please let me know your OS & Node version, and feel free to feel free to commit directly to this branch if you would like to use that `package-lock.json` instead.

Jira link: https://lite-farm.atlassian.net/browse/LF-3278

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

The easiest way to test is using Postman or similar on both `/sensor/reading/partner/1/farm/` + other API endpoints with raw JSON in the request body (you don't have to hit a valid route/controller since the file size limit is enforced prior to that). I attached a big JSON just short of 1MB to the Jira ticket if you would like to test with that.

If you have a sensor webhook set up with ngrok, you could also let accumulate some hours of data and check the request size in ngrok; that method is kind of slow though!

**Checklist:**


Duncan-Brain (commented):
@kathyavini 

Looks pretty much like I thought express should be capable of. 

But yeah I am surprised that you have to set the rest as the default value.

My only question is can it be more specific to: `/sensors/reading/partner/<esci partner_id>/farm/`

kathyavini:
@Duncan-Brain I pondered that! My thinking in applying it to the general route was that Kevin often says we want to be able to accomodate many kinds of sensors eventually and I wasn't sure if the webhook endpoint design was going to change when that happens.

But then it makes more sense to change it as necessary when the time comes, right?? I'll scope it more specifically!

SayakaOno (approved):
Looks good to me!

SayakaOno (approved):
Looks good to me!

PR #2689, opened 2023-05-23
LF-3284 Modify visualization algorithm for sensor readings data

kathyavini:
**Description**

This moves the sensor charts to hourly time points. Dots are added to the line chart when fewer than 90% of hourly time points are present.

A small final adjustment uses a smoother curve on data containing more than 50%, but less than 90%, of hourly data points on small chart widths (i.e. mobile screens).

Please note the sparse data display conditions are a bit of a placeholder -- they should be revisited when real data feeds are available. The chart in general will also look best if viewed after the xAxis and yAxis changes in LF-3291 are applied.

Jira link: https://lite-farm.atlassian.net/browse/LF-3284

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/SensorReadingsLineChart/index.jsx
+++ b/packages/webapp/src/components/SensorReadingsLineChart/index.jsx
@@ -36,7 +36,6 @@ const PureSensorReadingsLineChart = ({
   xAxisDataKey,
   yAxisDataKeys,
   lineColors,
-  graphDatapoints,
 }) => {
   const { t } = useTranslation();
   const [legendsList, setLegendsList] = useState({});
@@ -110,7 +109,8 @@ const PureSensorReadingsLineChart = ({
     const displayDate = dateFormat.format(tickDate);
     const displayDay = dayFormat.format(tickDate);
 
-    if (index % graphDatapoints.length == graphDatapoints.length / 2) {
+    // Hourly datapoints
+    if (index % 24 == 12) {
       return (
         <>
           <text x={x} y={y - 16} textAnchor="middle">
@@ -123,7 +123,7 @@ const PureSensorReadingsLineChart = ({
       );
     }
 
-    if (index % graphDatapoints.length === 0 && index !== 0) {
+    if (index % 24 === 0 && index !== 0) {
       const pathX = Math.floor(x) + 0.5;
 
       return <path d={`M${pathX},${y - 22}v${-16}`} stroke="black" />;
--- a/packages/webapp/src/components/SensorReadingsLineChart/index.jsx
+++ b/packages/webapp/src/components/SensorReadingsLineChart/index.jsx
@@ -1,4 +1,4 @@
-import React, { useState, useEffect } from 'react';
+import React, { useState, useEffect, useRef } from 'react';
 import styles from './styles.module.scss';
 import {
   LineChart,
@@ -20,6 +20,7 @@ import { getLanguageFromLocalStorage } from '../../util/getLanguageFromLocalStor
 import { SensorReadingChartSpotlightProvider } from './SensorReadingChartSpotlightProvider';
 import produce from 'immer';
 import { useTranslation } from 'react-i18next';
+import useElementWidth from '../hooks/useElementWidth';
 
 const PureSensorReadingsLineChart = ({
   showSpotLight,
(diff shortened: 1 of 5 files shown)


packages/webapp/src/components/SensorReadingsLineChart/index.jsx
```diff
@@ -43,6 +44,9 @@ const PureSensorReadingsLineChart = ({
   const dateFormat = new Intl.DateTimeFormat(language, { day: '2-digit' });
   const dayFormat = new Intl.DateTimeFormat(language, { weekday: 'short' });
 
+  const chartRef = useRef(null);
+  const { chartWidth } = useElementWidth(chartRef);
```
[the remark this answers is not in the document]
kathyavini: Ack sorry for using your hook wrong! Well in this case also for just being wrong with JavaScript object destructuring 😆

I had originally been testing this with a `useEffect()` and a listener on the window resize before I switched to the hook. I don't know why the responsive container doesn't work in the hook case 🤔 

But I think I'll just move the ref to the title wrapper

`<div className={styles.titleWrapper} ref={chartRef}>`

as that seems the easiest workaround (and they share width)?

packages/webapp/src/components/SensorReadingsLineChart/index.jsx
```diff
@@ -207,8 +227,16 @@ const PureSensorReadingsLineChart = ({
...
                   isAnimationActive={false}
+                  activeDot={{ r: 5 }}
+                  dot={dataDensity(sensorReadings, attribute.value) < 0.9}
+                  type={
+                    dataDensity(sensorReadings, attribute.value) >= 0.9 ||
+                    (chartWidth < 400 && dataDensity(sensorReadings, attribute.value) > 0.5)
+                      ? 'natural'
+                      : 'linear'
+                  }
+                  connectNulls
```
Duncan-Brain, line 239: @kathyavini I am not sure I like the connected nulls myself as it doesn't necessarily reflect whats happening well. I forget if this was discussed.

Screenshot 2023-05-24 at 9 52 42 AM
Screenshot 2023-05-24 at 9 52 13 AM

packages/webapp/src/components/SensorReadingsLineChart/index.jsx
```diff
@@ -207,8 +227,16 @@ const PureSensorReadingsLineChart = ({
                   strokeWidth={2}
                   dataKey={attribute.value}
                   stroke={attribute.color}
-                  activeDot={{ r: 6 }}
                   isAnimationActive={false}
+                  activeDot={{ r: 5 }}
+                  dot={dataDensity(sensorReadings, attribute.value) < 0.9}
```
Duncan-Brain, line 232: @kathyavini I am not sure this acts as intended. With the following graph it is missing datapoints in one 24 hour period but the density for the others is still ~100%.

Screenshot 2023-05-24 at 9 52 42 AM

packages/webapp/src/components/SensorReadingsLineChart/index.jsx
kathyavini: Hmm! The connectNulls was used because I was asked to handle sparse data feeds (feeds with only 3-4 data points per day). Without connectNulls, no line is drawn at all and those feeds become a sequence of dots.

I'm leaning towards connectNulls being a property we just hard-code to certain feeds that we know can be sparse, but if you can think of the right conditional to check so we can conditionally apply it, we can do that instead! I can't think of one off the top of my head.

Duncan-Brain: Yeah true that makes sense. I wish we knew what kind of data to expect. I think with a dot density function that assessed the frequency/distribution of nulls could also be used to make the correct tickmarks. But that is too much for this PR. I am good with it!

Duncan-Brain (dismissed):

SayakaOno (commented):
Looks good!
You have small conflicts :)

kathyavini:
@SayakaOno thank you! Merge conflicts are resolved.

@Duncan-Brain if you could kindly re-review 🙏

Duncan-Brain (approved):
Excited for this change nice work!

SayakaOno (approved):

PR #2694, opened 2023-05-24
LF-3283 Modify sampling algorithm for sensor_readings data

kathyavini:
Note: Please run the migration before loading the sensor reading chart

```
npm run migrate:dev:db
```

And afterwards
```
npm run migrate:make:rollback
```

**Description**

This updates the SQL function called by the `getSensorReadingsByLocationIds()` method in our sensorReading model.

The previous function returned the most recent raw datapoint in the 1-hour window previous to the specified time.

The new function returns the average of all sensor readings in the 1-hour window centred on the specified time (i.e. 30 minutes before to 30 minutes after). The data is still returned in hourly increments.

To view the function, I recommend copying and pasting the raw SQL string in the migration into a `.sql` file to get SQL syntax highlighting 🙂  The `avg_value` Common Table Expression (CTE) is new, plus the overall function was simplified slightly without the need to rank order the timestamps. I left the null calculations unchanged.

Jira link: https://lite-farm.atlassian.net/browse/LF-3283

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

I tested this using the debugger in pgAdmin4 with the synthetic data in my actual `sensor_reading` table. I didn't test with an SQL insert -- if you create one could you please add it to the Jira ticket? 🙂 

**Checklist:**


Duncan-Brain:
So I am generally on board with a new sampling algorithm. But I think I am hesitant about having this done right away. I might prefer to maybe wait until we have some more info on a few things. ( a lot of maybes in that sentence lol)

I have a lot of questions on this one (mostly around the requirements) so I have tried to separate the Jira requirements questions (on the Jira ticket) from the implementation ones here. 

For the implementation:
1. This function was kind of intense to begin with -- when I altered it last it was really just to make the graph work properly I didn't even try to assess if this was the best way to do things. How confident are you that we should continue using this custom function approach compared to a knex method or something else? To use your line should we ask the tech lead... ? haha 

2. When Anant originally shared this with me I got the impression he asked someone about how best to do this and landed on the ranking method in the custom function as an efficient method. I decided I should test the latency of each function before mentioning so I just tested the three functions your version, my version and Anants and I am seeing them take ~600ms, ~400ms, and ~700ms respectively just using the query tool so I think that is all in the same ballpark and ok.

kathyavini:
@Duncan-Brain

1. I'm fine with asking Ivàn 😆 Having now looked more closely, I don't think the SQL function is as bad as all that, but I'm not particularly attached to it. SQL is going to be a context shift and a learning curve whenever someone has to pick this up to work on it/understand it, and I agree that's not great. 

The table itself is HUGE (almost half a million records on beta!) but I guess we're only processing... 7200 records at 
a time, depending on how we query? That seems manageable to process on our backend, even though the SQL will probably remain more efficient. So yes... tech lead decision! Re-implementing in ORM might not be in scope for this ticket though.

2. I think averaging might be even less complex than ranking but yes, shouldn't be too big of a difference at this scale.

As I wrote on Jira, I do think that as a data processing method, averaging is better than our current implementation, which is arbitrarily selecting a single raw data point out of every 60. But it's only an incremental improvement and kind of a placeholder until we know what the real data will look like.

SayakaOno (commented):
I took a look at the query, but I don't think I will understand it very soon.
Since we've got the tech lead, I will refrain from approving/making comments at this time. (I just can't...)
Sorry!!!

Duncan-Brain:
@kathyavini 

I was just thinking about the partial time zone problem (ie India, Australia, Newfoundland) and wondering if this should be incorporated into the new resampling algorithm. 

Potentially two users living in different timezones but viewing the same farm could see different averaged data. 

Just wanted to confirm that currently this updated function averages over "an hour" ( ie Data-30mins - Data +30 mins) and not "the hour" (ie 1pm -2pm UTC).  I think we may want to go with "the hour" so that all users see the same averaging.

kathyavini:
@Duncan-Brain I'm pretty sure that the averaging will work the same as the ranking regarding the changes you made in PR https://github.com/LiteFarmOrg/LiteFarm/pull/2710 

The dates will be passed in with an offset, so when the one-hour window is calculated relatively, it will be centred relative to an offset (so centred on "the hour") and indeed it will be the same for users viewing the same farm at different locations -- the timestamp will be different in India but the hour of time being averaged is the same.

But let's get your PR merged first so I can re-write this migration relative to those changes, and then we can double-check 🙂 The migrations from these two PRs are not going to work in sequence since you're modifying a function I deleted...

kathyavini:
@Duncan-Brain and @iperdomo -- I've just adapted the migration to refer to Duncan's updated SQL function from this PR; otherwise there are no other changes (other than fixing the GNU License!)
- https://github.com/LiteFarmOrg/LiteFarm/pull/2710

@Duncan-Brain I just checked and you can please confirm that the averaging windows are the same in all time zones 🙂

iperdomo (approved):

Duncan-Brain (approved):
@kathyavini Looking good to me!

PR #2702, opened 2023-05-25
LF-3026 Pad start date for sensor data query

kathyavini:
**Description**

The time range for the data requested from the database is set using a Date object created on the backend (in `getAllSensorReadingsByLocationIds`) using a time string sent from the frontend. Conversions happen correctly on the frontend for display in the chart once the data is received, but for time zones set ahead of server time, the data stream does not correctly begin at midnight local time -- presumably because the API didn't send back any data for midnight local time, as it is outside of the window requested by the server code.

This PR, therefore, pads the server-requested data with an extra day to handle midnight in all time zones, which re-aligns the data start with midnight for all users and fixes the tick mark offset issue.

_However, we still don't know why only some combinations of system time settings cause this issue in the first place...!_

Jira link: https://lite-farm.atlassian.net/browse/LF-3026

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/api/src/models/sensorReadingModel.js
+++ b/packages/api/src/models/sensorReadingModel.js
@@ -101,7 +101,7 @@ class SensorReading extends Model {
   ) {
     const durationType = '1 hour';
     const startDate = new Date(endDate);
-    startDate.setDate(endDate.getDate() - 5);
+    startDate.setDate(endDate.getDate() - 6);
     const sensorReadingsResponsePromises = [];
     for (const readingType of readingTypes) {
       sensorReadingsResponsePromises.push(


SayakaOno (approved):
I think this is a reasonable solution. I wish we could update the API to only return data displayed on the graph...

In my local, "read_time" of the first item of the response is 12am in PT, but in beta, it is 12am in GMT.
It looks like my local DB's time zone is PT, and beta DB's time zone is GMT. I'm not sure if my assumption is correct, but I think 12am should be based on user's timezone...

kathyavini:
@SayakaOno you're right and it's true! It's the `endDate` string sent to the backend that should probably be something like a Unix timestamp, or maybe a timezone offset instead. But I think that @Duncan-Brain was saying that there was a complication passing something different there?

Maybe in the morning, we could take another look to see if we can create the correct solution; I have no more coding left in me today though 😅

Duncan-Brain:
@kathyavini  

Yeah it looks like I may have over-exaggerated the passing timestamp thing. Looked at it again and for some reason we are formatting it in the saga with moment, then using the JS date function to format it in the controller tooo, and then again using the JS date function in the model. I think I just didn't want to touch the date manipulation.

Also I looked into how I broke it and I think it was a combination of the "-4" and the ordering of the readings array for just the temperature graph -- I think the Soil Water Potential may have had this issue existing. The weather data built the array used in the temperature reduce function so its points were used as the basis for the returned data.

Duncan-Brain (approved):

PR #2714, opened 2023-06-02
Lf 3296 user cannot complete field work task without making change where it is supposed to be made

MwayaB:
**Description**

This PR fixes the bug where a user cannot complete a fieldwork task after selecting the "yes" radio under "did you make any changes to this task" in the task completion flow.

Jira link: https://lite-farm.atlassian.net/browse/LF-3296

**Type of change**

**How Has This Been Tested?**
Tested according to this test plan(https://lite-farm.atlassian.net/wiki/spaces/LITEFARM/pages/1317470213/LF-3296+Test+Plan)

- [x] Passes test case
- [ ] Other (please explain)

**Checklist:**


Duncan-Brain (commented):
@MwayaB 

So this is allowing the completion to successfully occur.

My thoughts so far are:
1. In the Jira ticket the comments mentioned disabling the continue button on the frontend if there are no changes to the fields when yes is selected. I think that is the pattern we want as well. (But we will still need your changes here for direct api calls)

2. checkForCompleteCustomFieldWork() is almost an exact copy of checkAndAddCustomFieldWork(). It would probably be best just to correct that function.

3. The changes to the fieldWorkTypeModel: If possible I think that we want all variable manipulation to occur in the controller. My understanding is that the controllers function is to put the data in the correct form for use with the database.

iperdomo (dismissed):

MwayaB:
@Duncan-Brain My fix is i think what @SayakaOno suggests here "My suggestion would be either let the user complete the task". I think "or disable “Continue” button when the user selects “Yes” but does not make any changes." is bug another possible fix for this bug.
@iperdomo Yes i can write tests for the scenario

Duncan-Brain:
@MwayaB 

Jest tests are passing. And I can now successfully complete a task without making changes even when I select that I have made changes. (If you could please make that Jira ticket to correct the frontend that would be cool)

- Thanks for taking a look at the lowerCase that clarification helped me understand - much appreciated
- Reduced function looks good

Question about the new tests:
- Do we need to disable other tests (the `xtest()` and commented `//expect `statement) in order to get a new test to work? It seems strange to me. Any other info you have on that?

iperdomo (changes requested):
Agree with @Duncan-Brain, we shouldn't have skipped tests

MwayaB:
@Duncan-Brain https://lite-farm.atlassian.net/browse/LF-3397 ticket created.
Re: the skipped test I couldn't quite figure out why those tests were failing and attempted several fixes before skipping them, I found myself spending quite a bit of time on that and figured that maybe I could create a ticket to explore that?

packages/api/tests/user.test.js
```diff
@@ -367,8 +367,8 @@ describe('User Tests', () => {
       });
 
       test('Should post then get a valid user and user spotlight', async (done) => {
-        const fakeUser = mocks.fakeUser();
-
+        jest.setTimeout(90000);
```
iperdomo, line 370: Why are we modifying the default timeout interval 🤔  [setTimeout](https://jestjs.io/docs/jest-object#jestsettimeouttimeout)?

iperdomo:
I'm not able to run tests locally so I can't debug _why_ there is a need to increase the timeout 😓 ...

iperdomo:
Still can't run the tests. After checking with @Duncan-Brain, he is able to test

iperdomo (approved):
As mentioned, I can't check this branch. `npm run test` fails for me.
[test.log](https://github.com/LiteFarmOrg/LiteFarm/files/11898583/test.log)

packages/api/tests/user.test.js
```diff
@@ -367,8 +367,8 @@ describe('User Tests', () => {
       });
 
       test('Should post then get a valid user and user spotlight', async (done) => {
-        const fakeUser = mocks.fakeUser();
-
+        jest.setTimeout(10000);
+        const fakeUser = await mocks.fakeUser();
```
Duncan-Brain, line 371: @iperdomo @MwayaB

Tested this without this both with and without the change and it seems to work for me. Generally, I think if we have to change another test something fishy is going on. I am wondering if it is this `await` that doesn't need to be there that necessitates the timeout? 

Screenshot 2023-06-28 at 1 47 59 PM

iperdomo: @Duncan-Brain now that I can run test tests 😅 , same here. I can run the tests without the call to `jest.setTimeout()`

Duncan-Brain (approved):
Looks good to me ! 

Thanks for adding that field work type code too. If you could link the new Jira ticket for the frontend piece in this Jira ticket that would be cool!

MwayaB:
@Duncan-Brain Already have it linked in the relates to section

iperdomo (approved):

Duncan-Brain:
@MwayaB sorry I missed that!

PR #2729, opened 2023-06-13
LF-3335 Initial implementation of "Dockerized" dev environment

iperdomo:
**Description**

Initial implementation of `docker compose up` for development environment. This is a working example for @kathyavini then we can continue the work for the `imaginary` part.

Jira link: https://lite-farm.atlassian.net/browse/LF-3335

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/api/dev.sh
+++ b/packages/api/dev.sh
@@ -1,0 +1,9 @@
+#!/usr/bin/env bash
+
+set -eu
+
+# We assume we're running using `node` user
+[[ ! -d "node_modules" ]] && { echo "Installing modules..."; npm install; }
+
+npm run migrate:dev:db
+npm run nodemon
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,62 @@
+version: "3.7"
+
+volumes:
+  postgres-data:
+
+services:
+  db:
+    container_name: litefarm-db
+    image: postgres:13
+    ports:
+      - "5433:5432"
+    environment:
+      POSTGRES_PASSWORD: "postgres"
+    volumes:
+      - ./initdb.d:/docker-entrypoint-initdb.d
+      - postgres-data:/var/lib/postgresql/data
+  api:
+    container_name: litefarm-api
+    image: node:16.15.0
+    volumes:
+      - ./packages/shared:/shared
+      - ./packages/api:/api
+    working_dir: /api
+    entrypoint: ./dev.sh
+    user: node
+    restart: unless-stopped
+    environment:
+      - DEV_DATABASE_HOST=db
+      - DEV_DATABASE=pg-litefarm
+      - DEV_DATABASE_USER=postgres
+      - DEV_DATABASE_PASSWORD=postgres
+      - TEST_DATABASE_HOST=db
+      - TEST_DATABASE=test_farm
+      - TEST_DATABASE_USER=postgres
+      - TEST_DATABASE_PASSWORD=postgres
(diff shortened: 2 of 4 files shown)


packages/api/dev.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+set -eu
+
+# We assume we're running using `node` user
+[[ ! -d "node_modules" ]] && { echo "Installing modules..."; npm install; }
```
kathyavini, line 6: Are you using a Linux machine? For those of us on MacOS it would be preferable to still be able to run locally with our system's `node_modules` intact.

Duncan-Brain:
So we could use this as a base to work off of... but it didn't quite work for me:

1. I received this error on backend during npm install _**on docker**_:
Screenshot 2023-06-13 at 3 28 14 PM

I can remove the problem by running npm install _**on my local machine**_  and taking advantage of this line `[[ ! -d "node_modules" ]]`

This check is maybe problematic for that reason.

2. Once that was fixed and all containers were running my frontend could not connect to the backend because in .env's I have:

```
// webapp/.env

VITE_API_URL=http://localhost:5001
```
and
```
// api/.env

PORT=5001
```

So we may want a single source of truth for shared variables I think .env could be good if we can use that context in docker files.

3. should the /shared package.json or the root ~package.json be installed somewhere?

kathyavini (commented):
I like it pretty well... it's just very, very slow for me!

I didn't get the named volumes working so I trashed my `node_modules` and let Docker write to the host system. It resulted in a horrendously slow `pnpm install` for me (18 minutes the first time!) and after some initial timeout errors from Vite initial load times of around 30s - 2.5m on main route changes.

My computer is a little older than Duncan's but it's still a fairly beefy machine. I'm curious how others (particularly other MacOS users) find the responsiveness of this setup. There's [some indication](https://www.github.com/docker/roadmap/issues/7) (also [here](https://www.github.com/pnpm/pnpm/issues/2479) for `pnpm install` specifically) that this is a known issue bind mounting with Docker Desktop for mac, but I'm surprised because LiteFarm is just not that big. 

With the small change mentioned below the debugging experience is on par with local, and the setup is of course so fun (well, if the source of @Duncan-Brain's install errors could be pinpointed and also if it didn't take so long). But the lack of snappiness is the main thing that would keep me from wanting to make this the primary dev environment.

Hopefully figuring out the `node_modules` volumes would be enough to speed it up?

iperdomo:
Yes, `node_modules`  must be removed before trying out this branch. It's good that we're trying out on macOS, it is a known issue that systems with lots of files (nodejs and php ecosystems) have a performance issue. 
Can you try enabling _virtiofs_ ? https://www.docker.com/blog/speed-boost-achievement-unlocked-on-docker-desktop-4-6-for-mac/

iperdomo:
It seems that [VirtioFS is now GA since Dec, 2022](https://www.docker.com/blog/docker-desktop-4-15-improved-usability-and-performance/). Availalble for macOS 12.5+

SayakaOno:
Here is the error I got for the webapp:
Screenshot 2023-06-15 at 2 42 55 AM

for the api:
Screenshot 2023-06-15 at 2 44 13 AM

iperdomo:
Closing this in favor of only dockerizing supporting services (PostgreSQL, Minio, Redis) - for now

PR #2738, opened 2023-06-21
Lf 3229 introduce iso country code to countries table and refactor to use that rather than name for google maps identification

gursimran-singh:
**Description**

This PR tries to solve the issue caused by the country name change ( as in case of Turkey to Türkiye) which throws an error while using google maps api as it is not found in DB and needs to be added every time if there is a change. So, to over come this issue, a more standard way of identifying the countries by using 2 character code rather than name is implemented. This change resolves the issue even when google uses different country name than present in our DB.

Jira link: https://lite-farm.atlassian.net/browse/LF-3229

**Type of change**

**How Has This Been Tested?**

Adding new farm does not cause any errors. 

- [x] Passes test case
- [ ] Other (please explain)

**Checklist:**


kathyavini:
Thank you @gursimran-singh!!

@SayakaOno has been working with both the countries table and ISO codes recently, so she is going to be the best person to review this for sure!

However, she is currently on vacation this week, so I just wanted to give you a heads up that the final review might end up a little delayed for that reason.

gursimran-singh:
Thanks Joyce for letting me know.

iperdomo (changes requested):
@gursimran-singh Thanks for your contribution. 

There has been a _small_ change in requirements for this table. Would you be keen on adding `alpha-3` code for countries too? The alpha-3 is needed for a small UI change we're introducing.

Again, thanks for your time.

gursimran-singh:
@iperdomo Sure, i will make the necessary changes.

SayakaOno:
I am not sure how I missed this, but what I need should be `alpha-2` as well. ([react-phone-number-input](https://gitlab.com/catamphetamine/react-phone-number-input#country-code))
I don't think extra change is required. I'm sorry and thank you @gursimran-singh !

iperdomo (approved):

SayakaOno (approved):
I'm sorry for being late for giving a review... changes look good!
I am wondering... what is "country-code" in `country_iso_codes.json` for?

gursimran-singh:
I downloaded the file form internet and made modification to fit our purpose. Its a numeric code for each country. Here is the  reference https://en.wikipedia.org/wiki/List_of_ISO_3166_country_codes.  We can ignore it as we are not using in our system or do you want me to remove it?

SayakaOno:
@gursimran-singh Thank you, I understand. I think it's fine if they are there. I'll merge the PR!

PR #2749, opened 2023-06-29
LF-3266 - Enhancers | Inactive users: display 

thxv3n0lvl:
**Description**

Added a couple validations to TaskCard so when a revoked user still assigned to a task, their name will be displayed with a "(inactive)" mark and also greys out the icon and name. This was an issue in terms of not handling the behavior described in the ticket  but it was discussed in the same ticket this should be the behavior although I might be loosing some bits of the conversation.

Jira link:
https://lite-farm.atlassian.net/browse/LF-3266

**Type of change**

**How Has This Been Tested?**

3 scenarios were tested on the Task view:
1. Task assigned to an active user -> name displays as always
1. Task unassigned -> unassigned format displays as always
1. Task assigned to inactive user -> new format in place

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

Screenshot of the implementation:
image
image

### DIFF
--- a/packages/webapp/src/components/CardWithStatus/TaskCard/TaskCard.jsx
+++ b/packages/webapp/src/components/CardWithStatus/TaskCard/TaskCard.jsx
@@ -67,13 +67,12 @@ export const PureTaskCard = ({
     onClickCompleteOrDueDate?.();
   };
 
+  const isAssigneeInactive = assignee?.status === 'Inactive';
   let assigneeName = '';
-  let isAssigneeInactive = false;
   if (assignee !== null) {
-    isAssigneeInactive = assignee.status === 'Inactive';
-    assigneeName = `${assignee.first_name} ${
-      assignee.last_name.length > 0 ? assignee.last_name.toUpperCase().charAt(0) + '.' : ''
-    }`;
+    const lastName =
+      assignee.last_name.length > 0 ? assignee.last_name.toUpperCase().charAt(0) + '.' : '';
+    assigneeName = `${assignee.first_name} ${lastName}`;
   }
 
   return (
--- a/packages/webapp/src/components/CardWithStatus/TaskCard/TaskCard.jsx
+++ b/packages/webapp/src/components/CardWithStatus/TaskCard/TaskCard.jsx
@@ -67,6 +67,15 @@ export const PureTaskCard = ({
     onClickCompleteOrDueDate?.();
   };
 
+  let assigneeName = '';
+  let isAssigneeInactive = false;
+  if (assignee !== null) {
+    isAssigneeInactive = assignee.status === 'Inactive';
+    assigneeName = `${assignee.first_name} ${
+      assignee.last_name.length > 0 ? assignee.last_name.toUpperCase().charAt(0) + '.' : ''
+    }`;
+  }
+
   return (
     <CardWithStatus
       data-cy="taskCard"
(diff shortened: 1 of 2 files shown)


Duncan-Brain (approved):
Sorry it took a while to respond Alan!

Everything looks great. Not sure whether the optional chaining operator (ie assignee?.status) could have saved a line or two of code or not. But that is the only thing I might've considered.

Thanks!

thxv3n0lvl:
> Sorry it took a while to respond Alan!
> 
> Everything looks great. Not sure whether the optional chaining operator (ie assignee?.status) could have saved a line or two of code or not. But that is the only thing I might've considered.
> 
> Thanks!

@Duncan-Brain correct me if I'm wrong but I think if you refer to do  this instead
`const isAssigneeInactive = assignee?.status === 'Inactive';`
and removing the if condition `if(assignee !== null)` it will lead to an error as the assignee object is actually null when the "Unassigned" option is selected for the task. Let me know if this is not what you mean.

On the other hand, I'm not sure how to proceed with this PR given [this](https://litefarm.slack.com/archives/C041EMXTBPY/p1688407643866929) conversation.

Duncan-Brain:
@thxv3n0lvl 

I see.. it looks like we had a duplicate issue haha That can be confusing sorry about that! 

Let me see if we can clear that up on the other ticket. It seems pretty small to reuse so I will inquire further. But I think in the meantime we could merge and they could always update this implementation with a reusable component once that is made.

As for the null condition -- really don't worry about it haha.  It is probably more readable as-is. Its just fun for me to think about how few lines of code could be used. Here is a little test of what I believed would happen in that case:
Screenshot 2023-07-05 at 9 26 29 AM

Duncan-Brain (changes requested):
Maybe this is actually more full stack than I realized! Re-reading again says maybe two other things could be completed:

watching the video you can see the task is unassigned when clicked -- presumably we want to keep the task assigned to the inactive user I am getting from the conversation around it
Also in finances section under labour expenses the same (inactive) should be applied*

packages/webapp/src/components/CardWithStatus/TaskCard/TaskCard.jsx
```diff
@@ -67,6 +67,14 @@ export const PureTaskCard = ({
     onClickCompleteOrDueDate?.();
   };
 
+  const isAssigneeInactive = assignee?.status === 'Inactive';
+  let assigneeName = '';
+  if (assignee !== null) {
```
thxv3n0lvl, line 72: @Duncan-Brain I left the null validation just to make sure we don't miss the unassigned state of a task. Let me know if you would llike to see this one differently.

packages/webapp/src/components/CardWithStatus/TaskCard/TaskCard.jsx
```diff
@@ -67,6 +67,14 @@ export const PureTaskCard = ({
...
+    const lastName =
+      assignee.last_name.length > 0 ? assignee.last_name.toUpperCase().charAt(0) + '.' : '';
+    assigneeName = `${assignee.first_name} ${lastName}`;
+  }
+
```
thxv3n0lvl, line 77: for this one I left the `lastName`  in a separate constant just to make it legible as there's a bit of logic to present the last name in a "capital first letter and dot" format (not sure how to call it 😂 ). Also is possible that a username doesn't have a last name, so I left a validation for that.

packages/webapp/src/components/CardWithStatus/TaskCard/styles.module.scss
```diff
@@ -88,3 +88,12 @@
...
 }
+
+.inactive {
+  color: var(--grey500);
+}
+
+.firstInitial.inactive {
+  background-color: var(--grey500);
+  color: white;
+}
```
thxv3n0lvl, line 99: @Duncan-Brain I don't have access to figma but maybe someone can confirm the correct color for the inactive label on the task card view.

thxv3n0lvl:
I updated the name of this PR to avoid confusions with the other tickets.

packages/webapp/src/components/CardWithStatus/TaskCard/TaskCard.jsx
Duncan-Brain: looks great!

Duncan-Brain (approved):
Thanks Alan looks great, if there is any chance you could update the Jira link in the description too for the next reviewer that would be cool!

PR #2751, opened 2023-06-30
created directory and added script

MwayaB:
**Description**

This PR creates a new directory(ci) in the monorepo to help facilitate CI activities primarily e2e testing. It also adds a shell script to build back and front end images to facilitate the same

Jira link:
https://lite-farm.atlassian.net/browse/LF-3400

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


kathyavini (approved):
Worked for me! As I am already building as AMD64 on my Intel mac the flag made no difference on my system, and I'm very glad to see this is enough to fix the ARM64 issue for M1 users.

Also thank you for testing development `Docker compose up` locally! 🥰  It was good to verify the Puppeteer/ARM64 issue doesn't cause errors with that.

packages/ci/build-prod-images.sh
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 echo "Building backend docker image..."
-docker build -t litefarm/backend:latest -f ./packages/api/prod.Dockerfile ./packages/
+docker build --platform linux/amd64 -t litefarm/backend:latest -f ./packages/api/prod.Dockerfile ./packages/
```
iperdomo, line 4: * Let's add the `--platform` option to both `docker build [...]` calls
* I suggest to use long version of parameters, `--tag` instead of `-t`, `--file` instead of `-f`
  * The rationale behind using long parameter names is to increase the readability of the script, specially for those not familiar with the concepts behind the program, e.g. What's a _tag_ when building a Docker image?

packages/ci/build-prod-images.sh
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
```
iperdomo, line 2: The first line of a bash script should be at least `set -e`, so the script fails if any program exists with non-zero value. 
Background: https://wizardzines.com/comics/bash-errors/

iperdomo (changes requested):
Let's add a couple of changes to the script

iperdomo (changes requested):
I expect that we have a top level `ci` folder, not child one, under `packages/*`

kathyavini (approved):

PR #2743, opened 2023-06-27
LF-3403 Number Input component with stepper

kathyavini:
**Description**

This PR adds the option of a stepper control (up and down buttons) to the number `<Input />` component. This will be used within the Repeat Crop Plan modal and potentially elsewhere in the app.

The stepper control has been given a pleasing hover effect 😁 

Jira link: https://lite-farm.atlassian.net/browse/LF-3403

**Type of change**

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

**How Has This Been Tested?**

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [ ] Passes test case
- [x] UI components visually reviewed on desktop view
- [x] UI components visually reviewed on mobile view
- [x] Other (please explain)

This was created and tested in Storybook on the Node 18 branch with -- my understanding is that this is important! -- a pnpm lockfile generated with pnpm v.8.5.3.

Because of complications with pnpm + React Icons, Storybook may or may not be working for you on integration. At the moment this Input with stepper does not occur in the application so Storybook is the way to test if you can figure out the combination of pnpm + Node to get it working 😬 

**Checklist:**

- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [x] The precommit and linting ran successfully
- [ ] I have added or updated language tags for text that's part of the UI
- [ ] I have added "MISSING" for all new language tags to languages I don't speak
- [ ] I have added [the GNU General Public License](https://lite-farm.atlassian.net/l/cp/BT0Dd7WW) to all new files

PR #2752, opened 2023-07-04
LF-3403a Fixes and tests for number input stepper

kathyavini:
**Description**

This is Part II of #2743 (Number Input Component with Stepper)

The original new component lacked the correct handling of min and max (which I misunderstood as referring to validation; they are actually the hard reset of the number value onBlur in a normal number `<Input />`). This has now been added.

After @SayakaOno kindly helped me understand better the rationale for interaction testing this component, I have added play tests for the Number Input with Stepper with and without min/max.

In support for play testing, I had also unintentionally broken the interaction on the Input with password!! (See LF-3417) I added an interaction test for that as well, but please note the bug was caused by the icon being covered by another component (in the CSS layer sense); this is not something that can be detected by @storybook/testing-library and so the proper automated test would belong in Cypress.

Jira links:
- Number Input: https://lite-farm.atlassian.net/browse/LF-3403
- Password Bug: https://lite-farm.atlassian.net/browse/LF-3417

**Type of change**

**How Has This Been Tested?**

- [x] Passes test case
- [ ] Other (please explain)

To run the automated Storybook tests, first start storybook and wait for it to load:

```
pnpm storybook
```

Then run these tests with:

```
pnpm test-storybook src/stories/Form/Input/Input.stories.jsx
```

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/Form/Input/index.jsx
+++ b/packages/webapp/src/components/Form/Input/index.jsx
@@ -135,6 +135,7 @@ const Input = ({
             right: 0,
             transform: inputType === 'date' ? 'translate(-26px, 15px)' : 'translate(-17px, 15px)',
             cursor: 'pointer',
+            zIndex: 1,
           }}
         />
       )}
--- a/packages/webapp/src/components/Form/Input/index.jsx
+++ b/packages/webapp/src/components/Form/Input/index.jsx
@@ -142,9 +142,17 @@ const Input = ({
       {isPassword &&
         !showError &&
         (showPassword ? (
-          <MdVisibility className={styles.visibilityIcon} onClick={setVisibility} />
+          <MdVisibility
+            aria-label="hide-password"
+            className={styles.visibilityIcon}
+            onClick={setVisibility}
+          />
         ) : (
-          <MdVisibilityOff className={styles.visibilityIcon} onClick={setVisibility} />
+          <MdVisibilityOff
+            aria-label="show-password"
+            className={styles.visibilityIcon}
+            onClick={setVisibility}
+          />
         ))}
       {unit && <div className={styles.unit}>{unit}</div>}
       {currency && (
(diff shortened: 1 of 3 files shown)


iperdomo (approved):

Duncan-Brain:
Things I am seeing:

1. Stepper clicking still works when disabled = true. 
2. Up/down arrows (not clicks) allow one to go beyond max and min visually --> clicking out returns input to set max/min -- maybe thats ok
3. Setting unit together with stepper may have the same issue as password with the overlap -- is there a use case for both?
4. Where did we land on if should we add prop exclusions for the developers sake (ie if both password and stepper are set there is an error)
5. If we add #4 is it worth a storybook test?
6. z-index on password visibility -- it seems strange to me that we need to change password styling to accomodate stepper I think I would prefer we exclude the possibility of stepper && password over z-index (but I am also pretty suspicious of all z-index : 
https://www.codegram.com/blog/stop-ab-using-z-index/)

kathyavini:
Thank you @Duncan-Brain for the thorough investigation, I appreciate it!

We discussed this is tech daily, but let me cover each the points so we have a record of what we decided:

1. Thank you!!! Fixed and Storybook play test added for that case
2. The keyboard up and down arrow + `onBlur()` reset is the how the other number `<Input />`s behave, so I think it will be safe to keep it, even as the arrow buttons enforce a hard stop. For the use case of this view which will have no validation error messages), I think we decided this hybrid behaviour is going to be ok, but we may want to revisit validation, min, and max in the future as we flesh out this component in different contexts.
3. No use case yet (for unit + stepper) so for the time being I won't enforce but will let the developer kindly _not_ set those two props together on their component! It is an existing "feature" of Input that you set props that will render over each other (e.g. password and unit).
4. Those two can't render together and although I swear we saw this as a Storybook artifact at one point, I can't seem to re-create it now, even in Storybook. This doesn't affect the above, but I did remember to add `stepper` to the propTypes this time, which will make sure it renders under the Storybook controls for all variants.
5. See above
6. I tried a few other ways to fix this rendering layer issue, but re-ordering isn't so feasible for this component and I couldn't z-index specifically the wrapper after all, which I would have prefered. Agreed in principle; for instance I found through further testing that the error cross is also affected and needed the same fix. I did start using `z-index: 1` though as a nod to the article's point about random and unnecessarily large values 🙂

Duncan-Brain (approved):
Thanks for looking into those! Looks good!

iperdomo (approved):

PR #2755, opened 2023-07-06
Lf 3317 the type of field work wont be stored in the drop down if anyone has already created the same type of field work earlier

Duncan-Brain:
### **WARNING: This PR runs an irreversible migration**

**Description**

For custom field work task. The first farmer to come up with the unique custom name wins ownership in the field_work_type database. Other farmers may use that other farmer's custom type but will not see the option available in their farm's field work task option list. 

This PR allows the creation of duplicate custom task names between farms while maintaining uniqueness per farm.

The migration also creates new task types for farms -- so users will now see their created custom types.

Jira link: [LF-3317](https://lite-farm.atlassian.net/browse/LF-3317)

**Type of change**

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain):

This SQL query may be run to assess your db exposure to this issue (this will not fix the issue):
```
select
	location_tasks.task_id,
	field_work_task.field_work_type_id,
	location.farm_id,
	field_work_type.farm_id,
	field_work_type.field_work_name,
	field_work_type.field_work_type_translation_key
from location
join location_tasks ON location.location_id = location_tasks.location_id
join task ON location_tasks.task_id = task.task_id
join field_work_task ON task.task_id = field_work_task.task_id
join field_work_type ON field_work_task.field_work_type_id = field_work_type.field_work_type_id
where field_work_type.farm_id IS NOT NULL
AND location.farm_id != field_work_type.farm_id
```

You may test repeatedly by running migration, then manually changing field_work_task_type_id in field_work_task back to original type number and manually deleting the associated new type in the field_work_type table

**Checklist:**

[LF-3317]: https://lite-farm.atlassian.net/browse/LF-3317?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Duncan-Brain:
One thing I was unsure about in my migration was the use of promises. Typically the migration shares that it likes to return a promise. So I may be missing something there.

iperdomo:
I'm wondering if this is a _data_ issue. Why a migration? 🤔 e.g. last time we had data issue was done directly in the database. Since is not possible to rollback?

Duncan-Brain:
@iperdomo 

I think our practice in the past has been:  "If the problem is just on beta no migration is necessary"

It seems sometimes also: if it is an easy fix like: Turkey -> Turkiye or value 3 -> 4.

But this is a bit more complex of a query since I am using the location of the task get a farm_id used to compare the farm id's. I _think_ this will facilitate testing before release too so we can take a dump of the db and run the migration. 

I think this is a good way to document the fix as well since it is not simple. 

Open to ideas.

iperdomo:
> I think our practice in the past has been: "If the problem is just on beta no migration is necessary"

We have executed _data fix_ in production too, e.g. https://lite-farm.atlassian.net/browse/LF-3307

Duncan-Brain:
@iperdomo I think that falls into the or...

> "if it is an easy fix like: Turkey -> Turkiye or value 3 -> 4."

kathyavini:
@Duncan-Brain personally I love having this kind of thing in JavaScript, not SQL (I think this was one of our take-away lessons from the sensor work), so I'm happy with a migration file for that reason.

So two errors, one of which is minor and the other I'll just put forward because I'm not sure the best way to address:

1) Minor one: please update your migration directory to the state of integration because running `migrate:dev:db` on you branch gives this error, which makes testing a little annoying: "The migration directory is corrupt, the following files are missing: 20230609181254_add_country_code_column_to_coutries_table.js"

2) Not sure about this one... I get the following error when I run your migration:

```
migration file "20230630185257_backfill_incorrect_field_work_task_types.js" failed
migration failed with error: insert into "field_work_type" ("farm_id", "field_work_name", "field_work_type_translation_key") values ($1, $2, $3) returning "field_work_type_id" - duplicate key value violates unique constraint "field_work_type_field_work_name_field_work_type_translation_key"
```

The reason is that in setting up my database to test your migration, I used the same Field Work Task Type in Farm 2 in two ways -- once exactly shown in Denis's video, and once the same way except I associated it with a management plan. For whatever reason, in the second case (association with a management plan) the field_work_type is created correctly and associated with a farm correctly.

Is this enough of an edge case we should just run the migration on the db dump and only worry about it in the off chance that it fails like this?

kathyavini (approved):
Ack, I'm going to blame my sick brain from yesterday morning on this one 🙂 The database entry was not from association with a management plan, but from doing my testing on your branch _before_ running the migration. It's good! It's safe!

Duncan-Brain:
Thanks ! 1. Integration is merged in now for future reviewers. 2. As you said: could not replicate so will hold off on pushing changes to handle that situation. It doesn't seem possible for the time being that that situation could have occurred (unless manual migrations were done). If @iperdomo has access to read prod data and thinks its valuable to check the field_work_task_type for duplicates we could do that too to assess further.

iperdomo (approved):

PR #2761, opened 2023-07-13
Fix timezone handling in isNotInFuture util function

Cali0707:
**Description**

Jira link: https://lite-farm.atlassian.net/browse/LF-3393

The problem seems to have been with the default behaviour of `new Date(string)` and `Date.parse(string)`, which parses dates as if they were in UTC timezone if there is no timezone in the date string. This causes an issue when compared with `new Date()` which creates a date in your local timezone.

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/Form/Input/utils.js
+++ b/packages/webapp/src/components/Form/Input/utils.js
@@ -11,7 +11,7 @@ const parseISOStringToLocalDate = (dateString) => {
   }
   // this string does not have a specified offset, so we should treat it as if it is in local time
   // note: by default, new Date(string) and Date.parse(String) WILL parse in UTC
-  const split = dateString.replace(/[zZ]$/, '').split(/\D/);
+  const split = dateString.split(/\D/);
   return new Date(split[0], split[1] - 1, ...split.slice(2));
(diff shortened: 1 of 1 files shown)


Cali0707:
Note that when I worked on this, there wasn't a good timezone to check the date boundary with UTC as it had just became a new day in UTC. So, I did investigate the dates being used on the crop management completion plan page and found that they were being parsed into UTC instead of local time. 

I have verified that my fix seems to parse dates correctly into the right local timezone value, but it would be worth testing this change at a time when the UTC date boundary is nicer. Also, it might be worth unit testing the `parseISOStringToLocalDate` function for some more edge cases, but I didn't have time.

kathyavini (approved):
This is actually _exactly_ the utility function I needed for the Repeat Crop Plan view ([LF-3362](https://lite-farm.atlassian.net/browse/LF-3362))! I exposed it as an export and have been testing it in that form to get dates and weekdays for `rrule` from the date `<Input />`. It works great!! Thank you! 👍

PR #2769, opened 2023-07-17
LF-3362 Repeat crop plan | Details of the repeating series

kathyavini:
**Description**

This is the pure form component of the Repeat Crop Plan flow.

Jira link: https://lite-farm.atlassian.net/browse/LF-3362

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

Tested in Storybook & in app with Repeat Crop Plan container (LF-3431) -- draft PR has been opened with that container.

In Storybook under Form > RepeatCropPlan > PureRepeatCropPlan -- no play tests as of yet.

**Checklist:**

@SayakaOno there is a new vite warning about dynamic import on the rrule translation files I will message you about on Slack!


packages/webapp/src/components/RepeatCropPlan/index.jsx
```diff
@@ -0,0 +1,378 @@
...
+    >
+      <MultiStepPageTitle
+        onGoBack={onGoBack}
+        onCancel={historyCancel}
+        cancelModalTitle={t('REPEAT_PLAN.REPEAT_PLAN_FLOW')}
+        title={t('REPEAT_PLAN.TITLE')}
+        value={33}
+        style={{ marginBottom: '24px' }}
+      />
+      <Main>{[t('REPEAT_PLAN.SUBTITLE')]}</Main>
```
SayakaOno, line 208: This doesn't have to be an array...?

packages/webapp/src/components/RepeatCropPlan/index.jsx
```diff
@@ -0,0 +1,378 @@
...
+  const intervalOptions = [
+    { value: 'day', label: t('REPEAT_PLAN.INTERVAL.DAY') },
+    { value: 'week', label: t('REPEAT_PLAN.INTERVAL.WEEK') },
+    { value: 'month', label: t('REPEAT_PLAN.INTERVAL.MONTH') },
+    { value: 'year', label: t('REPEAT_PLAN.INTERVAL.YEAR') },
+  ];
+
+  const [monthlyOptions, setMonthlyOptions] = useState([]);
+
+  const totalCount = useRef();
```
SayakaOno, line 76: Why is this ref? I'll just keep that in mind when I work on my ticket.

packages/webapp/src/components/RepeatCropPlan/index.jsx
```diff
@@ -0,0 +1,378 @@
...
+    });
+
+    totalCount.current = count;
+
+    trigger('finish_on_date');
+  }, [planStartDate, repeatFrequency, repeatInterval, finishOnDate, monthRepeatOn]);
+
+  // Custom validation for "finish on" date input
+  const validateFinishOnDate = (value) => {
+    if (getValues('finish') === 'after') {
```
SayakaOno, line 142: Is this how you get the latest value?
`const finish = watch('finish');` This can't be used??

packages/webapp/src/components/RepeatCropPlan/utils.js
```diff
@@ -0,0 +1,187 @@
...
+      label: dateString,
+    },
+    {
+      value: { ordinal, weekday },
+      label: dayWeekString,
+    },
+  ];
+};
+
+export const countOccurrences = ({
```
SayakaOno, line 111: Do you think you can return the array of occurrences here and do `setValue()` in the component so that I can use it in my component? (I plan to forward it to the backend)
I might have a few more requests like this, but I'm not sure if I should just make some updates after this PR is merged...

SayakaOno (commented):
Great job on this tough task!! I still need to come back and review the component itself but will submit review for today...

Here are easy ones that I noticed.
- I'm sorry I didn't spot this when I reviewed `DaysOfWeekSelect`, but the cursor becomes `default` when the mouse is on the labels.
- We need to add the license to the rruleTranslation files, sorry I didn't do it in my POC.

kathyavini:
Thank you for the careful read @SayakaOno!! 🙏 

I was trying to understand and finish the container yesterday so I will be working on those code changes today -- but I'll try to start with answering the questions that look like they impact on your component.

packages/webapp/src/components/RepeatCropPlan/utils.js
kathyavini: Definitely! But I only generate an array of occurrences for the finish on date situation right now (not for the "finish after _x_ repetitions") and I tie it specifically to the validation of the date picker, so there would have to be some changes to that timing 🤔 

I am thinking then running the count onSubmit would be best then? Since beyond validation there is no need for it to run onChange. So the count function could be generalized and called in two situations -- the useEffect and onSubmit.

And YES you are so very welcome to adjust the heck out of it for your own uses and just change whatever you like for the hookup to your component!! That might be easier.

packages/webapp/src/components/RepeatCropPlan/index.jsx
kathyavini: I don't think it will need to be a ref in your view -- I needed the value to not be tied to the render cycle to clear the validation on the "finish on" date component immediately upon interaction with the other inputs; when I kept the count in state the validation function wasn't getting the new value in time.

But it was only for this live feedback on the max count validation; otherwise I don't think it would be necessary.

packages/webapp/src/components/RepeatCropPlan/index.jsx
```diff
@@ -0,0 +1,378 @@
...
+
+  const totalCount = useRef();
+
+  const planStartDate = watch('plan_start_date');
+  const repeatFrequency = watch('repeat_frequency');
+  const repeatInterval = watch('repeat_interval');
+  const daysOfWeek = watch('days_of_week');
+  const monthRepeatOn = watch('month_repeat_on');
+  const finish = watch('finish');
+  const finishOnDate = watch('finish_on_date');
```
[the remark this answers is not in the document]
kathyavini: Sure! I'm going to ping you on Slack about this though because I see a lot of different patterns for form constants in our codebase...

packages/webapp/src/components/RepeatCropPlan/utils.js
```diff
@@ -0,0 +1,187 @@
...
+  return date.toLocaleString('en', { weekday: 'long' });
+};
+
+export const getDate = (planStartDate) => {
+  const date = new Date(parseISOStringToLocalDate(planStartDate));
+
+  return date.getDate();
+};
+
+const calculateWeekdayOrdinal = (date) => {
```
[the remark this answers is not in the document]
kathyavini: I'm not sure what would be the expected behaviour. If the 4th Monday is also the last, it looks like Google Calendar actually lets the user pick which rule to follow, which makes me think neither is the default:

Screenshot 2023-07-17 at 3 41 19 PM

We could have only the 5th week default to "last", but I think that sometimes people _do_ want to do something on e.g. the last Friday regardless of 4th/5th week!

Do you think maybe we have to offer both options??

packages/webapp/src/util/rruleTranslation/fr.js
```diff
@@ -0,0 +1,48 @@
...
+  times: 'fois',
+  time: 'temps',
+  minutes: 'minutes',
+  hours: 'heures',
+  weekdays: 'jours de la semaine',
+  weekday: 'jour de la semaine',
+  months: 'mois',
+  month: 'mois',
+  years: 'ans',
+  year: 'an',
```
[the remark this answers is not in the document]
kathyavini: I don't know! Did Calum respond? We should probably check Spanish with Iván sooner rather than later too.

packages/webapp/src/components/RepeatCropPlan/index.jsx
kathyavini: Correct. Right now the watch value can't be used for the same reason as above (the validation does not clear on interaction with the `finish` radio group if you use the watch value).

But this is only because of how I set up my repetition count useEffect to return when the "finish after _x_  repetitions" is selected... otherwise there is a `trigger('finish_on_date')` within that useEffect which would be sufficient, so depending on when/how the count useEffect is run, this could be changed!

packages/webapp/src/util/rruleTranslation/fr.js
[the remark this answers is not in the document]
SayakaOno: Oh okay, I thought Chat GPT knew those were not necessary! I will wait for Calum to respond :)

packages/webapp/src/components/RepeatCropPlan/index.jsx
SayakaOno: I understand, thank you for the explanation!

packages/webapp/src/components/RepeatCropPlan/utils.js
SayakaOno: Thank you Joyce! I wouldn't have felt good about modifying your code, but since you are good with that, you can leave the function as is! I will make adjustments later and ask you for a review.

I realized that my request was not as easy as I thought, so I will take what you mentioned into account and make a decision when I work on my ticket 🙏 🙏 🙏

SayakaOno (commented):
Thank you for addressing my comments!
It would be great if you could look into the missing validation. (I left a comment!)

packages/webapp/src/components/RepeatCropPlan/index.jsx
```diff
@@ -0,0 +1,387 @@
...
+                    <Input
+                      type="date"
+                      hookFormRegister={register(FINISH_ON_DATE, {
+                        validate: validateFinishOnDate,
+                      })}
+                      onChange={() => setValue(FINISH, 'on')}
+                    />
+                  </div>
+                ),
+              },
```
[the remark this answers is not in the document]
kathyavini: Ahh, it's because of the first interaction (`dirtyFields`) requirement; I didn't like the UX of getting an error message as soon as you pick the radio button (and before you have had a chance to interact with the date picker), but of course yes, you could submit at that point so that check needs to go!

But sorry I don't understand what you mean about the error message; if I remove the `dirtyFields` check, this is the UI as soon as you pick the "finish === 'on'". It's the same as when you clear the date picker input now. Do you not see that error message?

Screenshot 2023-07-19 at 12 05 17 PM

packages/webapp/src/util/rruleTranslation/fr.js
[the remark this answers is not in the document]
SayakaOno: Update:
Calum did some research for us:

> The main issue is that "1st" maps to "premiere", but  "21st" maps to "vingt et unième", and similarly "31st" maps to "trente et unième". You can see a good list of these here: https://francais.lingolia.com/fr/vocabulaire/nombres-date-et-heure/les-nombres-ordinaux

> So I think this will work, you just might have to work on some edge cases depending on which options you aim to support in the UI

> Although, you may also notice that with the exception of 1st (aka 1er/1re), the rest of the numbers are all of the form <digits>e, where the e is raised (see the table on the link above). I'm not sure how to type/render that into html though, but if you could get that to work then you should be good to go!

> But, from [this source](https://www.btb.termiumplus.gc.ca/tpv2guides/guides/clefsfp/index-fra.html?lang=fra&lettr=indx_catlog_n&page=91agpPDdxblQ.html), you should avoid just adding -ième, -ème, or -ère to the end of the number

The last solution seems good, so I'll set "st", "nd", "rd", "th" to an empty string!

packages/webapp/src/components/RepeatCropPlan/index.jsx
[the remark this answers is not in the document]
SayakaOno: > I didn't like the UX of getting an error message as soon as you pick the radio button
I agree...

I had not tried it by removing the dirtyFields check, but I confirmed that the error message would be shown! So your solution is good :)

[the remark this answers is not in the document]
kathyavini: Fixed! Thank you for finding that!! ❤️

SayakaOno (approved):
Looks good to me! Great job!!

iperdomo (approved):

PR #2772, opened 2023-07-18
LF-3308 Assign all tasks to one assignee pre-populate only user 

Tbrid:
**Description**

Missed the functionality when there is only 1 user, then pre-populate that user in input when assigning all tasks

Jira link: https://lite-farm.atlassian.net/browse/LF-3308

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/Crop/ManagementPlanName/index.jsx
+++ b/packages/webapp/src/components/Crop/ManagementPlanName/index.jsx
@@ -48,7 +48,6 @@ export default function PureManagementPlanName({
     shouldUnregister: false,
     user: user,
     users: users,
-    isAssigned: false,
     defaultAssignee:
       users.length === 1
         ? { label: `${user.first_name} ${user.last_name}`, value: user.user_id, isDisabled: false }
--- a/packages/webapp/src/components/Crop/ManagementPlanName/index.jsx
+++ b/packages/webapp/src/components/Crop/ManagementPlanName/index.jsx
@@ -4,7 +4,6 @@ import PropTypes from 'prop-types';
 import { useTranslation } from 'react-i18next';
 import Input, { getInputErrors } from '../../Form/Input';
 import Form from '../../Form';
-import { useForm } from 'react-hook-form';
 import { useSelector } from 'react-redux';
 import { userFarmsByFarmSelector, userFarmSelector } from '../../../containers/userFarmSlice';
 import MultiStepPageTitle from '../../PageTitle/MultiStepPageTitle';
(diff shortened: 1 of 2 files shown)


SayakaOno:
@Tbrid
We merged PRs, and the PR got conflicts... could you please resolve them?  Thank you!! 🙏

packages/webapp/src/components/Crop/ManagementPlanName/index.jsx
```diff
@@ -47,9 +49,13 @@ export default function PureManagementPlanName({
...
     users: users,
-    defaultAssignee: { label: t('TASK.UNASSIGNED'), value: null, isDisabled: false },
-    disableUnAssignedOption: false,
+    isAssigned: false,
+    defaultAssignee: unassigned,
     additionalFields: {
+      [ASSIGNEE]:
+        users.length === 1
+          ? { label: user.first_name, value: user.user_id, isDisabled: false }
+          : unassigned,
```
SayakaOno, line 58: Can we have these as `defaultAssignee`(line 53)? I think it will be the same thing...
and user's `last_name` is missing: `${user.first_name} ${user.last_name}`

packages/webapp/src/components/Task/AssignTask/useTaskAssignForm.jsx
```diff
@@ -109,7 +111,8 @@ const useTaskAssignForm = ({
   const hourlyWage = watch(HOURLY_WAGE);
 
   const disabled = !selectedWorker || !isValid;
-  const assigned = selectedWorker.label !== unAssignedOption.label;
+  const assigned =
+    isAssigned === undefined ? selectedWorker.label !== unAssignedOption.label : isAssigned;
```
SayakaOno, line 115: Does `assigned` need to be false when a user is selected by default??

packages/webapp/src/components/Crop/ManagementPlanName/index.jsx
Tbrid: Added. Yes, don't think that was necessary.

packages/webapp/src/components/Task/AssignTask/useTaskAssignForm.jsx
Tbrid: The thing here is: when I pre-fill the input with the only user then assigned becomes true, just think this is not the right way.

SayakaOno: Oh sorry, I guess I should have given the variable a better name. `assigned` actually means "user is being selected", so when there is a default assignee, we still want `assigned` to be true.

JimmyRowland: So isAssigned can be removed and the original line should just work, right?

SayakaOno: Yes, that's correct!

JimmyRowland (approved):

SayakaOno (approved):

PR #2776, opened 2023-07-18
LF-3385: Notification is not created once the user creates transplant planting or harvest task

SayakaOno:
**Description**

Send a notification to the assignee of transplant/harvest task or management plan on its creation. (notification had not been implemented)

Jira link: https://lite-farm.atlassian.net/browse/LF-3385

**Type of change**

**How Has This Been Tested?**

Create a transplant/harvest task or a management plan and assign it. The assignee should see a badge on the notification icon.

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/managementPlanController.js
+++ b/packages/api/src/controllers/managementPlanController.js
@@ -59,11 +59,12 @@ const managementPlanController = {
           if (!req.body.crop_management_plan.already_in_ground) {
             const due_date =
               req.body.crop_management_plan.plant_date || req.body.crop_management_plan.seed_date;
-            const { planting_management_plan_id } =
-              management_plan.crop_management_plan.planting_management_plans.find(
-                (planting_management_plan) =>
-                  planting_management_plan.planting_task_type === 'PLANT_TASK',
-              );
+            const {
+              planting_management_plan_id,
+            } = management_plan.crop_management_plan.planting_management_plans.find(
+              (planting_management_plan) =>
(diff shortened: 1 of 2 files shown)


packages/api/src/controllers/managementPlanController.js
```diff
@@ -173,6 +174,25 @@ const managementPlanController = {
             tasks.push(fieldWorkTask);
           }
 
+          if (req.body.assignee_user_id) {
+            await Promise.all(
+              tasks.map(async (task) => {
```
iperdomo, line 179: Why `.map()` is used here? Would it be better to use `.forEach()`?
My understanding is that you want to use [.map()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map) when you want to modify a collection and want to hold to the modified values. It seems that this time we're only interested in the _side-effect_ of `sendTaskNotification`? [.forEach()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach) is better suited for this?

iperdomo (changes requested):
Let's check if `forEach` is better suited in this case

packages/api/src/controllers/managementPlanController.js
SayakaOno: Thank you, I was trying to use `Promise.all()` so that notifications would be sent at once and realized that notification was actually being sent by the side-effect... I will fix it!

iperdomo (approved):

PR #2784, opened 2023-07-25
Fixing issue where 403 caused infinite loop

kikemarquez8:
This was caused because the first condition to check the 403 error was not met so the "else" branch just went home, which triggered another "getAlerts" request which returned 403 which went home and so on.

**Description**

Jira link:

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/webapp/src/containers/ErrorHandler/Forbidden/Forbidden.jsx
```diff
@@ -4,15 +4,18 @@ import { useDispatch } from 'react-redux';
...
-  }
- }, []);
- return <Spinner />;
+  const dispatch = useDispatch();
+  useEffect(() => {
+    if (
+      history?.location?.state?.error?.response?.status === 403 ||
+      history?.location?.state?.error?.message?.includes('403')
+    ) {
+      dispatch(handle403());
```
JimmyRowland, line 13: image
I think handle403 need to check userFarm status === 'Active'
This component should probably be re-evaluated. Failure case shouldn't have caused infinite loops

kikemarquez8: I agree, the component itself never renders because going into this state would get you logged out. Which would also get you to the login page. 
Do you think it's better if we do a bigger review of this before merging?

JimmyRowland: Does this work?
That Forbidden component was a horrible mistake :1st_place_medal: 
https://github.com/LiteFarmOrg/LiteFarm/pull/2790
https://lite-farm-7lsmxoqps-jimmyrowland.vercel.app/

kikemarquez8: I completely missed this @JimmyRowland. This works fine!

JimmyRowland (approved):

PR #2803, opened 2023-07-31
LF-3489 On mobile devices decimal values can be input for crop plan repetitions, creating scary output on confirmation screen

kathyavini:
**Description**

Putting a decimal input into the repeat crop plan `repetitions` Input (which can be done *only on mobile* due to the existing and known bug that `integerOnKeyDown` doesn't function on Chrome Android) causes a crazy number of repetitions to be generated by RRule.

There are probably several ways to fix this, including flooring/sanitizing the value when it's used in the calculation functions, but I have taken the basic approach of flooring the Input value `onBlur` for the two Number with Stepper inputs in this form.

I think the same approach could probably be used for the more general case of decimals on Chrome Android, but that is a separate ticket so should be addressed if/when it comes up in the release queue 🙂 

Jira link: https://lite-farm.atlassian.net/browse/LF-3489

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/RepeatCropPlan/index.jsx
+++ b/packages/webapp/src/components/RepeatCropPlan/index.jsx
@@ -266,6 +266,7 @@ export default function PureRepeatCropPlan({
               })}
               type="number"
               stepper
+              onBlur={(e) => (e.target.value = Math.floor(e.target.value))}
               onKeyDown={integerOnKeyDown}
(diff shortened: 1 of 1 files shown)


SayakaOno (approved):
Works great!

PR #2783, opened 2023-07-25
[LF-3001] Adding Harvest to date field for management plans.

kikemarquez8:
Also renaming the i18next parser config which was broken.

All of the changes on the translation files are automatic movements and additions done by i18n-parser. It seems we hadn't run that for a month or 2

**Description**

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Jira link:

**Type of change**

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

**How Has This Been Tested?**

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [x] Passes test case
- [ ] UI components visually reviewed on desktop view
- [ ] UI components visually reviewed on mobile view
- [x] Other (Added unit test for the backend controller affected)

**Checklist:**

- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] The precommit and linting ran successfully
- [x] I have added or updated language tags for text that's part of the UI
- [x] I have added "MISSING" for all new language tags to languages I don't speak
- [ ] I have added [the GNU General Public License](https://lite-farm.atlassian.net/l/cp/BT0Dd7WW) to all new files

PR #2809, opened 2023-08-01
[LF-3001] Adding harvest to date to date field for management plans

kikemarquez8:
**Description**

Refert  to LF-3001 prev PR https://github.com/LiteFarmOrg/LiteFarm/pull/2783
**Type of change**

**How Has This Been Tested?**

- [x] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/managementPlanController.js
+++ b/packages/api/src/controllers/managementPlanController.js
@@ -25,6 +25,7 @@ import PlantTaskModel from '../models/plantTaskModel.js';
 import UserFarmModel from '../models/userFarmModel.js';
 import objection, { raw } from 'objection';
 import _pick from 'lodash/pick.js';
+import knex from '../util/knex.js';
 import { sendTaskNotification, TaskNotificationTypes } from './taskController.js';
 import {
   getDatesFromManagementPlanGraph,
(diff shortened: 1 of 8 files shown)


Duncan-Brain (changes requested):
@kikemarquez8 Jest tests are passing.

But when I complete a harvest task on frontend I am not seeing the data reflected on the management plan detail page -- just '0'.  I am thinking it may need some editing of the slice or something to accommodate the new value.

I think the unit does also need to be chosen. Maybe @SayakaOno can correct me if I am wrong but I think the getDefaultUnit function in util might be the best way to do that for this. Since it could be mixed units like kg/mt.

Translation files for other languages should be set to MISSING.

kikemarquez8:
> @kikemarquez8 Jest tests are passing.
> 
> But when I complete a harvest task on frontend I am not seeing the data reflected on the management plan detail page -- just '0'. I am thinking it may need some editing of the slice or something to accommodate the new value.
> 
> I think the unit does also need to be chosen. Maybe @SayakaOno can correct me if I am wrong but I think the getDefaultUnit function in util might be the best way to do that for this. Since it could be mixed units like kg/mt.
> 
> Translation files for other languages should be set to MISSING.
Hmm I will take a look at that. 

In terms of the units, I had a discussion with David about it, I think we just want to use the default display guideline, from what I saw in the Unit component just passing null as the unit and the correct system yields exactly that, but I could be wrong.

Duncan-Brain (approved):
Looks good to me! Tested with different magnitude units and swapping systems so I think you were right about the unit.

PR #2810, opened 2023-08-01
LF-3486: Going < from repeat confirmation does not maintain custom day of the week

SayakaOno:
**Description**

- call `setPersistedPaths` in RepeatCropPlan component so that the form data will persist when coming back from the confirmation view.
- remove `setPersistedPaths` in ManagementPlanTask since it would happen twice because of the change above.
- check previous state of start date and repeat interval in RepeatCropPlan's useEffect so that "weekday" will not be reset when coming back from the confirmation view.

Jira link: https://lite-farm.atlassian.net/browse/LF-3486

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/Crop/ManagementDetail/ManagementPlanTasks.jsx
+++ b/packages/webapp/src/components/Crop/ManagementDetail/ManagementPlanTasks.jsx
@@ -8,8 +8,6 @@ import PropTypes from 'prop-types';
 import styles from './styles.module.scss';
 import IncompleteTaskModal from '../../Modals/IncompleteTaskModal';
 import RouterTab from '../../RouterTab';
-import { useDispatch } from 'react-redux';
-import { setPersistedPaths } from '../../../containers/hooks/useHookFormPersist/hookFormPersistSlice';
 import DeleteBox from '../../Task/TaskReadOnly/DeleteBox';
 import { FiAlertTriangle } from 'react-icons/fi';
 import { ReactComponent as TrashIcon } from '../../../assets/images/document/trash.svg';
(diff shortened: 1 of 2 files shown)


kathyavini:
@SayakaOno could we request Jimmy's review on this one you think? As he just fixed some persisted paths stuff in the other ticket

kathyavini:
I give it a big thumbs up in terms of behaviour though. It works great! 👍

JimmyRowland:
Does this work?
Persisted paths was not added to the tempStateReducer.
I have read that useEffect should be avoided when onClick handler can do the same.

https://lite-farm-git-3486-jimmyrowland.vercel.app/?vercelToolbarCode=GSJTFFfs2IthMqZ
https://github.com/LiteFarmOrg/LiteFarm/tree/LF-3486/Going_back_from_repeat_confirmation_does_not_maintain_custom_day_of_the_week

SayakaOno:
@JimmyRowland 
I'll test it in your branch later, thank you!
When I called `setPersistedPaths` in the container, paths got removed right after it was added, so I delayed it by moving `setPersistedPaths` to the component, and it was working. Your change to `useHookFormPersist` enabled paths to be set in the container??

JimmyRowland (approved):

JimmyRowland:
> When I called setPersistedPaths in the container, paths got removed right after it was added, so I delayed it by moving setPersistedPaths to the component, and it was working.
This might be caused by resetFormState in `useHookFormPersist` useLayoutEffect hook.
In the previous PR, I moved resetFormState before clearing history stack, maybe that is the reason why setPersistedPaths in container is working now.

PR #2815, opened 2023-08-02
LF-3486

JimmyRowland:
add persisted path to repeatCropPlan container.
Replace day of week useEffect reset hook with onClick callbacks to avoid unintended resets


kathyavini (approved):
Awesome, thank you @JimmyRowland!

PR #2838, opened 2023-08-04
LF-3468: Send notifications of the deleted tasks on crop plan delete

MMnemonic:
**Description**

Problem: Users were supposed to receive a notification when a task that they were assigned to is deleted. When a plan crop is deleted, all of the crop tasks are also deleted and the users assigned to it do not receive notifications.

The solution that was implemented: Use the function taskController.sendTaskNotification() to send the notifications of the deleted tasks in the managementPlanController.delManagementPlan() function.

Jira link: https://lite-farm.atlassian.net/browse/LF-3468

**Type of change**

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/api/src/controllers/managementPlanController.js
```diff
@@ -396,6 +398,30 @@ const managementPlanController = {
...
             .whereIn('task_id', taskIdsRelatedToOneManagementPlan)
             .delete();
+
+          await Promise.all(
+            taskIdsRelatedToOneManagementPlan.map(async (task_id) => {
+              const { task_translation_key } = await TaskModel.getTaskType(task_id);
+              const { assignee_user_id } = await TaskModel.query(trx)
+                .select('assignee_user_id')
+                .where({ task_id })
+                .first();
```
guilhermedeborba, line 408:
```suggestion
const [{ task_translation_key }, { assignee_user_id }] = await Promise.all([
  TaskModel.getTaskType(task_id),
  TaskModel.query(trx).select('assignee_user_id').where({ task_id }).first()
]);

```

guilhermedeborba: might be faster running in parallel

Duncan-Brain (changes requested):
@MMnemonic Thanks for your patience sorry it was super busy the last few weeks!

It seems to be working well for me when I tried!

My main concern is the ordering. In the off chance the management plan deletion fails and the transaction rolls back, then the notifications may have already sent. I think it might be best to move those to the end.

javiscript13:
> @MMnemonic Thanks for your patience sorry it was super busy the last few weeks!
> 
> It seems to be working well for me when I tried!
> 
> My main concern is the ordering. In the off chance the management plan deletion fails and the transaction rolls back, then the notifications may have already sent. I think it might be best to move those to the end.

@Duncan-Brain, I did keep the original logic and changed the order, please review.

packages/api/src/controllers/managementPlanController.js
```diff
@@ -434,13 +436,29 @@ const managementPlanController = {
...
+
+          if (delPlan) {
+            return deletedTasks;
+          } else {
+            return delPlan;
+          }
         });
 
         if (result) {
+          result.map(async (task) => {
```
Duncan-Brain, line 453: Does this need to be in a Promise.all?

Duncan-Brain (commented):
Thanks @javiscript13 for picking up this ticket! Just left one question about a Promise.all maybe being used? I am not confident

javiscript13:
> Thanks @javiscript13 for picking up this ticket! Just left one question about a Promise.all maybe being used? I am not confident

Thank you @Duncan-Brain ,  I wasn't sure about it, so I investigated a little bit about it, and ended up adding it, and also the async error handling for the notifications, so if there is any error on the notifications, it will not affect the snack bar of the plan being deleted.

Duncan-Brain (approved):
Looks great thank you!

packages/api/src/controllers/managementPlanController.js
```diff
@@ -434,13 +436,29 @@ const managementPlanController = {
...
+          const delPlan = await ManagementPlanModel.query(trx)
             .context(req.auth)
             .where({ management_plan_id })
             .delete();
+
+          if (delPlan) {
+            return deletedTasks;
+          } else {
+            return delPlan;
+          }
```
SayakaOno, line 449: @javiscript13 Thank you for taking over this task!
I don't think this will make any difference on the API itself. (What is returned here will be used on line 451 `if (result)`) What are these changes for??

javiscript13: Hey @SayakaOno , thank you for reviewin my pull request!.

I'm not sure if I understand the question, but, on the quoted code, the `deletedTasks` array will be passed as a result to send the notifications if everything went ok (so `result` will have all the needed information to send notifications).  If there is an issue on the `ManagementPlanModel.query(trx)` that issue will be passed and status 404 will be sent by the api (as original).  

I did'nt find on knex documentation if a query fails does it sends a faulty value or a pure false, so I decided to pass that value.  

Do you think I should simplify the quoted code to return `delPlan?deletedTasks:false;` ?

SayakaOno: Is there a case `await ManagementPlanModel.query(trx)` succeeds, but `deletedTasks` becomes faulty? Is that what you were trying to catch?
When `delPlan` is truthy, I thought `result` will always be truthy too, so I asked the question! I might have missed something though...

javiscript13: @SayakaOno,  the changes I'm proposing are doing the following:

I'm creating an array called `deletedTasks`, as there was a procedure to identify which tasks are related to the managentPlan that is going to be deleted, and as in this procedure all the information needed to fill the notifications is available (on the original PR this place was used to send the notifications), but as Duncan suggested if there was an error on the deletion the notifications would be already sent, .. so I'm just using this place to fill objects with the needed data to send notifications and pushing each object on the `deletedTasks` array, so the notifications are sent later when the transaction is confirmed to be ok.

Later in the code (line 440) is the query that deletes the management plan, and is the last procedure on the transaction.  So, if this query executes ok, I'm sending the `deletedTasks` as a result of the whole transaction if the deletion went ok it will send the notifications based on the gathered data (line 452) (on `deletedTasks`, and then in the other context as `result`).

The promise.all where the notifications are sent (line 453), has its own catch, so if there is an error sending the notifications it will only log the error, and will not affect the outer catch because that would inform the user that the managementPlan could not be deleted, but would be a mistake beacuse the managementePlan would be deleted but the error would be on the notification.

SayakaOno: Ahhhhhhhh! I'm so sorry, for some reason I kept missing `result.map`(line 454). I didn't notice `result` was being used there.
Let me buy you lunch 😭

SayakaOno (approved):

PR #2841, opened 2023-08-17
LF-3552: Create reusable tile view component

SayakaOno:
**Description**

- Create `LabelIconTile` and `Tiles` component that can be used for:
  - task type view (refactor [ticket](https://lite-farm.atlassian.net/browse/LF-3567))
  - expense type view
  - revenue type view
- [IconLabelTile stories](http://localhost:6006/?path=/docs/components-tile-iconlabeltile--docs)
- [Tiles stories and tests](http://localhost:6006/?path=/docs/components-tile-tiles--docs)
- (the scss file was copied from `packages/webapp/src/components/Task/PureTaskTypeSelection/styles.module.scss`)

Jira link: https://lite-farm.atlassian.net/browse/LF-3552

**Type of change**

**How Has This Been Tested?**

storybook tests (the link is in the description)

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/Tile/Tiles.jsx
+++ b/packages/webapp/src/components/Tile/Tiles.jsx
@@ -1,0 +1,55 @@
+/*
+ *  Copyright 2023 LiteFarm.org
+ *  This file is part of LiteFarm.
+ *
+ *  LiteFarm is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  LiteFarm is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ */
+import React, { useMemo } from 'react';
+import PropTypes from 'prop-types';
+import IconLabelTile from './IconLabelTile';
+import { tyleTypes } from './constants';
+import styles from './styles.module.scss';
+
+const tileComponents = {
+  [tyleTypes.ICON_LABEL]: (props) => <IconLabelTile {...props} />,
+};
+
+/**
+ * A component that places tiles so that the empty space is evenly distributed for any window sizes.
+ * Either "children" or "tileType" and "tileData" props are required.
+ * See packages/webapp/src/stories/Tile/Tiles.stories.jsx for examples.
+ */
+export default function Tiles({ children, tileType, tileData, formatTileData, ...props }) {
+  const tiles = useMemo(() => {
+    if (children || !tileType || !tileData) {
+      return children || null;
+    }
+
+    return tileData.map((data) => {
+      const tileProps = formatTileData ? formatTileData(data) : data;
+      return tileComponents[tileType](tileProps);
+    });
+  }, [children, tileType, tileData, formatTileData]);
+
+  return (
+    <div className={styles.matrixContainer} {...props}>
+      {tiles}
+    </div>
+  );
+}
+
+Tiles.propTypes = {
+  children: PropTypes.node,
+  tileType: PropTypes.oneOf(Object.keys(tileComponents)),
+  tileData: PropTypes.array,
+  /* formatTileData must return an object that has "key" */
+  formatTileData: PropTypes.func,
+};
--- a/packages/webapp/src/components/Tile/Tiles.jsx
+++ b/packages/webapp/src/components/Tile/Tiles.jsx
@@ -15,11 +15,11 @@
 import React, { useMemo } from 'react';
 import PropTypes from 'prop-types';
 import IconLabelTile from './IconLabelTile';
-import { tyleTypes } from './constants';
+import { tileTypes } from './constants';
 import styles from './styles.module.scss';
 
 const tileComponents = {
-  [tyleTypes.ICON_LABEL]: (props) => <IconLabelTile {...props} />,
+  [tileTypes.ICON_LABEL]: (props) => <IconLabelTile {...props} />,
 };
 
 /**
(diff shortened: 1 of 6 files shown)


packages/webapp/src/components/Tile/styles.module.scss
```diff
@@ -0,0 +1,137 @@
...
+}
+
+.taskTypeLabelContainer {
+  overflow: hidden;
+  text-overflow: ellipsis;
+  width: 100%;
+}
+
+.matrixContainer {
+  width: 100%;
```
SayakaOno, line 62: ⬆️ is the only line added to the copy of `/packages/webapp/src/components/Task/PureTaskTypeSelection/styles.module.scss`

packages/webapp/src/stories/Tile/Tiles.stories.jsx
```diff
@@ -0,0 +1,101 @@
...
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ */
+import React from 'react';
+import { expect } from '@storybook/jest';
+import { within } from '@storybook/testing-library';
+import { componentDecorators } from '../Pages/config/Decorators';
+import Tiles from '../../components/Tile/Tiles';
+import IconLabelTile from '../../components/Tile/IconLabelTile';
+import { tyleTypes } from '../../components/Tile/constants';
```
Duncan-Brain, line 21: Just a misspelling  - tyle -> tile

packages/webapp/src/components/Tile/Tiles.jsx
```diff
@@ -0,0 +1,55 @@
...
+  return (
+    <div className={styles.matrixContainer} {...props}>
+      {tiles}
+    </div>
+  );
+}
+
+Tiles.propTypes = {
+  children: PropTypes.node,
+  tileType: PropTypes.oneOf(Object.keys(tileComponents)),
```
Duncan-Brain, line 51: Is this correct? I am Getting a weird error when trying to change this value in the storybook page.

Screenshot 2023-08-18 at 3 40 49 PM
Screenshot 2023-08-18 at 3 41 50 PM

Duncan-Brain:
Hi @SayakaOno, I think everything seems good! I left a few inline comments.

I have one questions:
1. Why support both children and tileData?

packages/webapp/src/components/Tile/Tiles.jsx
```diff
@@ -0,0 +1,55 @@
...
+  [tyleTypes.ICON_LABEL]: (props) => <IconLabelTile {...props} />,
+};
+
+/**
+ * A component that places tiles so that the empty space is evenly distributed for any window sizes.
+ * Either "children" or "tileType" and "tileData" props are required.
+ * See packages/webapp/src/stories/Tile/Tiles.stories.jsx for examples.
+ */
+export default function Tiles({ children, tileType, tileData, formatTileData, ...props }) {
+  const tiles = useMemo(() => {
```
Duncan-Brain, line 31: Just wondering if you know when is best to use useMemo? I am still not 100% comfortable the best situations and would be interested in the decision here. 

I have only read this article : https://blog.bitsrc.io/when-using-usememo-is-a-really-bad-idea-a2bdeb909812

packages/webapp/src/stories/Tile/Tiles.stories.jsx
```diff
@@ -0,0 +1,101 @@
...
+          onClick={() => console.log('clicked!')}
+          selected={false}
+        />
+      );
+    }),
+  },
+  play: tileLengthTest(6, ['Label']),
+};
+
+export const IconLabeTilesAsData = {
```
kathyavini, line 60: Another small typo -- just missing the `l` in `Label` for each of these

kathyavini (commented):
Very nice demonstration with Storybook!!

I guess I had the same exact questions as @Duncan-Brain 😄 

1. So this has a very interesting and flexible API which I like! What was the rationale for supporting so many different input props (including with the format function)?
2. I was curious about useMemo as well... how did you decide the optimisation was needed?

Looks great, and I was just curious to learn about the rationale for those two 🙁

SayakaOno:
@Duncan-Brain @kathyavini 
Thank you for reviewing and catching typos!

1. Why does `Tiles` component support 3 different types of props?
    1. `children`
   This was my least favourite option. There is `CardWithStatusContainer` that places `ManagementPlanCard` and it takes `children`. When I used it, I had to create children using `ManagementPlanCard` and thought it would be more efficient if one component could do both aligning cards and rendering each card. (we know `CardWithStatusContainer` is used with `ManagementPlanCard`, then why will we need to import `ManagementPlanCard` every time?)
    2. `tileData`
    This is the cleanest, but the performance would be bad; you would need to format data in its parent component, and then iterate the data again to render each piece of data.
    3. `tileData` + format function
    This is to improve 2. Formatting and rendering will happen at once.

     I wasn't sure which option I should go with when I created the component, so I made it flexible. I am going to go with 1 or 3, but now I know option 1 is enough 😅 

2. I don't have a clear rule when to/not to use `useMemo`, so I need to learn more about it. But since Jimmy gave me this [comment](https://github.com/LiteFarmOrg/LiteFarm/pull/2449#discussion_r1096813900), I use it for components.
     > https://twitter.com/0xca0a
This guy always preaches not to declare components in the component body since components will always be recreated on each render.

Duncan-Brain (approved):

Duncan-Brain:
So what I got was :

You are planning on deleting one of the two options when you decide which way you prefer? -- at a later date

And

useMemo() is maybe beneficial in the situation where you declare a component in the constructor. That way you do not re-render those components on every render. I think that makes sense and relating to the two options I just found this on the useMemo() documentation. Thank you for explaining!

Screenshot 2023-08-21 at 3 14 27 PM

SayakaOno:
@Duncan-Brain 
Sorry I wasn't clear, I wasn't planning on updating the component. I am still not sure which option I want to go with in the future for a different component. But I would delete anything that shouldn't be there.

Thank you for sharing the screenshot! I didn't know the official document had such a lot of information.

> "let it [accept JSX as children](https://react.dev/learn/passing-props-to-a-component#passing-jsx-as-children)"
This made me realize that I passed `children` as `prop` not wrapping it in one of the stories... I need to fix it.

PR #2844, opened 2023-08-21
LF-3506/Wrong date in tasks list

Tbrid:
**Description**

When user entered due-date year that is smaller than 1970 the displaying date for the task is incorrect. 
Fixed with setting the year manually and with UTC timezone.

Jira link: https://lite-farm.atlassian.net/browse/LF-3506

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/CardWithStatus/TaskCard/TaskCard.jsx
+++ b/packages/webapp/src/components/CardWithStatus/TaskCard/TaskCard.jsx
@@ -33,7 +33,11 @@ export const taskStatusTranslateKey = {
 };
 
 const getDate = (date, language = 'en') => {
-  return new Intl.DateTimeFormat(language, { dateStyle: 'medium' }).format(new Date(date));
+  const d = new Date(date);
+  d.setFullYear(date.split(', ')[1]);
+  return new Intl.DateTimeFormat(language, { timeZone: 'UTC', dateStyle: 'medium' }).format(
+    new Date(d),
+  );
 };
 
 export const PureTaskCard = ({
--- a/packages/webapp/src/components/CardWithStatus/TaskCard/TaskCard.jsx
+++ b/packages/webapp/src/components/CardWithStatus/TaskCard/TaskCard.jsx
@@ -33,10 +33,8 @@ export const taskStatusTranslateKey = {
 };
 
 const getDate = (date, language = 'en') => {
-  const d = new Date(date);
-  d.setFullYear(date.split(', ')[1]);
   return new Intl.DateTimeFormat(language, { timeZone: 'UTC', dateStyle: 'medium' }).format(
-    new Date(d),
+    new Date(date),
   );
 };
 
--- a/packages/webapp/src/util/moment.js
+++ b/packages/webapp/src/util/moment.js
@@ -65,7 +65,7 @@ export const getManagementPlanTileDate = (date) =>
   moment(date).locale(getLanguageFromLocalStorage()).format(`MMM DD,'YY`);
 
 export const getTaskCardDate = (date) =>
-  moment(date).locale(getLanguageFromLocalStorage()).format('MMM D, YYYY');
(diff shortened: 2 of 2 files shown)


Duncan-Brain:
@Tbrid 

It looks like the date formats being sent to the task cards are not consistent. That means we can't assume the format to parse. 
Screenshot 2023-08-21 at 2 15 18 PM

So for me the task list is crashing when I have abandoned tasks present (maybe also completed tasks).

On first look I can't figure out where the date format differ's from the slice or controller -- sorry about this preexisting bug!

Duncan-Brain:
I think this is the offending function for the crash -- 
Screenshot 2023-08-21 at 2 35 02 PM

Duncan-Brain (changes requested):
Either fixing that function or choosing a different solution. Its too bad we can't set an option to parse numeric vs two-digit format. 

Seems like Intl.DateTimeFormat() lets you set it that way but not parse 🤷

Tbrid:
@Duncan-Brain  Thanks for finding the bug and the function, I changed the date format for the getTaskCardDate to be yyyy-mm-dd

packages/webapp/src/components/CardWithStatus/TaskCard/TaskCard.jsx
```diff
@@ -33,7 +33,9 @@ export const taskStatusTranslateKey = {
 };
 
 const getDate = (date, language = 'en') => {
-  return new Intl.DateTimeFormat(language, { dateStyle: 'medium' }).format(new Date(date));
+  return new Intl.DateTimeFormat(language, { timeZone: 'UTC', dateStyle: 'medium' }).format(
```
Duncan-Brain, line 36: @Tbrid just wondering about the addition of the timeZone here - I am not sure we should be setting the timeZone

Duncan-Brain:
I like the new solution, just curious about the timeZone setting though.

Tbrid:
> I like the new solution, just curious about the timeZone setting though.

I used that from a stackoverflow post when I was researching. I tried without the UTC timezone and it will be off.

iperdomo:
> I used that from a stackoverflow post when I was researching.

As general comment for all contributions (including core team @LiteFarmOrg/reviewers ). If you include snippets of code from StackOverflow, you must do a proper [attribution](https://stackoverflow.blog/2009/06/25/attribution-required/), e.g. `// Attribution: https://meta.stackexchange.com/a/126432`

Duncan-Brain (commented):
Sorry this took a while to think about. These stored dates (not datetimes) are treated quite weirdly in LiteFarm backend. 

I tried setting the date in Anchorage Alaska and Auckland New Zealand to get a large date difference and it seemed to work fine. 

I am at least okay with fallout from displaying in UTC. But @kathyavini mentioned that we should check the copy crop plan functionality as well. And it seems to assume 1901 instead of 2001 like the initial issue. 

This was Joyces idea but I think I agree that maybe this should be more about setting epoch as the earliest selectable date throughout the app?

kathyavini (commented):
Yes as @Duncan-Brain said I am seeing the 1902 date assumed in repeat crop plan. This same date displays as 2 (AD I guess?!) in the task card, and 0002 in the input field, and '02 in the management plan card (which is correct for any of the above, but reads like 2002 because that is how we interpret '02 currently 🙂 )

My feeling is that we built too many views around different variations of recent date ranges to smoothly adapt to accepting these inputs, and the usefulness to the user of accepting dates before 1970 has to be really, really small. I think putting a minimum validation constraint on the date (= must be after 1970) is going to make a lot more sense that refactoring all of those views including the 2-digit year.

Screenshot 2023-08-29 at 2 58 52 PM

Screenshot 2023-08-29 at 3 14 32 PM

Screenshot 2023-08-29 at 2 58 27 PM

PR #2848, opened 2023-08-22
Lf 3556 create custom expense type read only view

Duncan-Brain:
**Description**

CONSIDER MERGING THIS AFTER API AND AND OTHER VIEWS MADE 

This PR will include the containers for add, edit and readonly expense type. It is also required as part of the ticket to add sagas and necessary selectors.

Also:
- Adds updates to maxChars validation requested.
- add prop for input validation
- Adds type=button for routing issues
- update url name
- fixes broken expense details on expense detail and edit page

A lot of code was copied from LF-3553 (#2851) and I made some bad merges that fixed the mistakes. Relies upon current state of LF-3551 (#2849) so this PR will not function entirely without a few of these pieces.

**Design decisions**
- Creates a separate container/route for each view of add, edit, readonly - it might be nice to combine at least edit + readonly in the future
- Validation for naming is added to frontend

Jira link: [LF-3556](https://lite-farm.atlassian.net/browse/LF-3556),  [LF-3560](https://lite-farm.atlassian.net/browse/LF-3560), [LF-3557](https://lite-farm.atlassian.net/browse/LF-3557)

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

[LF-3556]: https://lite-farm.atlassian.net/browse/LF-3556?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LF-3560]: https://lite-farm.atlassian.net/browse/LF-3560?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LF-3557]: https://lite-farm.atlassian.net/browse/LF-3557?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/webapp/src/containers/Finances/CustomExpenseType/constants.js
```diff
@@ -0,0 +1 @@
+export const CUSTOM_EXPENSE_NAME = 'expense_name';
```
Duncan-Brain, line 1: Note to self add License to this file before/after next merge commit

SayakaOno:
Could you resolve conflicts again? 😓
I hope it's simple... I failed to resolve it for my PR and had to make a commit to fix it...

packages/webapp/src/containers/Finances/saga.js
```diff
@@ -148,6 +151,69 @@ export function* getFarmExpenseTypeSaga() {
...
+  const { expenseTypeUrl } = apiConfig;
+  let { user_id, farm_id } = yield select(loginSelector);
+  const header = getHeader(user_id, farm_id);
+  const { expense_type_id } = action;
+
+  try {
+    const result = yield call(axios.delete, `${expenseTypeUrl}/${expense_type_id}`, header);
+    if (result) {
+      yield put(enqueueSuccessSnackbar(i18n.t('message:EXPENSE_TYPE.SUCCESS.DELETE')));
+      yield call(getFarmExpenseTypeSaga);
```
kathyavini, line 208: Wow, it's so simple this way 😆 Just a fresh GET and full set every time!

kathyavini (commented):
It looks great!! And works well for me with the current API branch.

Two questions:

1. If I fail the unique name constraint by inputting the name of a deleted type, the save button is disabled but no validation error message shows to tell me why the save is disabled. Is that intentional?

2. Does the deleted state need to be handled in any way (e.g. hiding the retire button and/or the edit button or making the Edit input readonly)? The API returns an error on edit and the retire just "successfully" re-retires so nothing damaging happens. But it's not too hard to get to a deleted type via back button so I'm curious if it needs addressing.

Also I think I saw some comments in another PR about persisted paths being reviewed in tech daily next week. I think I need that 😆 I can't actually tell what the persisted behaviour is supposed to be here (for instance, is the value in the Edit view supposed to be saved if you return to the readonly view and use the edit button again?) But I thought this is a nice set of views to discuss paths with if you don't mind bringing it up as a demo!

packages/webapp/src/containers/Finances/saga.js
```diff
@@ -148,6 +151,69 @@ export function* getFarmExpenseTypeSaga() {
...
+export function* updateCustomExpenseTypeSaga(action) {
+  const { expenseTypeUrl } = apiConfig;
+  let { user_id, farm_id } = yield select(loginSelector);
+  const header = getHeader(user_id, farm_id);
+  const { expense_type_id } = action;
+  const { expense_type_name } = action.custom_expense_type;
+
+  try {
+    const result = yield call(
+      axios.put,
```
kathyavini, line 182: Odd; I could have sworn this worked before (maybe this PR or the other changed since I first tested?) but with most recent of version of the API PR, this needs to be `patch`; there is no `put` route.

packages/webapp/src/containers/Finances/saga.js
```diff
@@ -148,6 +151,69 @@ export function* getFarmExpenseTypeSaga() {
...
+    yield put(enqueueErrorSnackbar(i18n.t('message:EXPENSE_TYPE.ERROR.ADD')));
+  }
+}
+
+export function* updateCustomExpenseTypeSaga(action) {
+  const { expenseTypeUrl } = apiConfig;
+  let { user_id, farm_id } = yield select(loginSelector);
+  const header = getHeader(user_id, farm_id);
+  const { expense_type_id } = action;
+  const { expense_type_name } = action.custom_expense_type;
```
kathyavini, line 178: And this and the API call right below need to both be `expense_name`, not `expense_type_name`. Then the route will work!

packages/webapp/src/containers/Finances/saga.js
Duncan-Brain: Yup good call thanks!

packages/webapp/src/containers/Finances/saga.js
Duncan-Brain: Thank you!

Duncan-Brain:
@SayakaOno thanks, the merge was not bad! I have not had so many merges since working here -- it was a strange one I think I did something wrong when copying others code haha.

@kathyavini I made those two small fixes. As for the re-delete and persistedForm good call, I think the way I have it is not ideal. Some solutions I might try:

- Persisted forms I just copied from tasks, I might try to delete this and see what happens -- I don't think it is necessary here (and I am really not sure  its complete functionality). In theory the usefulness is that the form state will remain if one navigates away and then returns -- but for one single input I do not think its very useful. It might also help with the re-delete.

- The old selector style without the adapters do not utilize upsertOne, and deleteOne -- if I can add a custom action for that and then it should show a record not found page (which I can add in to the routing). I think I have been stuck in trying to keep the old style for this since we didn't want to update Finances that I didn't add the extra functionality our new style handles.

packages/webapp/src/containers/Finances/saga.js
```diff
@@ -349,6 +415,9 @@ export default function* financeSaga() {
   yield takeLeading(ADD_OR_UPDATE_SALE, addSale);
   yield takeLatest(GET_EXPENSE, getExpenseSaga);
   yield takeLatest(GET_FARM_EXPENSE_TYPE, getFarmExpenseTypeSaga);
+  yield takeLeading(ADD_CUSTOM_EXPENSE_TYPE, addCustomExpenseTypeSaga);
+  yield takeLeading(UPDATE_CUSTOM_EXPENSE_TYPE, updateCustomExpenseTypeSaga);
+  yield takeLeading(RETIRE_CUSTOM_EXPENSE_TYPE, retireCustomExpenseTypeSaga);
```
SayakaOno, line 420: Just FYI, when a reducer is not necessary for the action, you could create an action in this file:
ex. `export const addCustomExpenseType = createAction('addCustomExpenseTypeSaga');`
then `yield takeLeading(addCustomExpenseType,type, addCustomExpenseTypeSaga);`
and use the action in the component. So you would not have to create constants and actions in the actions file. (I believe so...)

packages/webapp/src/containers/Finances/CustomExpenseType/EditSimpleCustomExpense.jsx
```diff
@@ -0,0 +1,70 @@
...
+  const validateUniqueTypeName = (value) => {
+    const expenseNameExists = expenseTypes.some((type) => {
+      return type.expense_name === value;
+    });
+
+    if (expenseNameExists) {
+      return t('EXPENSE.ADD_EXPENSE.DUPLICATE_NAME');
+    }
+    return true;
+  };
```
SayakaOno, line 50: Is this exactly the same as the one in the creation component??
I'm just wondering why you didn't make it reusable...

SayakaOno (commented):
> The old selector style without the adapters do not utilize upsertOne, and deleteOne...

`upsertOne` and `deleteOne`  could be converted like this in `/packages/webapp/src/containers/Finances/reducer.js`
(I didn't test it...)

```
case UPSERT_EXPENSE_TYPE:
      return {
          ...state,
          expense_types: [...state.expense_types, action.expense_type]
      }
case DELETE_EXPENSE_TYPE:
      return {
          ...state,
          expense_types: state.expense_types.filter(({ expense_type_id }) => {
             return expense_type_id !== action.expense_type_id
          }) 
      }
...
```

packages/webapp/src/containers/Finances/saga.js
Duncan-Brain: Thanks yes I get confused between what is new style and old style. I thought I had to create an action this way to be consistent. I can fix it.

packages/webapp/src/containers/Finances/CustomExpenseType/EditSimpleCustomExpense.jsx
Duncan-Brain: No reason haha yes I should make it generic.

packages/webapp/src/containers/Finances/saga.js
Duncan-Brain: Just copying tasks haha. Might change though

SayakaOno:
Sorry to add comments one after another...
`persistedPaths` sent to `SimpleCustomType` are not being used, correct?

It works great by the way!

Duncan-Brain:
@SayakaOno -- I am not entirely sure about the persisted paths but I think you are right. I am actually going to try removing the persisted path portions.

Duncan-Brain:
@SayakaOno I have removed persisted paths, and used createAction. ...I tried the upsert and delete but I ended up getting into a rabbit hole of needing to make some creative functions to deal with the fact that unlike entities they are not indexed by id, just an array of objects also needed to add the pick utility since it was returning all properties and it seemed like at that point I should just use the entity adapter if I wanted to do this. 

@kathyavini I started trying to deal with the what to do if the type is deleted. I think I want to put it in another ticket since it is not hurting anything as is. Is that cool? I have a couple solutions ready but I think a seperate ticket would be better for discussing that.

kathyavini (approved):
I don't mind the deleted check getting added in a separate ticket. Perhaps the API should be merged without the deleted checks too then? Otherwise the user will get into a failing snackbar state without indication of why, which _is_ worse than the current no-check situation.

SayakaOno:
@Duncan-Brain sounds and looks good! I don't think we should use `history.push()` for back arrows, but let's discuss the approach on Monday. (thank you for working on this while blocked by my tile view!)

SayakaOno (approved):

PR #2849, opened 2023-08-23
Lf 3551 create api's for custom expense types

gursimran-singh:
**Description**

To create Api's to add, edit, delete custom expense types. Following are the response types of each endpoint

1. `POST /expense_type` to create new expense type

- If for combination of expense name and farm id exists but is either deleted or not, then no new record is created and deleted is set to false for the existing record. 204 status code is sent in this case.
- If for above combination, no record is present in DB, then new expense type is created and 201 status code is sent.

2. `PUT /expense_type/[expense_type_id]` to update expense type

- While editing, if we change custom expense type to which is already present for that particular farm, then 409 status code is sent.
- On successful update 204 status code is sent.

3. `DELETE /expense_type/[expense_type_id]` to delete expense type

- If delete is successful, then 200 status code is send.
- If expense_type_id is wrong, then 404 status code is sent.

Note: Farm workers(role 3) has permission to edit custom expense type which needs to be changed and is addressed in this ticket https://lite-farm.atlassian.net/browse/LF-3580

Jira link: https://lite-farm.atlassian.net/browse/LF-3551

**Type of change**

**How Has This Been Tested?**

Test cases written to support each case such as Owners, Managers can add, edit, delete expense type in their own farms and workers or unauthorized users cannot add, edit, delete expense type.

- [x] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/farmExpenseTypeController.js
+++ b/packages/api/src/controllers/farmExpenseTypeController.js
@@ -23,23 +23,26 @@ const farmExpenseTypeController = {
     return async (req, res) => {
       const trx = await transaction.start(Model.knex());
       try {
+        const farm_id = req.headers.farm_id;
         const data = req.body;
         data.expense_translation_key = data.expense_name;
 
-        // check if records exixts in DB
-        const record = await ExpenseTypeModel.query()
-          .where({
-            expense_name: data.expense_name,
-            farm_id: data.farm_id,
-          })
-          .first();
-
-        //if record exists then set 'deleted' column to false
+        const record = await ExpenseTypeModel.existsInFarm(farm_id, data.expense_name);
+        // if record exists in db
         if (record) {
-          record['deleted'] = false;
-          await baseController.put(ExpenseTypeModel, record.expense_type_id, record, req, { trx });
-          await trx.commit();
-          res.status(204).send();
+          // if not deleted, means it is a active expense type
+          // throw conflict error
+          if (record.deleted === false) {
+            return res.status(409).send();
+          } else {
+            // if its deleted, them make it active
+            record.deleted = false;
+            await baseController.put(ExpenseTypeModel, record.expense_type_id, record, req, {
+              trx,
+            });
+            await trx.commit();
+            res.status(201).send(record);
+          }
         } else {
           const result = await baseController.postWithResponse(ExpenseTypeModel, data, req, {
             trx,
@@ -122,21 +125,13 @@ const farmExpenseTypeController = {
     return async (req, res) => {
       const trx = await transaction.start(Model.knex());
       const { expense_type_id } = req.params;
+      const farm_id = req.headers.farm_id;
       const data = req.body;
       data.expense_translation_key = data.expense_name;
 
       try {
-        // check if records exixts in DB
-        const record = await ExpenseTypeModel.query()
-          .where({
-            expense_name: data.expense_name,
-            farm_id: data.farm_id,
-          })
-          .whereNot(ExpenseTypeModel.idColumn, expense_type_id)
-          .first();
-
         // if record exists throw Conflict error
-        if (record) {
+        if (await ExpenseTypeModel.existsInFarm(farm_id, data.expense_name, expense_type_id)) {
           return res.status(409).send();
         }
 
--- a/packages/api/src/controllers/farmExpenseTypeController.js
+++ b/packages/api/src/controllers/farmExpenseTypeController.js
@@ -100,6 +100,12 @@ const farmExpenseTypeController = {
         res.sendStatus(403);
       }
       try {
+        // do not allow operations to deleted records
+        if (await this.isDeleted(req.params.expense_type_id)) {
+          return res.status(404).send();
+        }
+
+        // soft delete the record
         const isDeleted = await baseController.delete(
           ExpenseTypeModel,
           req.params.expense_type_id,
@@ -127,17 +133,21 @@ const farmExpenseTypeController = {
       const { expense_type_id } = req.params;
       const farm_id = req.headers.farm_id;
       const data = req.body;
-      data.farm_id = farm_id;
 
       try {
-        // if record exists throw Conflict error
+        // do not allow update to deleted records
+        if (await this.isDeleted(expense_type_id)) {
+          return res.status(404).send();
+        }
+
+        // if record exists then throw Conflict error
         if (await this.existsInFarm(farm_id, data.expense_name, expense_type_id)) {
           return res.status(409).send();
         }
 
         data.expense_translation_key = baseController.formatTranslationKey(data.expense_name);
 
-        const result = await baseController.put(ExpenseTypeModel, expense_type_id, data, req, {
+        const result = await baseController.patch(ExpenseTypeModel, expense_type_id, data, req, {
           trx,
         });
         await trx.commit();
@@ -168,6 +178,24 @@ const farmExpenseTypeController = {
     }
 
     return query.first();
+  },
+
+  /**
+   * To check if record is deleted or not
+   * @param {number} expense_type_id - Expesnse type id
+   * @async
+   * @returns {Boolean} - true or false
+   */
+  async isDeleted(expense_type_id) {
+    const expense = await ExpenseTypeModel.query()
+      .context({ showHidden: true })
+      .where({
+        expense_type_id,
+      })
+      .select('deleted')
+      .first();
+
+    return expense.deleted;
   },
 };
 
--- a/packages/api/src/controllers/farmExpenseTypeController.js
+++ b/packages/api/src/controllers/farmExpenseTypeController.js
@@ -25,9 +25,9 @@ const farmExpenseTypeController = {
       try {
         const farm_id = req.headers.farm_id;
         const data = req.body;
-        data.expense_translation_key = data.expense_name;
+        data.expense_translation_key = baseController.formatTranslationKey(data.expense_name);
 
-        const record = await ExpenseTypeModel.existsInFarm(farm_id, data.expense_name);
+        const record = await this.existsInFarm(farm_id, data.expense_name);
         // if record exists in db
         if (record) {
           // if not deleted, means it is a active expense type
@@ -127,13 +127,14 @@ const farmExpenseTypeController = {
       const { expense_type_id } = req.params;
       const farm_id = req.headers.farm_id;
       const data = req.body;
-      data.expense_translation_key = data.expense_name;
 
       try {
         // if record exists throw Conflict error
-        if (await ExpenseTypeModel.existsInFarm(farm_id, data.expense_name, expense_type_id)) {
+        if (await this.existsInFarm(farm_id, data.expense_name, expense_type_id)) {
           return res.status(409).send();
         }
+
+        data.expense_translation_key = baseController.formatTranslationKey(data.expense_name);
 
         const result = await baseController.put(ExpenseTypeModel, expense_type_id, data, req, {
           trx,
@@ -145,6 +146,27 @@ const farmExpenseTypeController = {
         return res.status(400).send(error);
       }
     };
+  },
+
+  /**
+   * Check if records exists in DB
+   * @param {number} farm_id
+   * @param {String} expense_name
+   * @param {number} expense_type_id - Expesnse type id to be excluded while checking records
+   * @async
+   * @returns {Promise} - Object DB record promise
+   */
+  existsInFarm(farm_id, expense_name, expense_type_id = '') {
+    let query = ExpenseTypeModel.query().context({ showHidden: true }).where({
+      expense_name,
+      farm_id,
+    });
+
+    if (expense_type_id) {
+      query = query.whereNot({ expense_type_id });
+    }
+
+    return query.first();
   },
 };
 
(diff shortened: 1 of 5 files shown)


packages/api/src/controllers/farmExpenseTypeController.js
```diff
@@ -25,9 +25,28 @@ const farmExpenseTypeController = {
       try {
         const data = req.body;
         data.expense_translation_key = data.expense_name;
-        const result = await baseController.postWithResponse(ExpenseTypeModel, data, req, { trx });
-        await trx.commit();
-        res.status(201).send(result);
+
+        // check if records exixts in DB
+        const record = await ExpenseTypeModel.query()
```
JalilArfaoui, line 30: suggestion: maybe we could add a `ExpenseTypeModel.existsInFarm(expense_name, farm_id)` method and just call it here so that 
1/ it’s more clear what we’re doing and we don’t need the comment 
2/ we don’t leak database schema detail in our http controller
3/ we can reuse it maybe

packages/api/tests/expense_type.test.js
```diff
@@ -419,4 +434,82 @@ describe('Expense Type Tests', () => {
...
+        async (err, res) => {
+          expect(res.status).toBe(403);
+          expect(res.error.text).toBe(
+            'User does not have the following permission(s): add:expense_types',
+          );
+          done();
+        },
+      );
+    });
+    test('Unauthorized user should get 403 if they try to update expense type', async (done) => {
```
JalilArfaoui, line 494: **suggestion**: we should avoid using both `async` and `done` in a test function, because
1/ It can lead to confusion or mistake : is test done as soon as `done()` is called or as soon as promise is completed ? or maybe when both ?
2/ It is forbidden starting with Jest 27, it will throw the error `Test functions cannot both take a 'done' callback and return something. Either use a 'done' callback, or return a promise.`

You can see [this diff](https://github.com/LiteFarmOrg/LiteFarm/pull/2852/commits/38322e32672176c59075fe9a03101b2443ed3c4e#diff-5d0091677c17d96ab4243cdfe9f0c6f7872171ec4ebfc75a67489cb076221d1b) for an example of how we can get rid of `done` callback and go full `async/await` for instance …

packages/api/src/controllers/farmExpenseTypeController.js
```diff
@@ -25,9 +25,28 @@ const farmExpenseTypeController = {
       try {
         const data = req.body;
         data.expense_translation_key = data.expense_name;
```
Duncan-Brain, line 27: I think we will need to refactor this to handle the translation key snake case formatting

packages/api/src/controllers/farmExpenseTypeController.js
```diff
@@ -25,9 +25,28 @@ const farmExpenseTypeController = {
...
+        // check if records exixts in DB
+        const record = await ExpenseTypeModel.query()
+          .where({
+            expense_name: data.expense_name,
+            farm_id: data.farm_id,
+          })
+          .first();
+
+        //if record exists then set 'deleted' column to false
+        if (record) {
```
Duncan-Brain, line 38: Should this be `if (record && record['deleted'] == true)` ?

Else we are changing updated_at in the database if it already exists and is already false.

packages/api/src/routes/farmExpenseTypeRoute.js
```diff
@@ -43,4 +43,11 @@ router.delete(
   farmExpenseTypeController.delFarmExpenseType(),
 );
 
+router.put(
+  '/:expense_type_id',
+  hasFarmAccess({ params: 'expense_type_id' }),
```
Duncan-Brain, line 48: I think this should be farm_id otherwise you are able to change the farm_id using the body param to another farm (that you may nat have access to). I think you can also have multiple hasfarmAccess calls if this one is needed.

packages/api/tests/expense_type.test.js
```diff
@@ -307,7 +322,7 @@ describe('Expense Type Tests', () => {
     test('Owner should get 403 if they try to delete default expense type', async (done) => {
       const { mainFarm, user } = await returnUserFarms(1);
       const expense = await returnDefaultExpenseType();
-      console.log(expense);
+      //console.log(expense);
```
Duncan-Brain, line 325: Could possibly just delete this

Duncan-Brain (changes requested):
Added a couple of inline comments. Seems to be working well for the most part.

Things I noticed were:
1. The translation key is not being formatted properly
2. We maybe shouldn't update the record if nothing changes, if record already exists and is not deleted. I am not sure the best response code for that case.
3. Route authorization is not correct for put endpoint allows farm_id to change
4. (edit) I think we want to remove `.whereNotDeleted()` from the get endpoint since we still want to allow filtering expenses where the type has been deleted.

Design questions:
1. Is using put correct? Should the update route be able to update columns like "deleted" -- would a patch work? Since a type is really only editable for name and translation key (based on name)
2. New error codes - I am interested in your use of new error codes since they are probably correct. If you could elaborate further I would appreciate though since they are not present in other parts of the codebase. 409 I think makes sense, 204 I am less sure about since we may (or may not) need to update state on the frontend with the updated record.

packages/api/src/controllers/farmExpenseTypeController.js
```diff
@@ -98,6 +117,40 @@ const farmExpenseTypeController = {
...
+      const { expense_type_id } = req.params;
+      const data = req.body;
+      data.expense_translation_key = data.expense_name;
+
+      try {
+        // check if records exixts in DB
+        const record = await ExpenseTypeModel.query()
+          .where({
+            expense_name: data.expense_name,
+            farm_id: data.farm_id,
```
Duncan-Brain, line 133: Also should we use the request header farm_id here?

Duncan-Brain:
Hey @gursimran-singh just wondering if this is ready for re-review or if you are still refactoring?

If it is ready for re-review I am still seeing the following unexpected behaviour on the patch endpoint:

When I use postman to test and 

1. When req.header.farm_id != req.body.farm_id the patch is capable of changing the farm_id. This should be a not authorized error since farm_ids should match. Adding a second hasFarmAccess [here](https://github.com/LiteFarmOrg/LiteFarm/pull/2849#discussion_r1307888874) should prevent that when verifying the body farm id. There may be other ways to do this also.

2. When patching without any changes to the name the endpoint should return conflict 409 (since it already exists). I think the existsInFarm() is supposed to cover this case but I am not sure it is working properly.

gursimran-singh:
@Duncan-Brain .

1. 

- I got your point. Ideally user should not be able to update the farm_id while updating expense type as we are not giving them option to do that. But, Model has `farm_id` field as required ` required: ['expense_name', 'farm_id'],` so we need to pass it while updating. I have changed the code to use `farm_id` passed in header  `const farm_id = req.headers.farm_id; data.farm_id = farm_id;` 
- And in my opinion using `hasFarmAccess({ body: 'farm_id' })` in the patch route will not server our purpose as it is just equating the both(headers and body) passed farm_id. So, it would return true if both are same even if they do not exists in database. 
  `farm_id: (farm_id) => ({ farm_id }),`
  `sameFarm(farmIdObjectFromEntity, farm_id)`
  `function sameFarm(object, farm) {
    return object.farm_id === farm;
  }`
     Its actually `checkScope(['edit:expense_types'])` which is checking the authenticity by matching the database records with the passed `farm_id` in headers.
     So, this is the reason i decided to use `farm_id` passed in headers while patching expense type as it's authenticity would be checked by `checkScope` middleware

2. The endpoint should not return conflict 409 (even its already exists and we are not changing) because we are checking with the other records in the same farm. This is what `existsInFarm` is doing by checking other records then the passed expense_type_id `if (expense_type_id) {
      query = query.whereNot({ expense_type_id });
    }`. This is normal behaviour and is used widely everywhere. For example you are able to update you social profiles(LinkedIn, instagram, twitter etc) without making any changes. This would not return any conflict error. To verify, Just try to update your LinkedIn profile URL without making any changes. You will get success response instead of conflict. 

3. Yes, this is ready for re-reviewing. Let me know if any change is required.

kathyavini (dismissed):
This branch works great for me alongside Duncan's frontend container branch (#2848) and I find your helper functions very readable @gursimran-singh!

My only question is equally for you and for @Duncan-Brain (because the same operations are permitted on the frontend) regarding deleted records; right now both "re-delete" (it does nothing but triggers another successful snackbar) and edit (which actually will edit the deleted record) are permitted. I am indifferent about re-delete (particularly on the backend) but what do you guys think about edit? Should it be prohibited?

gursimran-singh:
@kathyavini No, delete and edit operations should not be allowed on already deleted records. Nice catch. I will update the code to handle this case.

Duncan-Brain:
1. Thanks for taking a look at the farm_id. The way I shared is just the way that was shared with me by previous employees... It may not be the best practice. But I am concerned that now we are now re-writing the body instead of returning some sort of bad request response.

2. I get your point, maybe 409 is the incorrect response -- as mentioned before this is the first time we have used it in the code! Using your example, if you save your instagram profile with no changes would instagram alter the database by updating the updated_at column in the database? I agree that whichever response that is sent back would be handled to show no issues to the user on the frontend. My main point is that if theres no change to the user data then maybe the db should not be altered at all -- Its not a sticking point for me and I am happy to approve whatever you choose for this point.

@kathyavini I agree with @gursimran-singh that deleted records should not be updated/edited. As for re-deleted records I think it is fine to show the snackbar but I am not sure anything should change on the db data level.

kathyavini (commented):
Sorry @gursimran-singh for more notes (!), but in today's tech daily, we had some time to delve deeper into point (1) about `farm_id` body/header and gathered some additional context. While recognizing that we haven't always followed a consistent pattern in the past, we want to now always return an error code when there's a discrepancy between the `farm_id` in the body and header of a request.

The preferred way to do that (as @Duncan-Brain mentioned above) is to make a second `hasFarmAccess()` call. But just to clarify, this doesn't mean replacing the current one (which we still need), but rather calling them in sequence:

```js
router.patch(
  '/:expense_type_id',
  hasFarmAccess({ body: 'farm_id' }), //  only to check body against header
  hasFarmAccess({ params: 'expense_type_id' }), // to check record ownership
  checkScope(['edit:expense_types']),
  farmExpenseTypeController.updateFarmExpenseType(),
);
```

For reference, this dual `hasFarmAccess` approach is on two of our newest routes (only!) which are [here](https://github.com/LiteFarmOrg/LiteFarm/blob/214ff95637bb4eeb99e3fe60cc8842d4b48ea96f/packages/api/src/routes/nominationRoute.js#L32) and [here](https://github.com/LiteFarmOrg/LiteFarm/blob/214ff95637bb4eeb99e3fe60cc8842d4b48ea96f/packages/api/src/routes/sensorRoute.js#L60) and moving forward we'd like to keep doing this.

gursimran-singh:
@kathyavini @Duncan-Brain I'm ok with what you guys prefer in this regard, but i would strongly recommend to have a discussion regarding the use of `hasFarmAccess({ body: 'farm_id' })`  as its very easy to spoof this check, just pass any string (doesn't matter if its a valid farm id or not), let say 'abc' for both header and body, and you can bypass this check easily. The question to ask ourself here is that, should this be acceptable?.

(2) I'm not sure what DB fields instagram is updating as we do not have access to their codebase, so i can not comment. But generally speaking, i expect them to have updated the timestamp in the DB. As it is a user initialized action which is to update the profile and the server responded with success, means, the request got authenticated, authorized, validated and data was saved in DB without any error.

Duncan-Brain:
@gursimran-singh if you are free to chat about (1) in this mornings meeting we want to make sure we understand your concerns. If it were just the single check I would agree but there are multiple checks (the second hasFarmAccess, the checkScope etc) beforehand. I am now seeing that the post request is actually not correct because as you say, the single check is not sufficient. I personally kind of like the old pattern functionality Joyce shared better to be honest but it does seem that the other pattern of multiple hasFarmAccess was preferred by the code writers.

2. Thanks for sharing your perspective here: I ask these questions to learn!  This is my first experience with a production db and so I think I agree with your point about a no change save request being:  "a user initiated save request which is successful". I was thinking in terms of "a user initiated save request which changes no user data and sends an appropriate response".

packages/api/src/controllers/farmExpenseTypeController.js
```diff
@@ -127,6 +127,7 @@ const farmExpenseTypeController = {
       const { expense_type_id } = req.params;
       const farm_id = req.headers.farm_id;
       const data = req.body;
+      data.farm_id = farm_id;
```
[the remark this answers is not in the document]
gursimran-singh: @kathyavini I have created 'Patch' method in base controller which allows to update single field in database and do not require farm_id in body of the request and we can skip the above suggested check. It is still validated but an error isn't thrown if the patch object doesn't contain all required properties. Additionally, deleted records will throw 404 error if either tried update or delete.

kathyavini (approved):
I love the `patch`, thank you! I'm glad you figured out how to get it working that way!

I'm enjoying no longer sending the `farm_id` without having to remove it from required properties 😄

Duncan-Brain (approved):
Thanks Gursimran -- will add the hasFarmAccess ourselves. The new patch is pretty sweet !

Fix for the farm_id overwriting here if interested: https://github.com/LiteFarmOrg/LiteFarm/pull/2872

decided against the hasFarmAccess so we could choose not to provide it

PR #2513, opened 2023-03-23
Docker with litefarm cli for development

mickenorlen:
## Updated 2023-04-20
 
# Docker compose
This PR includes a new setup for developing with `docker-compose`. As it doesn't modify any previous setup files, it can be merged in a beta state, and tested in parallel to previous setups. 

- The docker commit contains more information about the compose project structure. 
- How to actually work with the new setup is explained by the cli through it's verbose logging and documentation. 
- Use the provided cli commands to manage the project directly, or, if you want inspiration to experiment with your own compose commands, check the verbose logs for explanations of the commands that the cli runs internally.

# LiteFarm cli
As it isn't convenient for every developer to write complex compose commands, I created the LiteFarm command line interface `lfcli` to help working with the setup.

>  Instead of writing documentation about which commands to run,
>  why not create a script that actually runs those commands,
>  and generates its own documentation.

- The cli is completely self-contained in the cli directory and is not required for managing the docker setup.
- It is a helper that runs, logs and documents all the commands needed to manage the project. My hope is that this can help everyone get started and learn about how the new setup works as efficiently as possible, without the need for further documentation.
- The `web` and `api` commands provide a more convenient development server startup than the conventional 
`docker-compose up` by starting the services in an idle state and running the server commands explicitly.

 ## As light and simple as possible, yet scalable.
- The cli core is about 50 lines of code.
- All the documentation is generated in about 100 lines.
- Add 20 more lines to bring some color to the logs.
- The rest is command files, many of which are one-liners.
- To add a new command, just create a new script file in a commands directory.

## An interface between the command line and regular script files
- `lfcli` functions as an interface between the command line and regular script files.
- Available commands are defined by, and run their corresponding script files, found in the commands directory.
- Developers can experiment with their own commands by setting the `LITEFARM_USER_COMMANDS_DIR` variable in `cli/.env`

| Commands  | File structure |
| ------------- | ------------- |
| image | image |

## Command documentation, straight from the command file
Each individual command is documented at the beginning of the file between two docmarks `#####`. This documentation is then presented in a pretty format when calling `lfcli help my_command`.

| `lfcli help install`  | Install command file doc comment |
| ------------- | ------------- |
| image | install_help |

## The cli manual
The cli now also has its own manual following the same principle as the individual commands. Read the full manual with `lfcli man`

| Part 1  | Part 2 |
| ------------- | ------------- |
| image |image |

## Verbose logging
[...]

PR #2850, opened 2023-08-23
chore: add Makefile for easier dev environnment setup

JalilArfaoui:
**Description**

This PR aims to ease with setting up the stack in development environment.

Usage examples :
```sh
make setup # creates the .env files, mount the docker, install node_modules, run migrations … everything needed to get working
make reset # properly remove every node_modules, drops the database, and setup all again
make serve_api # serves api
make serve_webapp # serves webapp
```

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [X] Other : testing the commands on terminal

**Checklist:**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,76 @@
+# Copyright 2023 LiteFarm.org
+# This file is part of LiteFarm.
+#
+# LiteFarm is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# LiteFarm is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+
+include .env.secrets
+
+secrets:
+	@if [ ! -f .env.secrets ]; then echo "Please copy .env.secrets.default to .env.secrets and fill the variables"; exit 1; fi;
+	@if grep -Fq "=?" .env.secrets ; then echo "Please fill all the secrets in .env.secrets"; exit 1; fi;
+
+db_dropped:
+	@docker compose down
+	@docker volume prune -a -f
+
+api_env_file: ./packages/api/.env.default secrets
+	@cp ./packages/api/.env.default ./packages/api/.env
+	@sed -i 's/GOOGLE_API_KEY=?/GOOGLE_API_KEY=$(GOOGLE_API_KEY)/g' ./packages/api/.env
+	@sed -i 's/GOOGLE_OAUTH_CLIENT_ID=?/GOOGLE_OAUTH_CLIENT_ID=$(GOOGLE_OAUTH_CLIENT_ID)/g' ./packages/api/.env
+	@sed -i 's/OPEN_WEATHER_APP_ID=?/OPEN_WEATHER_APP_ID=$(OPEN_WEATHER_APP_ID)/g' ./packages/api/.env
+	@sed -i 's/DEV_GMAIL=/DEV_GMAIL=$(DEV_GMAIL)/g' ./packages/api/.env
+	@sed -i 's/DEV_GMAIL_APP_PASSWORD=/DEV_GMAIL_APP_PASSWORD=$(DEV_GMAIL_APP_PASSWORD)/g' ./packages/api/.env
+	@echo "Generated ./packages/api/.env"
+
+webapp_env_file: ./packages/webapp/.env.default secrets
+	@cp ./packages/webapp/.env.default ./packages/webapp/.env
+	@sed -i 's/VITE_GOOGLE_MAPS_API_KEY=?/VITE_GOOGLE_MAPS_API_KEY=$(GOOGLE_API_KEY)/g' ./packages/webapp/.env
+	@sed -i 's/VITE_GOOGLE_OAUTH_CLIENT_ID=?/VITE_GOOGLE_OAUTH_CLIENT_ID=$(GOOGLE_OAUTH_CLIENT_ID)/g' ./packages/webapp/.env
+	@sed -i 's/VITE_WEATHER_API_KEY=?/VITE_WEATHER_API_KEY=$(OPEN_WEATHER_APP_ID)/g' ./packages/webapp/.env
+	@echo "Generated ./packages/webapp/.env"
+
+env_files: api_env_file webapp_env_file
+
+docker_up:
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -0,0 +1,80 @@
...
+	@cp ./packages/webapp/.env.default ./packages/webapp/.env
+	@sed -i 's/VITE_GOOGLE_MAPS_API_KEY=?/VITE_GOOGLE_MAPS_API_KEY=$(GOOGLE_API_KEY)/g' ./packages/webapp/.env
+	@sed -i 's/VITE_GOOGLE_OAUTH_CLIENT_ID=?/VITE_GOOGLE_OAUTH_CLIENT_ID=$(GOOGLE_OAUTH_CLIENT_ID)/g' ./packages/webapp/.env
+	@sed -i 's/VITE_WEATHER_API_KEY=?/VITE_WEATHER_API_KEY=$(OPEN_WEATHER_APP_ID)/g' ./packages/webapp/.env
+	@echo "Generated ./packages/webapp/.env"
+
+env_files: api_env_file webapp_env_file
+
+change_port_to_9091:
+	@sed -i 's/9090/9091/g' docker-compose.yml
```
iperdomo, line 43: Why the port change? You can override ports locally with `docker-compose.override.yml`

Makefile
```diff
@@ -0,0 +1,80 @@
...
+	@echo "Minio port changed to 9091"
+
+docker_up: change_port_to_9091
+	@docker-compose up -d db minio create_minio_bucket redis imaginary
```
iperdomo, line 47: * `docker-compose` is not longer getting updates, use `docker compose` instead
* > From July 2023 Compose V1 stopped receiving updates. It's also no longer available in new releases of Docker Desktop. https://docs.docker.com/compose/

iperdomo (changes requested):
We have had [previous attempts](https://github.com/LiteFarmOrg/LiteFarm/pull/2513) to improve the local development. 

Adding `make` to the local tools makes us type less (fewer keystrokes), but hides some interesting to know concepts/flags when dealing with `docker compose`

This `Makefile` assume that all project dependencies are installed locally, e.g I use a _Dockerized_ development environment.

Requesting contributors to install `make` is just another dependency to deal with.

If you're OK with my arguments, I'll vote for closing the PR and not merge the changes.

Thanks,

Makefile
JalilArfaoui: Yep, of course ! I’ve removed that !

Makefile
JalilArfaoui: Good catch, thank you ! I’ve changed that

JalilArfaoui:
I’m not sure that I agree with your reasoning here @iperdomo … 

It’s not about fewer keystrokes but about automating stuff, less error-prone operations and abstracting (hiding indeed) implementation details.

I think that the whole point of writing software is all about abstracting and hiding details. Of course, it’s interesting to know about Docker concept. But as I think that a farmer using Litefarm should not **have to** worry about React principles, I think that some developer wanting to work on the UI for instance should not **have to** master Docker or Postgres … 

Actually, Docker itself is a tool that is build over Linux containers to abstract (and hide) their complexity … And we could talk about Lerna … or even React … 

I agree that it adds an optional dependency (even though `make` is pretty common on a developer machine) … but since it’s optional and anyone can’t still run the commands by hand, I’m not sure if it does any harm to have the option … This PR does not break any existing option … We would not be requesting contributors to install `make`, as well as [we don’t request contributors to install `docker` at all](https://github.com/LiteFarmOrg/LiteFarm/#preliminaries).

But but but !  The real question here I guess is … Does that file seem useful to any other contributors ? If I’m the only one saving time/trouble with this, I’ll be happy to close this PR and keep it only on my machine :-)

Makefile
```diff
@@ -0,0 +1,76 @@
...
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# LiteFarm is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+
+include .env.secrets
```
kathyavini, line 14: I had to make this an optional include (`-include`) to prevent crashing on the first run with this error. After that change it worked!

Screenshot 2023-08-29 at 3 44 44 PM

Makefile
```diff
@@ -0,0 +1,76 @@
...
+	@if [ ! -f .env.secrets ]; then echo "Please copy .env.secrets.default to .env.secrets and fill the variables"; exit 1; fi;
+	@if grep -Fq "=?" .env.secrets ; then echo "Please fill all the secrets in .env.secrets"; exit 1; fi;
+
+db_dropped:
+	@docker compose down
+	@docker volume prune -a -f
+
+api_env_file: ./packages/api/.env.default secrets
+	@cp ./packages/api/.env.default ./packages/api/.env
+	@sed -i 's/GOOGLE_API_KEY=?/GOOGLE_API_KEY=$(GOOGLE_API_KEY)/g' ./packages/api/.env
```
kathyavini, line 26: With default MacOS `sed` (I had to look this up when I got an error) the syntax needs to be `@sed -i ''`, so I think a conditional to check the platform and format the sed command is needed. (A mac user could also `brew install gnu-sed` and update the PATH but for onboarding I don't think that should be required!)

kathyavini (commented):
I also like the native docker-compose syntax as in my opinion it's already very succinct and readable even for beginners. I agree with @iperdomo that exposing as much docker as possible also lets new devs build some transferable knowledge which is an implicit goal of our open invitation to contributors.

That said, I think our current handling of `.env` variables is a painful slog (in two over-long files in two disparate directories) that doesn't give any real benefit to new or experienced developers. I really like the command-line feedback and ease of 

`make secrets`

and

`make env_files`

that you built here, and I *really* love the division of `.env` variables into secret and those that can be be copied verbatim (although I think @Duncan-Brain pointed out better file organization on our part could also achieve that), and that contributors are populating one file with keys rather than two.

I had to make a few adjust to the Makefile to make it work for me (noted below), and a few more variables are needed in `env.secrets.default`, but fundamentally I think this is a much better solution for `.env` files that what we do currently and I would like to adopt it... as long as `make` is feasible to get and get working on Windows (which I can't test).

So moving to `make` at least for `.env` file management currently has my vote, but I'd love to leave the discussion open for more devs and contributors to chime in!

JalilArfaoui:
Thanks @kathyavini 

I do agree on the fact that this is mainly a workaround for the existing onboarding steps needed in the README … but it would be better to have a single .env.

But, with current state of the repo, this is what I made to help myself and though I would share :-)

PR #2851, opened 2023-08-23
LF-3553: Updated multi select expense type view

SayakaOno:
**Description**

Screenshot 2023-08-23 at 2 52 41 PM

- Backend
  - migration: soft-delete expense type "Other"
  - update `/expense_type` API to:
      - return soft-deleted types as well
      - include `deleted` property in the response (override `hidden()` of BaseModel)
- Frontend
   - update icons
   - create `PureFinanceTypeSelection` component that will be used for revenue as well
   - update `ExpenseCategories` container
      - use `PureFinanceTypeSelection`
      - get `selectedExpense` from Redux store to persist selected types
    - Redux
      - create an action `getFarmExpenseType` and `getFarmExpenseTypeSaga` for getting all types including soft-deleted ones
      - create a selector `allExpenseTypeSelector` that returns types including soft-deleted ones
      - rename the misleading variable names "default expense type" that are used for "expense type"
      - update `expense_type` state to include soft-deleted types
      - update `expenseTypeSelector` to exclude soft-deleted types (we didn't have to exclude deleted ones before because the API was not returning them)
      - create `allExpenseTypeSelector` to be used for expense history (deleted types should still show)
      - create `expenseTypeTileContentsSelector` that returns sorted types (the logic was moved from the component)
   - some improvements on `IconLabelTile` (add className) and Tile's css (truncate texts as we want to)
   - create a story (http://localhost:6006/?path=/docs/page-finance-purefinancetypeselection--docs)
   - fix Tiles story to render `children` properly

NOTE:
- page navigation (cancel, back arrow buttons) will not work until the next view is complete
- "Manage custom type" page is not ready, so you will be navigated to the home page

Jira link: https://lite-farm.atlassian.net/browse/LF-3553

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/webapp/src/components/Tile/styles.module.scss
```diff
@@ -56,6 +56,9 @@
   overflow: hidden;
   text-overflow: ellipsis;
   width: 100%;
+  display: -webkit-box;
+  -webkit-line-clamp: 2;
+  -webkit-box-orient: vertical;
```
SayakaOno, line 61: [Text overflow ellipsis on two lines](https://stackoverflow.com/questions/15909489/text-overflow-ellipsis-on-two-lines)

packages/webapp/src/containers/Finances/NewExpense/ExpenseCategories/index.jsx
```diff
@@ -1,66 +1,47 @@
...
-      unSelectedStyle: {
-        width: '80px',
-        height: '80px',
-        borderRadius: '50px',
-        margin: '0 auto',
-        background: '#82CF9C',
-        boxShadow: '0px 2px 4px rgba(0, 0, 0, 0.08)',
-      },
-      selectedTypes: [],
+      selectedTypes: this.props.selectedExpense,
```
SayakaOno, line 36: get `selectedExpense` from the store to persist `selectedTypes`.

packages/webapp/src/containers/Finances/NewExpense/ExpenseCategories/styles.module.scss
```diff
@@ -30,25 +15,11 @@
...
-  margin: auto;
-}
+.labelIcon {
+  font-weight: 600;
 
-.continueButton {
-  margin-top: 10px;
+  svg {
+    border-top: solid 3px transparent;
+    border-bottom: solid 3px transparent;
```
SayakaOno, line 23: Solution for making icons smaller.

packages/webapp/src/containers/Finances/selectors.js
```diff
@@ -23,7 +24,33 @@ const selectedSaleSelector = createSelector(financeSelector, (state) => state.se
 
 const expenseSelector = createSelector(financeSelector, (state) => state.expenses);
 
-const expenseTypeSelector = createSelector(financeSelector, (state) => state.expense_types);
+const allExpenseTypeSelector = createSelector(financeSelector, (state) => state.expense_types);
+
+const expenseTypeSelector = createSelector(financeSelector, (state) => {
+  return state.expense_types.filter((type) => !type.deleted);
+});
```
SayakaOno, line 31: This was returning `state.expense_types` that did not include deleted types.
Now `state.expense_types includes deleted one, and they will be filtered out.
The result won't change.

packages/webapp/src/containers/Finances/selectors.js
```diff
@@ -23,7 +24,33 @@ const selectedSaleSelector = createSelector(financeSelector, (state) => state.se
...
+  });
+
+  return [
+    ...defaultTypes.sort((typeA, typeB) =>
+      i18n
+        .t(`expense:${typeA.expense_translation_key}`)
+        .localeCompare(i18n.t(`expense:${typeB.expense_translation_key}`)),
+    ),
+    ...customTypes.sort((typeA, typeB) =>
+      typeA.expense_translation_key.localeCompare(typeB.expense_translation_key),
```
SayakaOno, line 50: The logic was moved from `ExpenseCategories` component.
I followed the same pattern as `taskType` (sorted default types -> sorted custom types):
https://github.com/LiteFarmOrg/LiteFarm/blob/0dced6cb926689a6b625b0307f88f3def0e10426/packages/webapp/src/components/Task/PureTaskTypeSelection/PureTaskTypeSelection.jsx#L121-L176

packages/webapp/src/containers/Finances/NewExpense/ExpenseCategories/index.jsx
```diff
@@ -80,90 +61,37 @@ class ExpenseCategories extends Component {
...
-              disabled={!selectedTypes.length}
-              className={styles.continueButton}
-            >
-              {this.props.t('common:CONTINUE')}
-            </Button>
-          </div>
-        </div>
+        <HookFormPersistProvider>
+          <PureFinanceTypeSelection
+            title={this.props.t('EXPENSE.ADD_EXPENSE.TITLE')}
```
JalilArfaoui, line 69: **question**: Why not doing the translation inside the component instead of passing title as a prop, since we're using `useTranslation` inside anyway ?

Do we plan to use different titles for this component depending of where it's called ?

SayakaOno: Yes, that's correct! We are going to use `PureFinanceTypeSelection` for revenue type as well.

Thank you so much for the translations Jalil!!!

JalilArfaoui: Of course, that makes sense … thank you for answering

packages/webapp/public/locales/en/translation.json
```diff
@@ -488,7 +488,10 @@
       "MIN_ERROR": "Please enter a value greater than ",
       "REQUIRED_ERROR": "Expense is required",
       "TITLE_1": "New Expense (1 of 2)",
-      "TITLE_2": "New Expense (2 of 2)"
+      "TITLE_2": "New Expense (2 of 2)",
```
Duncan-Brain, line 491: Can we delete old translations? Tile 1 and/or Title 2?

packages/webapp/src/components/Finances/PureFinanceTypeSelection/index.jsx
```diff
@@ -0,0 +1,92 @@
...
+      <Tiles tileType={tileTypes.ICON_LABEL} tileData={types} formatTileData={formatTileData} />
+      <div className={styles.manageCustomTypeLinkContainer}>
+        <IconLink
+          id={iconLinkId}
+          icon={<GearIcon style={{ transform: 'translate(0px, 2px)' }} />}
+          onClick={onGoToManageCustomType}
+          isIconClickable
+          underlined={false}
+        >
+          {t('SALE.FINANCES.MANAGE_CUSTOM_EXPENSE_TYPE')}
```
Duncan-Brain, line 72: I think the text is generic - but should the name of the translation be generic too?

packages/webapp/src/containers/Finances/actions.js
```diff
@@ -20,10 +20,11 @@ import {
   DELETE_EXPENSES,
   DELETE_SALE,
   GET_DEFAULT_EXPENSE_TYPE,
```
Duncan-Brain, line 22: Do we use default types anymore?

Duncan-Brain (changes requested):
Very nice! I have leftover custom types from the API PR and it works well to add even custom expenses not yet defined!

My questions :

1. migration: Do we actually want to delete "other" or just rename it to miscellaneous (at least for now)? 
2. controller, actions, selector etc: Do we use the getDefaultTypes anymore? (can we delete default types references?)
3. I like the ellipsis thing - could be nice to consider adding a long expense type name to the story?
4. For some reason the tile is selectable on mobile but not on desktop (in storybook) - I am not sure the desired behaviour. The select on mobile is weird in the actual app too  (see video) but it doesn't seem to be working right.


https://github.com/LiteFarmOrg/LiteFarm/assets/30075973/e0af4379-6b2a-4153-9f68-9af2cc34719a
5. French translations: It is nice we received these - but I always remember the one time when Ivan was translating he mentioned seeing errors in Spanish translation and he didn't realize we were using inclusive language. Do we know if this meets our translation guidelines?

6. Small thing: does the Tile label seem "too bold" to you?

packages/webapp/src/containers/Finances/actions.js
SayakaOno: I can delete them.

SayakaOno:
Thank you @Duncan-Brain !

1. migration: updated to "miscellaneous". I will finalize the order of types and the icon once David confirms.
2. default types: deleted
3. ellipsis: story added
4. selectable tiles on mobile:
    This was interesting, and good catch! In storybook, onClick function is just a console.log, so tiles cannot be selected. After tapping a tile on mobile, the tile is considered as "hovered". The style of "hovered" state is the same as "selected" style so the weird thing happened. I referred this [page](https://stackoverflow.com/questions/23885255/how-to-remove-ignore-hover-css-style-on-touch-devices) (should I do the attribution?) and added @media to the css.

> Small thing: does the Tile label seem "too bold" to you?

It does! It really stands out... maybe it's too dark?

packages/webapp/src/components/Finances/PureFinanceTypeSelection/index.jsx
Duncan-Brain: he text for this translation is :
```
...
"MANAGE_CUSTOM_EXPENSE_TYPE": "Manage custom type",
"OTHER_EXPENSES_LABEL": "Other Expenses",
...
 ```
 
 It looks like you did not add this translation so feel free to ignore. Just thought since the text is already generic. We should rename it "MANAGE_CUSTOM_TYPE"

Comment (SayakaOno) [reply packages/webapp/src/components/Finances/PureFinanceTypeSelection/index.jsx:72]: Ah, I see. I am adding real "Manage custom expense types" later, so it would make it less confusing, too.
I don't like that it's under SALE because it will be used for revenue type as well, so I will fix it!

File: packages/webapp/src/containers/Finances/NewExpense/ExpenseCategories/index.jsx [hunk 10]
```diff
@@ -27,7 +28,15 @@ export const icons = {
   SEEDS: <SeedIcon />,
   OTHER: <OtherIcon />,
   LAND: <LandIcon />,
-  MISCELLANEOUS: <OtherIcon />,
+  MISCELLANEOUS: (
+    <MiscellaneousIcon
+      style={{
+        border: 'solid 10px transparent',
+        filter:
+          'invert(30%) sepia(94%) saturate(787%) hue-rotate(136deg) brightness(103%) contrast(98%)',
```
SayakaOno, line 36: I believe this icon is temporary, but this is to change the colour of the svg.
https://stackoverflow.com/questions/22252472/how-can-i-change-the-color-of-an-svg-element

SayakaOno:
In addition to addressing Duncan's comments, these were done:
- add Miscellaneous icon ("other" icon will be reused)
- fixed ManageCustomExpenseTypesSpotlight (I accidentally broke it). Now it's in the component not the container.
- add null check to the new selectors. (the app was crashing on logout)

packages/api/db/migration/20230817213340_soft_delete_other_expense_type.js
```diff
@@ -0,0 +1,36 @@
...
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ */
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+export const up = async function (knex) {
```
kathyavini, line 20: Just wanted to check if you wanted to rename the file now that it's just a column name change and not removal of "other".

packages/api/src/models/expenseTypeModel.js
```diff
@@ -24,6 +24,10 @@ class ExpenseType extends baseModel {
     return 'expense_type_id';
   }
 
+  static get hidden() {
```
kathyavini, line 27: Thank you for explaining what this does re:deleted! I'll add it to revenueTypes too!

packages/webapp/public/locales/fr/translation.json
```diff
@@ -497,10 +497,13 @@
...
     "ADD_EXPENSE": {
       "ALL_FIELDS_REQUIRED": "Tous les champs sont requis",
+      "FLOW": "Création d'une dépense",
       "MIN_ERROR": "Veuillez saisir une valeur supérieure à ",
       "REQUIRED_ERROR": "Une dépense est requise",
+      "TITLE": "Ajouter une dépense",
       "TITLE_1": "Nouvelle dépense (1 sur 2)",
-      "TITLE_2": "Nouvelle dépense (2 sur 2)"
+      "TITLE_2": "Nouvelle dépense (2 sur 2)",
+      "WHICH_TYPES_TO_RECORD": "Quel type de dépense voulez-vous enregistrer ?"
```
kathyavini, line 506: I think this needs the non-breaking space code \u00a0 before the ? as used elsewhere in the fr translation file or the ? can end up floating alone on mobile, which I think is meant to be avoided:

Screenshot 2023-08-31 at 9 05 55 AM

packages/webapp/src/containers/Finances/selectors.js
```diff
@@ -23,7 +24,33 @@ const selectedSaleSelector = createSelector(financeSelector, (state) => state.se
 
 const expenseSelector = createSelector(financeSelector, (state) => state.expenses);
 
-const expenseTypeSelector = createSelector(financeSelector, (state) => state.expense_types);
+const allExpenseTypeSelector = createSelector(financeSelector, (state) => state.expense_types);
+
+const expenseTypeSelector = createSelector(financeSelector, (state) => {
```
kathyavini, line 29: I like this; it's so readable compared to the taskType version!!

packages/webapp/src/containers/Finances/selectors.js
```diff
@@ -23,7 +24,33 @@ const selectedSaleSelector = createSelector(financeSelector, (state) => state.se
...
+  return state.expense_types?.filter((type) => !type.deleted);
+});
+
+const expenseTypeTileContentsSelector = createSelector(financeSelector, (state) => {
```
kathyavini, line 33: I think I see why this logic has to go here (since the class-based React container means you can't useMemo and put it in the container as done in taskTypes), but is there another reason to prefer using a selector here? Specifically for custom revenue could this filtering logic go back into the container in a useMemo hook? Or should it go into the slice?

kathyavini (commented):
This was so good @SayakaOno!! I am really impressed at how easily you navigated the old and new codebase styles and still managed to refactor out a new pure component in the updated style 👌 

I saw only one thing that I thought needed correcting (in the translation files) and my other notes were just comments or questions!

packages/api/db/migration/20230817213340_soft_delete_other_expense_type.js
SayakaOno: Good catch, thank you!

packages/webapp/public/locales/fr/translation.json
SayakaOno: I wasn't aware of this, thank you!

packages/webapp/src/containers/Finances/selectors.js
SayakaOno: I have been wondering if selector should do this these days! (it will not be just a selector...)
I just followed the pattern of `filteredTaskCardContentSelector` here without taking class/functional component into account.
https://github.com/LiteFarmOrg/LiteFarm/blob/78cd987dc9846c439c31f9c14b2b6092640acdd2/packages/webapp/src/containers/Task/taskCardContentSelector.js#L72-L79

We could create a hook like `useCropCatalogue` (extract useMemo that you mentioned)
https://github.com/LiteFarmOrg/LiteFarm/blob/integration/packages/webapp/src/containers/CropCatalogue/useCropCatalogue.js

For custom revenue, I think the logic can be written in the container.

SayakaOno:
@kathyavini Thank you for reviewing! I have made some updates.

@Duncan-Brain 
I changed the manage custom type page path to `/manage_custom_expense` as discussed this morning. It would be great if you could double check!

kathyavini (approved):
Awesome!

packages/webapp/src/containers/Finances/selectors.js
kathyavini: Okay thank you 🙏  Yes our crashing selectors have gotten me thinking a lot recently about if less logic can go into them...!

SayakaOno: Same!!!!!

Duncan-Brain (approved):
Looks good to me!

PR #2853, opened 2023-08-29
Fixing and running i18n script 

kikemarquez8:
**Description**
The i18n script was failing, I think because the i18n library requires it now to be explicitely a commonjs module. 
The script purpose is to go through all of the files under src/ and check for language tags ( something like **t('TAG.SUBTAG')** and add it to its corresponding place in the translation file. 

So let's say for example you make a change in components/AddNewCrop/index.js
and add a tag 
image

now running the script will add that tag into its corresponding json object in the translation file. The script also orders the tags alphabetically so that's why this change is so large. 

The only thing I would be careful of is that it also removes unused tags (though I think this can be modified) 

Jira link:

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


kathyavini (approved):
All those "MISSING" tags we had been inserting by hand... 🥲 

Thank you. The auto-insert is great.

Duncan-Brain (approved):
Thanks for this! Should definitely be a nice time saver!

SayakaOno (approved):

PR #2857, opened 2023-08-31
LF-3581 Create revenue types table and model and add references to other tables where appropriate

kathyavini:
**Description**

This PR covers the database and model changes necessary to introduce custom revenue types:
1. Creation of the `revenue_type` table and model
2. Permissions for revenue types
3. Addition of revenue types to the sale table and model
4. Pre-population of table with one type (`crop_sale`) which is assigned to all current sale records
5. Tiny automated tests update (to reflect updated schema in saleModel.js)

Jira link: https://lite-farm.atlassian.net/browse/LF-3581

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

There are no UI changes; the migration and rollback were tested in my local database.

**Checklist:**


SayakaOno (approved):
Looks good to me!

packages/api/db/migration/20230831015743_create_revenue_type_table.js
```diff
@@ -0,0 +1,85 @@
...
+  });
+
+  await knex.schema.alterTable('sale', (table) => {
+    table.integer('revenue_type_id').references('revenue_type_id').inTable('revenue_type');
+  });
+
+  // Prepopulate with one revenue type (crop_sale)
+  await knex('revenue_type').insert({
+    revenue_type_id: 1,
+    revenue_name: 'crop_sale',
```
Duncan-Brain, line 36: I think this should be plain english ? 'Crop sale' or something without snake case? I am seeing it I think in both task_type and farmExpenseType.

packages/api/db/migration/20230831015743_create_revenue_type_table.js
```diff
@@ -0,0 +1,85 @@
...
+    table.string('revenue_translation_key').notNullable();
+  });
+
+  await knex.schema.alterTable('sale', (table) => {
+    table.integer('revenue_type_id').references('revenue_type_id').inTable('revenue_type');
+  });
+
+  // Prepopulate with one revenue type (crop_sale)
+  await knex('revenue_type').insert({
+    revenue_type_id: 1,
```
Duncan-Brain, line 35: Did we decide we want to carry forward the (possibly bad) practice of manually setting the id for all tables? Since farmers will be able to add to this table the numbers will not line up past this addition. (unlike permissions which is not farmer add-able)

packages/api/tests/sale.test.js
```diff
@@ -388,6 +388,7 @@ describe('Sale Tests', () => {
             crop_variety_id: cropVariety2.crop_variety_id,
           },
         ],
+        revenue_type_id: 1,
```
Duncan-Brain, line 391: should this be added to mocks.fakeSale()? Or maybe not unless renamed fakeCropSale()..

Also to above point about  setting type id as 1 -- manually setting type id versus retrieving it.. this style will only work for this one default sale type if we add CSA box or Processed or something in the future it could be a bad pattern.

Duncan-Brain (changes requested):
Left a couple of inline comments. 

The only one I think actually needs attention is the revenue_name of crop sale.

packages/api/db/migration/20230831015743_create_revenue_type_table.js
kathyavini: It should indeed!

packages/api/db/migration/20230831015743_create_revenue_type_table.js
kathyavini: You mean since we will not be establishing all our default revenue types now (before introducing the creation of custom revenue types)? If so, yes, in the future (= data migrations that add default revenue types after this release) we will have to use the database auto-increment, and we'll have to just live having our default revenue types have different IDs in each database. I don't love it, but I guess the only other option is uuids, right?

But I'm not sure how it matters for this first record because it's going to be `1` whether we auto-increment or hard code. And since we are setting all existing records to this type I have no issue with hard-coding -- but again I'm still not understanding how it matters for this migration, can you clarify that part?

packages/api/tests/sale.test.js
kathyavini: I'll add a comment. This is a placeholder to not crash the Jest tests; I don't think it makes any sense to touch fakeSale() now, but rather it needs to be updated along with the sale controller (and flow) when we actually have a way to specify which revenue type is associated with a sale. Obviously we do not want to hardcode `1` at that point!

packages/api/db/migration/20230831015743_create_revenue_type_table.js
Duncan-Brain: Just asking the intention there ... 1 will definitely have the intended effect for this first default types migration but no others -- so anyone copying your code in the future to add new values should not hard code the number!

packages/api/tests/sale.test.js
Duncan-Brain: Maybe a note on the Jira ticket?

Duncan-Brain (approved):
Crop Sale! nice

packages/api/tests/sale.test.js
kathyavini: That's what I was thinking -- do you know which one aligns closest with integrating revenue type into sales?

Duncan-Brain: APIs? since it requests jest tests? LF-3584

SayakaOno (approved):

PR #2858, opened 2023-08-31
LF-3582: Refactor the add expense container to updated style

SayakaOno:
**Description**

- create components
   - ExpenseItemInputs
   - ExpenseItemsForType ([useFieldArray](https://react-hook-form.com/docs/usefieldarray) is used)
   - AddExpense component (entire form)
   Screenshot 2023-08-31 at 3 27 15 PM
- create story (http://localhost:6006/?path=/story/page-finance-addexpense--primary)
- move a date util function in RepeatCropPlan to the common date utils file
- update AddExpense container
   - move away from react-redux-form and use react-hook-form in the child components
   - update `handleSubmit` function to take form data from react-hook-form
   - format `date` before calling the post API
   - update `getTypeName` method to show custom types properly

Jira link: https://lite-farm.atlassian.net/browse/LF-3582

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


Duncan-Brain (commented):
I think everything looks good to me -- I noticed just two little things which may have been intentional.

1. Did you change how the date is calculated for expense_date? 

It looks to me like it is now choosing midnight as the time instead of a timestamp. It might be better the way you have it -- but just wondering if its intentional.
Screenshot 2023-09-05 at 5 09 54 PM

2. The second little thing is when I blur the value input without entering a value -- it automatically sets the value at 0.01. This enables the save button which could lead to an accidental save click.

SayakaOno:
@Duncan-Brain 
Thank you for the careful eye on this!

1. I noticed that the default was the current time (`moment()`), but once you change the date, it would be midnight. I found that strange and changed it to always make it midnight.
2. I wasn't really sure about that behaviour too, but I found that was the default behaviour of `<Input />`, so I didn't override it.
Screenshot 2023-09-05 at 4 31 08 PM

Duncan-Brain (approved):
Thank you Sayaka for explaining!

I am guessing the difference was due to if you are recording sales the day they happen the time may be accurate as moment(). But previous days it would be unlikely to know that detail. 

Weird that is the default - I have not seen it on other forms before.

kathyavini (commented):
It looks beautiful and very nice with the `useFieldArray`!

I guess my only question is why not also refactor `AddExpense` into a functional component at this point? I think you did all the difficult work and it might be AI-able level trivial at this point to convert it! 🙂

Unless I'm mistaken I think there aren't going to be any more tickets touching this part of the flow in Finances 2.0? So it's kind of our chance!

SayakaOno:
@kathyavini 
Thank you Joyce! Personally I find it easier to review an existing file seeing the differences than to review a new file, so I usually try to make minimal changes. (but a new file is easier if you are not so familiar with class components?) I didn't see the need for converting it to a functional component this time, but If we make the change, I prefer another PR as a reviewer. If everyone doesn't mind reviewing more changes, I will make changes at one go next time!

kathyavini:
@SayakaOno that's a good point! I usually also I prefer a more explicit diff, but I am definitely one of those who has never seen class-based components before so I find them wildly unreadable 😆 But yes probably a separate conversion ticket/PR is better!

kathyavini (approved):

PR #2859, opened 2023-09-01
LF-3554: Create custom expense types tile view

SayakaOno:
**Description**

**UPDATE**
After we discussed about persistedPaths, I made updates.
- separate two forms. (remove `/manage_custom_expenses` from the first `persistedPaths`)
- when a tile is clicked in `/manage_custom_expenses`, call `setPersistedPaths()`.
- when "+ Add custom expense type" is clicked, call `setPersistedPaths()`.
- wrap the components that are part of the second form with `HookFormPersistProvider`
- call `useHookFormPersist` in `PureSimpleCustomType`
- add `history.listen` in `ManageCustomExpenseTypes` so that the user goes to `/expense_categories` when the user clicks the back button of the form or browser's back button. Since `/expense_categories` is within the first form, call `setPersistedPaths`
- fix the behaviour that tiles are selected in `/manage_custom_expenses`

Original description 🔽 
- create useCustomExpenseTypeTileContents hook
- create ManageExpenseTypes component, container and [story](http://localhost:6006/?path=/docs/form-managecustomtypes--docs)
- update setPersistedPaths called on clicking "Add New Expense" to add `/manage_custom_expense`
- ~~create `addPersistedPaths` action and update persistedPaths for manage custom expense views (`persistedPaths` is updated when "Manage custom expense types" page is opened. Please let me know if this is not a good idea.)~~

Jira link: https://lite-farm.atlassian.net/browse/LF-3554

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/webapp/src/Routes.jsx
```diff
@@ -586,6 +587,7 @@ const Routes = () => {
             <Route path="/expense_detail" exact component={ExpenseDetail} />
             <Route path="/expense_categories" exact component={ExpenseCategories} />
             <Route path="/add_expense" exact component={AddExpense} />
+            <Route path="/manage_custom_expense" exact component={ManageExpenseTypes} />
```
Duncan-Brain, line 590: `/manage_custom_expenses` plural form may be good idea ?

packages/webapp/src/components/Forms/ManageCustomTypes/index.jsx
```diff
@@ -0,0 +1,89 @@
...
+    defaultValues: persistedFormData,
+    mode: 'onChange',
+  });
+
+  useHookFormPersist(getValues);
+  register(customTypeFieldName);
+  const selectedTileId = watch(customTypeFieldName);
+
+  return (
+    <Form>
```
Duncan-Brain, line 48: Do we need this to be a form? 

Right now the manage custom types view only needs to be capable of single-click I think. So the click behaviour shouldn't need to be saved in form format.

packages/webapp/src/components/Forms/ManageCustomTypes/index.jsx
```diff
@@ -0,0 +1,89 @@
...
+          const tileProps = formatTileData(data);
+          const { tileKey } = tileProps;
+          const selected = getIsSelected ? getIsSelected(tileKey) : selectedTileId === tileKey;
+
+          return (
+            <IconLabelTile
+              {...tileProps}
+              key={tileKey}
+              selected={selected}
+              onClick={() => {
```
Duncan-Brain, line 64: So the way I have the readonly and edit views set up is that the click should just lead to `/readonly_custom_expense/:expense_type_id`. Let me know if I did something wrong there.

[PR-2848](https://github.com/LiteFarmOrg/LiteFarm/pull/2848/files)

Duncan-Brain:
So you asked in your description if setting the persisted paths is a good idea -- it might be worth a tech daily chat. I can't claim to know exactly how it works. 

My understanding is that persisted paths is really useful for multi-page form data. I think it may be unnecessary in these cases because unlike adding an expense (which is multi-select) managing a custom type is single-select and a seperate flow altogether, and so does not need to be part of any form (in my opinion). 

So I think much of the persisted paths logic could be deleted and replaced with just a `onclick=history.push(/readonly_custom_expense/${expense_type_id}`

packages/webapp/src/containers/Finances/ManageCustomExpenseTypes/index.jsx
```diff
@@ -0,0 +1,87 @@
...
+  const dispatch = useDispatch();
+
+  const selectedExpenseTypes = useSelector(selectedExpenseSelector);
+
+  const onTileClick = (typeId) => {
+    if (!selectedExpenseTypes.includes(typeId)) {
+      dispatch(setSelectedExpenseTypes([...selectedExpenseTypes, typeId]));
+    }
+    const { readOnly } = getPaths(typeId);
+    history.push(readOnly);
```
SayakaOno, line 45: This is actual `onTileClick`!

packages/webapp/src/components/Forms/ManageCustomTypes/index.jsx
SayakaOno: I tested it and thought it was working perfect!
Actual page navigation is [here](https://github.com/LiteFarmOrg/LiteFarm/pull/2859/files#r1316520904).

packages/webapp/src/components/Forms/ManageCustomTypes/index.jsx
SayakaOno: Definitely not. I didn't check what we had in the component for task type and just copied it...😅🙏

SayakaOno:
@Duncan-Brain
Thank you! I will look into how persisted paths really works and bring it up in tech daily (maybe next week?).

packages/webapp/src/components/Forms/ManageCustomTypes/index.jsx
Duncan-Brain: Sorry yes I saw the other function after the previous comment! It does indeed go where needed. I think I was thrown off by the complexity of the persisted paths. It seemed like it could possibly be a simple prop to pass down. Maybe [here](https://github.com/LiteFarmOrg/LiteFarm/pull/2859/files/3eb45a65ce3fd9eeac4eda6506e1488546318418#r1317323400) is what my brain thought of

SayakaOno: Ah understood!
I'll do some more experiment with your branch #2848 after the conflicts are resolved. (I will be working on something else, so no need to rush!)

packages/webapp/src/containers/Finances/ManageCustomExpenseTypes/index.jsx
```diff
@@ -0,0 +1,91 @@
...
+
+    history.push(readOnly);
+  };
+
+  useEffect(() => {
+    // Manipulate page navigation by pushing "/expense_categories" on top of "/Finances".
+    // When browser's back button or form's back button is clicked, we want to
+    // navigate the user to "/expense_categories" not "/Finances".
+    const unlisten = history.listen(() => {
+      if (history.action === 'POP' && history.location.pathname === '/Finances') {
```
kathyavini, line 48: For reasons that I don't fully understand, I get the double-back bug _sometimes_ (I think it depends on previous pathing??) after delete or edit only when this `history.location.pathname === '/Finances'` check is included. I don't seem to get it when the check is not included (nor do I see any other bugs) but I wanted to check if you saw the same. Is it absolutely necessary?

kathyavini (approved):
Yay it feels so good to complete the flow! And thank you for fixing the tile highlight behaviour!!

The only issue I had was with our favourite thing -- pathing! I was seeing the double-back bug (which we have in other places so I'll approve anyway) with the more elaborated `history.listen()` code -- could you check if you see the same?

packages/webapp/src/containers/Finances/ManageCustomExpenseTypes/index.jsx
SayakaOno: The double-back bug is seen after completing the form (edit, create, retire).
When the user comes to `/manage_custom_expenses` from `/home` directly for some reason and click the back button after completing the form, we want the user to go to `/home`, not to be forced to go '/expense_categories`. So the check is for that case.
(If we need to fix the double-back bug, I could add some code here, but I would prefer to fix the hook...)

By the way, I see a strange behaviour after editing the type. I'm redirected to the edit view... I'm looking into it.

kathyavini: I think I saw that bug too! There was also an extra page refresh. I recorded it because it was so strange but then a reload fixed it!

https://github.com/LiteFarmOrg/LiteFarm/assets/82857964/4fa0dfa2-846b-41b7-b4bf-029ec544fe01

SayakaOno: It was a form button's submit again that adds "?" to the URL.
I added a type to the button and believe it's working now!

kathyavini: Ahhh that pesky submit!!

kathyavini: I think it can end up being a **3x** back button situation if you go Add and then Edit the one you just added 😅  But the behaviour is smooth and the routing is otherwise 👌

SayakaOno: 3x back button is quite annoying 😅 Why don't we create a function which works the same as `historyCancel` and always call it when completing forms in sagas instead of `history.push(<the previous page of form's first page>)`?

kathyavini: Can it be done?? I have had no success trying to interact with the hookFormPersist stuff from sagas!

SayakaOno: Sorry 😣 I meant something like this:
 
Screenshot 2023-09-15 at 2 42 04 PM
Screenshot 2023-09-15 at 2 42 18 PM

This seems to work fine...

packages/webapp/src/containers/Finances/ManageCustomExpenseTypes/index.jsx
```diff
@@ -0,0 +1,91 @@
...
+import PureManageCustomTypes from '../../../components/Forms/ManageCustomTypes';
+import { setPersistedPaths } from '../../hooks/useHookFormPersist/hookFormPersistSlice';
+import { icons } from '../NewExpense/ExpenseCategories';
+import labelIconStyles from '../NewExpense/ExpenseCategories/styles.module.scss';
+import useCustomExpenseTypeTileContents from '../useCustomExpenseTypeTileContents';
+
+const addCustomTypePath = '/add_custom_expense';
+
+const getPaths = (typeId) => ({
+  readOnly: `/readOnly_custom_expense/${typeId}`,
```
Duncan-Brain, line 27: readonly should be all lower case

Duncan-Brain (approved):
Looks great ! Cool stuff with the persisted paths -- while I don't yet fully understand.. I definitely now better understand! 

I have approved and if someone wants to take on the case issue in a future PR I am good with that!

PR #2861, opened 2023-09-06
LF-3586 make the revenue type slice and add the revenue types selector and sagas

kathyavini:
**Description**

This PR creates the new Redux Toolkit (RTK) slice and the saga functions for revenue types, corresponding to the db table created in PR #2857.

I took the custom expense type API PR (#2849) by @gursimran-singh as the model for how I expect the new API endpoints to look.

Jira link: https://lite-farm.atlassian.net/browse/LF-3586

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

I tested these rendering by the selectors in the client (with some dummy JSX) and some temporary API routes on the backend -- please ping on Slack if you would like the dummy code for frontend or backend. Ideally these should be tested with the actual controllers and API 🙂 

**Checklist:**

### DIFF
--- a/packages/webapp/src/containers/Finances/saga.js
+++ b/packages/webapp/src/containers/Finances/saga.js
@@ -35,6 +35,7 @@ import i18n from '../../locales/i18n';
 import history from '../../history';
 import { enqueueErrorSnackbar, enqueueSuccessSnackbar } from '../Snackbar/snackbarSlice';
 import { createAction } from '@reduxjs/toolkit';
+import { getRevenueTypesSuccess } from '../revenueTypeSlice';
 
 export function* getSales() {
   const { salesURL } = apiConfig;
@@ -240,6 +241,23 @@ export function* tempEditExpenseSaga(action) {
   }
 }
 
+export const getRevenueTypes = createAction('getRevenueTypesSaga');
+
+export function* getRevenueTypesSaga() {
+  const { revenueTypeUrl } = apiConfig;
+  let { user_id, farm_id } = yield select(loginSelector);
+  const header = getHeader(user_id, farm_id);
+
+  try {
+    const result = yield call(axios.get, `${revenueTypeUrl}/farm/${farm_id}`, header);
+    if (result) {
+      yield put(getRevenueTypesSuccess(result.data));
+    }
+  } catch (e) {
+    console.log('failed to fetch revenue types from database');
+  }
+}
+
 export const patchEstimatedCropRevenue = createAction(`patchEstimatedCropRevenueSaga`);
 export function* patchEstimatedCropRevenueSaga({ payload: managementPlan }) {
   const { managementPlanURL } = apiConfig;
@@ -267,6 +285,7 @@ export default function* financeSaga() {
   yield takeLeading(ADD_OR_UPDATE_SALE, addSale);
   yield takeLatest(GET_EXPENSE, getExpenseSaga);
   yield takeLatest(GET_FARM_EXPENSE_TYPE, getFarmExpenseTypeSaga);
+  yield takeLatest(getRevenueTypes.type, getRevenueTypesSaga);
   yield takeLeading(ADD_EXPENSES, addExpensesSaga);
   yield takeLeading(DELETE_SALE, deleteSale);
   yield takeLeading(DELETE_EXPENSES, deleteExpensesSaga);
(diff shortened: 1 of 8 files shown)


packages/webapp/src/containers/Finances/saga.js
```diff
@@ -240,6 +246,82 @@ export function* tempEditExpenseSaga(action) {
...
+export const addCustomRevenueType = createAction('addRevenueTypeSaga');
+
+export function* addRevenueTypeSaga({ payload: { revenue_name } }) {
+  const { revenueTypeUrl } = apiConfig;
+  let { user_id, farm_id } = yield select(loginSelector);
+  const header = getHeader(user_id, farm_id);
+
+  const body = {
+    revenue_name,
+    farm_id: farm_id,
```
Duncan-Brain, line 290: I think this may no longer be needed and headers is used.

packages/webapp/src/containers/revenueTypeSlice.js
```diff
@@ -0,0 +1,119 @@
...
+    'farm_id',
+    'deleted',
+  ]);
+};
+
+const addManyRevenueTypes = (state, { payload: revenueTypes }) => {
+  state.loading = false;
+  state.error = null;
+  state.loaded = true;
+  revenueTypeAdapter.upsertMany(
```
Duncan-Brain, line 35: This copies the taskTypeSlice but for some reason I thought we figured that get requests should be setAll ?

Duncan-Brain:
Added a couple of minor inline comments. 

Mainly curious about how you felt about this being broken up from the containers. It seemed like a good idea at the time but since there is not much to test on this is kind of a Pure Code review -- do you think this or part of this should be lumped into another ticket? and if yes which one?

packages/webapp/src/containers/Finances/saga.js
kathyavini: I know we _just_ discussed this last week with Gursimran but I've already forgotten the details! As far as I can see it's going to still be needed in the POST for `hasFarmAccess` (right @gursimran-singh?) Or at least that's how it is currently still set up in the expense PR:

```
router.post(
  '/',
  hasFarmAccess({ body: 'farm_id' }),
  checkScope(['add:expense_types']),
  farmExpenseTypeController.addFarmExpenseType(),
);
```

Did I miss something from our discussion Friday?

Duncan-Brain: Oh yea I might have missed that piece.

packages/webapp/src/containers/revenueTypeSlice.js
Duncan-Brain: Yeah I guess I thought we figured that the way you found for task and managementPlan was the best way forward.

kathyavini: I'm actually thinking of backtracking on both `task` and `managementPlan` in favour of upsert with soft delete, given the crashing selectors! I'll organize my thoughts on this and bring it up in tech daily for discussion.

kathyavini:
> Mainly curious about how you felt about this being broken up from the containers. It seemed like a good idea at the time but since there is not much to test on this is kind of a Pure Code review -- do you think this or part of this should be lumped into another ticket? and if yes which one?

@Duncan-Brain I thought about this a lot since while working the sagas since I wanted to test the whole flow too. I didn't mind it being broken up from the containers because throwing some selectors and dispatch into a frontend container is trivial but yes right, it can't really be tested then unless the PR reviewer wants to do the same too (or has a container ticket in progress to test with). Having to make an API was a pain though; ideally the API should come first.

Not sure what is best. I somewhat like the idea of different people tackling each part though because I think it will make review easier and more effective with multiple people looking at how/where things are joined together.

Duncan-Brain (approved):
Cool yeah it looked like you made all the things you needed there so maybe it becomes more work. But also see the point about helps with review hmm okay. 

Approving since you are good with your upsertMany and farm_id

packages/webapp/src/containers/Finances/saga.js
gursimran-singh: @kathyavini yes, `hasFarmAccess` middleware needs this to check farm access.

SayakaOno (approved):
Looks good to me!

PR #2870, opened 2023-09-14
LF-3627 Update the edit individual expense flow

kathyavini:
**Description**

This PR creates the new readonly and edit flow for expense. The flow utilizes one container and one component for both read-only and edit, but two routes.

Notes:
- The original flow used selectors and Redux manipulation to store the target expense; the new flow follows the current pattern of storing the `expense_id` in the URL parameter.
- I removed a fair amount of code, including the old Redux actions, reducers, and selectors that handled saving the current expense to the store
- The edit and delete sagas were already functional. I removed their prefix "TEMP", which it appears they have had since 2021
- The class-based container was made into a functional one
- I tried to harmonize the value Input between Add Expense, Edit Expense, and Add Sale by modelling the newer two after Add Sale; note this doesn't fix the Portuguese-language bug in Add Sale that David showed us; there are now three tickets in the technical document to cover Input behaviour to address next sprint
- Storybook for the Pure Component is viewable at: [PureExpenseDetail](http://localhost:6006/?path=/story/page-finance-pureexpensedetail)

Jira link: https://lite-farm.atlassian.net/browse/LF-3627

**Type of change**

**How Has This Been Tested?**

- [x] Passes test case (Storybook)
- [ ] Other (please explain)

**Checklist:**


packages/webapp/src/components/Finances/PureExpenseDetail/index.jsx
```diff
@@ -0,0 +1,177 @@
...
+    defaultValues: {
+      [NOTE]: expense.note,
+      [DATE]: getDateFromDateTimeString(expense.expense_date),
+      [TYPE]: expenseTypeReactSelectOptions.find(
+        (option) => option.value === expense.expense_type_id,
+      ),
+      [VALUE]: expense.value,
+    },
+  });
+  const readonly = view === 'read-only' || false;
```
SayakaOno, line 62: `|| false` could be removed 😉

packages/webapp/src/components/Finances/PureExpenseDetail/index.jsx
```diff
@@ -0,0 +1,177 @@
...
+        control={control}
+        name={TYPE}
+        rules={{ required: true }}
+        render={({ field: { onChange, value } }) => (
+          <ReactSelect
+            label={t('EXPENSE.TYPE')}
+            options={expenseTypeReactSelectOptions}
+            onChange={(e) => {
+              onChange(e);
+            }}
```
SayakaOno, line 107: This could be `onChange={onChange}`

packages/webapp/src/containers/Finances/ExpenseDetail/index.jsx
```diff
@@ -1,222 +1,103 @@
...
+
+        return {
+          value: type.expense_type_id,
+          label: type.farm_id
+            ? type.expense_name + retireSuffix
+            : t(`expense:${type.expense_translation_key}`),
+        };
+      }
+    })
+    .filter(Boolean);
```
SayakaOno, line 60: I think we could do `expenseTypeTileContentsSelector` (sorry for the bad name) + `expenseTypeByIdSelector` (if the type is deleted one) and map...?

packages/webapp/src/containers/Finances/ExpenseDetail/index.jsx
```diff
@@ -1,222 +1,103 @@
...
-  return {
-    dispatch,
+  const handleGoBack = () => {
+    history.back();
   };
+
+  return (
+    expense && (
+      <PureExpenseDetail
+        key={isEditing ? 'editing' : 'readonly'} // remount the component
```
SayakaOno, line 89: Which trait is used for remounting? Does it behave differently compared to passing `view={isEditing ? 'edit' : 'read-only'}`?

SayakaOno (commented):
I think it's working great!
I added a few comments and a question :)

packages/webapp/src/components/Finances/PureExpenseDetail/index.jsx
kathyavini: HA! Thank you

packages/webapp/src/containers/Finances/ExpenseDetail/index.jsx
kathyavini: My first thought was to use a map! But then got nervous because I've never actually used one before 😆 I will for sure try it!!

packages/webapp/src/containers/Finances/ExpenseDetail/index.jsx
kathyavini: It's the `key` prop specifically! So if you pass just view and remove this line, the uncontrolled form inputs (having not unmounted) will store their value when you hit the back button, even if you don't hit submit! It's for that behaviour specifically I added the key.

packages/webapp/src/components/Finances/PureExpenseDetail/index.jsx
kathyavini: I think I'm going to remove it from the source component where that code comes from, too.

kathyavini: Oops deleted a comment, but... I'm thinking I should remove it from the source component where that code comes from too?

packages/webapp/src/containers/Finances/ExpenseDetail/index.jsx
SayakaOno: Sorry for the lack of the explanation, the result of `sort()` and `filter()` is the same as not deleted types + the type of the current expense (if the type is retired):
```
[
   ...useSelector(expenseTypeTileContentsSelector),
   useSelector(expenseTypeByIdSelector(expense.expense_type_id))
]
```
correct?

So you can do:
```
theArrayAbove.map((type) => {
    const retireSuffix = type.deleted ? ` ${t('EXPENSE.EDIT_EXPENSE.RETIRED')}` : '';

     return {
       value: type.expense_type_id,
       label: type.farm_id
         ? type.expense_name + retireSuffix
        : t(`expense:${type.expense_translation_key}`),
    };
})
```
My intention was to remove the loops, sort and filter!

kathyavini: Okay, I think that's about the same as what I figured (I was hoping to use a new datatype but I guess it's not necessary!) and yes it's WAY shorter!

But you would still need the deleted check right before appending to the array?

```js
  const expenseTypeReactSelectOptions = [
    ...sortedExpenseTypes,
    currentExpenseType.deleted && currentExpenseType,
  ].map((type) => {
    // as above
  });
```

packages/webapp/src/containers/Finances/ExpenseDetail/index.jsx
SayakaOno: Oh I see! It's tricky to reuse the container 🤣 Wise solution!

packages/webapp/src/containers/Finances/ExpenseDetail/index.jsx
SayakaOno: Mine was just an example so you can chain `map`!
Other than that, the code above looks great!

SayakaOno (approved):
Looks good!

packages/webapp/src/Routes.jsx
```diff
@@ -603,7 +600,8 @@ const Routes = () => {
             <Route path="/estimated_revenue" exact component={EstimatedRevenue} />
             <Route path="/labour" exact component={Labour} />
             <Route path="/other_expense" exact component={OtherExpense} />
-            <Route path="/expense_detail" exact component={ExpenseDetail} />
+            <Route path="/expense/:expense_id/" exact component={ExpenseDetail} />
```
Duncan-Brain, line 603: Is the trailing backslash needed?

Duncan-Brain (approved):
Looks great! I love a negative code PR!

Its approved so if you think the following is not important to address right now if you could merge it @kathyavini 

- Saw two little things I was unsure about I left inline comments (an extra backslash on the route and a double parseFloat)
- Do you like the naming for "...ExpenseDetail" vs ReadOnly/Edit or some such?
- PureExpenseDetail looks like it will be able to be reused possibly for Revenues

kathyavini:
@Duncan-Brain Actually the naming of the read-only view is the only thing I have a _very_ strong feeling about -- I wanted to avoid using "read-only" at all costs 😆 Reason is that it's already a decision point each time it's referred to as how to name the files and imports (`readonly` vs `readOnly` vs `read-only` vs `read_only`) and we already saw the downside of that that in the TypeError that crashed beta deploy, and the cleaning up still to be done in routes.

Calling the non-editable view "details" is a pattern elsewhere in the app (and also the existing one here anyway) so I think it's a safe bet. And that way we avoid the cognitive overhead of having to search and confirm each time that we are using the correct naming convention, or the danger of crashing the app when we mess it up.

packages/webapp/src/Routes.jsx
kathyavini: No, it's not! Thanks! It's a holdover from when the endpoint was `/:expense_id/details`. I'll make a note to fix that in a subsequent PR (router accepts it works but it's technically incorrect).

packages/webapp/src/containers/Finances/ExpenseDetail/index.jsx
```diff
@@ -1,222 +1,96 @@
...
+        : t(`expense:${type.expense_translation_key}`),
+    };
+  });
+
+  const handleSubmit = (formData) => {
+    let data = {
+      expense_date: moment(formData.date),
+      expense_type_id: formData.type.value,
+      note: formData.note,
+      value: parseFloat(parseFloat(formData.value).toFixed(2)),
```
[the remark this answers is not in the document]
kathyavini: I suppose maybe if the `formData.value` couldn't be coerced into a number the inner is purposeful? I didn't touch this part of the flow (just moved it from `TempEditExpense` so I'll probably leave it, having not tested all combinations. But my impulse would be to write it with just the outer one.

Duncan-Brain:
Cool! Yeah I agree readonly is not great, I am just thinking about the names if I need to copy the pattern. Its weird to be both readonly and edit. I think I am on board with details vs edit.

Going to merge this then!

PR #2875, opened 2023-09-16
First version of end to end tests

mika-robots:
**Description**

Jira link:

**Type of change**
- Non breaking change that enable automated cypress tests
- There are no changes to any files outside of end-to-end folder

**How Has This Been Tested?**

I run it both locally and on my branch

**Checklist:**

- [ x] I have commented my code, particularly in hard-to-understand areas
- [ x] My changes generate no new warnings
- [ x] I have added tests that prove my fix is effective or that my feature works
- [ x] New and existing unit tests pass locally with my changes
- [ x] The precommit and linting ran successfully
- [ x] I have added or updated language tags for text that's part of the UI
- [ x] I have added "MISSING" for all new language tags to languages I don't speak
- [ x] I have added [the GNU General Public License](https://lite-farm.atlassian.net/l/cp/BT0Dd7WW) to all new files


.github/workflows/cypress_tests.yml
```diff
@@ -54,51 +53,52 @@ jobs:
...
           PORT: 5000
 
-      - name: Install pnpm
-        uses: pnpm/action-setup@v2
+      - uses: pnpm/action-setup@v2
         with:
           version: latest
-
-      - name: Run and test app
+      - name: Run Webapp
```
MwayaB, line 59: I wonder here if adding a step to build the webapp, would ensure our product under test more closely resembles what is on production

packages/end-to-end/cypress/e2e/specs/01-onboarding.js
```diff
@@ -0,0 +1,35 @@
...
+            user.name,
+            user.language,
+            translation['SLIDE_MENU']['CROPS'],
+          );
+        },
+      );
+    });
+  });
+
+  it.only('01_Onboard new user if not already created', () => {});
```
antsgar, line 34: Could we remove the only here to ensure all tests run if we ever add any new ones to this file? 🙏

packages/end-to-end/cypress/e2e/specs/02-farm_people.js
```diff
@@ -0,0 +1,144 @@
...
+          cy.get('[data-cy=continueGoogle]').should('exist');
+          cy.loginOrCreateAccount(
+            user.email,
+            user.password,
+            user.name,
+            user.language,
+            translation['SLIDE_MENU']['CROPS'],
+          );
+        },
+      );
```
antsgar, line 31: This block of code seems to be shared across all the different test files. Could we abstract this to a custom command?
I'm thinking that this block here is actually what's reused between tests, while `loginOrCreateAccount`, `addFarm` and `onboardCompleteQuestions` are not reused in isolation, but just in the scope of this block. I'd suggest making those last three regular functions instead, and having this whole "user/farm setup" be the reusable command. Thoughts?

packages/end-to-end/cypress/support/commands.js
```diff
@@ -1,25 +1,159 @@
...
+        // cy.get('[id$=react-select-3-input]').type(language+'{enter}');
+        cy.contains('English')
+          .parent()
+          .find('input')
+          .type(language + '{enter}');
+
+        cy.get('[data-cy=createUser-password]').type(password);
+        cy.get('[data-cy=createUser-create]').should('exist').and('be.enabled').click();
+
+        cy.intercept('POST', '**/user').as('createUser');
```
antsgar, line 54: Is the intention behind this intercept to wait for the request to resolve? I think we'd need to add `cy.wait(@createUser)` below this line for it to wait if that's the case.

packages/end-to-end/cypress/e2e/specs/01-onboarding.js
```diff
@@ -0,0 +1,35 @@
+describe.only('Onboarding', () => {
+  let users;
+  let translation;
+
+  beforeEach(() => {
+    // Load the users fixture before the tests
+    cy.fixture('e2e-test-users.json').then((loadedUsers) => {
+      users = loadedUsers;
+      const user = users[Cypress.env('USER')];
```
antsgar, line 9: Where would this env variable come from?

packages/end-to-end/cypress/support/commands.js
```diff
@@ -1,25 +1,159 @@
...
+  'loginOrCreateAccount',
+  (email, password, fullName, language, crop_menu_name) => {
+    //Login page
+    cy.get('[data-cy=email]').type(email);
+    cy.intercept('GET', '**/login/user/' + email, (req) => {
+      delete req.headers['if-none-match'];
+    }).as('emailLogin');
+    cy.get('[data-cy=continue]').should('exist').and('be.enabled').click();
+
+    cy.waitUntilAnyVisible('[data-cy=createUser-email]', '[data-cy=enterPassword-password]');
```
antsgar, line 38: Instead of waiting for any of the elements to be visible here, we could wait for the above request to resolve, `cy.wait(@emailLogin)`

.github/workflows/cypress_tests.yml
```diff
@@ -54,51 +53,52 @@ jobs:
...
+        with:
+          name: logs
+          path: packages/api/logs
 
       - name: set code coverage badge
-        run: npx -p check-code-coverage update-badge --from '${GITHUB_WORKSPACE}/packages/webapp/coverage/coverage-summary.json'
+        working-directory: "."
+        run: npx -p check-code-coverage update-badge --from '${GITHUB_WORKSPACE}/packages/end-to-end/coverage/coverage-summary.json'
         env:
           GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
```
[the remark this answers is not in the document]
mika-robots: Looks like we don't need to set it up it should automatically be created. 

https://docs.github.com/en/actions/security-guides/automatic-token-authentication

I am not sure if it is failing for me because of my permissions on GitHub. Will try on my own fork.

packages/end-to-end/cypress/e2e/specs/01-onboarding.js
mika-robots: This variable is coming from command line argument or CI workflow. Essentially it indicates which user from fixture to use. The intent is to call cypress command multiple times with different users (e.g. Spanish, English, etc)

Duncan-Brain:
It looks like a great start to me -- I noticed that some scope may have changed to include locations and tasks since you showed us in person. If it is causing issues I would vote to roll back to what was working well before and you could tackle increased coverage in a separate PR.

In case it is not obvious I would ask to remove commented code, and revert the Action file to use the environment secrets and move the run scope back to integration -- but this is probably your last commit once tests pass again. 

Main question I had was:
- should we be just be mocking Google maps interactions? Mocks are hard but testing interactions with maps may also be quite hard.

PR #2877, opened 2023-09-18
Lf 3584 create apis for revenue types

gursimran-singh:
**Description**

To create Api's to add, edit, delete custom revenue types. Following are the response types of each endpoint.

1. `POST /revenue_type` to create new expense type

- If for combination of revenue name and farm id exists and is deleted, then no new record is created and deleted is set to false for the existing record. 204 status code is sent in this case.
- If for above combination, no record is present in DB, then new revenue type is created and 201 status code is sent.

2. `PUT /revenue_type/[revenue_type_id]` to update revenue type

- While editing, if we change custom revenue type name to which is already present for that particular farm, then 409 status code is sent.
- On successful update 204 status code is sent.
- If tried to update a deleted record, then 404 status code is sent.

3. `DELETE /revenue_type/[revenue_type_id]` to delete revenue type

- If delete is successful, then 200 status code is send.
- If revenue_type_id is wrong or already deleted, then 404 status code is sent.

Jira link: https://lite-farm.atlassian.net/browse/LF-3584

**Type of change**

**How Has This Been Tested?**

Test cases written to support each case such as Owners, Managers can add, edit, delete revenue type in their own farms and workers or unauthorized users cannot add, edit, delete revenue type.

- [x] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/api/tests/mock.factories.js
SayakaOno, line 1: Could we remove commented out code?

packages/api/src/controllers/revenueTypeController.js
```diff
@@ -0,0 +1,188 @@
...
+          revenue_name: data.revenue_name,
+          farm_id,
+        });
+
+        // if record exists in db
+        if (record) {
+          // if not deleted, means it is a active revenue type
+          // throw conflict error
+          if (record.deleted === false) {
+            return res.status(409).send();
```
kathyavini, line 40: Sorry for missing it in the previous PR but where the transactions are explicitly started with `transaction.start` they need to be committed or rolled back in every logical branch or hanging transactions will freeze the server)

(See PR https://github.com/LiteFarmOrg/LiteFarm/pull/2880)

packages/api/src/controllers/baseController.js
```diff
@@ -190,5 +190,40 @@ export default {
...
+   * Check if records exists in table
+   * @param {object} model - Database model instance
+   * @param {object} where - 'Where' condition to fetch record
+   * @param {object} whereNot - 'WhereNot' condition to fetch record
+   * @returns {Promise} - Object DB record promise
+   */
+  existsInTable(model, where, whereNot = {}) {
+    let query = model.query().context({ showHidden: true }).where(where);
+
+    if (Object.keys(whereNot)) {
```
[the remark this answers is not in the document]
gursimran-singh: Thanks for finding out. Done

packages/api/src/controllers/revenueTypeController.js
```diff
@@ -0,0 +1,188 @@
...
+          } else {
+            // if its deleted, them make it active
+            record.deleted = false;
+            await baseController.put(RevenueTypeModel, record.revenue_type_id, record, req, {
+              trx,
+            });
+            await trx.commit();
+            res.status(201).send(record);
```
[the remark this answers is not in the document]
gursimran-singh: Done. Now it would return status 204 as per description.

packages/api/src/controllers/revenueTypeController.js
```diff
@@ -0,0 +1,188 @@
...
+        return res.status(400).json({
+          error,
+        });
+      }
+    };
+  },
+
+  delType() {
+    return async (req, res) => {
+      // debugger;
```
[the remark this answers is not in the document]
gursimran-singh: Removed.

packages/api/tests/revenue_type.test.js
```diff
@@ -0,0 +1,436 @@
...
+ *  LiteFarm is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  LiteFarm is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ */
```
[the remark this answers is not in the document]
gursimran-singh: Done.

packages/api/src/controllers/revenueTypeController.js
gursimran-singh: Thanks for finding out that. I have added relevant code where needed to close the transaction before sending response and to be consistent with current design.

But, i strongly believe that transactions do not provide any benefits over single database operations (add, update, delete) and i'm still scratching my head to find the purpose of using transactions in read queries, and that too from a single table as authored [here](https://github.com/LiteFarmOrg/LiteFarm/pull/2880/commits/035a752322f1e4359b95cb1d4452026be612510d). If a transaction fails, will it un-read the read 🤔?

[Here](https://github.com/Vincit/objection.js/issues/809) is a healthy discussion on official `objection.js` repo were user is having similar problem and wants to have a transaction in every query. It is not related exactly to our case but discussion explains that we might not need them for single DB operations.

SayakaOno (approved):
Looks good, thank you!!

packages/api/src/controllers/revenueTypeController.js
kathyavini: We had this conversation yesterday and I was on your side 😄 I came around only reluctantly supposing we might want to lock the record to other users between when it's read and when it's updated, on the off-chance that two users were trying to query for and `insert/update` absolutely simultaneously. But the difference in time the record is locked by making the `query` and `patch/insert` a single transaction (instead of their database native two) must be so tiny it's basically a theoretical benefit. But still, that's only the conceivable benefit I could fathom! 🤷 Yes rolling back a query is meaningless 😁 

It was at least interesting to learn a bit about transaction behaviour in the process. Thank you for the link!

kathyavini (approved):
Dhanwaad!

packages/api/src/controllers/revenueTypeController.js
Duncan-Brain: One of the other things we discussed was the number of transactions made. 

If you make the trx object then by definition you are making 2 transactions (one for the queries that use trx, and one for the queries it is not passed to). This [article at the very end](https://gregberge.com/blog/knex-transactions) shows you how to set the pool to min1 max 1. If you use that setting and test the two apis the add expense type that passes the trx will succeed and the add revenue types one that does not will fail as it exceeds the number of transactions (which max is 1). You may need to restart the db in between api calls to get a clean response since our app may have other transactions going on.

If you removed the startTransaction altogether from this endpoint that is fine and it will succeed.

Main point is: my understanding is we do not want to make excess transactions.

kathyavini: That's a good point @Duncan-Brain. I wasn't thinking about filling up the pool but that's right, two transactions made sequentially is not the same from the perspective of the pool as opening one transaction and then creating a separate one before concluding the first.

Which doesn't itself support of the need for a transaction, but definitely supports the need to pass `trx` if a transaction has been started 👍

gursimran-singh: @Duncan-Brain 
>If you make the trx object and dont pass it to one query then by definition you are making 2 transactions (one for the queries that use trx, and one for the queries it is not passed to).

Transactions in `Objection.js` are explicit in nature and not implicit. Queries does not start transaction on their own. If we need transaction, then we have to start them and then pass to other queries which we wish to be part of that transaction, and if we miss, then queries won't start a new transaction. [This comment support this](https://github.com/Vincit/objection.js/issues/809#issuecomment-366976046). 

>if you use that setting and test the two apis the add expense type that passes the trx will succeed and the add revenue types one that does not will hang as it exceeds the number of transactions (which max is 1)

That primarily happened because in certain cases the response was sent without closing the started transaction, not because a new transaction was started by the query in the process. Code was updated to handle that when Joyce found this.

I totally agree with setting the pool to min1 max 1 in the development environment to find out all the un-handled transaction and close them properly to avoid hanging of server which need a restart. 

@kathyavini 
> I came around only reluctantly supposing we might want to lock the record to other users between when it's read and when it's updated, on the off-chance that two users were trying to query for and insert/update absolutely simultaneously. 

If this is our concern, which is a genuine one 👍 , that other process should not be able to modify table records while the operation that started first has not concluded yet, then transaction alone might not be the answer as other process can modify the records while it is still running. Transactions are atomic in nature but doesn't provide locking of tables by default. For that we need to use [forUpdate](https://vincit.github.io/objection.js/api/query-builder/find-methods.html#forupdate), read more [here](https://github.com/Vincit/objection.js/issues/57)

PR #2879, opened 2023-09-19
LF-3585: Create custom revenue types tile view

SayakaOno:
**Description**

Create the custom revenue types tile view by following the same pattern for the expense types.
- create `useSortedCustomRevenueTypes` hook
- create `ManageCustomRevenueTypes` container
- fix the readonly custom expense path (Thank you for your [comment](https://github.com/LiteFarmOrg/LiteFarm/pull/2859#discussion_r1329283248) Duncan!)
- add CSS to scss file for `Tiles` to bold Icon tile label
   - update some files imports to use it
   - delete one scss file

Jira link: https://lite-farm.atlassian.net/browse/LF-3585

**Type of change**

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

**How Has This Been Tested?**

- [ ] Passes test case
- [x] UI components visually reviewed on desktop view
- [ ] UI components visually reviewed on mobile view
- [ ] Other (please explain)

**Checklist:**

- [ ] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [x] The precommit and linting ran successfully
- [ ] I have added or updated language tags for text that's part of the UI
- [x] I have added "MISSING" for all new language tags to languages I don't speak
- [x] I have added [the GNU General Public License](https://lite-farm.atlassian.net/l/cp/BT0Dd7WW) to all new files

PR #2881, opened 2023-09-19
LF-3586 Add pathing to manage_custom_revenues

kathyavini:
**Description**

This completes the custom revenue types sagas by adding the pathing to the container newly created in:

- https://github.com/LiteFarmOrg/LiteFarm/pull/2879

Jira link: https://lite-farm.atlassian.net/browse/LF-3586

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/containers/Finances/saga.js
+++ b/packages/webapp/src/containers/Finances/saga.js
@@ -342,6 +342,7 @@ export function* deleteRevenueTypeSaga({ payload: id }) {
 
     yield put(deleteRevenueTypeSuccess(id));
     yield put(enqueueSuccessSnackbar(i18n.t('message:REVENUE_TYPE.SUCCESS.DELETE')));
+    history.push('/manage_custom_revenues');
   } catch (e) {
     yield put(enqueueErrorSnackbar(i18n.t('message:REVENUE_TYPE.ERROR.DELETE')));
   }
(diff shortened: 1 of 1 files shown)


SayakaOno (commented):
Thank you Joyce, this looks good!

A few things I noticed:
- I used Gursimran's branch for the first time, and I couldn't create a new type on the first attempt. (is it the index issue?)
  Screenshot 2023-09-19 at 3 13 04 PM
- The API to get types is now `/revenue_type`. Do you think we should make it the same as the one for expense types or just update the saga?
   https://github.com/LiteFarmOrg/LiteFarm/blob/53eaa0105c6fa45784b2d60754a354ddeb7b74bd/packages/api/src/routes/revenueTypeRoute.js#L29

kathyavini:
@SayakaOno It's the index issue!! I don't know how we want to address solving it. Could use a migration, could... just insert twice I guess?! Do we all see it only once? (As in, did subsequent inserts work for you?)

> Do you think we should make it the same as the one for expense types or just update the saga?

Not sure that I follow this. Both `expense_type` and `revenue_type` as API routes are in the singular on both frontend and backend so I don't think either saga needs updating.

Screenshot 2023-09-19 at 3 36 38 PM


The check permissions table entry (from your screenshot) IS in the plural (that seemed to be the convention in that table) but there is no enforced relationship between the permissions table and the routes, they just loosely refer to each other 🙂

SayakaOno:
The index issue was seen on the second attempt right after the first, too. After a while, it worked! I did some research but I could not find a good answer... my wild solution would be to delete revenue types (as well as associated sales) in the table manually, and write a migration to insert the type without `id`? I don't really know what to do...!

No! It's my bad, let's discuss it tomorrow :)

SayakaOno (approved):

Duncan-Brain (approved):

PR #2888, opened 2023-09-22
LF-3623 Build saga, api route, controller and utility function to generate and return an xlsx file

kathyavini:
**Description**

This PR introduces the code skeleton for generating downloadable financial reports and saving the fact of the download (to evaluate user interaction & uptake) to a new database table. There is an interim UI (a big button) and an interim Excel report (with just the three values in the old finance homepage "summary" area).

The following can then be added in future PRs:
- filtering the transaction data that should be put into the report. Ticket is here: [LF-3622](https://lite-farm.atlassian.net/browse/LF-3622)
- styling the Excel file -- this can be done using the [extensive ExcelJS API](https://github.com/exceljs/exceljs). Ticket for that is here: [LF-3755](https://lite-farm.atlassian.net/browse/LF-3755)
- (the nice to have) adding PDF export: [LF-3754](https://lite-farm.atlassian.net/browse/LF-3754)

Jira link: https://lite-farm.atlassian.net/browse/LF-3623

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

Please use the interim UI (the big button) to test the file and database record.

**Checklist:**

[LF-3622]: https://lite-farm.atlassian.net/browse/LF-3622?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LF-3755]: https://lite-farm.atlassian.net/browse/LF-3755?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LF-3754]: https://lite-farm.atlassian.net/browse/LF-3754?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/api/db/migration/20231010170020_create_finance_report_table.js
```diff
@@ -0,0 +1,46 @@
...
+ *  LiteFarm is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ */
+
+export const up = async function (knex) {
+  await knex.schema.createTable('finance_report', (table) => {
+    table.increments('finance_report_id').primary();
+    table.uuid('farm_id').references('farm_id').inTable('farm').defaultTo(null);
```
antsgar, line 19: Should this be not nullable too, or is there a way a user could generate a report not belonging to any farm?

antsgar (dismissed):
Tested it and it's working great, excellent job @kathyavini! 🎉

packages/api/src/util/generateFinanceReport.js
SayakaOno, line 1: License is missing.
Other than that, it works great!!

packages/api/src/util/generateFinanceReport.js
```diff
@@ -0,0 +1,15 @@
+import ExcelJS from 'exceljs';
+
+const generateFinanceReport = (data) => {
```
Duncan-Brain, line 3: Not really necessary to address this comment now... but I am thinking it could be beneficial to separate the creation of the workbook from the making of the "Sheet 1".  Such that when you pass in a workbook to a sheet making function you can add sheets (or not). If in the future we have multiple reports like "balance sheet", "transactions", "expense report", "tax report" etc. the user might check off which ones to generate and we just pass along the workbook to each?

Just a thought.

Duncan-Brain:
Also tested and works great! 

My questions are: 
1. Do you see this moving to a jobs queue at some point in the future like certifications?
2. I was able to click it a lot of times in a row and it worked well and added many records to the database -- should there be a limit?
3. One inline comment about the generateReport function.

None of these are things I think needed to be sorted now. I just figured you were going to add the licence to that one file, and and so I was waiting for that!

packages/api/src/util/generateFinanceReport.js
```diff
@@ -0,0 +1,15 @@
...
+
+  // Add data to worksheet
+  worksheet.addRow(['Revenue', data.revenue]);
+  worksheet.addRow(['Expenses', data.expenses]);
+  worksheet.addRow(['Balance', data.balance]);
+
+  return workbook;
+};
+
+export { generateFinanceReport };
```
Duncan-Brain, line 15: could save a line and export the function in the declaration line.

packages/api/db/migration/20231010170020_create_finance_report_table.js
kathyavini: No, it should always refer to a farm! I'll fix that, thanks 🙂

packages/api/src/util/generateFinanceReport.js
kathyavini: Yeah, I could totally see passing the workbook to a different function for each sheet type here! I don't think using sheets in that way has been discussed yet (as opposed to just creating different kinds of reports downloads entirely) but it's something to bring up with product when we get there and I think it makes sense. It's definitely technically feasible.

kathyavini:
@Duncan-Brain regarding your questions:

1. Hehe I will do anything possible to avoid moving to a job queue 😆 I find the export server queue hard to read and even harder to troubleshoot. For a single file like this I don't anticipate it ever becoming necessary -- even if we wanted to move to emailing the file I don't think there is any reason not to just do it on the API server as with farm map 🤞 
2. Hmm! I think the interim UI (big clicky button) makes this behaviour easier than will be possible in the final UI -- like there will be a new view in-between and a form to fill out -- so I don't think it should be too easy. Makes me want to test though and make sure someone wouldn't be able to go through the whole submission again while e.g. waiting for the file on a slow connection.
3. Fixed, thanks!

SayakaOno (approved):

antsgar (approved):

PR #2891, opened 2023-09-22
LF-3662: "< " Back arrow redirectes to the wrong page when the user saves the "Custom Revenue Type" 

SayakaOno:
**Description**

- Wrap PureSimpleCustomType in revenue views to trigger page navigation.
- Fix ManageRevenueTypes path in useEffect.

Jira link: https://lite-farm.atlassian.net/browse/LF-3662

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/containers/Finances/CustomRevenueType/AddCustomRevenue.jsx
+++ b/packages/webapp/src/containers/Finances/CustomRevenueType/AddCustomRevenue.jsx
@@ -19,6 +19,7 @@ import { addCustomRevenueType } from '../saga';
 import { revenueTypesSelector } from '../../revenueTypeSlice';
 import { CUSTOM_REVENUE_NAME } from './constants';
 import { hookFormUniquePropertyValidation } from '../../../components/Form/hookformValidationUtils';
+import { HookFormPersistProvider } from '../../hooks/useHookFormPersist/HookFormPersistProvider';
 
 function AddCustomRevenue({ history }) {
   const { t } = useTranslation();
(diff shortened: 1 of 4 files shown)


kathyavini (commented):
This takes care of clearing the edit view from the history, but now after editing a type, the back arrow returns to `/Finances` rather than to `/revenue_types` (in contrast to how expense types return to `/expense_categories`)... that's a bug, right?

SayakaOno:
@kathyavini That's a bug... I'm sorry I thought it was working as expected. Let me check it again!

kathyavini (approved):
Ah, the pesky singular vs plural pathnames!! 😁

Duncan-Brain:
I am still needing to click twice to get back.

kathyavini:
@Duncan-Brain try adding/retiring 5+ types and see how many times you have to click "<" 😁 

Which is to say it's a quite serious bug that has to be addressed but it's the same behaviour as in expense types and has its own ticket here: https://lite-farm.atlassian.net/browse/LF-3647

Duncan-Brain (approved):
@kathyavini @SayakaOno 

Just asking for if LF-3647 is then a blocker to this one. Since the expected result is not achieved. 

I am down to approve and merge but curious if I should move the ticket back to development after Jira automation.

kathyavini:
@Duncan-Brain I say let's ping Denis to update LF-3647 description so it explicitly covers both types, and let this one be validated based on the edit view no longer being part of the history stack, i.e. brought up to the same (dys-)functionality as expense types. Doesn't make sense to block revenue but not expenses when they are in the same state with the same bug.

PR #2893, opened 2023-09-25
LF-3190-change-too-many-password-resets-to-be-in-line-error-rather-than-snack-bar

navDhammu:
**Description**

Replaced the error snackbar with an inline error message by creating a new state variable in redux to hold the error message. 

Modified modal component to accept an error prop to be rendered underneath the button.

Jira link: https://lite-farm.atlassian.net/browse/LF-3190

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/Modals/ModalComponent/v1/index.jsx
+++ b/packages/webapp/src/components/Modals/ModalComponent/v1/index.jsx
@@ -1,7 +1,7 @@
 import styles from './styles.module.scss';
 import Button from '../../../Form/Button';
 import React from 'react';
-import { Title } from '../../../Typography';
+import { Error, Title } from '../../../Typography';
 import PropTypes from 'prop-types';
 
 export default function ModalComponent({
@@ -40,7 +40,7 @@ export default function ModalComponent({
       >
         {buttonLabel}
       </Button>
-      {error && <span className={styles.error}>{error}</span>}
+      {error && <Error className={styles.error}>{error}</Error>}
     </div>
   );
 }
--- a/packages/webapp/src/components/Modals/ModalComponent/v1/index.jsx
+++ b/packages/webapp/src/components/Modals/ModalComponent/v1/index.jsx
@@ -12,6 +12,7 @@ export default function ModalComponent({
   descriptions,
   buttonLabel,
   buttonColor,
+  error,
 }) {
   const buttonStyles = {
     letterSpacing: '0.4005px',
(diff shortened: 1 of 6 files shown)


packages/webapp/src/components/Modals/ModalComponent/v1/index.jsx
```diff
@@ -26,7 +27,9 @@ export default function ModalComponent({
         {title}
       </Title>
       {descriptions.map((description) => (
-        <Title style={{ marginBottom: 0, textAlign: 'center' }}>{description}</Title>
+        <Title key={description} style={{ marginBottom: 0, textAlign: 'center' }}>
```
navDhammu, line 30: added the key here because pre-commit / linting was giving an error

Duncan-Brain (commented):
Looks good to me! I am just wondering though if it should dismiss the error on click so that the user knows it is retrying?

navDhammu:
> Looks good to me! I am just wondering though if it should dismiss the error on click so that the user knows it is retrying?

Would it be better to just disable the resend button when theres an error? Because clicking the button would result in the same error again and again.

kathyavini:
@navDhammu and @Duncan-Brain personally I like the idea of disabling the button!

Duncan-Brain:
I like the idea too -- could be worth asking/tagging Kevin and Loic on Jira ticket?

packages/webapp/src/components/Modals/ModalComponent/v1/index.jsx
```diff
@@ -37,6 +40,7 @@ export default function ModalComponent({
       >
         {buttonLabel}
       </Button>
+      {error && <span className={styles.error}>{error}</span>}
```
antsgar, line 43: I'd suggest using the `Error` component from the `Typography` file here which will already give you the right styling (the font color is a deeper red than the regular red used here), and then using custom styling only for the margin.

antsgar (changes requested):

packages/webapp/src/components/Modals/ModalComponent/v1/index.jsx
navDhammu: Thanks for the suggestion - I went ahead and replaced it with the Error component.

Duncan-Brain (approved):
Looking good to me! Thanks for updating

packages/webapp/src/components/Modals/ModalComponent/v1/index.jsx
antsgar: Thank you for updating this and for the contribution! 🙂

antsgar (approved):

PR #2894, opened 2023-09-27
LF-3595 Add option on create custom revenues to select if it is a crop generated revenue

kathyavini:
**Description**

This PR adds two properties to custom revenue types (`agriculture_associated` and `crop_generated`) and sets them both to true for our single default type (`'Crop Sale'`).

Notes:
- `agriculture_associated` has no particular functionality or behaviour associated with it (other than conditionally rendering the `crop_generated` radio in the Add Custom Revenue form), it is just something recorded in the database
- Following our whole-team discussion on Friday September 22, 2023, these properties are unalterable on the template (type) after creation, AND the revenue type is uneditable on the instance (sale) after creation. Therefore I did not update the patch controller for sale to handle moving between revenue types, or the patch controller on type to change these properties, as both of these changes are prohibited
- ~~I didn't make either property required on the model level but they are set by database defaults so will be assigned if not supplied~~ As discussed in tech daily the model now mirrors the React Hook Form validation. `agriculture_associated` is always required, and `crop_generated` is required if `agriculture_associated` is true.

Summary of PR:
- addition of properties to the `revenue_type` table, model, factory (for tests), slice, and add saga, and setting them appropriately for 'Crop Sale'
- updating conditional rendering of the appropriate general vs crop sale form (frontend) and handling of the `crop_variety_sale` update (backend). **Thank you @SayakaOno for meticulously indicating each place in the codebase where the previous General Sale vs Crop Sale logic needed to be updated -- and _especially_ for making `getRevenueFormType`!!**
- updating the UI appropriately for the setting and reading of these properties. Thank to you @Duncan-Brain for suggesting that I don't ruin `<PureSimpleCustomType />` to do this 😄 The radio groups are now passed as render props from the containers
- also updated the read-only radio per Loïc's specifications -- now grey rather than green

Jira link: https://lite-farm.atlassian.net/browse/LF-3595

**Type of change**

**How Has This Been Tested?**

- [x] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/webapp/src/containers/Finances/CustomRevenueType/CustomRevenueRadios.jsx
SayakaOno, line 1: License is missing :)

One thing I noticed is that `crop_generated` could be `true` even if `agriculture_associated` is `false`. (I think it's an edge case, please check the recording for the steps to make it happen)

https://github.com/LiteFarmOrg/LiteFarm/assets/33141219/1e64177c-5c77-4056-aafa-6d39b6a124d1



But it looks great overall!

kathyavini: Thank you for catching that @SayakaOno!! I didn't expect React Hook Form would retain the value when the form component was not rendered -- and thus I learned today for the first time that this behaviour is not default, but rather comes from including `shouldUnregister: true` in the `useForm()` 😄

SayakaOno: I feel like I have done something more complicated to get it to work right... `shouldUnregister` is so handy!

Sorry, I found one more thing...
1. create a type with `crop_generated` `true` and retire
2. create a type with the same name but with `crop_generated` `false`

The type's `deleted` becomes true, and I don't think other fields are updated.
As the user, it feels like a bug, but if a sale is created with a type, the type is retired and restored with a different `crop_generated` value, the sale will be messed up I guess... I think we should discuss it later...

kathyavini: That's a very good point and an issue with how we secretly un-retire under the hood, a behaviour which it sounds like is not Kevin's preference based on his comments here: https://lite-farm.atlassian.net/browse/LF-3666

Let's definitely discuss the right way to handle this!

packages/api/db/migration/20230922175919_add-crop-generated-boolean-to-revenue-type-table.js
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+export const up = async function (knex) {
+  await knex.schema.alterTable('revenue_type', (table) => {
+    table.boolean('agriculture_associated').defaultTo(true);
+    table.boolean('crop_generated').defaultTo(false);
```
Duncan-Brain, line 23: Good question about should we use defaults here. I do not know what is right for architecture so I think it will be fine either way. 

But I _thiiiink_  I prefer having required fields since I don't like assuming things. (One persons opinion that makes sense to me : https://stackoverflow.com/a/6409322) 

But also if undefined can sneak through like this mongoose example that could be not great: https://stackoverflow.com/questions/76114812/should-i-set-a-required-field-if-there-is-already-a-default-field#:~:text=If%20we%20try%20to%20save,required%20field%20is%20not%20redundant.

packages/api/src/controllers/saleController.js
```diff
@@ -53,8 +54,9 @@ const SaleController = {
         saleData.sale_date = sale_date;
       }
 
-      // TODO: LF-3595 - properly determine if value and note need to be updated
-      if (!req.body.crop_variety_sale?.length) {
+      const revenueType = await RevenueTypeModel.query().findById(revenue_type_id);
```
Duncan-Brain, line 57: So if we don't add this to the try/catch with trx then we are opening 2 trx in this method I think.

packages/api/src/middleware/validation/sale.js
```diff
@@ -17,11 +17,11 @@ import RevenueTypeModel from '../../models/revenueTypeModel.js';
 async function validateSale(req, res, next) {
   // TODO replace upsertGraph
   const { crop_variety_sale, revenue_type_id } = req.body;
-  // TODO: implement properly once LF-3595 is complete
-  const cropSaleRevenueType = await RevenueTypeModel.query()
-    .where('revenue_name', 'Crop Sale')
-    .first();
-  const isCropRevenue = revenue_type_id === cropSaleRevenueType.revenue_type_id;
+
+  const revenueType = await RevenueTypeModel.query().findById(revenue_type_id);
```
Duncan-Brain, line 21: Not your code but it is weird that this is middleware. Its kind of similar to a schema check but also not really. 

Don't feel obligated to take action but it seems like this check should be in the controller for the same trx reasons we have been discussing lately.

packages/api/src/models/revenueTypeModel.js
```diff
@@ -42,6 +42,8 @@ class RevenueType extends baseModel {
         revenue_name: { type: 'string', minLength: 1, maxLength: 100 },
         farm_id: { type: 'string' },
         revenue_translation_key: { type: 'string' },
+        agriculture_associated: { type: 'boolean' },
```
Duncan-Brain, line 45: Bringing up again the required property. I think we want required because we are then saying to developers/api users that these are "optional" but not giving users the opportunity to make it optional on the frontend.

packages/webapp/src/components/Form/Radio/radio.module.scss
```diff
@@ -78,8 +78,12 @@
 }
 
 .container input[disabled]:checked ~ .checkmark svg circle:first-child {
-  fill: var(--teal500);
-  stroke: var(--teal700);
+  fill: var(--grey200);
```
Duncan-Brain, line 81: nice find ❤️

packages/webapp/src/components/Forms/SimpleCustomType/index.jsx
```diff
@@ -66,17 +68,25 @@ const PureSimpleCustomType = ({
   inputMaxChars = 100,
   validateInput,
   useHookFormPersist = () => ({}),
+  customRadioGroup,
```
Duncan-Brain, line 71: Can we make this even more generic-er? Like customizedFormInputs or customFormChildren or something? Radio group seems somewhat specific.

packages/api/src/middleware/validation/sale.js
Duncan-Brain: For a minute this line crashed my db -- then I cleared cache and it worked normal. 

So revenue_type_id wasn't passed just noting in case anyone else has it. 
Screenshot 2023-09-29 at 2 09 54 PM

But technically a savvy api user could crash our db by not supplying it.

Duncan-Brain (changes requested):
Most things look great, and tests ran! Things I noticed:

inline comments about:
- db crashing -- only one that needs to fixing for approve
- required in model?
- more generic-er for SimpleCustomType?

packages/api/src/middleware/validation/sale.js
kathyavini: Oh gnarly, `findById()` sucks 😆 I'll use `.where()` and `.first()` instead then

packages/webapp/src/components/Forms/SimpleCustomType/index.jsx
kathyavini: Yeah that would be much better!

packages/api/src/middleware/validation/sale.js
Duncan-Brain: a simple check beforehand is probs fine too -- I am guessing it is a performant function

kathyavini: Ok. I guess since it needs a check anyway might as well put it first. We use the `.where()` pattern a lot more but I imagine you're right about performance.

packages/api/src/models/revenueTypeModel.js
kathyavini: Ok let's make it required on the model (to also cover API users), but keep default values on the database level. From what I read on the StackOverflow links you shared, that should be fine. I took advantage of the default database inserts to keep more tests running without modification (and as a bonus they also keep the custom types already on beta or local behaving as before so I quite like them). But I agree the application level requirements should be the same via API or frontend so it belongs on the model.

Duncan-Brain: Yeah I was thinking I was glad this was not live because it would be hard to update the user data

kathyavini: @Duncan-Brain When coding it up I realized that we do not require `crop_generated` on the frontend (if `agriculture_associated` is false on the frontend nothing is sent for `crop_generated` and the database defaults it to false). So I matched the model requirements to that so API and front behave the same.

packages/api/src/controllers/saleController.js
kathyavini: Sequentially though, right? So it wouldn't have any effect on filling the transaction pool -- this one is closed before the try/catch one is opened.

Duncan-Brain (approved):
Database no longer crashes, and everything still seems to be working.

No suggestion here from me but I still don't love that we default/optional crop associated revenue on the backend, but require it for form completion on the frontend. Consistency there would be cool but I get that the one is hierarchical over the other so its kind of weird.

SayakaOno (dismissed):

packages/webapp/src/containers/Finances/CustomRevenueType/CustomRevenueRadios.jsx
```diff
@@ -0,0 +1,89 @@
...
+          {
+            label: t('common:YES'),
+            value: true,
+          },
+          { label: t('common:NO'), value: false },
+        ]}
+        required
+        disabled={view === 'edit' || view === 'read-only'}
+      />
+      {view === 'add' && <Info>{t('REVENUE.ADD_REVENUE.CANNOT_BE_CHANGED')}</Info>}
```
antsgar, line 56: Should we show this information on the Edit view too so that it's clearer to the user why it's disabled and can't be edited?

packages/webapp/src/containers/Finances/CustomRevenueType/CustomRevenueRadios.jsx
```diff
@@ -0,0 +1,89 @@
...
+
+      {isAgricultureAssociated && (
+        <>
+          <Main style={{ paddingTop: '32px', paddingBottom: '12px' }}>
```
antsgar, line 60: Same here, I'd suggest moving this to CSS

antsgar (approved):
Looks good, awesome work! 👏 👏

packages/webapp/src/containers/Finances/CustomRevenueType/CustomRevenueRadios.jsx
kathyavini: Yes, I like that idea! The given wording only makes sense for add, though, I think:

Screenshot 2023-10-12 at 9 58 04 AM

For edit, do you think just "This cannot be changed" is enough? Or "This cannot be changed once set?"

Duncan-Brain (commented):
Great find with the optional required in the model!

I am sorry I found one more thing though -- using postman I can set agriculture_associated to false and crop_generated to true. I don't think this case should be allowed.

Otherwise everything including tests continue to function well for me.

kathyavini:
@Duncan-Brain well anything can go into the schema, but I'm not sure it's worth it. Here is what it looks like to extend the schema logic to say if `agriculture_associated` is false, nothing can be provided for `crop_generated` (which is how the frontend behaves). We could also say `crop_generated` cannot be true, but that wouldn't strictly mirror the frontend:

```
if: {
  type: 'object',
  properties: {
    agriculture_associated: { const: true },
  },
},
then: {
  type: 'object',
  required: ['crop_generated'],
},
else: {
  if: {
    properties: {
      agriculture_associated: { const: false },
    },
  },
  then: {
    not: {
      required: ['crop_generated'],
    },
  },
},
```

I think it is a bit verging into unreadable with the third logical block, but this is the most accurate. Do you prefer this one?

Duncan-Brain:
Up to you on the schema, I am fine with a simple if => response in the controller. Just would be pretty weird to not have that possibility on the frontend but allow it on the backend. Could mean we create bad data and then make assumptions based on it in the future in finances and forms etc.

Duncan-Brain (dismissed):
Nice LGTM!

packages/webapp/src/containers/Finances/CustomRevenueType/CustomRevenueRadios.jsx
antsgar: "This cannot be changed" makes sense to me!

antsgar (dismissed):

SayakaOno (approved):

antsgar (approved):

Duncan-Brain (approved):

PR #2897, opened 2023-10-04
LF-3637 (1): Update the date inputs used in date range selector [UI]

SayakaOno:
This is the first PR for [LF-3637](https://lite-farm.atlassian.net/browse/LF-3637).
1. **[new component](https://github.com/LiteFarmOrg/LiteFarm/pull/2897)**
2. [date range class](https://github.com/LiteFarmOrg/LiteFarm/pull/2898) (will be rebased once this PR is merged)
3. integration (replacing the existing component with the new one)

**Quick summary on how the new date range selector works**
- When inputting custom date range,
  - "< back" button will not be enabled while the date range is invalid
  - the value on the window is "yyyy-mm-dd - yyyy-mm-dd" while the date range is invalid
  - if the user clicks the outside the selector while the date range is invalid, the date range option will be reset to ~~"This year"~~"Year to date"

**Description**
- update `DateContainer`
  - to show "yyyy-mm-dd" when there is no value
  - to add `fromProps` and `toProps` so that `min` and `max` can be set for the date input
- create `DateRangeSelector` component (the selector itself)
- create `CustomDateRangeSelector` component (part of `DateRangeSelector` that opens when "Pick a custom range" option is selected
   - `FromToDateContainer` is reused. (so `moment` is still being used although we would like to move away from it)
   - override styles of `FromToDateContainer` for smaller screen size to prevent the inputs from becoming two lines. (the solution is not beautiful, but I assume the style of the component will be with Loïc's new design)
- [stories](http://localhost:6006/?path=/docs/components-daterangeselector--docs)

Jira link: https://lite-farm.atlassian.net/browse/LF-3637

**Type of change**

**How Has This Been Tested?**

```
pnpm test-storybook DateRangeSelector
```

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

[LF-3637]: https://lite-farm.atlassian.net/browse/LF-3637?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/webapp/src/components/DateRangeSelector/CustomDateRangeSelector.jsx
```diff
@@ -0,0 +1,81 @@
...
+      <FromToDateContainer
+        onStartDateChange={changeStartDate}
+        onEndDateChange={changeEndDate}
+        endDate={endDate}
+        startDate={startDate}
+        fromProps={{ max: fromDateMax, ...inputStyle }}
+        toProps={{ min: toDateMin, ...inputStyle }}
+      />
+      {startDate && endDate && !validRange && (
+        <Semibold style={{ textAlign: 'center', color: 'red' }}>
```
SayakaOno, line 62: I copied the existing code. I don't like `color: 'red'` but will leave it for now.

packages/webapp/src/stories/DateRangeSelector/DateRangeSelector.stories.jsx
```diff
@@ -0,0 +1,172 @@
...
+
+    await userEvent.type(input2, '2022-11-01');
+    expect(selectedOptionText).toBeInTheDocument();
+
+    const errorMessage = await canvas.findByText(
+      `End date must be after start date to return results`,
+    );
+    expect(errorMessage).toBeInTheDocument();
+
+    // This does not pass when rerunning on storybook UI
```
kathyavini, line 99: I get a failure at this step in both CLI (i.e. running `pnpm test-storybook DateRangeSelector`) and the Storybook UI. The failure seems legitimate because the placeholder `yyyy-mm-dd yyyy-mm-dd` is still shown after clicking away rather than 'Year to date' (until the dropdown is re-selected). Is that behaviour not correct?

Duncan-Brain: Me too ^^

packages/webapp/src/components/DateRangeSelector/index.jsx
```diff
@@ -0,0 +1,198 @@
...
+  const changeEndDate = (date) => {
+    setValidRange(customFromDate <= date);
+    setCustomToDate(date);
+    changeDateMethod('end', date);
+  };
+
+  return (
+    <ClickAwayListener onClickAway={onClickAway}>
+      <div className={styles.wrapper}>
+        <ReactSelect
```
Duncan-Brain, line 152: Feel free to ignore this :) 

I think the only question I have is do we have a different non-form element other than ReactSelect?

This is related to my question of tech debt for Form elements vs View elements. I can see value in separating the use of form elements that can be used react-hook-form. And view elements like filters since this (mostly) acts like a filter for data. (But I do get that an argument can be made that it is also a form element for document export -- that does not use react-hook-form)

Duncan-Brain (commented):
Everything looks great to me too.

But I think I am a third wheel reviewing this so I might remove myself. Since I had the same question as Joyce. I left one other nothing comment haha

packages/webapp/src/stories/DateRangeSelector/DateRangeSelector.stories.jsx
SayakaOno: I guess it was mysteriously passing...
I realized that I should not have used `useSelector` in the component to make it work in storybook, so I fixed it. However, click away behaviour couldn't be tested still...! So I gave up and removed the test this time 😢 
Thank you so much for testing!

packages/webapp/src/components/DateRangeSelector/index.jsx
SayakaOno: There is an input that is a non-form element that is used for the custom date range. (`FromToDateContainer` -> `DateContainer`)
Screenshot 2023-10-12 at 11 12 25 PM
Now I feel that you are asking if we have a select element for non-form... - I don't think we do.

I thought I understood your question at that time, but now I am confused... you don't want us to use the same ReactSelect component when it's not part of a form?

SayakaOno (commented):
Thank you everyone for reviewing!

I addressed most of Antonella's comments and also made a few updates on the component and the stories so that the component works as expected in the stories.

I might make a few more commits depending on your reply to my comments.
Thank you 🙏

packages/webapp/src/components/DateRangeSelector/index.jsx
Duncan-Brain: Yes you are right that is what I was asking! Since we don't have one now I don't think anything should be done. All I could find is DropDown and it is unused.

I am wondering if that makes sense that we _should_ then maybe have a non-form react select in the future. One for taking/managing user input data into a RHF, one for filtering data to a view from redux. And should elements like this DateRangeSelector live under /components/Filter/DateRangeSelector or something?

packages/webapp/src/components/DateRangeSelector/CustomDateRangeSelector.jsx
```diff
@@ -0,0 +1,81 @@
...
+  fromDateMax,
+  toDateMin,
+}) {
+  const { t } = useTranslation();
+  const inputStyle = { classes: { container: { minWidth: 'calc((100% - 24px) / 2)' } } };
+
+  return (
+    <div className={styles.customDateRangeSelector}>
+      <div className={styles.buttons}>
+        <div
```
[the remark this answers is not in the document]
antsgar: Oh I clearly copy pasted the wrong thing 😂 sorry about that! This is the right link https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role

packages/webapp/src/components/DateRangeSelector/CustomDateRangeSelector.jsx
```diff
@@ -0,0 +1,81 @@
...
+  validRange,
+  startDate,
+  endDate,
+  changeStartDate,
+  changeEndDate,
+  fromDateMax,
+  toDateMin,
+}) {
+  const { t } = useTranslation();
+  const inputStyle = { classes: { container: { minWidth: 'calc((100% - 24px) / 2)' } } };
```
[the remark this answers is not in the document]
antsgar: Makes sense, thank you for clarifying! 🙂

packages/webapp/src/components/DateRangeSelector/constants.jsx
```diff
@@ -0,0 +1,26 @@
...
+ *  the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  LiteFarm is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ */
+
+export const dateRangeOptions = {
```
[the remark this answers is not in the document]
antsgar: If the rest of the code is consistent with another convention, we don't need to change this, preferable to keep it consistent for now

antsgar:
@SayakaOno changes look great to me! There are a few of my comments that I don't think we need to address, I'd only make the change from `div` to `button` (linked the right doc now, sorry about that 😂) but other than that one this is perfect

packages/webapp/src/components/DateRangeSelector/index.jsx
SayakaOno: Thank you for clarifying! By taking a quick look, I don't feel like a ReactSelect component for non-form will be very different from the existing ReactSelect (I need to look into the taking/managing user input data part), but I think this could be an interesting discussion!

I initially created this component using `Form/DateRangePicker` that takes properties from RHF without knowing the difference from `DateRangeSelector, and I ended up almost re-doing to switch to the one without RHF 😢  So in short, I agree with setting a clear rule (like where to place files) for forms and non-forms and documenting it!

SayakaOno:
@antsgar
Thank you for the link! I changed the div to button in this [commit](https://github.com/LiteFarmOrg/LiteFarm/pull/2897/commits/3ce0ab0e2082cf646aca64780a9a8c2e15e43d06), and after reading the doc, I added `type="button"` ([b5c5ca0](https://github.com/LiteFarmOrg/LiteFarm/pull/2897/commits/b5c5ca0f771ac051798c4695e9c6b1e99cea3b28)).
It would be great if you could double-check, thank you!

packages/webapp/src/components/DateRangeSelector/CustomDateRangeSelector.jsx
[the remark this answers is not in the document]
kathyavini: Sorry I wasn't sure exactly where this landed (there were a few threads on button) but I see

> I just learned default button style is reset :)

Just checking that this means this will be updated back to how it used to look?

Screenshot 2023-10-13 at 10 25 30 AM

packages/webapp/src/stories/DateRangeSelector/DateRangeSelector.stories.jsx
kathyavini: Oh it still didn't pass for you? Because now the clickAway behaviour *is* working for me in the interactive Storybook 😆! It's the store slice being mocked state that did it right? But I'm fine with skipping that part of testing 🙂 Thanks for writing the interaction tests!

packages/webapp/src/components/DateRangeSelector/CustomDateRangeSelector.jsx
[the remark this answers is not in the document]
SayakaOno: 😨😨😨!!!
I guess I needed to restart storybook, it looked normal after I switched to `button`.
Thank you for catching it, I will fix it!

antsgar (approved):

packages/webapp/src/stories/DateRangeSelector/DateRangeSelector.stories.jsx
SayakaOno: Storybook doesn't click away for me for some reason...

I was doing `const dateRange = useSelector(dateRangeSelector)`, but `dateRange` was becoming `undefined` in storybook and the error (`undefined.option`) was happening in `onClickAway` function.
Screenshot 2023-10-13 at 11 21 19 AM

Thank you 😄

SayakaOno:
I created `TextButton` component and replaced the native `<button />` with it. It would be appreciated if you could re-review!
(I'm sorry I staled your approval Antonella 🙏 )

packages/webapp/src/stories/DateRangeSelector/DateRangeSelector.stories.jsx
kathyavini: Yes, me too! Exactly like that! But not since you updated this part:

```js
// DateRangeSelectorStories.jsx

const [dateRange, setDateRange] = useState({ option: dateRangeOptions.YEAR_TO_DATE });
  
return (
  <Suspense fallback={'Loading...'}>
    <DateRangeSelector
        dateRange={dateRange}
        changeDateMethod={() => ({})}
        setDateRangeOptionValue={(option) => setDateRange({ option })}
 ```
 
Now `dateRange.option` is always defined correctly for me even in Storybook so the clickAway has begun happening. I thought that was a very clever solution!

kathyavini (approved):
It all looks great! I threw it in the app and had fun watching it interact with the dateRange slice 🙂 

Looking forward to seeing how it integrates with the rest.

packages/webapp/src/stories/DateRangeSelector/DateRangeSelector.stories.jsx
SayakaOno: Thank you Joyce! ❤️

PR #2900, opened 2023-10-05
LF-3675 improve biodiversity calculation performance

antsgar:
**Description**

Changes to how we query the data from GBIF. Essentially, instead of querying for all occurrences and then filtering out those ones that match the kind of species we’re interested in, I’ve re-written it so that we make a query for each class/kingdom (plants, birds, mammals, etc) and then filter out how many unique occurrences there are for each of those classes. This will mean that we’re less likely to hit the 100,000 limit from their API. I’ve also limited concurrency so that we only make calls for one specific boundary concurrent. So if we have one boundary, we’ll do 1 chunk of calls, if we have two we’ll make 2 chunks, and so on. This also means that if we do hit that API limit, the user could solve the problem by splitting their boundary into several smaller ones.

Jira link:
https://lite-farm.atlassian.net/browse/LF-3675

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

Tested manually with boundaries of various sizes.

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/insightHelpers.js
+++ b/packages/api/src/controllers/insightHelpers.js
@@ -299,23 +299,21 @@ export const getBiodiversityAPI = async (pointData, countData) => {
       // Do the subsequent API calls concurrently
       for (let i = 300; i < count; i += 300) {
         apiCalls.push(
-          new Promise((resolve, reject) => {
-            rp({
-              uri: endPoints.gbifAPI,
-              qs: {
-                ...qs,
-                offset: i,
-              },
+          rp({
+            uri: endPoints.gbifAPI,
+            qs: {
+              ...qs,
+              offset: i,
+            },
+          })
+            .then((data) => {
+              const { results } = JSON.parse(data);
+              processBiodiversityResults(results, label);
+              return Promise.resolve();
             })
-              .then((data) => {
-                const { results } = JSON.parse(data);
-                processBiodiversityResults(results, label);
-                resolve();
-              })
-              .catch((error) => {
-                reject(error);
-              });
-          }),
+            .catch((error) => {
+              return Promise.reject(error);
+            }),
         );
       }
     }
--- a/packages/api/src/controllers/insightHelpers.js
+++ b/packages/api/src/controllers/insightHelpers.js
@@ -271,9 +271,7 @@ export const getBiodiversityAPI = async (pointData, countData) => {
   });
 
   for (const polygon of polygons) {
-    console.log(polygon);
     for (const label in filterDictionary) {
-      console.log(label);
       const qs = {
         geometry: polygon,
         limit: 300,
@@ -286,9 +284,7 @@ export const getBiodiversityAPI = async (pointData, countData) => {
           offset: 0,
         },
       });
-      console.log(`first call done for ${label}`);
       const { count, results } = JSON.parse(data);
-      console.log(count);
       processBiodiversityResults(results, label);
       const apiCalls = [];
       for (let i = 300; i < count; i += 300) {
@@ -312,22 +308,15 @@ export const getBiodiversityAPI = async (pointData, countData) => {
           }),
         );
       }
-      console.log(`${apiCalls.length} calls for ${label}`);
       await Promise.allSettled(apiCalls);
-      console.log(`api calls done for ${label}`);
     }
 
     for (const label in speciesCount) {
-      console.log(`parsing ${label}`);
-      console.log(speciesCount[label]);
-      console.log(parsedSpecies.length);
-      console.log((speciesCount[label] / parsedSpecies.length) * 100);
       resultData.data.push({
         name: label,
         count: speciesCount[label],
         percent: (speciesCount[label] / parsedSpecies.length) * 100,
       });
-      console.log(resultData.data);
     }
     resultData.preview = parsedSpecies.length;
     return resultData;
--- a/packages/api/src/controllers/insightHelpers.js
+++ b/packages/api/src/controllers/insightHelpers.js
@@ -272,10 +272,10 @@ export const getBiodiversityAPI = async (pointData, countData) => {
     return makePolygon(points.grid_points.sort(compareLatLong));
   });
 
-  // Handle one polygon and one species type at a time
+  const apiCalls = [];
+
   for (const polygon of polygons) {
     for (const label in filterDictionary) {
-      // First API call to get the count
       const qs = {
         geometry: polygon,
         limit: 300,
@@ -297,7 +297,6 @@ export const getBiodiversityAPI = async (pointData, countData) => {
       processBiodiversityResults(results, label);
 
       // Do the subsequent API calls concurrently
-      const apiCalls = [];
       for (let i = 300; i < count; i += 300) {
         apiCalls.push(
           new Promise((resolve, reject) => {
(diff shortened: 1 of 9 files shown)


packages/api/src/controllers/insightHelpers.js
```diff
@@ -236,105 +238,100 @@ export const getBiodiversityAPI = async (pointData, countData) => {
...
-    Insecta: 'Insects',
-    Plantae: 'Plants',
-    Amphibia: 'Amphibians',
-    Mammalia: 'Mammals',
+  const filterDictionary = {
+    Birds: { classKey: 212 },
+    Insects: { classKey: 216 },
+    Plants: { kingdomKey: 6 },
+    Amphibians: { classKey: 131 },
+    Mammals: { classKey: 359 },
```
SayakaOno, line 246: Where are these keys from?

packages/api/src/controllers/insightHelpers.js
```diff
@@ -236,105 +238,100 @@ export const getBiodiversityAPI = async (pointData, countData) => {
...
     Insects: 0,
     Plants: 0,
     Amphibians: 0,
-    CropVarieties: 0,
+    CropVarieties: parseInt(countData),
     Mammals: 0,
   };
-  speciesCount['CropVarieties'] = parseInt(countData);
+
+  const parsedSpecies = [];
```
SayakaOno, line 258: `Set` seems better to me here...

packages/api/src/controllers/insightHelpers.js
```diff
@@ -236,105 +238,100 @@ export const getBiodiversityAPI = async (pointData, countData) => {
...
+
+  const processBiodiversityResults = (results, label) => {
+    results.map((result) => {
```
SayakaOno, line 261: Nothing is returned from this function, so I think `forEach` should be used instead.

packages/webapp/public/locales/es/translation.json
```diff
@@ -856,10 +856,15 @@
       "AMPHIBIANS": "Anfibio",
       "BIRDS": "Aves",
       "CROP_VARIETIES": "Variedades de cultivos",
-      "ERROR": {
+      "API_OFFSET_LIMIT_ERROR": {
+        "BODY": "Los límites de su granja son demasiado amplios, no podemos proveerle un reporte de Biodiversidad. Intente crear límites más pequeños.",
+        "PREVIEW": "Unavailable",
```
SayakaOno, line 861: "PREVIEW" is in English in all files :)

packages/webapp/public/locales/pt/translation.json
```diff
@@ -856,7 +856,12 @@
       "AMPHIBIANS": "Anfíbios",
       "BIRDS": "Aves",
       "CROP_VARIETIES": "Variedades de cultivos",
-      "ERROR": {
+      "API_OFFSET_LIMIT_ERROR": {
+        "BODY": "MISSING",
+        "PREVIEW": "Unavailable",
+        "TITLE": "MISSING"
```
SayakaOno, line 862: I think you forgot to add the title "Tinha um problema" as you did for the other languages.

SayakaOno (commented):
I went through the code but could not figure out the GBIF API part. Could you share the link to the doc that was useful? I am confused with "limit 300" and count... I will come back and re-review once I understand how the API works!

A few more things...
- For translations, you can run `pnpm i18n` in `webapp` after adding tags (ex. `t('TAG.SUBTAG')`). The script updated in this [PR](https://github.com/LiteFarmOrg/LiteFarm/pull/2853) inserts  "MISSING" tags automatically and also orders translations alphabetically.
  It would be great if you could run it!
- You duplicated "PREVIEW" and "TITLE" for the API error. I always get indecisive if I should duplicate one or try to use the existing one. (I feel like I should duplicate but am hesitant to do it) Do you have a clear rule to make this kind of decision?

packages/api/src/controllers/insightHelpers.js
antsgar: Unfortunately I couldn't find anywhere in the docs where the keys are explicitly specified -- I ended up doing several queries to the API directly (e.g. https://api.gbif.org/v1/occurrence/search?taxonKey=1) and figuring out from the responses which class or kingdom keys matched the class or kingdom names we were previously using

packages/api/src/controllers/insightHelpers.js
antsgar: Good idea! I'll update this

packages/webapp/public/locales/es/translation.json
antsgar: I think this key is actually not being used anywhere, maybe I should just go ahead and delete it?

packages/webapp/public/locales/pt/translation.json
antsgar: Good call! 🙂

antsgar:
@SayakaOno thank you for the review! I should have linked the GBIF docs in the PR description, here they are https://www.gbif.org/developer/occurrence. If you go to the "Searching" section you'll find the relevant information. In summary requests are paginated and you can query max 300 records per page, and no more than 100,000 records total.

I didn't know we had a script to update the language tags, that's awesome! Thank you for letting me know, I'll run it.

And as for duplicating the title and preview, I think either way is fine. In this case since they are the same copy for both modals, so it'd be probably simpler to not duplicate to avoid having extra conditions in the component.

SayakaOno (approved):
@antsgar I was hoping there would be examples in the document... great job on figuring things out! Thanks to your comment, I understood the logic too :)
It works great!

kathyavini (commented):
I'm very impressed with both the solution @antsgar and the nice use of Set @SayakaOno 👌

It's working well for me for boundaries with counts under 300 (and fixes the percentages too so they add up to 100% -- I didn't see that mentioned in the PR text but it's a good bonus!), and also handles the too large boundaries gracefully.

However I have a farm in a sightings-dense area with a count of 91,545 and `const { results } = JSON.parse(data);` is crashing for me on one of my iterations because the API has returned this at the end of some normal JSON data:

```
Could not write JSON: (was java.lang.ArrayIndexOutOfBoundsException); 
nested exception is com.fasterxml.jackson.databind.JsonMappingException: 
(was java.lang.ArrayIndexOutOfBoundsException) 
(through reference chain: org.gbif.api.model.common.search.SearchResponse["results"]->java.util.ArrayList[19]->org.gbif.api.model.occurrence.Occurrence["eventDate"])
```

Have you seen this error before? I don't know if there is just some bunk data in this particular location, or this is related to the size of the dataset? It doesn't seem size-related from the error message, but my `i` was at 50100 at the point of the crash.

It did bring down the API for me, and if we don't mind bringing down beta I can throw the same boundary up on beta to look at. It's not a very big boundary, just in an eBird-sightings-dense location.

antsgar:
> I'm very impressed with both the solution @antsgar and the nice use of Set @SayakaOno 👌
> 
> It's working well for me for boundaries with counts under 300 (and fixes the percentages too so they add up to 100% -- I didn't see that mentioned in the PR text but it's a good bonus!), and also handles the too large boundaries gracefully.
> 
> However I have a farm in a sightings-dense area with a count of 91,545 and `const { results } = JSON.parse(data);` is crashing for me on one of my iterations because the API has returned this at the end of some normal JSON data:
> 
> ```
> Could not write JSON: (was java.lang.ArrayIndexOutOfBoundsException); 
> nested exception is com.fasterxml.jackson.databind.JsonMappingException: 
> (was java.lang.ArrayIndexOutOfBoundsException) 
> (through reference chain: org.gbif.api.model.common.search.SearchResponse["results"]->java.util.ArrayList[19]->org.gbif.api.model.occurrence.Occurrence["eventDate"])
> ```
> 
> Have you seen this error before? I don't know if there is just some bunk data in this particular location, or this is related to the size of the dataset? It doesn't seem size-related from the error message, but my `i` was at 50100 at the point of the crash.
> 
> It did bring down the API for me, and if we don't mind bringing down beta I can throw the same boundary up on beta to look at. It's not a very big boundary, just in an eBird-sightings-dense location.
> 
> Edit: I should also add that since it's the GBIF API that's messing up I guess it doesn't really matter why its failing as long as we keep our server from crashing on that line 🙂

@kathyavini ooooh interesting catch! I don't think I've seen that one before. Yeah I think it's okay that we try that out in beta, we'd need to figure out if it's related to these changes or not. Can you send the coordinates for your boundary over so I can repro on my side?

kathyavini (approved):
It took me so many tries to get that weird flaky error from GBIF that I thought I couldn't test your PR 😆 

But eventually it came and was cleanly caught 🙂👍

SayakaOno (approved):

PR #2903, opened 2023-10-10
LF-3524 allow user to download all files inside a document at once in a zip from the document tile view preview link

antsgar:
**Description**

- Added `jszip` package to webapp.
- Added capability to download a document that has several files in it as a ZIP file. I was originally going to do this in the BE, but since we already have the URLs for all the files, I figured it'd be faster to do it on the client and download it directly, and also not add load to the server with a potentially somewhat long request.

Jira link:
https://lite-farm.atlassian.net/browse/LF-3524

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

Tested locally, checked that downloads for documents with several files work and that there are no regressions for regular single image/file downloads and those still work.

**Checklist:**


packages/webapp/src/containers/MediaWithAuthentication/index.jsx
```diff
@@ -18,30 +21,62 @@ export function MediaWithAuthentication({
...
+          );
+          const content = await zip.generateAsync({ type: 'base64' });
+          setZipContent(content);
         } else {
-          const url = new URL(fileUrl);
-          url.hostname = 'images.litefarm.workers.dev';
-          const response = await fetch(url.toString(), config);
-          const blobFile = await response.blob();
-          subscribed && setMediaUrl(URL.createObjectURL(blobFile));
+          const fileUrl = fileUrls[0];
```
antsgar, line 47: Kept the logic in this "else" block basically the same that it was before for single file downloads.

package.json
```diff
@@ -23,5 +23,8 @@
     "verbose": true,
     "testURL": "http://localhost/",
     "testEnvironment": "node"
+  },
+  "dependencies": {
```
kathyavini, line 27: Your PR text said adding the package to webapp but it's installed here in root. Either functions, but I think webapp as your note said makes more sense.

package.json
```diff
@@ -23,5 +23,8 @@
...
+    "jszip": "^3.10.1"
```
kathyavini, line 28: Your PR text said adding the package to webapp but it's installed here in root. Either functions, but I think webapp as your note said makes more sense.

kathyavini (commented):
It works great!! Just had one comment about the package install.

packages/webapp/src/containers/MediaWithAuthentication/index.jsx
```diff
@@ -1,14 +1,17 @@
 import { useEffect, useState } from 'react';
 import { mediaEnum } from './constants';
 import { ReactComponent as Download } from '../../assets/images/farmMapFilter/Download.svg';
+import * as JSZip from 'jszip';
```
SayakaOno, line 4: It looks like this does not work with storybook ([story](http://localhost:6006/?path=/story/components-documenttile--variety)). Do we have a solution?

SayakaOno (commented):
I noticed one story was broken, but it's working great!

package.json
antsgar: I probably literally run "npm install" without checking where I was at first 😅 thank you, that's a great call!

kathyavini: I think the new `pnpm-lock.yaml` in root has to be deleted too? 🙂 I think we only use `npm` in root is why it was freshly generated.

antsgar: Done!

SayakaOno (approved):

PR #2912, opened 2023-10-17
changed symbol_native property for the right one

javiscript13:
**Description**

The issue: when a farm in India added an expense an incorrect symbol for the INR currency was being used.  The problem was traced to the json file with all currency symbols, where the one for INR was wrong.

The fix: the INR symbol was fixed on the json file that contains all the supported currency symbols.

Jira link: https://lite-farm.atlassian.net/browse/LF-3669?atlOrigin=eyJpIjoiNmNhMGQyOWM1MWJhNDczNWI2ZTY5OGU3YmJhMzFiZGYiLCJwIjoiaiJ9

**Type of change**

**How Has This Been Tested?**

The usage of the currency symbol was tested on all types of expenses

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/containers/AddFarm/currency/commonCurrency.json
+++ b/packages/webapp/src/containers/AddFarm/currency/commonCurrency.json
@@ -434,7 +434,7 @@
   "INR": {
     "symbol": "Rs",
     "name": "Indian Rupee",
-    "symbol_native": "টকা",
+    "symbol_native": "₹",
     "decimal_digits": 2,
     "rounding": 0,
     "code": "INR",


kathyavini (approved):
Waaaait, so we have a 1000-line json file which is completely redundant with the `symbol` column of the `countries` table??! And it stores full objects for each currency key when we're only pulling out the symbol??! 😅

I think the system needs a bit of a rethink! But as far as this bug goes... 👍 Thanks Javier 🙂

antsgar (approved):

PR #2913, opened 2023-10-17
LF-3637(3): Update the date inputs used in date range selector [integration]

SayakaOno:
**Description**

Update the date range selector using the new one ([PR](https://github.com/LiteFarmOrg/LiteFarm/pull/2897)) and the new class ([PR](https://github.com/LiteFarmOrg/LiteFarm/pull/2898)).

- update `date_range` data structure in the store.
  Before:  
  ```
  { startDate: ..., endDate: ... }
   ```
   After:
   ```
   {
      startDate: ...,
      endDate: ...,
      option: ...,
      customRange: { startDate: ..., endDate: ... }
   }
   ```
   - ~~`option` becomes "custom range" only when `customRange` is valid.~~ The top level `startDate` and `endDate` are ~~also~~ updated to be the same as `customRange` ~~at the same time~~ when `customRange` is valid.
   - The top level `startDate` and `endDate` are the dates for filtering, `customRange` is just for `<CustomDateRangeSelector />`
   Screenshot 2023-10-20 at 12 24 56 PM

- create `useDateRangeSelector` hook
- update `FinanceDateRangeSelector` to use the new DateRangeSelector (the component was renamed)
- update `Finances` and `OtherExpense` using `useDateRangeSelector`
- update `Labour` to use the new selector (it is a class component, so the hook was not be able to be used...)
- update `ActualRevenue` and `EstimatedRevenue` to use the new selector (they were using a different date picker for some reason)
- delete `LegacyEstimatedRevenue` component that should not be there
- update `DateRangeSelector` and [stories](http://localhost:6006/?path=/docs/components-daterangeselector--docs)
- fix file names' extension (`jsx` -> `js`, thank you Joyce!)

Jira link: https://lite-farm.atlassian.net/browse/LF-3637

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/webapp/src/components/DateRangeSelector/index.jsx
```diff
@@ -27,10 +27,9 @@ import styles from './styles.module.scss';
 export default function DateRangeSelector({
```
Duncan-Brain, line 27: Small thing I didn't notice before -- there is a slight but handled naming conflict with dateRangeSelector from the state selectors in Finances.   `/components/Finances/DateRangeSelector/index.jsx` is at least one file that imports both {dateRangeSelector} and DateRangeSelector.

I don't have any great name alternatives haha

Duncan-Brain:
This is looking great!

I love that the forms have been removed from the plain data views -- but I kind of expected the stored range to clear at some point when leaving a "persisted area". For example if I select a range in finances and go to map and then back -- the range is still there.

I also saw these weird situations:
- On custom range: clicking on the dropdown and then "out" instead directly of "out of range picker" does not return the selection to what is shown, or apply the filter. 

https://github.com/LiteFarmOrg/LiteFarm/assets/30075973/2a48c94e-018b-4b2e-9d6a-6d73349be08e

- Possibly related to above: if there is a startDate and endDate "stored". It does not apply when swapping between options and clicking out normally:

https://github.com/LiteFarmOrg/LiteFarm/assets/30075973/7f7a881d-c76c-43a2-b804-bdf74e645691

SayakaOno:
@Duncan-Brain Thank you so much for catching those! It looks like the last changes I made weren't good... I will fix them!

The date range not being cleared is the existing behaviour, so it might be expected...? I will try asking David!

SayakaOno:
I fixed [bugs](https://github.com/LiteFarmOrg/LiteFarm/pull/2913#issuecomment-1775520157) that Duncan found.
I updated the description, but the date range option in the store is now updated to what the user selected immediately (it was waiting for the valid range to be added before).

packages/webapp/src/components/DateRangeSelector/index.jsx
SayakaOno: Is it known to cause any problems??

Duncan-Brain: It's just a best practice to avoid having the same name in one file. The main problem with non unique names is developer confusion (in javascript). As you know this will work. Communication becomes harder.

SayakaOno: Understood.

antsgar (commented):
Left a couple of comments, but it works great 🔥 awesome work!

The only additional feedback I have is that it's a bit jarring to me that when I click on the selector and start typing to "search" for an option, the cursor shows up where the icon is, instead of where the text starts -- but I'm not sure that's easily fixable with our current select component. Is it required that the date range selector should be "searchable"?

packages/webapp/src/components/DateRangeSelector/useDateRangeSelector.js
```diff
@@ -0,0 +1,46 @@
...
+  const dateRange = useSelector(dateRangeSelector);
+  const dispatch = useDispatch();
+
+  const dateRangeUtil = new DateRange(new Date(), weekStartDate);
+  const option = dateRange.option || dateRangeOptions.YEAR_TO_DATE;
+
+  useEffect(() => {
+    if ((!dateRange.startDate || !dateRange.endDate) && option !== dateRangeOptions.CUSTOM) {
+      const { startDate, endDate } = dateRangeUtil.getDates(option);
+      dispatch(setDateRange({ option, startDate, endDate }));
```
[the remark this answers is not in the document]
SayakaOno: It's possible, but we might want to change "start date" of the week (Sunday or Monday) depending on the user in the future, so I decided not to set the default in the reducer.

packages/webapp/src/components/DateRangeSelector/index.jsx
```diff
@@ -111,13 +115,16 @@ export default function DateRangeSelector({
...
-    if (dateRange.option !== rangeOptions.CUSTOM) {
+    if (!isCustomDatePickerOpen) {
       return;
     }
-    if (!isValid) {
+    setIsCustomDatePickerOpen(false);
+  };
+
+  const cleanup = () => {
+    if (!isValidRef?.current) {
```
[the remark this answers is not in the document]
SayakaOno: I think state used in the cleanup function references the value at the first render, so I guess I would need to create a ref in the child component as well... but I will do some cleanup.

SayakaOno (commented):
@antsgar thank you for reviewing!
I don't believe the date range selector needs to be searchable. I will disable the functionality!

Duncan-Brain (dismissed):
Nice! Changes looks good to me!

Duncan-Brain (dismissed):

packages/webapp/src/components/DateRangeSelector/CustomDateRangeSelector.jsx
```diff
@@ -33,10 +33,25 @@ export default function CustomDateRangeSelector({
   changeEndDate,
   fromDateMax,
   toDateMin,
+  resetDateRangeOption,
 }) {
   const { t } = useTranslation();
   const inputStyle = { classes: { container: { minWidth: 'calc((100% - 24px) / 2)' } } };
 
+  const isValidRef = useRef(isValid);
```
antsgar, line 41: Ohhh, you're right, this won't work without a ref either. I do think we should try to find a solution here not involving refs, since refs are usually meant to be used as a last resource kind of thing because they step out of the "normal" React workflow, so in general they're used for logic that won't affect rendering which isn't the case here.

I'm curious about what wasn't working with your first approach where you checked `isValid` on the `onClickAway` method?

```js
const onClickAway = () => {
    if (!isCustomDatePickerOpen) {
      return;
    }
    if (!isValid) {
      setSelectedDateRangeOption(options[0]);
    }
    setIsCustomDatePickerOpen(false);
  };
```

That seems to be working well in my testing but I could be missing something. Alternatively we could also setup an effect like so

```js
useEffect(() => {
  if (!isValid & !isCustomDatePickerOpen) {
    setSelectedDateRangeOption(options[0]);
  }
}, []);
```

antsgar (commented):
Minor thing but I'm wondering if in the custom date range option we should show `yyyy/mm/dd - yyyy/mm/dd` as a placeholder instead of `yyyy-mm-dd - yyyy-mm-dd` to keep it consistent with the pickers below and avoid the double dash which is a bit harder to read. I think Loïc used `yyyy.mm.dd` in his designs so maybe that's a good option too for both cases.

packages/webapp/src/components/DateRangeSelector/CustomDateRangeSelector.jsx
SayakaOno: In addition to clicking away, the picker can be closed by clicking the selector as you can see in Duncan's first screen recording [here](https://github.com/LiteFarmOrg/LiteFarm/pull/2913#issuecomment-1775520157) (0:55~).
That was what I had to handle also, and I used a ref.

SayakaOno (commented):
Is the placeholder `yyyy/mm/dd` for you? Mine is `yyyy-mm-dd`...
I don't think we have used `yyyy.mm.dd` format, so I wasn't sure if we would want to go for it, but I guess it's not a big deal, so I will change it.
Screenshot 2023-10-25 at 12 07 30 PM

Duncan-Brain (approved):
I like dots

packages/webapp/src/components/DateRangeSelector/CustomDateRangeSelector.jsx
antsgar: Ahhh gotcha, thank you for clarifying! Would the `useEffect` alternative work for that scenario? It should actually be something like

```js
useEffect(() => {
  if (!isValid & !isCustomDatePickerOpen) {
    setSelectedDateRangeOption(options[0]);
  }
}, [isValid, isCustomDatePickerOpen]);
```

I had missed the dependencies on the other comment!

SayakaOno: It didn't work with the validation that I had, but I removed it, so it should be good. The validation was like an insurance, so we should not have a problem without it.

antsgar (approved):

PR #2919, opened 2023-10-20
LF-3742 create hook to return list of all farm transactions

antsgar:
**Description**

- Added a couple of util functions to extract the logic that was currently being used to show the labour expenses and the revenue items. I needed to reuse a lot of this logic for the transactions, so I moved it to a utility to be consumed both by my new hook and by the components that had this logic previously.
- Added `useTransactions` hook that contains the logic to group revenues and expenses by transaction and returning the data in the right format to be consumed by the transaction list in the dashboard.

Jira link:
https://lite-farm.atlassian.net/browse/LF-3742

**Type of change**

**How Has This Been Tested?**

- [x] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/containers/Finances/useTransactions.js
+++ b/packages/webapp/src/containers/Finances/useTransactions.js
@@ -103,6 +103,7 @@ const buildExpenseTransactions = ({ expenses, expenseTypes, dateFilter, expenseT
           : i18n.t(`expense:${expenseType?.expense_translation_key}.EXPENSE_NAME`),
         amount: -roundToTwoDecimal(expense.value),
         note: expense.note,
+        relatedId: expense.farm_expense_id,
       };
     });
 };
@@ -133,10 +134,11 @@ const buildRevenueTransactions = ({
       transactionType: transactionTypeEnum.revenue,
       typeLabel: revenueType?.farm_id
         ? revenueType?.revenue_name
-        : i18n.t(`revenue:${revenueType?.revenue_translation_key}`),
+        : i18n.t(`revenue:${revenueType?.revenue_translation_key}.REVENUE_NAME`),
       amount: item.totalAmount,
       note: item.sale.customer_name,
       items: item.financeItemsProps,
+      relatedId: item.sale.sale_id,
     };
   });
 };
--- a/packages/webapp/src/tests/buildTransactions.test.jsx
+++ b/packages/webapp/src/tests/buildTransactions.test.jsx
@@ -236,9 +236,10 @@ const allResults = [
     icon: 'INFRASTRUCTURE',
     date: '2023-10-28T03:00:00.000Z',
     transactionType: 'EXPENSE',
-    typeLabel: 'INFRASTRUCTURE',
+    typeLabel: 'INFRASTRUCTURE.EXPENSE_NAME',
     amount: -20,
     note: 'Expense 2',
+    relatedId: 'c239a46a-69e6-11ee-8e6f-0242ac180006',
   },
   {
     date: '2023-10-24T00:00:00.000',
@@ -265,15 +266,16 @@ const allResults = [
     icon: 'INFRASTRUCTURE',
     date: '2023-10-19T03:00:00.000Z',
     transactionType: 'EXPENSE',
-    typeLabel: 'INFRASTRUCTURE',
+    typeLabel: 'INFRASTRUCTURE.EXPENSE_NAME',
     amount: -100,
     note: 'Expense 3',
+    relatedId: '39bc4644-6e8a-11ee-a583-0242ac180005',
   },
   {
     icon: 'CROP_SALE',
     date: '2023-10-18T00:00:00.000Z',
     transactionType: 'REVENUE',
-    typeLabel: 'CROP_SALE',
+    typeLabel: 'CROP_SALE.REVENUE_NAME',
     amount: 30,
     note: 'Customer 2',
     items: [
@@ -294,6 +296,7 @@ const allResults = [
         quantityUnit: 'kg',
       },
     ],
+    relatedId: 16,
   },
   {
     icon: 'CUSTOM',
@@ -303,6 +306,7 @@ const allResults = [
     amount: 200,
     note: 'Customer 3',
     items: [{ key: 18, title: 'Custom type', amount: 200 }],
+    relatedId: 18,
   },
   {
     date: '2023-10-14T00:00:00.000',
@@ -317,15 +321,16 @@ const allResults = [
     icon: 'EQUIPMENT',
     date: '2023-10-13T03:00:00.000Z',
     transactionType: 'EXPENSE',
-    typeLabel: 'EQUIPMENT',
+    typeLabel: 'EQUIPMENT.EXPENSE_NAME',
     amount: -10,
     note: 'Expense 1',
+    relatedId: 'aff01e2e-69e6-11ee-9b0c-0242ac180006',
   },
   {
     icon: 'CROP_SALE',
     date: '2023-10-12T00:00:00.000Z',
     transactionType: 'REVENUE',
-    typeLabel: 'CROP_SALE',
+    typeLabel: 'CROP_SALE.REVENUE_NAME',
     amount: 100,
     note: 'Customer 1',
     items: [
@@ -338,6 +343,7 @@ const allResults = [
         quantityUnit: 'kg',
       },
     ],
+    relatedId: 9,
   },
 ];
 
@@ -347,7 +353,6 @@ const filterResultsByIndex = (indexList) =>
 describe('buildTransactions test', () => {
   test('Should return all transactions if no filters applied', () => {
     const transactions = buildTransactions(testData);
-    console.log(transactions.length);
     expect(transactions).toEqual(allResults);
   });
 


packages/webapp/package.json
```diff
@@ -16,7 +16,7 @@
     "build-storybook": "storybook build",
     "test-storybook": "test-storybook",
     "precommit": "lint-staged",
-    "test": "vitest"
+    "test": "vitest --environment happy-dom"
```
antsgar, line 19: Had to add this here because I use `i18n` in the function and it doesn't work unless it runs in a browser environment.

packages/webapp/src/containers/Finances/useTransactions.js
```diff
@@ -0,0 +1,241 @@
...
+        amount: revenueItems.reduce((sum, item) => sum + item.totalAmount, 0),
+        items: revenueItems,
+      });
+    });
+  });
+
+  return groupedTransactions;
+};
+
+export const buildTransactions = ({
```
antsgar, line 162: I exported this function so that I could unit test it by passing mock data for sales, expenses, etc., without having to mock the Redux store or selectors.

packages/webapp/vitest.config.ts
```diff
@@ -0,0 +1,12 @@
+import { defineConfig, mergeConfig } from 'vitest/config';
+import viteConfig from './vite.config';
+
+export default mergeConfig(
+  viteConfig,
+  defineConfig({
+    test: {
+      include: ['src/tests/**/*.test.js?(x)'],
+      environment: 'happy-dom',
```
antsgar, line 9: Added this so that the tests run in a browser environment. In the case of my function since it uses i18n, it fails unless run in a browser environment.

.github/workflows/webapp_unit_tests.yml
```diff
@@ -0,0 +1,24 @@
+name: "Webapp Unit Tests"
```
antsgar, line 1: Added this so that unit tests will run automatically on every PR (like the BE ones).

packages/webapp/package.json
```diff
@@ -45,6 +45,7 @@
     "i18next-xhr-backend": "^3.2.2",
     "immer": "^9.0.16",
     "jszip": "^3.10.1",
+    "lodash-es": "^4.17.21",
```
antsgar, line 48: I'm using this as fallback because some browsers/the testing environment doesn't have an implementation for the new Object.groupBy functionality. As it starts being more used, we should be able to get rid of this eventually.

packages/webapp/src/tests/util.test.js
```diff
@@ -13,19 +13,25 @@
  *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
  */
 
+import { expect, describe, test } from 'vitest';
```
antsgar, line 16: This is totally unrelated to this PR but since I added automatically running tests in the CI I had to fix this one so it wouldn't fail.

packages/webapp/src/containers/Finances/util.js
```diff
@@ -116,3 +121,108 @@ export function calcActualRevenue(sales, startDate, endDate, revenueTypes) {
...
+    });
+    labourItemGroups[option.label] = items;
+  });
+
+  return labourItemGroups;
+};
+
+export const mapSalesToRevenueItems = (sales, revenueTypes, cropVarieties) => {
+  const revenueItems = sales.map((sale) => {
+    const revenueType = revenueTypes.find(
```
kathyavini, line 184: ActualRevenue is passing functions (`getRevenueTypes`, `getCropVarieties`) not arrays.

Did you mean to use the passed functions here

```
const revenueType = getRevenueTypes(sale.revenue_type_id)
```

Or is the function call incorrect?

antsgar: Thank you for catching this! I changed course so many times that this was just a lingering thing from a previous approach 😅 should be fixed now, I'm passing the arrays instead of the functions

kathyavini: Thank you, all good now!

packages/webapp/src/containers/Finances/useTransactions.js
```diff
@@ -0,0 +1,241 @@
...
+        tasks,
+        expenses,
+        expenseTypes,
+        revenueTypes,
+        taskTypes,
+        cropVarieties,
+        users,
+        dateFilter,
+        expenseTypeFilter,
+        revenueTypeFilter,
```
kathyavini, line 223: I wonder if it would make sense to return all transactions (rather than return no transactions) if `revenueTypeFilter` and `expenseTypeFIlter` are not provided. I noticed this is how `dateFilter` behaves, and it will also save type selectors having to get imported in the transaction list container to populate the default (unfiltered) view when they are already also called here.

Maybe this isn't the best way to write it, but the behaviour could be equivalent to something like this:

```suggestion
        expenseTypeFilter: expenseTypeFilter || [
          ...expenseTypes.map(({expense_type_id}) => expense_type_id),
          'LABOUR_EXPENSE',
        ],
        revenueTypeFilter: revenueTypeFilter || revenueTypes.map(({revenue_type_id}) => revenue_type_id),
```

packages/webapp/src/containers/Finances/useTransactions.js
```diff
@@ -0,0 +1,241 @@
...
+    );
+
+    Object.keys(groupedByDateSales).forEach((date) => {
+      const revenueItems = mapSalesToRevenueItems(
+        groupedByDateSales[date],
+        revenueTypes,
+        cropVarieties,
+      );
+      groupedTransactions.push({
+        icon: revenueType?.farm_id ? 'OTHER' : revenueType?.revenue_translation_key,
```
kathyavini, line 147: This icon key is called `'CUSTOM'` in revenues.

This is lucky (or maybe @SayakaOno thought ahead and made it this way on purpose?!) because the revenue and expense icon keys won't conflict with each other when mixed in a single list.

kathyavini (commented):
I like this refactor so much!! I didn't quite grok the need for a new hook until I saw it in action, but I think this is big improvement in organization for all of finances and it's soooo usable for labour!! 🙂 

The main question for me was the grouping of revenues by revenue type, per day. Maybe the conversation around this has progressed beyond the last group discussion I participated in, but at that time we had made the decision to keep revenues grouped by sale, with customer name as the title in the transaction list (= equivalent of expense item name/`note`). This fits with what is [in Figma now](https://www.figma.com/file/96NZ02oYe3jpet1roUp0s0/Mockups?node-id=4438%3A97319&mode=dev) with the "View & Edit" link, because I believe that needs to direct to a single sale (the view that Duncan's big PR created).

Is that still accurate or did I miss a subsequent discussion?

packages/webapp/src/containers/Finances/util.js
```diff
@@ -116,3 +121,108 @@ export function calcActualRevenue(sales, startDate, endDate, revenueTypes) {
...
+      (revenueType) => revenueType.revenue_type_id === sale.revenue_type_id,
+    );
+    if (getRevenueFormType(revenueType) === revenueFormTypes.CROP_SALE) {
+      const quantityUnit = getMassUnit();
+      const cropVarietySale = sale.crop_variety_sale;
+      return {
+        sale,
+        totalAmount: cropVarietySale.reduce((total, sale) => total + sale.sale_value, 0),
+        financeItemsProps: cropVarietySale.map((cvs) => {
```
SayakaOno, line 193: Do we expect to use `financeItemsProps` for the desktop view as well in the future?
If so, I think:
- title and subTitle should be `crop` and `quantity` (they have dedicated columns in the desktop view) 
- dealing with numbers as string is a little tricky when sorting, so I think adding `quantityUnit` later is easier.

SayakaOno (commented):
I am not confident that I could review this perfectly, but the transaction list returned from the hook looks good overall!
I left one question regarding the hook :)

antsgar:
> I like this refactor so much!! I didn't quite grok the need for a new hook until I saw it in action, but I think this is big improvement in organization for all of finances and it's soooo usable for labour!! 🙂
> 
> The main question for me was the grouping of revenues by revenue type, per day. Maybe the conversation around this has progressed beyond the last group discussion I participated in, but at that time we had made the decision to keep revenues grouped by sale, with customer name as the title in the transaction list (= equivalent of expense item name/`note`). This fits with what is [in Figma now](https://www.figma.com/file/96NZ02oYe3jpet1roUp0s0/Mockups?node-id=4438%3A97319&mode=dev) with the "View & Edit" link, because I believe that needs to direct to a single sale (the view that Duncan's big PR created).
> 
> Is that still accurate or did I miss a subsequent discussion?
> 
> (The other two in-line comments are minor!)

Ohhh this is a really good point! We chatted about it with Loïc and he mentioned we should group revenues that have the same date and type in the same transaction, but you're right that I'm not sure how the "View & edit" link would make sense in that scenario. Grouping by sale is also a lot easier from the logic point of view in the code. I'll double check with Loïc but I'm pretty sure you're right that grouping by sale will make more sense here.

packages/webapp/src/containers/Finances/useTransactions.js
antsgar: I had originally implemented it this way, but then I figured we'd need a way for the user to enable all expense type filters and no revenue type filters to only view expenses and no revenues, or the other way round, and in that scenario we'd need some way for the UI to explicitly say we shouldn't show any revenue types or any expense types (which would be the empty arrays with the current implementation).

packages/webapp/src/containers/Finances/useTransactions.js
antsgar: Ohhh good catch, will update!

packages/webapp/src/containers/Finances/util.js
```diff
@@ -116,3 +121,108 @@ export function calcActualRevenue(sales, startDate, endDate, revenueTypes) {
...
+
+  return labourItemGroups;
+};
+
+export const mapSalesToRevenueItems = (sales, revenueTypes, cropVarieties) => {
+  const revenueItems = sales.map((sale) => {
+    const revenueType = revenueTypes.find(
+      (revenueType) => revenueType.revenue_type_id === sale.revenue_type_id,
+    );
+    if (getRevenueFormType(revenueType) === revenueFormTypes.CROP_SALE) {
```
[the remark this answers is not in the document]
antsgar: Yup good catch! I'll wait for Duncan's PR to be merged first and then update this 🙂

packages/webapp/src/containers/Finances/Labour/Employee/index.jsx
```diff
@@ -68,7 +47,7 @@ const Employee = ({ currencySymbol, tasks, startDate, endDate }) => {
     {
       id: 'labour_cost',
       Header: t('SALE.LABOUR.TABLE.LABOUR_COST'),
-      accessor: (d) => d.labour_cost,
+      accessor: (d) => `${currencySymbol}${d.labourCost.toFixed(2)}`,
```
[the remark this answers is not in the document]
antsgar: Ohhh, that's a great callout! Will try this out

packages/webapp/src/containers/Finances/useTransactions.js
```diff
@@ -0,0 +1,241 @@
...
+
+  const sortedTransactions = transactions.sort((a, b) => {
+    if (a.date > b.date) {
+      return -1;
+    }
+    if (a.date < b.date) {
+      return 1;
+    }
+    return 0;
+  });
```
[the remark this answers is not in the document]
antsgar: Niiiice! 🎉

packages/webapp/src/containers/Finances/util.js
antsgar: Yeah I think you're right, it'd be good to have this be adaptable to the desktop view too so it's less rework in the future. Will update this!

packages/webapp/src/containers/Finances/useTransactions.js
kathyavini: Empty arrays are truthy so they should still work with the above (they won't default to the righthand side)

antsgar: You're absolutely right! I've just updated this 🙂

SayakaOno (approved):
Looks good to me!

kathyavini (commented):
Looks good to me! Just needs the update from Duncan's PR merge.

And then I was wondering where the revenue organization by sale discussion ended up -- will that be part of this PR?

antsgar:
> Looks good to me! Just needs the update from Duncan's PR merge.
> 
> And then I was wondering where the revenue organization by sale discussion ended up -- will that be part of this PR?

@kathyavini yes! Loïc said thumbs up for grouping by sale, so I've already included this change in my latest commits. The hook no longer groups by date/type, just by sale.

kathyavini (approved):
> I've already included this change in my latest commits. The hook no longer groups by date/type, just by sale.

Ack sorry I thought that decision was still in limbo so I didn't even think to check for it in the PR comments commit! 🤦‍♀️

antsgar:
> > I've already included this change in my latest commits. The hook no longer groups by date/type, just by sale.
> 
> Ack sorry I thought that decision was still in limbo so I didn't even think to check for it in the PR comments commit! 🤦‍♀️

No worries, it's hard to keep up! 😅

packages/webapp/src/containers/Finances/useTransactions.js
```diff
@@ -0,0 +1,215 @@
...
+      const expenseType = expenseTypes.find(
+        (expenseType) => expenseType.expense_type_id === expense.expense_type_id,
+      );
+      return {
+        icon: expenseType?.farm_id ? 'CUSTOM' : expenseType?.expense_translation_key,
+        date: expense.expense_date,
+        transactionType: transactionTypeEnum.expense,
+        typeLabel: expenseType?.farm_id
+          ? expenseType?.expense_name
+          : i18n.t(`expense:${expenseType?.expense_translation_key}`),
```
kathyavini, line 103: For the same bug as here:

- https://github.com/LiteFarmOrg/LiteFarm/pull/2924

```suggestion
          : i18n.t(`expense:${expenseType?.expense_translation_key}.EXPENSE_NAME`),
```

SayakaOno (approved):

kathyavini (commented):
Just needs a tiny fix for this now! (So many conflicts from all working in the same files!)

packages/webapp/src/containers/Finances/useTransactions.js
```diff
@@ -0,0 +1,215 @@
...
+      revenueTypeFilter,
+    }),
+  ];
+
+  const sortedTransactions = transactions.sort(getComparator('desc', 'date'));
+
+  return sortedTransactions;
+};
+
+const useTransactions = ({ dateFilter, expenseTypeFilter, revenueTypeFilter }) => {
```
SayakaOno, line 174: Sorry @antsgar, I have a request that I should have made earlier...
Screenshot 2023-10-24 at 3 17 58 PM
I think we need `sale_id` and `farm_expense_id` to navigate to the detail page of each transaction. Would you be able to add them?

SayakaOno (approved):

kathyavini (approved):
❤️

PR #2720, opened 2023-06-07
Lf 3271 remove unused code

Duncan-Brain:
**Description**

This continues the work of #1920. It is a whopping 14,305 deletions and 94 files changed. Some of the feature deletion may extend beyond the other PR files touched but tries to keep the same scope.

The commits try to focus on one single feature at a time. This should allow the ability to git cherry pick and revert any feature that is wanted in the future. There are a few commits that apply to a few features.

The commits remove the following:
- unused crypto keypair utility
- export all farm data feature (GDPR?)
- Insight - People Fed
- Insight - Erosion
- Insight - Water Balance & Scheduler
- Insight - Nitrogen balance & Scheduler
- Insight - shared components between Water Balance and Nitrogen Balance
- Sales Summary
- unused getAuth0Url utility
- Shift and shift task feature
- continue removing shift and shift task feature
- unused userFarm enum
- unused Job - compress image
- unused URLs in apiConfig
- Insight - helper function
- non-existent routes
- merge integration

Things I did not keep from #1920:
- CustomSignUp/saga.js Url improvements
- deletion of:
  - station_sync/mapping
  - certifier_country Model
  - country Model
  
 Things/references that were not deleted:
- migrations - for obvious reasons
- seed table database - need to review if this affects migrations or testing
- Unused Tables in the database - Will make a jira ticket to review this
- Other unused files/features discovered along the way - Will make another Jira ticket for this as well

Using `git ls-files | xargs wc -l` we see there are about ~501,783 lines of code so this is a 3% of that lol.

Jira link: [LF-3271](https://lite-farm.atlassian.net/browse/LF-3271)

**Type of change**

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

**How Has This Been Tested?**

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [ ] Passes test case
- [ ] UI components visually reviewed on desktop view
- [ ] UI components visually reviewed on mobile view
- [ ] Other (please explain):

There are a few areas I would expect to see issues if they were being caused:
- Biodiversity insights
- Finances: labour expenses (there are existing console errors here with the dropdown)
- Authentication: (shift and shift task was the most intertwined but still unused)

Jest tests:
- still are working for me with less tests skipped
[...]

PR #2923, opened 2023-10-24
LF-3757 Remove associated with agriculture question and logic from revenue types

kathyavini:
**Description**

Kevin's instructions in the Slack thread (linked in Jira) read:

> I'm hoping we can just hide the question and null the values for now rather than remove it from the schema.

So I initially removed `agriculture_associated` everywhere except the database (because of the reference to schema), but the phrase "hide the question" (i.e. comment out, I think) makes me believe we want to keep this as reversible as possible. Therefore I have:

- nulled the value in the database (without changing underlying database schema) or default value
- ~~_commented out_~~ deleted all references to the dependent logic between `agriculture_associated` and `crop_generated` in backend and frontend
- retained `agriculture_associated` as a required property but set it to `null` in the model. This means anything other than a `null` value (as the frontend now supplies) will cause an error, but when we are ready to reverse this change we can still do a text string search on `agriculture_associated` to restore the previous behaviour

Jira link: https://lite-farm.atlassian.net/browse/LF-3757

**Type of change**

**How Has This Been Tested?**

- [x] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/api/db/migration/20231024164318_null_agricuture_associated.js
+++ b/packages/api/db/migration/20231024164318_null_agricuture_associated.js
@@ -1,0 +1,36 @@
+/*
+ *  Copyright (c) 2023 LiteFarm.org
+ *  This file is part of LiteFarm.
+ *
+ *  LiteFarm is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  LiteFarm is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ */
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+export const up = async function (knex) {
+  await knex('revenue_type').update({
+    agriculture_associated: null,
+  });
+};
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+export const down = async function (knex) {
+  /* Down function will set agricuture_associated to its default value, but any previous data in this column cannot be restored -- i.e. the data migration is irreversible */
+
+  await knex('revenue_type').update({
+    agriculture_associated: true,
+  });
+};
--- a/packages/api/src/controllers/revenueTypeController.js
+++ b/packages/api/src/controllers/revenueTypeController.js
@@ -27,11 +27,12 @@ const revenueTypeController = {
         const data = req.body;
         data.revenue_translation_key = baseController.formatTranslationKey(data.revenue_name);
 
-        if (!data.agriculture_associated && data.crop_generated) {
-          return res.status(400).json({
-            error: 'Revenues not associated with agriculture cannot be generated from crops',
-          });
-        }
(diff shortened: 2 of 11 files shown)


Duncan-Brain:
I might just note that it is still easily reversible by keeping a branch open with the changes in place or checking git history without leaving commented code in production.

kathyavini:
@Duncan-Brain What does this mean?

> by keeping a branch open, jira ticket blocked, with the changes in place

I don't mind deleting (that's what I started doing initially), I just questioned if that was in the spirit of what Kevin was asking for with "hiding" (vs "removing") the question. For April release he had us comment stuff out that was restored right afterwards... but not sure if this is one of those cases or more of an indefinite hide, in which case I would also prefer the comments gone 🙂

@antsgar what do you think?

Duncan-Brain:
I just meant you can keep an "add agricultural_associated back in" Jira ticket open with a branch that fixes it all ready to go. Then if you put that Jira ticket in the blocked column they may remember to revisit it.

I don't think Kevin cares how, just as long as it is easily undoable, and we keep your work done. I think we should follow good practice and remove the code since it could take a while for them to figure out what they want. This still fits the requirement of "as reversible as possible". Do you remember the feature graveyard that was "Remove unused code" https://github.com/LiteFarmOrg/LiteFarm/pull/2720 that was all on production haha

kathyavini:
@Duncan-Brain I remember the feature graveyard!!! 🤣 I'll delete the comments!!

packages/api/db/migration/20231024164318_null_agricuture_associated.js
```diff
@@ -0,0 +1,36 @@
+/*
```
Duncan-Brain, line 1: So, lol .. my interpretation of what Kevin was asking was -- "migrations are a pain/should be used sparingly so if we can remove this without affecting the db or schema that would be sweet" 

Just like the commented code and if we are doing a migration anyways... should we just remove it and add it back in with the rest of the "ready-to-go" PR since it is just as easy to dropColumn? We can set preexisting custom types to null at that point too if that has some usefulness. 

There are so many unused ghost tables and columns in the db that should be re-evaluated too.  For example in task table there is a "photo, and action_needed" alluding to unimplemented features of the past.

kathyavini: Hmmm, I'm not sure about that. If you read the full Slack thread, Kevin mentioned several times that he wanted the column to stay in the database specifically with value `null` value. I would prefer to write the migration that Kevin asked for, since it was the migration that Kevin asked for!

I was also thinking of the weird orphan unused columns that sprinkle the database (ugh), but the thread says this is a funded module and that question is relevant to research, so I don't think it's going to be abandoned entirely.

Duncan-Brain: I am good with @antsgar  deciding best practice here and me just approving. But I am seeing nothing about the importance of keeping the column for this rollback in that Slack thread so much as generally when making db columns providing the option for null vs default(true) -- and setting null when the value is unknown. 

Kevin:

> In the general case, I think it's really valuable to have null, false, true for data points. Otherwise you can make assumptions around the shape of the data that are actually just default values. .......  there's a lot of hands in the pot, so while we may "know" that this true doesn't mean anything, someone forking LiteFarm for their own uses may not.

Duncan-Brain (dismissed):

antsgar (commented):
Looks good to me! 🙌 only one minor detail -- since the "crop generated" question used to be below the agriculture one, it currently has a wider top padding of 32 px instead of 12px.

Duncan-Brain (approved):

antsgar (approved):

PR #2927, opened 2023-10-25
Lf 3679 add support for short description to custom expense revenue type creation read only and edit screens

Duncan-Brain:
**Description**

This adds the custom description to the edit, add, readonly views. 

Adds any updates to selectors, controller updates as needed to function.

Minor naming updates.

Fix expense and revenue name translations.

Note: For some reason I cant explain sometimes useTranslation needs the name of the file to use or it just displays the translation key.

Jira link: [LF-3679](https://lite-farm.atlassian.net/browse/LF-3679)

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

[LF-3679]: https://lite-farm.atlassian.net/browse/LF-3679?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/api/src/controllers/revenueTypeController.js
```diff
@@ -154,7 +155,10 @@ const revenueTypeController = {
       const trx = await transaction.start(Model.knex());
       const { revenue_type_id } = req.params;
       const farm_id = req.headers.farm_id;
-      const data = { revenue_name: req.body.revenue_name };
+      const data = {
+        revenue_name: req.body.revenue_name,
+        custom_description: req.body.custom_description ?? null,
```
Duncan-Brain, line 160: I need to change this to undefined so it passes tests and does not overwrite the description. 

But just figuring out why it is failing as is right now since null is valid.

Duncan-Brain: Classic forgot to migrate testing db goodbye 2 hours. 😢  Anyways this actual problem is fixed.

packages/api/src/controllers/revenueTypeController.js
```diff
@@ -148,7 +149,10 @@ const revenueTypeController = {
       const trx = await transaction.start(Model.knex());
       const { revenue_type_id } = req.params;
       const farm_id = req.headers.farm_id;
-      const data = { revenue_name: req.body.revenue_name };
+      const data = {
+        revenue_name: req.body.revenue_name,
+        custom_description: req.body.custom_description ?? undefined,
```
kathyavini, line 154: Why `undefined`? It looks like our ORM / db is going to coerce that into an empty string (is that what you see in your local db?) so wouldn't it make more sense to set it to an empty string explicitly here?

And this is an API check, right? Since the frontend will always send this field even if it's just an empty string?

packages/webapp/src/containers/Finances/AddSale/index.jsx
```diff
@@ -51,7 +55,7 @@ function AddSale() {
     <HookFormPersistProvider>
       <GeneralRevenue
         onSubmit={onSubmit}
-        title={t('common:ADD_ITEM', { itemName: revenueType?.revenue_name })}
+        title={t('common:ADD_ITEM', { itemName: translatedRevenueName })}
```
kathyavini, line 58: Ooops, how long did we have that bug in there 😆

packages/webapp/src/containers/Finances/CustomExpenseType/EditSimpleCustomExpense.jsx
```diff
@@ -46,10 +56,10 @@ function EditCustomExpense({ history, match }) {
         buttonText={t('common:SAVE')}
         pageTitle={t('EXPENSE.ADD_EXPENSE.CUSTOM_EXPENSE_TYPE')}
         inputLabel={t('EXPENSE.ADD_EXPENSE.CUSTOM_EXPENSE_NAME')}
-        customTypeRegister={CUSTOM_EXPENSE_NAME}
-        defaultValue={expense_name}
+        nameFieldRegisterName={CUSTOM_EXPENSE_NAME}
+        typeDetails={{ name: translatedExpenseName, description: translatedCustomDescription }}
         validateInput={hookFormUniquePropertyValidation(
-          expenseTypes,
+          expenseTypesWithoutSelectedType,
```
kathyavini, line 62: Subjective and completely optional but why not `otherExpenseTypes`?

packages/webapp/src/components/Forms/SimpleCustomType/constants.js
```diff
@@ -0,0 +1,17 @@
...
+ *  (at your option) any later version.
+ *
+ *  LiteFarm is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ */
+
+export const CUSTOM_DESCRIPTION = 'custom_description';
+export const DESCRIPTION_MAX_CHARS = 256;
```
kathyavini, line 17:
```suggestion
export const DESCRIPTION_MAX_CHARS = 255;
```

Submitting 256 chars produced this 400 response for me:

Screenshot 2023-10-26 at 3 35 59 PM

And I got this pasting the same 256 char string into my db directly:

Screenshot 2023-10-26 at 3 13 37 PM

Not sure where db got that 255 value though, I see it wasn't in the migration.

packages/webapp/src/stories/Pages/Finance/PureExpenseDetail.stories.jsx
```diff
@@ -30,7 +30,7 @@ const args = {
     console.log('Go back');
   },
   inputLabel: 'Custom Type Name',
-  customTypeRegister: 'custom_type_name',
+  nameFieldRegisterName: 'custom_type_name',
```
kathyavini, line 33: Oh no did I write this?? 😳 

It's a sloppy copy-paste from the story below! There's no need for those two props in this component (I was like... why would Expense Detail need to be Edited here??!) Sorry!! Can you delete both `inputLabel` and `nameFieldRegisterName`?

kathyavini (commented):
Looks good, works well!

The only thing that I though needed definite fixing is the max character (at least on my db; dunno if the limit on character varying might be one of those database-dependent default properties though, that would be kind of annoying!)

But I wanted to clarify the handling of the custom description column to make sure the intention was to have `null` in some cases and an empty string in others, and which cases in particular need each kind of "empty" datatype.

kathyavini (changes requested):
Discussed on our Slack thread with Loïc, but wanted to keep the GitHub updated too:

The description should be capped at 125 characters and the input for it should be a textarea input, not a one-line input (per Figma).

packages/api/src/controllers/farmExpenseTypeController.js
```diff
@@ -38,6 +38,7 @@ const farmExpenseTypeController = {
           } else {
             // if its deleted, them make it active
             record.deleted = false;
+            record.custom_description = data.custom_description ?? null;
```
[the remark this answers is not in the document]
Duncan-Brain: Good spotting!

Yeah I don't like how this is handled either with regards to un-retiring. But basically the logic is.. if they are un-retiring a type using the same name -- description supplied may be different so that should update. -- Really don't love it but if someone is making a new type that would normally succeed if there is a duplicate name than it should show all the inputted data. 

Any empty strings are definitely a mistake. And this should be undefined so as not to mess with the existing description. I just forgot to apply it to all the instances.

packages/api/src/controllers/revenueTypeController.js
Duncan-Brain: Yeah so it stays what it is before. I don't really like how this endpoint is not just using the raw request body and removing additional properties but yah.  Frontend should not send an empty string I hope but yeah mostly API check.

packages/api/src/controllers/farmExpenseTypeController.js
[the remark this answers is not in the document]
kathyavini: Don't worry about un-retiring via add as we are removing that before release ([Jira ticket](https://lite-farm.atlassian.net/browse/LF-3666)

packages/webapp/src/containers/Finances/AddSale/index.jsx
Duncan-Brain: 🙏  thanks! And no worries I might've done it too ahah I forget. I noticed it in the edit values which shouldn't really be able to see defaults but they can.

packages/webapp/src/stories/Pages/Finance/PureExpenseDetail.stories.jsx
Duncan-Brain: Good eye! I will check that out!

packages/webapp/src/components/Forms/SimpleCustomType/constants.js
kathyavini: Will do; I'll follow-up on the length if you don't get to connect with Loïc before end of day. With the move to textarea half of the main UX issue with the longer field is negated, and I agree with the value of easy-to-pick-out in a list.

Duncan-Brain: Thanks, appreciate that! -- such a minor detail but its fun to think about haha

packages/api/src/controllers/farmExpenseTypeController.js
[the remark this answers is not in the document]
Duncan-Brain: I like to keep things working in case it gets moved to nice to have haha.

But nullish coalescing got me in trouble here. Empty strings should not be permitted and I forget it works differently than ternary or regular logic.

packages/webapp/src/stories/Pages/Finance/PureExpenseDetail.stories.jsx
Duncan-Brain: gone!

Duncan-Brain:
Fixes required edits to InputAutoSize to match input placeholder. Cancel also added to match figma.

packages/api/src/controllers/farmExpenseTypeController.js
```diff
@@ -38,13 +38,16 @@ const farmExpenseTypeController = {
           } else {
             // if its deleted, them make it active
             record.deleted = false;
+            record.custom_description = data.custom_description || undefined;
```
kathyavini, line 41: **This is an optional suggestion, it doesn't change functionality!**

Now that I understand what the goal is with in the database column (`null` if it's not a string value, and no need to distinguish between API and frontend), I think the more common pre-existing pattern would be to clean it up between the `onSubmit` and the saga in the frontend, e.g.:

```js
  const onSubmit = (data) => {
    const payload = {
      ...data,
      custom_description: data.custom_description || undefined, // or null
    };
    dispatch(addCustomExpenseType(payload));
  };
  ```
 
If it stays in the controller you can move the `|| undefined` line to the top so it doesn't have to be duplicated in both logical branches.

File: packages/api/src/controllers/farmExpenseTypeController.js [hunk 9]
```diff
@@ -156,6 +159,8 @@ const farmExpenseTypeController = {
         }
 
         data.expense_translation_key = baseController.formatTranslationKey(data.expense_name);
+        //prevent empty strings
+        data.custom_description = data.custom_description || undefined;
```
kathyavini, line 163: **This is actual buggy behaviour, it will need fixing**

The user needs to be able to delete an existing description by clearing that text area, but using `undefined` means that the original description will always persist once it's set. It needs to be set explicitly to `null`.

If done in the frontend:
```jsx
  const onSubmit = (payload) => {
    dispatch(
      updateCustomExpenseType({
        ...payload,
        custom_description: payload.custom_description || null,
        expense_type_id,
      }),
    );
  };
  ```

Or leaving logic where it is:
```suggestion
        data.custom_description = data.custom_description || null;
```

packages/api/src/controllers/revenueTypeController.js
```diff
@@ -148,7 +151,10 @@ const revenueTypeController = {
       const trx = await transaction.start(Model.knex());
       const { revenue_type_id } = req.params;
       const farm_id = req.headers.farm_id;
-      const data = { revenue_name: req.body.revenue_name };
+      const data = {
+        revenue_name: req.body.revenue_name,
+        custom_description: req.body.custom_description || undefined,
```
kathyavini, line 156:
```suggestion
        custom_description: req.body.custom_description || null,
```

As above.

kathyavini (commented):
Changes look good and the text area in the form is 👌 There are just two remaining bits with the database/data handling -- one optional, and one that is a bug (I indicated which is which in the inline comments)

@antsgar with Duncan on leave, what is the best way to address these?

antsgar:
@kathyavini I've made the changes requested in this PR, can you take a look again when you have a minute?

kathyavini (approved):
Looks good, thanks! That's a good way to handle the un-retire if we indeed end up having to leave it in.

PR #2931, opened 2023-10-26
E2e test mika

mika-robots:
Description
This is another attempt at adding cypress e2e test.
Few notes:

- The coverage was removed. It didn't not seem to produce accurate results and we should spend some time on doing properly before re-enabling
- There are few explicit waits which I could't avoid. They are related to clicking on the map to create locations

Jira link:

Type of change

Non breaking change that enable automated cypress tests
There are no changes to any files outside of end-to-end folder
How Has This Been Tested?

I run it both locally and on my branch

Checklist:

[ x] I have commented my code, particularly in hard-to-understand areas
[ x] My changes generate no new warnings
[ x] I have added tests that prove my fix is effective or that my feature works
[ x] New and existing unit tests pass locally with my changes
[ x] The precommit and linting ran successfully
[ x] I have added or updated language tags for text that's part of the UI
[ x] I have added "MISSING" for all new language tags to languages I don't speak
[ x] I have added [the GNU General Public License](https://lite-farm.atlassian.net/l/cp/BT0Dd7WW) to all new files


kathyavini (approved):
I'm sorry for taking so long to review this!!

I think this looks awesome and the more modular test design is 👌 . Since we took so long to get to this, some selectors and behaviours are now out of date -- I noted a few below -- but I would love to merge first to get this into the repo, and then open an update PR or two to implement those needed changes, which will touch on `packages/webapp` as well.

packages/end-to-end/cypress/support/commands.js
```diff
@@ -1,25 +1,246 @@
...
+  //certification summary
+  cy.get('[data-cy=certificationSummary-continue]').should('exist').and('not.be.disabled').click();
+
+  // Outro
+  cy.url().should('include', '/outro');
+  cy.get('[data-cy=outro-finish]').should('exist').and('not.be.disabled').click();
+});
+
+Cypress.Commands.add('createFirstLocation', (fieldString) => {
+  cy.get('[data-cy=home-farmButton]').should('exist').and('not.be.disabled').click({ force: true });
```
antsgar, line 146: We could store these selector strings somewhere to avoid having to update them in several different files when there's a UI change, e.g.

`const FARM_BUTTON_SELECTOR = '[data-cy=home-farmButton]'`

packages/end-to-end/cypress/e2e/crops.js
```diff
@@ -0,0 +1,146 @@
...
+describe('Crops', () => {
+  let users;
+  let translation;
+  let crops;
+
+  beforeEach(() => {
+    // Load the users fixture before the tests
+    cy.fixture('e2e-test-users.json').then((loadedUsers) => {
+      users = loadedUsers;
+      const user = users[Cypress.env('USER')];
```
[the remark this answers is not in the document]
antsgar: Yes, these actually should be added either in the configuration file, or in its own separate env variables file. There's an overview here on the different methods and their pros/cons https://docs.cypress.io/guides/guides/environment-variables#Setting

packages/end-to-end/cypress/support/commands.js
```diff
@@ -1,25 +1,246 @@
...
+  cy.window()
+    .its('store')
+    .invoke('getState')
+    .its('entitiesReducer')
+    .its('gardenReducer')
+    .its('ids')
+    .should('not.be.empty');
+});
+
+Cypress.Commands.add('acceptSlideMenuSpotlights', (crop_menu_name) => {
```
antsgar, line 226: These last three commands, `addFarm`, `onboardCompleteQuestions` and `acceptSlideMenuSpotlights` are not reused in other test files but just by other commands here. With that in mind, they could probably just be regular functions instead of commands https://docs.cypress.io/api/cypress-api/custom-commands#1-Dont-make-everything-a-custom-command

antsgar (approved):

PR #2938, opened 2023-10-31
LF-3761: Implement crop sales expandable content

SayakaOno:
**Description**

- create `CropSale` transaction table
- pass `mobileView` `currencySymbol` to `ExpandedContent` (`mobileView` is for the future, and the style for the desktop view will not be shown and has not been finalized)

Some other improvements
- remove unused properties in TableV2 stories
- remove accessor prop that is incompatible with sorting from V2 table
- fix `ExpandableItem`'s `key`. (`index` was being used, and it was expanding a different transaction after the date range is updated)
- remove hovered colour of the expandable item when expanding
- update transactionDate's tag to `h4` (the title of the page "Finances" is `h3` so I chose `h4`)

Jira link: https://lite-farm.atlassian.net/browse/LF-3761

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/Finances/Transaction/ExpandedContent/CropSaleTable.jsx
+++ b/packages/webapp/src/components/Finances/Transaction/ExpandedContent/CropSaleTable.jsx
@@ -52,9 +52,9 @@ const getColumns = (t, mobileView, totalAmount, quantityTotal, currencySymbol) =
     id: 'amount',
     label: t('FINANCES.REVENUE'),
     align: 'right',
-    format: (d) => `${currencySymbol} ${Math.abs(d.amount).toFixed(2)}`,
+    format: (d) => `${currencySymbol}${Math.abs(d.amount).toFixed(2)}`,
     columnProps: {
-      style: { width: '250px', paddingRight: `${mobileView ? 9 : 75}px` },
+      style: { width: '100px', paddingRight: `${mobileView ? 9 : 75}px` },
     },
     Footer: mobileView ? null : <div className={styles.bold}>{totalAmount}</div>,
   },
@@ -74,7 +74,7 @@ export default function CropSaleTable({ data, currencySymbol, mobileView }) {
   const quantityUnit = items?.[0]?.quantityUnit;
   const quantityTotal = items.reduce((total, { quantity }) => total + quantity, 0);
   const quantityWithUnit = `${quantityTotal} ${quantityUnit}`;
-  const totalAmount = `${currencySymbol} ${amount.toFixed(2)}`;
+  const totalAmount = `${currencySymbol}${amount.toFixed(2)}`;
 
   if (!items?.length) {
     return null;
--- a/packages/webapp/src/components/Finances/Transaction/ExpandedContent/CropSaleTable.jsx
+++ b/packages/webapp/src/components/Finances/Transaction/ExpandedContent/CropSaleTable.jsx
@@ -72,8 +72,8 @@ export default function CropSaleTable({ data, currencySymbol, mobileView }) {
   const { t } = useTranslation();
   const { items, amount, relatedId } = data;
   const quantityUnit = items?.[0]?.quantityUnit;
-  let quantityTotal = items.reduce((total, { quantity }) => total + quantity, 0);
-  quantityTotal += ` ${quantityUnit}`;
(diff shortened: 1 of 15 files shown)


packages/webapp/src/components/Finances/Transaction/ExpandedContent/CropSaleTable.jsx
```diff
@@ -0,0 +1,97 @@
...
+    <div className={styles.bold}>{totalAmount}</div>
+  </div>
+);
+
+export default function CropSaleTable({ data, currencySymbol, mobileView }) {
+  const { t } = useTranslation();
+  const { items, amount, relatedId } = data;
+  const quantityUnit = items?.[0]?.quantityUnit;
+  let quantityTotal = items.reduce((total, { quantity }) => total + quantity, 0);
+  quantityTotal += ` ${quantityUnit}`;
```
antsgar, line 76: NIT, but I'd probably keep `quantityTotal` a constant and make the following line another const named something like `quantityWithUnit` to make it a bit easier to read and make both immutable

packages/webapp/src/components/Finances/Transaction/Mobile/styles.module.scss
```diff
@@ -42,7 +43,7 @@
     border-radius: 1px;
   }
 
-  &:hover {
+  &:hover:not(.expanded) {
```
antsgar, line 46: I kinda liked the hover effect on the expanded item because it showcases more clearly that the item can be clicked on to collapse, but not sure if Figma calls for anything specific in this situation

antsgar (dismissed):
It's working/looking great! 🎉

packages/webapp/src/components/Finances/Transaction/ExpandedContent/CropSaleTable.jsx
```diff
@@ -0,0 +1,97 @@
...
+    },
+    Footer: mobileView ? null : <div className={styles.bold}>{quantityTotal}</div>,
+  },
+  {
+    id: 'amount',
+    label: t('FINANCES.REVENUE'),
+    align: 'right',
+    format: (d) => `${currencySymbol} ${Math.abs(d.amount).toFixed(2)}`,
+    columnProps: {
+      style: { width: '250px', paddingRight: `${mobileView ? 9 : 75}px` },
```
kathyavini, line 57: Only noticed because I was to struggling to find the right column widths for my own table, but this seems like maybe too much width for mobile as it truncates the first column down to almost nothing. (I see passing the FooterCell separately fixes truncation there though... neat!)

Screenshot 2023-11-02 at 10 46 27 AM

kathyavini (commented):
Looks really good! I just saw one mobile view issue when I was trying to investigate the right way to use `shouldFixTableLayout`.

Also I noticed one bug -- code isn't from this PR, but the functionality is:

```
const componentKey = typeLabel === 'Crop Sale' ? 'CROP_SALE' : transactionType;
```

`typeLabel` is translated so this will only work in English. But maybe the componentKeys are due to be revisited in the next PR? I think association of custom crop-associated revenues with this table still has to be set up too, is that right?

SayakaOno (commented):
Thank you for reviewing! All comments should be addressed, and I fixed the transaction amount format ⬇️ to be "XXX.XX".
Screenshot 2023-11-02 at 2 11 55 PM
Re-review would be appreciated, thank you!

@kathyavini

> association of custom crop-associated revenues with this table still has to be set up too
True...!  Thank you so much for catching this!!! I added `cropGenerated` to the transaction!

antsgar (approved):

packages/webapp/src/components/Finances/Transaction/Mobile/styles.module.scss
SayakaOno: [Loïc's input](https://lite-farm.atlassian.net/browse/LF-3745?focusedCommentId=16353):
_It should not change colour on hover when the row is expanded, the reason is that it can be distracting when you process the info inside the expanded row._

kathyavini (approved):
I got a bit nervous when I noticed transactions weren't storing `crop_generated` but you fixed everything so easily!! Looks great!

PR #2940, opened 2023-11-01
LF-3748 Implement labour type transaction expandable content

kathyavini:
**Description**

This PR adds the labour table expandable content to the transaction list. I utilized an already built, but currently unused component in our codebase called `<StateTab />` to implement the table-type toggle.

The tables themselves are built from @SayakaOno's table code in Storybook (so please let me know if I have implemented it incorrectly!!) I have tried to also implement the patterns I saw in the [LF-3761 WIP PR](#2938). However, I did _not_:

- implement mobile vs. desktop view. If that should be implemented in this PR (vs when Desktop view is finalized), please let me know
- did not replicate the necessary changes to the `key` prop or with the properties passed in to the component (e.g `currencySymbol`) that are implemented in that PR. I think the `key` change can be auto-merged, but depending on which PR is merged first the `currencySymbol` handling could be updated to use the passed value. I did it this way just to not create another dependency between PRs (this one should stand alone -- the grouping by date does not affect it)

Jira link: https://lite-farm.atlassian.net/browse/LF-3748

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/webapp/src/components/Finances/Transaction/ExpandedContent/LabourTable.jsx
```diff
@@ -0,0 +1,173 @@
...
+import StateTab from '../../../RouterTab/StateTab';
+import { useCurrencySymbol } from '../../../../containers/hooks/useCurrencySymbol';
+
+export default function LabourTable({ transaction }) {
+  const { t } = useTranslation();
+  const currencySymbol = useCurrencySymbol();
+
+  const [activeTab, setActiveTab] = useState('employee');
+
+  const employeeTotalHours = useMemo(
```
kathyavini, line 30: It's quite possible -- @antsgar do you know off the top of your head? -- that the summation of hours in the `EMPLOYEE` and `TASK_TYPE` arrays _must_ be the same, but I didn't trust my knowledge of task edge cases enough to write it that way.

packages/webapp/src/components/Finances/Transaction/ExpandedContent/LabourTable.jsx
```diff
@@ -0,0 +1,173 @@
...
+
+  const tasksColumns = useMemo(
+    () => [
+      {
+        id: 'taskType',
+        label: t('SALE.LABOUR.TABLE.TASK'),
+        Footer: <div className={styles.footer}>{t('SALE.LABOUR.TABLE.DAILY_TOTAL')}</div>,
+        component: 'th',
+        columnProps: {
+          style: { width: '33%' },
```
SayakaOno, line 85: I'm sorry, but I didn't do a good job on writing stories, and the style is terrible... so it would be great if you could set whatever works best (if necessary)! If we need all columns to have the same width, 33% should be 33.33% or should be written using calc. (https://isotropic.co/how-to-1-3-of-100-33-333-in-css/)


https://github.com/LiteFarmOrg/LiteFarm/assets/33141219/4672eb31-3fd7-4181-b36d-1cbddcdbab07

To make the columns widths the same between the two tables, the same styles need to be added to `employeeColumns`, and `shouldFixTableLayout={true}` needs to be passed to the table component to make `width: 33%;` work. (it adds `table-layout: fixed;` to the table)
🙏🙏🙏

packages/webapp/src/components/Finances/Transaction/ExpandedContent/LabourTable.jsx
```diff
@@ -0,0 +1,173 @@
...
+
+  const employeeTotalHours = useMemo(
+    () => calculateTotalHours(transaction.items.EMPLOYEE),
+    [transaction],
+  );
+
+  const taskTotalHours = useMemo(
+    () => calculateTotalHours(transaction.items.TASK_TYPE),
+    [transaction],
+  );
```
SayakaOno, line 38: @antsgar 
I am curious about the approach you would take.
I would create variables for the activeTab only, so I would combine `employeeTotalHours` and `taskTotalHours` as `totalHours`, and I would create `columns` to return columns for the activeTab.
It would be less code and calculations for inactive tab would not happen.

However, Joyce's approach is good because re-calculation will not happen by changing the tab.

How would you make a decision?

SayakaOno (commented):
It looks great! I left a comment on the table layout, and...
- mobile/desktop views
   there will not be many differences between the two views, so we can work on it later :)
- `currencySymbol`
    I know, I did it by myself but dependency is frustrating...😓 I think we can just call `useCurrencySymbol()` in the tables as you did. Thank you for the note!

packages/webapp/src/components/Finances/Transaction/ExpandedContent/LabourTable.jsx
kathyavini: Thank you!! I actually thought there was some reason I didn't understand for applying the styling to only the tasks table so I kept it!! 😂 

I noticed that adding the styling to both sets of columns is enough to fix the shifting, and on desktop views adding `shouldFixTableLayout` doesn't do anything additional. However, on mobile it does truncate the column values rather than letting the table be horizontally scrollable. Do you know if that's the preferred behaviour? Is that what you will use on your tables?

packages/webapp/src/components/Finances/Transaction/ExpandedContent/LabourTable.jsx
```diff
@@ -0,0 +1,182 @@
...
+
+  const employeeColumns = useMemo(
+    () => [
+      {
+        id: 'employee',
+        label: t('SALE.LABOUR.TABLE.EMPLOYEE'),
+        Footer: <div className={styles.footer}>{t('SALE.LABOUR.TABLE.DAILY_TOTAL')}</div>,
+        columnProps: {
+          style: { width: '33.33%' },
```
kathyavini, line 47: @SayakaOno thank you for the 1/3rd width link! I couldn't figure out how to supply a fallback with a style object (specifically ESLint complained about duplicate keys)... do you know if it's possible? Otherwise I'll just keep the low-tech approach.

packages/webapp/src/components/Finances/Transaction/ExpandedContent/LabourTable.jsx
SayakaOno: Column widths might change depending on the length of the text that the columns hold on the desktop view...
The existing table which is being used in the labour page etc. truncates texts, so I tried not to change the behaviour when I created the new table. And yes, I added `shouldFixTableLayout` to my tables!

packages/webapp/src/components/Finances/Transaction/ExpandedContent/LabourTable.jsx
SayakaOno: I don't know if it's possible to set a fallback, sorry... but I believe calc is fine too!!

packages/webapp/src/components/Finances/Transaction/ExpandedContent/styles.module.scss
```diff
@@ -38,3 +37,30 @@ button.toDetail {
...
+.labourTable {
+  display: flex;
+  flex-flow: column nowrap;
+  justify-content: center;
+  align-items: center;
+  background: #f6fbfa;
+
+  .tabs {
+    margin: 24px 0 24px 0;
+    padding-block: 16px;
```
antsgar, line 50: The top margin seems to be around 8px according to Figma, looks closer to the "View" link

packages/webapp/src/components/Finances/Transaction/ExpandedContent/styles.module.scss
```diff
@@ -38,3 +37,30 @@ button.toDetail {
     margin-left: 8px;
   }
 }
+
+.labourTable {
+  display: flex;
+  flex-flow: column nowrap;
+  justify-content: center;
+  align-items: center;
+  background: #f6fbfa;
```
antsgar, line 46: I'd remove this background because it's already taking the background color from the parent container, and this one's causing an odd bit of green to show up when hovering over the item

packages/webapp/src/components/Finances/Transaction/ExpandedContent/LabourTable.jsx
antsgar: You could actually just read this from `transaction.amount`, which is the sum of the items amounts when the transaction is grouped by tasks and the number that shows up on the expandable item up top. I trusted that summing by task and by employee would give the same result so only did it, hope that's working correctly 😅

packages/webapp/src/components/Finances/Transaction/ExpandedContent/LabourTable.jsx
antsgar: Both are okay IMO! Although even though there are things in common between the two sets of columns, there's also enough differences that I think bundling them together might cause for the code to have a bunch of conditions which I think would make it a bit harder to read that it is right now. I wouldn't worry too much about the calculations in this case because the arrays are super straight forward and they should take virtually no time to be calculated

packages/webapp/src/components/Finances/Transaction/ExpandedContent/LabourTable.jsx
```diff
@@ -0,0 +1,182 @@
...
+            label: t('SALE.LABOUR.EMPLOYEES'),
+            key: 'employee',
+          },
+          {
+            label: t('SALE.LABOUR.TASKS'),
+            key: 'tasks',
+          },
+        ]}
+      />
+      {activeTab === 'employee' && (
```
antsgar, line 150: NIT, but I'd make an enum with the options for `activeTab`

antsgar:
Left a few comments but nothing major, it's looking awesome! 🙂

packages/webapp/src/components/Finances/Transaction/ExpandedContent/LabourTable.jsx
kathyavini: Well `transaction.amount` is the labour cost sum, while this refers to the hour sum!

packages/webapp/src/components/Finances/Transaction/ExpandedContent/styles.module.scss
```diff
@@ -43,10 +43,9 @@ button.toDetail {
   flex-flow: column nowrap;
   justify-content: center;
   align-items: center;
-  background: #f6fbfa;
 
   .tabs {
-    margin: 24px 0 24px 0;
+    margin: 4px 0px 24px 0px;
```
kathyavini, line 48: It looks as though I'm breaking the rule of 8 but it appears there was some margin/padding supplied by (I think) the table itself... pretty sure (at least by visual check) this ends up being 8px overall.

packages/webapp/src/components/Finances/Transaction/ExpandedContent/LabourTable.jsx
kathyavini: I'm not 100% sure I did that right; can you take a look and make sure it fits what you were thinking?

kathyavini:
@SayakaOno and @antsgar sorry I was a bit slow with the revisions, but I think everything is addressed. If you could please re-review when you get the chance 🙏

packages/webapp/src/components/Finances/Transaction/ExpandedContent/LabourTable.jsx
antsgar: Ah you're right, totally got confused there!

packages/webapp/src/components/Finances/Transaction/ExpandedContent/LabourTable.jsx
antsgar: Yup looks good! 🙂

antsgar (approved):

PR #2941, opened 2023-11-01
Lf 3752 implement filtering in dashboard

antsgar:
**Description**

- Implemented Finances dashboard filtering.
- Added new `FilterButton` component that can be used standalone (here) or as part of the search bar + filter.
- Added new `FilterGroup` component to be reused both in this filtering modal and in the `FilterPage` component.

Jira link:

**Type of change**

**How Has This Been Tested?**

- [x] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/webapp/public/locales/en/expense.json
```diff
@@ -36,8 +36,8 @@
     "CUSTOM_DESCRIPTION": "Recurring expenses related to electricity, gas, water (including irrigation), garbage collection, and other periodic services."
   },
   "LABOUR": {
-    "EXPENSE_NAME": "Labour",
-    "CUSTOM_DESCRIPTION": "Expenses related to employed and contracted individuals."
+    "EXPENSE_NAME": "Contract labour",
```
antsgar, line 39: I talked to Loïc and changed this to Contract labour so that we can have two separate filters, one for the regular Labour expenses derived from tasks and one for the one-offs. We were initially thinking to combine them together in the same filter, but since we're not summing them together in the total in the KPI card that seemed more confusing

packages/webapp/src/components/Filter/FilterGroup/index.jsx
```diff
@@ -0,0 +1,101 @@
+/*
```
antsgar, line 1: The contents of this component are derived from part of what used to be in the `FilterPage` component. I abstracted this to a separate one so it can be reused both for `FilterPage` and for my modal/drawer

packages/webapp/src/components/Filter/FilterPillSelect/index.jsx
```diff
@@ -46,50 +64,79 @@ const FilterPillSelect = ({
...
-      </fieldset>
-    </div>
+      {showIndividualControls && (
+        <FilterControls
+          updateFilters={updateAllFilters}
+          selectAllDisabled={selectAllDisabled}
+          clearAllDisabled={clearAllDisabled}
+        />
+      )}
+    </>
```
antsgar, line 139: This is the bulk of the change in this component, added the ability to display individual controls to clear/select all only in this filter

packages/webapp/src/containers/Filter/Transactions/index.jsx
```diff
@@ -0,0 +1,86 @@
+/*
```
antsgar, line 1: This file is one of the most relevant ones in terms of the actual feature implemented in this PR

packages/webapp/src/containers/filterSlice.js
```diff
@@ -28,7 +28,7 @@ const initialDocumentsFilter = {
   TYPE: {},
```
antsgar, line 28: Changes more relevant to this PR in this file too

packages/webapp/src/containers/Finances/useTransactions.js
```diff
@@ -44,7 +44,11 @@ const buildLabourTransactionsFromTasks = ({
   users,
```
antsgar, line 44: This file is important too, I updated the functions so they would work well with the actual filter which is an object rather than an array

packages/webapp/src/containers/Finances/index.jsx
```diff
@@ -38,21 +38,23 @@ import { downloadFinanceReport } from './saga';
 import FinancesCarrousel from '../../components/Finances/FinancesCarrousel';
```
antsgar, line 38: Relevant changes in this file as well, changing how KPIs are calculated and displaying the filter button

packages/webapp/src/containers/Finances/util.js
```diff
@@ -33,33 +33,17 @@ import {
   SALE_VALUE,
```
antsgar, line 33: Changes here for how the KPIs are calculated

packages/webapp/src/containers/Finances/TransactionFilter/index.jsx
```diff
@@ -0,0 +1,55 @@
...
+
+  const filterRef = useRef({});
+
+  return (
+    <>
+      <FilterButton onClick={() => setIsFilterOpen(true)} isFilterActive={isFilterActive} />
+      {/* TODO LF-3751 show contents on drawer instead of modal component for mobile view */}
+      {isFilterOpen && (
+        <ModalComponent
+          title={t('FINANCES.FILTER.TITLE')}
```
SayakaOno, line 36: Should this be black...?

SayakaOno (commented):
It's working well!
Thanks to your suggestion in the tech daily, reviewing was much more efficient and easier! ❤️ 

I left some comments, and I think the modal should be scrollable
Screenshot 2023-11-02 at 9 35 52 AM

packages/webapp/src/components/Finances/Transaction/Mobile/styles.module.scss
```diff
@@ -23,17 +23,17 @@
...
 .expandableItem {
-  padding: 12px 0;
+  padding: 12px 0px 12px;
   align-items: center;
 }
 
 .expandableItemWrapper {
   border: solid 1px transparent;
   border-bottom: solid 1px var(--grey200);
+  margin: 8px 0;
```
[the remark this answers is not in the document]
antsgar: In some of the Figmas there seems to be this spacing but I'm thinking now those might be Figmas just to demonstrate all of the possible "states" of an item and not an actual list. Will remove this!

packages/webapp/src/containers/Finances/useTransactions.js
```diff
@@ -44,7 +44,11 @@ const buildLabourTransactionsFromTasks = ({
...
   dateFilter,
   expenseTypeFilter,
+  expenseTypes,
 }) => {
+  const labourExpenseType = expenseTypes.find(
+    (expense) => expense.expense_translation_key === 'LABOUR',
+  );
```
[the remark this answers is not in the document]
antsgar: Oops yeah this is a leftover of when we thought we were going to filter one off and task labour expenses with the same filter

packages/webapp/src/stories/Filters/FilterButton.stories.jsx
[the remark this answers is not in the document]
antsgar: Ah yeah, good point! I ended up removing it because after lifting up the ownership of filters to the finances container, this ended up just rendering the PureTransactionList with a transactions prop. We can always re-add it when the time comes and we need the desktop/mobile logic

packages/webapp/src/containers/Finances/TransactionFilter/index.jsx
antsgar: Added this change!

antsgar:
@SayakaOno thank you so much for your thorough review! I think I've addressed all of your comments, please take a look when you can 🙂

packages/webapp/src/containers/Finances/TransactionFilter/styles.module.scss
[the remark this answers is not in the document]
SayakaOno: License is still missing!

SayakaOno (commented):
License is missing in one file, but everything else looks good!

packages/webapp/src/components/Filter/FilterButton/index.jsx
```diff
@@ -0,0 +1,27 @@
...
+import PropTypes from 'prop-types';
+import styles from './styles.module.scss';
+import TextButton from '../../Form/Button/TextButton';
+import clsx from 'clsx';
+
+const FilterButton = ({ isFilterActive, onClick }) => {
+  return (
+    <div className={clsx([styles.filterButtonContainer, isFilterActive && styles.active])}>
+      <TextButton className={styles.filterButton} onClick={onClick}>
+        <FiFilter data-cy="tasks-filter" className={styles.filterButtonIcon} />
```
kathyavini, line 11: NIT (just noting): this Cypress selector is for tests that we don't have anymore, and in any case the filter is not specific to tasks so it's incorrectly named

packages/webapp/src/components/Filter/FilterButton/styles.module.scss
```diff
@@ -0,0 +1,50 @@
+.filterButtonContainer {
+  position: relative;
+  height: 100%;
```
kathyavini, line 3: In Figma and in Storybook this results in a square button, but in app the button ends up rectangular in both Finances and the old PureSearchbarAndFilter (I think as height will be more than 40px). I'm okay with either and I'll match the search button, but just checking that it shouldn't stay square.

packages/webapp/src/components/Filter/FilterButton/styles.module.scss
```diff
@@ -0,0 +1,50 @@
...
+  width: 40px;
+  display: flex;
+  justify-content: center;
+  background-color: var(--grey200);
+  border-radius: 4px;
+  border: 1px solid var(--grey400);
```
kathyavini, line 9: Is this a compromise between Figma and the older app styles perhaps? The filter button is borderless in Figma but I see this border looks quite nice next to the old search input! Again just checking if it's intentional, and then I'll match with the collapsible search.

packages/webapp/src/components/Filter/FilterButton/styles.module.scss
```diff
@@ -0,0 +1,50 @@
...
+    background-color: var(--teal100);
+    border: 1px solid var(--teal500);
+
+    .filterButtonIcon {
+      color: var(--teal600);
+    }
+  }
+
+  &:hover {
+    background-color: var(--white, #FFF);
```
kathyavini, line 21: Just wanted to REALLY double-check on this 😂 It doesn't look like any other hover effect in Finances or the older app styles that I know of.

packages/webapp/src/components/Filter/FilterDateRange/index.jsx
```diff
@@ -13,6 +15,7 @@ export function FilterDateRange({
   filterRef,
   shouldReset,
   style,
+  className,
```
kathyavini, line 18: Thank you 😄 

NIT: Here and elsewhere the className should be added to PropTypes as well

packages/webapp/src/containers/Filter/Transactions/index.jsx
```diff
@@ -0,0 +1,86 @@
...
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ */
+
+import PropTypes from 'prop-types';
+import { useTranslation } from 'react-i18next';
+import { useSelector } from 'react-redux';
+import { EXPENSE_TYPE, REVENUE_TYPE } from '../constants';
+import { expenseTypeSelector } from '../../Finances/selectors';
```
kathyavini, line 20: I think we need to include retired types (`allExpenseTypeSelector`, `allRevenueTypeSelector`) as well, based on the fact that the retired modal mentions filtering. But then I wonder if there should be a visual difference between retired and active types on the pills?

Screenshot 2023-11-02 at 6 04 54 PM

packages/webapp/src/containers/Filter/Transactions/index.jsx
```diff
@@ -0,0 +1,86 @@
...
+  const revenueTypes = useSelector(revenueTypesSelector);
+
+  const filters = [
+    {
+      subject: t('FINANCES.FILTER.EXPENSE_TYPE'),
+      filterKey: EXPENSE_TYPE,
+      options: [
+        ...expenseTypes.map((type) => ({
+          value: type.expense_type_id,
+          default: transactionsFilter[EXPENSE_TYPE]?.[type.expense_type_id]?.active ?? true,
```
kathyavini, line 39: In reading the code (!) this was the hardest for me to comprehend because it didn't read logically that a default value would be conditional on the current filter state.

I guess it comes from what I would call `componentInitialState` being named `defaultFilterState` (I know it's not your naming convention!) Only if it was hard for you to catch the first time as well, it might be worth a comment here what `default` indicates.

packages/webapp/src/containers/Finances/TransactionFilter/index.jsx
```diff
@@ -0,0 +1,72 @@
...
+  const [isDirty, setIsDirty] = useState(false);
+  const isFilterActive = useSelector(isFilterCurrentlyActiveSelector('transactions'));
+
+  const transactionsFilter = useSelector(transactionsFilterSelector);
+  const dispatch = useDispatch();
+
+  const handleApply = () => {
+    dispatch(setTransactionsFilter(filterRef.current));
+    setIsFilterOpen(false);
+  };
```
kathyavini, line 42: Could we reset the filter to the initial state when all filters have been selected?

```suggestion
  const handleApply = () => {
    const allAreActive = () =>
      Object.values(filterRef.current).every((filter) =>
        Object.values(filter).every((subFilter) => subFilter.active),
      );

    if (allAreActive()) {
      dispatch(resetTransactionsFilter());
    } else {
      dispatch(setTransactionsFilter(filterRef.current));
    }
    setIsFilterOpen(false);
  };
```

kathyavini (commented):
The code looked good and everything worked so well!

I tried to spend a little bit more time with reading, but in the end I did end up with a primarily UX question. It relates to how the initial state (= no filters applied, filter button inactive) looks like this to the user:

Screenshot 2023-11-02 at 5 31 17 PM

but if the user subsequently puts the filters into that state again, the filter button remains active. In the other app filters, not applying any filtering (= the initial state here) would also make the button inactive. Could we replicate that by resetting the filter state when everything is selected? I left a code suggestion.

Relatedly, I don't love that `isFilterCurrentlyActiveSelector` returns false when both the expense and revenue filters are cleared (and ack, the naming is a bit confusing because clicking both "clear all" links is not equivalent to "clear all filters" in e.g. tasks!), because here in transactions, clearing both sets of filters results in the _opposite_ of the initial state. That one I'm not sure how to fix in code off the top of my head, but it does feel like it doesn't fit the pattern of the existing filters.

Other than that, just had a few inline questions. The nitpicks were just notes 🙂 

EDIT: I forgot to write down until I Sayaka's note above that I also think the modal needs to be scrollable!!

packages/webapp/src/components/Filter/FilterButton/styles.module.scss
antsgar: Ohhh, I didn't see the borderless filter button! Apparently it's borderless in the desktop view, but it does have a border in mobile which is the screen I was looking at

packages/webapp/src/components/Filter/FilterButton/styles.module.scss
antsgar: I kinda prefer it to be square too, I'll update!

packages/webapp/src/components/Filter/FilterButton/styles.module.scss
antsgar: I think it's entirely different for sure! But I followed the buttons spec here https://www.figma.com/file/96NZ02oYe3jpet1roUp0s0/Mockups?type=design&node-id=5385-32459&mode=design&t=b7fbFPa9zO43Y15x-0

packages/webapp/src/containers/Filter/Transactions/index.jsx
antsgar: Thank you for pointing this out! I didn't have any retired types so I didn't notice it

packages/webapp/src/containers/Filter/Transactions/index.jsx
antsgar: Yeah, this was confusing for me at first too!

packages/webapp/src/containers/Finances/TransactionFilter/index.jsx
antsgar: I ended up baking this in in the `isFilterCurrentlyActiveSelector` which should also take care of showing the filter as active now if all filters are cleared 🙂

packages/webapp/src/containers/Filter/Transactions/index.jsx
antsgar: Btw I checked with Loïc and he said we don't need a visual distinction for retired types in the pills

kathyavini (dismissed):
It's perfect!! 😍

SayakaOno:
Is it possible to fix the height?
https://github.com/LiteFarmOrg/LiteFarm/assets/33141219/e447a65d-ad2b-4180-84ed-11ef3034c9ef

packages/webapp/src/containers/Filter/Transactions/index.jsx
kathyavini: This is a reach goal -- if it gets implemented I think a new ticket is appropriate -- but do you think someday we should check for whether a retired type is associated with any transactions before showing it in the filters?

antsgar:
@SayakaOno good point! It's possible to fix the height, but I'm not sure we'd want to do that because we could end up with a lot of whitespace on bigger screens both when the expandable filters are collapsed or if the types list is relatively short. If the content was fixed it'd be a bit easier, but since the list is dynamic and will vary per user I think there's no way to tell how much whitespace we could end up with. Ideally we'd fix the height dynamically depending on the height of the fully expanded content, but I think that would take a fair amount of CSS tinkering to do

kathyavini (approved):

SayakaOno:
@antsgar
I was wondering if it was possible to not extend the height when the selector is opened because what I was looking at wasn't there and I got confused for a little bit. I agree to work on enhancement later on!

SayakaOno (approved):

PR #2945, opened 2023-11-03
LF-3750 Add search to dashboard

kathyavini:
**Description**

This PR creates the new `<PureCollapsibleSearchbar />` component ([Storybook](http://localhost:6006/?path=/docs/components-purecollapsiblesearch--docs)) and adds it to the Finances dashboard.

Note:
- The button background is correct in app but often -- not always! -- doesn't show up in Storybook (it looks like the `<TextButton />` native transparent takes precedent). If you know why that is, could you let me know 🙏 

Jira link: https://lite-farm.atlassian.net/browse/LF-3750

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/webapp/src/containers/Finances/styles.module.scss
```diff
@@ -51,8 +51,27 @@
...
   margin-bottom: 8px;
 }
+
+.filterBarButtons {
+  display: flex;
+  gap: 8px;
+}
+
+.dateRangeSelector {
+  max-width: 360px;
```
kathyavini, line 68: This is actually larger than what is shown in Figma, but I was worried about the text not fitting.

packages/webapp/src/components/PopupFilter/PureCollapsibleSearch/index.jsx
```diff
@@ -0,0 +1,136 @@
...
+              value={value}
+              onChange={onChange}
+              placeholder={placeholderText}
+            />
+          </form>
+        </Modal>
+      )}
+
+      <TextButton className={clsx(styles.searchButton, isSearchActive && styles.active)}>
+        <div className={styles.circleContainer}>
```
antsgar, line 111: @SayakaOno made me aware of this the other day, we shouldn't place a div inside a TextButton because it's a <button> tag, which doesn't allow nesting block elements inside it. Since the circle is positioned absolutely anyway, you can just remove it from the TextButton and place it outside. You'll need a container for both with `position: relative` for it to work properly.

packages/webapp/src/components/PopupFilter/PureCollapsibleSearch/index.jsx
```diff
@@ -0,0 +1,136 @@
...
+          <div className={styles.circle} />
+        </div>
+        <SearchIcon className={styles.searchIcon} onClick={onSearchOpen} />
```
antsgar, line 114: Here the `onClick` should actually be on the button instead of the icon -- currently if you try to click on the little edges of the button right outside of the icon, even though it seems clickable it won't open the search.

packages/webapp/src/components/PopupFilter/PureCollapsibleSearch/index.jsx
```diff
@@ -0,0 +1,136 @@
...
+    let resizeObserver;
+
+    const updateModalStyle = (ref) => {
+      if (ref) {
+        const rect = ref.getBoundingClientRect();
+
+        setModalStyle({
+          position: 'absolute',
+          top: `${rect.top}px`,
+          left: ref === searchRef.current ? undefined : `${rect.left}px`,
```
antsgar, line 56: I don't understand the bit where it checks if the ref is actually the searchRef and sets different values in that case, can you explain that case a bit more? 🙏

antsgar (commented):
Working great, it's beautiful! ❤️ I left a couple of small comments

packages/webapp/src/components/PopupFilter/PureCollapsibleSearch/index.jsx
kathyavini: THANK YOU!!! This is why the onClick was mysteriously failing at times 😁

packages/webapp/src/components/PopupFilter/PureCollapsibleSearch/index.jsx
kathyavini: If a `containerRef` isn't provided from the parent component, this sets the position of the modal searchbar directly on top of the search button -- so it grabs a vertical position, but does not grab a horizontal position or width (as the horizontal position and width of the button itself is irrelevant). Not supplying a `left` keeps the modal default of centering horizontally, and the searchbar width will always be `95vw`.

I think overlaying the button is a pretty good default (so that a container ref doesn't always have to be created in the parent), but if it makes the code too unreadable, I can remove it. Alternatively... if you have a suggestion for writing the same in a more readable style, that would be great!

antsgar: That makes sense to me! I'd keep it and if possible just add a small comment indicating if a container ref isn't passed in the position will be calculated to overlay the button 🙂

antsgar (dismissed):

packages/webapp/src/components/Finances/DateRangeSelector/index.jsx
```diff
@@ -78,4 +80,8 @@ const FinanceDateRangeSelector = () => {
   );
 };
 
+FinanceDateRangeSelector.PropTypes = {
```
SayakaOno, line 83: There is a warning about the "P" in "PropTypes" being capitalized!

kathyavini: Thank you!

SayakaOno (approved):

antsgar (approved):

PR #2948, opened 2023-11-06
LF-3545, LF-3634 Viewing the task list after another user has created new crop plans causes whitescreen type error crash

kathyavini:
**Description**

This undoes several code changes that were implemented to allow tasks to handle hard-deleted (in the Redux store) management plans, which nonetheless created a chain effect of type error bugs.

As the management plan entities are too intertwined with the task entities (which have a very particular order of loading as well), I moved management plan deletion to to soft delete _on the frontend_, filtered in the main selector (but not the entities selector), and reverted the changes to task.

Jira link: https://lite-farm.atlassian.net/browse/LF-3545

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

I tested this with two users logged onto the same farm, looking at the task list after the other user changed (added/deleted) management plans.

**Checklist:**


packages/webapp/src/containers/Task/index.jsx
```diff
@@ -27,7 +27,7 @@ import {
   userFarmSelector,
 } from '../userFarmSlice';
 import { resetAndUnLockFormData } from '../hooks/useHookFormPersist/hookFormPersistSlice';
-import { getCropsAndManagementPlans } from '../saga';
+import { getManagementPlansAndTasks } from '../saga';
```
kathyavini, line 30: The changes in this file are directly undoing the changes from this PR:

- https://github.com/LiteFarmOrg/LiteFarm/pull/2829

packages/api/src/controllers/taskController.js
```diff
@@ -679,7 +679,7 @@ const taskController = {
         .whereNotDeleted()
         .withGraphFetched(
           `[locations.[location_defaults], managementPlans, soil_amendment_task, field_work_task.[field_work_task_type], cleaning_task, pest_control_task, 
-            harvest_task.[harvest_use], plant_task, transplant_task, irrigation_task.[irrigation_type]]
+            harvest_task.[harvest_use], plant_task.[planting_management_plan.[bed_method, container_method, broadcast_method, row_method]], transplant_task, irrigation_task.[irrigation_type]]
```
kathyavini, line 682: This change doesn't actually impact on this specific bug.

However adding this part of the graph was based on `getPlantingTasksAndPlantingManagementPlansSuccessSaga` being set up to get and then forward the `planting_management_property` (which had previously always just not existed). There may be another task + plans bug that is therefore addressed by this; at the least it shouldn't hurt as it's not difficult to fetch this extra part of the graph.

packages/webapp/src/containers/managementPlanSlice.js
```diff
@@ -61,6 +62,22 @@ const addManyManagementPlan = (state, { payload: managementPlans }) => {
   );
 };
 
+const deleteOneManagementPlan = (state, { payload }) => {
```
JimmyRowland, line 65: {payload : management_plan_id}

packages/webapp/src/containers/Task/index.jsx
```diff
@@ -99,13 +99,10 @@ export default function TaskPage({ history }) {
   }, [activeUsers.length, taskTypes.length]);
 
   useEffect(() => {
-    dispatch(getTasks());
     dispatch(getTaskTypes());
```
JimmyRowland, line 102: getManagementPlansAndTasks calls getTaskTypes.
getTaskTypes can be removed

packages/webapp/src/containers/managementPlanSlice.js
```diff
@@ -61,6 +62,22 @@ const addManyManagementPlan = (state, { payload: managementPlans }) => {
...
+  const managementPlan = state.entities[payload];
+  if (managementPlan) {
+    managementPlan.deleted = true;
+  }
+};
+
+const deleteManyManagementPlans = (state, { payload }) => {
+  payload.forEach((id) => {
```
JimmyRowland, line 73: {payload : managementPlanIds}

JimmyRowland (approved):

JimmyRowland:
I think there should be a follow up PR
In the management plan selector, deleted ones can be filtered out.
image

In tasksSelector, complemented/abandoned tasks can be filtered out if all locations and management plans are deleted.

After, a proper fix can be implemented. Whether it's WebSocket or cache invalidation on each fetch

kathyavini:
Thank you for the suggestions @JimmyRowland!! For some reason putting the deleted check in `tasksSelector` didn't work -- it looked like maybe the deleted management plan wasn't in the task object at that point? -- and I gotta admit my comprehension of `taskSlice` is not 100% so I couldn't get to the bottom of that.

That said, filtering on the task cards (in `taskCardsContentSelector`) works and seems to hit all the UI places (including locations) where the inconsistency could be spotted so I'm hoping that is enough! 🤞

JimmyRowland:
does this work? `export const tasksSelector = createSelector(
  [taskEntitiesSelector, loginSelector],
  (taskEntities, { farm_id }) => {
    return Object.values(taskEntities).filter(({ locations, managementPlans, taskType }) => {
      if (
        (locations.length || managementPlans.length) &&
        [locations, managementPlans].reduce(
          (allDeleted, array) => allDeleted && array.every(({ deleted }) => deleted),
          true,
        )
      ) {
        return false;
      }
      for (const location of locations) {
        if (location.farm_id === farm_id) {
          return true;
        }
      }
      for (const managementPlan of managementPlans) {
        if (managementPlan.farm_id === farm_id) {
          return true;
        }
      }
      if (taskType.farm_id === farm_id) {
        return true;
      }
      return false;
    });
  },
);`

JimmyRowland (dismissed):

packages/webapp/src/containers/Task/index.jsx
```diff
@@ -99,13 +98,9 @@ export default function TaskPage({ history }) {
   }, [activeUsers.length, taskTypes.length]);
 
   useEffect(() => {
-    dispatch(getTasks());
-    dispatch(getTaskTypes());
-    dispatch(getProducts());
-    dispatch(getHarvestUseTypes());
     dispatch(getAllUserFarmsByFarmId());
+    dispatch(getManagementPlansAndTasks());
```
antsgar, line 102: I've tested it with this PR, and this is the change that fixes my current ticket -- with the previous code we were fetching the tasks first and the management plans at the end, here we're fetching all of the dependencies first and then the tasks

antsgar (approved):

kathyavini:
> I am not sure about the requirements. It's probably better to move filter in tasksSelector
> 
> does this work?

It works, the `every` was the key!! 👌 I think we do need to filter on every task type that belongs to only a deleted management plan but works well that way too. Yes definitely better in this selector @JimmyRowland thank you!!! 🙏

JimmyRowland (approved):

PR #2951, opened 2023-11-07
LF-3788 The user is not able to type out a product name during the creation of the soil amendment task or clean task

kathyavini:
**Description**

This is a fun bug!!

This [commit](https://github.com/LiteFarmOrg/LiteFarm/pull/2913/files#diff-9b510ffac5e2f98e7a188cd29fe6409be58bfd51d9a33be2b4cec8aade222fb5) by @SayakaOno fixed the logic assigning `isSearchable` in the `<ReactSelect />`.

Currently we have only one "creatable" React Select in App (for Product). Here the `isSearchable` prop was never passed, but the faulty original logic:

`isSearchable={options?.length > 8 || isSearchable}`

meant that if there were fewer than 8 options, `isSearchable={undefined}` was passed, and the `<CreateableSelect />` took its default value of `isSearchable={true}`

Now that the logic has been fixed, `isSearchable` was false when the number of options was less than 8.

However a `<CreateableSelect />` needs to have `isSearchable={true}` in order to accept input (and it should not have anything to do with the number of options), so I am assigning `isSearchable={true}` and leaving a note about what the prop does.

Jira link: https://lite-farm.atlassian.net/browse/LF-3788

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/Form/ReactSelect/index.jsx
+++ b/packages/webapp/src/components/Form/ReactSelect/index.jsx
@@ -219,7 +219,7 @@ const ReactSelect = React.forwardRef(function ReactSelect(
             ),
             ...components,
           }}
-          isSearchable={isSearchable ?? options?.length > 8}
+          isSearchable={true} // CreatableSelects must be searchable to accept input
           ref={ref}
           defaultValue={defaultValue}
           isDisabled={isDisabled}


SayakaOno (approved):
I'm glad Denis caught this...!
I think I missed checking `CreatableSelect`. Thank you for fixing!

PR #2957, opened 2023-11-09
 LF-3758 When the user creates an expense, the date will not be displayed correctly in Expense detailed view

javiscript13:
**Description**

The bug is that when viewing an expense that was created with timezone UTC+X, where X>1, it shows the date of the day before.  

This was caused by the use of a function that truncated the timezone of the data retrieved from the server, so to fix it, I just used a function that converted to the user timezone considering the TZ of the data passed by the backend.

Jira link: https://lite-farm.atlassian.net/browse/LF-3758

**Type of change**

**How Has This Been Tested?**

The change was tested by changing to different timezones that are on the day after the one that my database is storing the data, and creating expenses on these timezones to check if it would take the right time.  (My db stores on UTC-6 and now is 22h40m, so I change to UTC+3, and UTC+7, on integration branch I can replicate the error and when I change to this branch it works ok)

- [x] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/Finances/PureExpenseDetail/index.jsx
+++ b/packages/webapp/src/components/Finances/PureExpenseDetail/index.jsx
@@ -27,7 +27,7 @@ import { ReactComponent as TrashIcon } from '../../../assets/images/document/tra
 import { hookFormMaxCharsValidation } from '../../Form/hookformValidationUtils';
 import { useCurrencySymbol } from '../../../containers/hooks/useCurrencySymbol';
 import ReactSelect from '../../Form/ReactSelect';
-import { getDateFromDateTimeString } from '../../../util/moment';
+import moment from 'moment';
 import { NOTE, VALUE, DATE, TYPE } from '../AddExpense/constants';
 
 const PureExpenseDetail = ({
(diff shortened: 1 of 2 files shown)


SayakaOno (commented):
@javiscript13 it works great!
I don't want us to keep `getDateFromDateTimeString` function, I guess it could cause the same bug, right...? Would you be able to modify the function rather than fixing the code directly? `getDateFromDateTimeString` is not used anywhere, so it should be a safe change.

javiscript13:
> @javiscript13 it works great! Thank you for fixing! I don't want us to keep `getDateFromDateTimeString` function, I guess it could cause the same bug, right...? Would you be able to modify the function rather than fixing the code directly? `getDateFromDateTimeString` is not used anywhere, so it should be a safe change.

Hey @SayakaOno , thanks for reviewing!   So, I deleted the `getDateFromDateTimeString` function, and used another one existing on the LF moment.js file,  the library can receive string or date object as parameter so I used the existing "getDateInputFormat"  ... would that be ok, or should I use the existing `getDateFromDateTimeString` ??

SayakaOno:
@javiscript13 that's perfect, thank you!

SayakaOno (approved):

PR #2958, opened 2023-11-09
LF-3755 Style xlsx finance report with ExcelJS API

kathyavini:
**Description**

This takes the transaction list and filter configuration created in

- https://github.com/LiteFarmOrg/LiteFarm/pull/2942

and generates a two-sheet workbook with (1) the list of transactions and the (2) selected filters. The filter configuration is also stored in the `finance_report` table in the `filter_config` column.

Additional utility functions were added to the frontend for generating the translated strings used in the spreadsheet column headers and worksheet names.

NOTE: There is a pre-existing bug with the data the frontend passes where if the report is generated without interacting with the filters in the dashboard or the report modal, the filter configuration will be empty. This will not affect the transactions but there will be no listed expense/revenue types in `filter_config` or the "Export Settings" worksheet.

Jira link: https://lite-farm.atlassian.net/browse/LF-3755

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/webapp/src/containers/Finances/index.jsx
```diff
@@ -98,7 +98,10 @@ const Finances = ({ history }) => {
   };
 
   const makeTransactionsSearchableString = (transaction) =>
-    [transaction.note, transaction.typeLabel || t('SALE.FINANCES.LABOUR_EXPENSE')]
+    [
```
kathyavini, line 101: This fixes an error from https://github.com/LiteFarmOrg/LiteFarm/pull/2945 where the translation strings were mangled and confused 🙂 Running the `i18n` script uncovered this error.

packages/webapp/public/locales/en/common.json
```diff
@@ -63,6 +63,7 @@
   "SUBMITTING": "Submitting...",
   "THATS_FINE": "That’s fine",
   "TODAY": "Today",
+  "TYPE": "Type",
```
kathyavini, line 66: I didn't include the refactor to this common translation (it was too big to include) but it should be undertaken; there are 14 occurrences in `translation.json`

antsgar (dismissed):
Works great! 🎉 the only minor thing I noticed is that I don't see a border in the amount cells that have a background color, but that might be because I'm using Numbers. Also the negative amounts show up with a double sign `--` for me, but might be Numbers too!

Screenshot 2023-11-09 at 11 39 54

kathyavini:
Oh dear, it does look a mess in Numbers! The double negative makes it look like Numbers is having an issue with the Excel negative sign, which is really weird/unfortunate 😆 I'll see if something can be done!

> Works great! 🎉 the only minor thing I noticed is that I don't see a border in the amount cells that have a background color, but that might be because I'm using Numbers. Also the negative amounts show up with a double sign `--` for me, but might be Numbers too!

antsgar:
@kathyavini the new report is looking great! I'm good with keeping it as is for Numbers, it's not unreadable or anything, just a little less aesthetically pleasing. I tried sorting and it still works well regardless of the double --.

Did you end up filing a bug ticket for the `undefined` filters sent by the Report screen?

antsgar (dismissed):

kathyavini:
Thank you @antsgar!

> Did you end up filing a bug ticket for the `undefined` filters sent by the Report screen?

I did not! I had been thinking of fixing it today but I think it's going to be a bit of work compared to the payoff (since the transaction list itself is unaffected). I will make a fresh ticket now.

antsgar (approved):

PR #2962, opened 2023-11-09
LF-3799-Transactions-added-with-today-s-date-are-shown-as-Yesterday

javiscript13:
**Description**

Issue is that on finance view, today's sales appeared on "Yesterday" section, this was because the POST sent was only sending the date without TZ and the backend was making a mess when interpreting the time.

So in this PR I changed the sales POST to have the TZ (and the backend to accept the timezoned date).

Jira link: https://lite-farm.atlassian.net/browse/LF-3799

**Type of change**

**How Has This Been Tested?**

I have test to create sales on a custom revenew type and on the default one, configuring my computer to have a TZ where the day is the same as my backend configuration, and a TZ where it is the next day, and as the TZ is now being taken into account, it seems to work ok.

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/api/src/server.js
+++ b/packages/api/src/server.js
@@ -198,6 +198,7 @@ app.set('json replacer', (key, value) => {
     'transition_date',
     'transplant_date',
     'valid_until',
+    'sale_date',
   ];
 
   if (value && pgDateTypeFields.includes(key)) {


SayakaOno:
@javiscript13 Wow, this is a great find...!
So basically the existing sales' `sale_date` in the DB are all wrong? If we want to fix the date, we need to convert it to the one using user's timezone? (I guess timezone we can detect wouldn't be 100% accurate though)

javiscript13:
> @javiscript13 Wow, this is a great find...! So basically the existing sales' `sale_date` in the DB are all wrong? The existing data is user's local date (string), and if we want to correct it, we need to convert it to UTC using user's timezone? (I guess timezone we can detect wouldn't be 100% accurate though)

Thanks @SayakaOno for reviewing!,   What I can see in my database is the following: on integration at it is right now, if I make a sale it would be sent from the FE POST as "2023-11-09", and as my backend+db is configured to my local time (UTC-6), it would be stored as "2023-11-08 18:00:00.000 -0600" since it is interpreting it as being on UTC.  So, I guess that all entries on `sale_date` from users that are not in UTC are wrong.    The PR as it is right now fixes future insertions providing the timezone (it will get stored as "2023-11-09 00:00:00.000 -0600"),  should something else be done on existing data before merging this?, If so, could yo provide some guidance?

SayakaOno:
Thank you @javiscript13 !
I think Kevin will make a decision on this. It would be great if you could add some information to the ticket, like what the problem is (just copy and paste would be good), possible fix on the existing data and estimate if you can. If you can make it to the tech daily tomorrow, we can discuss the solution, otherwise, I will bring it up :)

javiscript13:
@SayakaOno , so I investigated a little more and, the function `mapRevenueFormDataToApiCallFormat` has been modified or created 3 weeks ago, so, not sure if before those 3 weeks the date was sent in any other way. 

The backend saleModel didn't expect the date to have date-time format since 3 years ago, BUT, I tried to keep the backend as it is RN on integration and changing only how the FE sends the data, and if FE sends timezone it stores the values in the right way.  So, this means that if on the published version the date is sent with TZ, the data should be ok.

SayakaOno:
@javiscript13 I agree that your fix is good! I guess we need to check if wrong data exists in the production DB! (but I'm assuming `mapRevenueFormDataToApiCallFormat` was created by modifying the old code...)

javiscript13:
> @javiscript13 I agree that your fix is good! I guess we need to check if wrong data exists in the production DB! (but I'm assuming `mapRevenueFormDataToApiCallFormat` was created by modifying the old code...)

Just to mantain conversation updated, I reviewed the main branch on my local and it does have the same issue, it just sends the date without timezone and when it reaches database it was interpreted as being on UTC.

kathyavini:
> What I can see in my database is the following: on integration at it is right now, if I make a sale it would be sent from the FE POST as "2023-11-09", and as my backend+db is configured to my local time (UTC-6), it would be stored as "2023-11-08 18:00:00.000 -0600" since it is interpreting it as being on UTC

@javiscript13 just to confirm, you saw `sale_date` values like this in your database _before_ using this PR branch? Both my local and on beta I only see `sale_date` entries at UTC midnight: `2023-11-09 00:00:00+00`, never with that kind of timezone-based offset... those I only saw on `farmExpense`.

> as my backend+db is configured to my local time (UTC-6)

Did you specifically configure your backend to be in local time? The default `.env` file includes `TZ=UTC` so mine is always on UTC. I'm 100% sure the beta db is on UTC too (just checked with `SHOW timezone`), and the server... I dunno, 90% sure? 😆 The Docker container is on UTC at least.

javiscript13:
> %

thank you for reviewing @kathyavini , great questions!!  My .env on api folder has TZ=UTC, but if I run SHOW timezone on the database it has "America/Guatemala", so my local timezone (no idea how I got to that configuration tbh).

So I wrote this before reading your message on Jira, I will leave it as explanation, BUT, I do agree that what you proposed is a lot better solution, and will avoid us the problem with the dates as they are on the db.

~~So, at this moment I understand it this way:  even if the DB is on UTC the data is not stored correctly, since the browser is sending the date on localtime but the backend thinks it is on UTC, so if you are on UTC-6 it is `2023-11-09 19h00m` and you add a sale, the FE will send it as `2023-11-09` (which is "today" since we haven't pass midnight in UTC-6), the server being at `2023-11-10 01h00m`, will understand that the date value is `2023-11-09 00h00m` when the right value should be: `2023-11-09 06h00m`.  So, when converting back to UTC-6 the date will be: `2023-11-08 18h00m`.~~

kathyavini:
@javiscript13 I don't know how my postgres db got its timezone configured either 😂 But interesting to know that the db time being different is enough for those wonky timestamps (regardless of server time).

I think how you fixed `farmExpense` was perfect and should be our goal moving forward when we don't have legacy data to deal with like this. I actually wouldn't have guessed that (I would have reached for plain `dates` instead of timezoned `datetimes` before reading up on that other Jira thread!) but good to know.

javiscript13:
I have changed the commit to what @kathyavini very wisely suggested.  @SayakaOno , @antsgar the re-review button is not enabled, but I think this PR is ready to be reviewed again.

Thank you all!!

kathyavini (approved):

PR #2964, opened 2023-11-10
LF-3800 Fixing White screen when user tries to add custom revenue.

kikemarquez8:
What I think its going on is the persistedFormData is not being saved fast enough for the AddSale component to retrieve it on the first render and since it's trying to destructure an object on that render it fails. As a temporary fix, just using the object instead of destructuring works.

 Also removing <= validation on finances that should be "<"

Jira link:

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/containers/Finances/AddSale/index.jsx
+++ b/packages/webapp/src/containers/Finances/AddSale/index.jsx
@@ -39,7 +39,7 @@ function AddSale() {
   const revenueTypeReactSelectOptions = mapRevenueTypesToReactSelectOptions(revenueTypes);
   const translatedRevenueName = revenueType?.farm_id
     ? revenueType?.revenue_name
-    : t(`revenue:${revenueType?.translation_key}.REVENUE_NAME`);
+    : t(`revenue:${revenueType?.revenue_translation_key}.REVENUE_NAME`);
 
   const onSubmit = (data) => {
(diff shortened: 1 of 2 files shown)


packages/webapp/src/containers/Finances/AddSale/index.jsx
```diff
@@ -37,10 +37,9 @@ function AddSale() {
   const revenueType = useSelector(revenueTypeByIdSelector(revenue_type_id));
   const revenueTypes = useSortedRevenueTypes();
   const revenueTypeReactSelectOptions = mapRevenueTypesToReactSelectOptions(revenueTypes);
-  const { revenue_name, revenue_translation_key, farm_id } = revenueType;
-  const translatedRevenueName = farm_id
-    ? revenue_name
-    : t(`revenue:${revenue_translation_key}.REVENUE_NAME`);
+  const translatedRevenueName = revenueType?.farm_id
```
kathyavini, line 40: This is definitely the code that's crashing, but unfortunately with just the optional chaining check I'm seeing this then on the Add Sale title:

Screenshot 2023-11-10 at 2 46 44 PM

kikemarquez8: You are right! Just fixed this

kathyavini (approved):
Thank you Kike!!! Sorry for completely missing that it was the key and not the entity that was undefined!

Works great! 👌

PR #2970, opened 2023-11-10
LF-3666 An error message will be displayed failed to update revenue type once the user will try to rename the custom revenue type

kathyavini:
**Description**

This will (most likely) be the first of two PRs for handling retired finance types differently. This PR:

- Provides a frontend validation message (which will be triggered by the case described in the linked Jira ticket) when the user tries to re-use the name of a retired type
- Removes the "under-the-hood" un-retire we had been doing in the Add controllers, in which a retired type is reactivated by creating a new type using the same name. Instead, trying to re-use the same name as a retired type will now throw a conflict error

The subsequent PR (unless it grows too complicated for the current scope) will add a distinction between retired (associated with previous records, filterable, namespace reserved) and deleted (associated with no records, invisible to the user, not namespace reserved) on the database level, and move the frontend retired functionality to the new `retired` property

Jira link: https://lite-farm.atlassian.net/browse/LF-3666

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/webapp/src/components/Form/hookformValidationUtils.js
```diff
@@ -54,3 +54,36 @@ export const hookFormUniquePropertyValidation = (objArr, property, message) => {
...
+ * @param {string} messageFalse - The error message to return if the status property is false or the object is not found.
+ * @returns {(value: any) => string|boolean} A validation function that takes a value to validate and returns
+ * either an error message or `true` if the value is unique
+ */
+export const hookFormUniquePropertyWithStatusValidation = ({
+  objArr,
+  property,
+  status = 'deleted',
+  messageTrue,
+  messageFalse,
```
antsgar, line 75: Since these two are returned depending on the status value, but they both are returned when the item is found, I'd rename them to something a bit more specific like `trueStatusMessage` or similar

antsgar (approved):
Works great! 🎉

PR #2974, opened 2023-11-11
LF-3608: Although the success pop up message is displayed, the harvest task won't be created, it actually creates an empty harvest task.

SayakaOno:
**Description**

Screenshot 2023-11-10 at 3 36 55 PM

In this page above, the user is not allowed to "continue" to the next page without selecting a plan. The bug is that a plan that is not available as an option on this page is "selected" under the hood. This PR adds a check if the plan is available as an option before "selecting" it.

Details:
When creating a harvest task from a crop plan, the crop plan you selected is selected as a "plan that will be affected" by default. If you change the location to create a harvest task for another plan, the crop plan you selected should be deselected because the plan is not for the location you selected. (you will not see the plan as an option, but it remains "selected")

I added a check if the crop plan selected in the previous page is available for the location before setting it as "selected plan". 
(Also, console.log on the same page is removed)

Jira link: https://lite-farm.atlassian.net/browse/LF-3608

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/Task/PureTaskCrops/index.jsx
+++ b/packages/webapp/src/components/Task/PureTaskCrops/index.jsx
@@ -215,7 +215,9 @@ const PureTaskCrops = ({
   const disabled = isRequired && !selectedManagementPlanIds?.length;
 
   useEffect(() => {
-    defaultManagementPlanId && setSelectedManagementPlanIds([parseInt(defaultManagementPlanId)]);
+    if (defaultManagementPlanId && managementPlanIds.includes(+defaultManagementPlanId)) {
+      setSelectedManagementPlanIds([parseInt(defaultManagementPlanId)]);
+    }
   }, []);
 
   return (
--- a/packages/webapp/src/containers/Task/TaskDate/index.jsx
+++ b/packages/webapp/src/containers/Task/TaskDate/index.jsx
@@ -51,21 +51,13 @@ function TaskDate({ history, match, location }) {
         },
       );
       if (transplantTask) {
-        console.log('transplantTask');
         location.state.location = transplantTask.locations[0];
-        console.log(location.state);
       } else if (plantingTask) {
-        console.log('plantingTask');
         location.state.location = plantingTask.locations[0];
-        console.log(location.state);
       } else if (mostRecentlyCompleted) {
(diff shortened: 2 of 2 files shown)


packages/webapp/src/components/Task/PureTaskCrops/index.jsx
```diff
@@ -215,7 +215,9 @@ const PureTaskCrops = ({
   const disabled = isRequired && !selectedManagementPlanIds?.length;
 
   useEffect(() => {
-    defaultManagementPlanId && setSelectedManagementPlanIds([parseInt(defaultManagementPlanId)]);
+    if (defaultManagementPlanId && managementPlanIds.includes(+defaultManagementPlanId)) {
```
antsgar, line 218: Is the + a typo?

packages/webapp/src/containers/Task/TaskDate/index.jsx
```diff
@@ -51,21 +51,13 @@ function TaskDate({ history, match, location }) {
...
         location.state.location = plantingTask.locations[0];
-        console.log(location.state);
       } else if (mostRecentlyCompleted) {
-        console.log('mostRecentlyCompleted');
         location.state.location = mostRecentlyCompleted.locations[0];
-        console.log(location.state);
       } else if (nextUpcoming) {
-        console.log('nextUpcoming');
         location.state.location = nextUpcoming.locations[0];
-        console.log(location.state);
```
antsgar, line 68: Logs need to be removed 🙂

packages/webapp/src/components/Task/PureTaskCrops/index.jsx
SayakaOno: No, `managementPlanIds` was an array of numbers and `defaultManagementPlanId` was string, so I converted it to number :)

packages/webapp/src/containers/Task/TaskDate/index.jsx
SayakaOno: I removed them...?

antsgar (approved):

PR #2982, opened 2023-11-14
Lf 2200 allow the user to upload a custom image for the lf homepage background

navDhammu:
**Description**

In the 'Farm info' form, an option to upload an image is added as well as to change or delete existing image. On successful form submission, image also appears on home page. 

Jira link: https://lite-farm.atlassian.net/browse/LF-2200

**Approach:**

Initially, I attempted to follow a similar approach used for crop and document uploads (seperate from rest of the form, as discussed on slack) but ran into some issues with deleting the image that way. So I went with another approach where the image is sent along with rest of the form fields. Added benefit of this is that it reduces the amount of unnecessary images in storage in case users upload but don't submit. 

**Overview of changes:**
- Changed request format for the putFarm request from application/json to multipart/form-data to allow the image file to be sent. 
- Created a FarmImagePicker component that handles image selection with preview as well as drag and drop.
- Added middleware to handle image related actions that is responsible for creating thumbnail images, uploading both thumbnail and original images to s3, deleting existing images from s3 and passing control to the updateFarm controller.
- Tests: I'm still new to testing so wasn't sure how to add tests for image uploading but I considered using chatHttp's attach method to attach files as part of the tests, however this meant I would need to add some dummy images to the tests directory and I'm not sure if thats the correct approach.

For local testing, imaginary and minio services should be running and minio bucket should be set to public. In beta and production image requests will be sent to a cloudflare worker hosted at 'images.litefarm.workers.dev' that is responsible for authenticating the request and retrieving images from the private bucket. 

**Additional improvements to consider:**
- Currently the file size isn't being restricted on either frontend or backend and it might be good to do so.
- We are using v2 of the aws-sdk but there is a newer version (v3) available.
- When clicking 'save', the button should probably show some type of loading state as there is a slight delay. Also, after submission the button should go back to original disabled state.
- When selecting images, it might be good to provide users with a reset option to discard any changes they made. For example, as a user if I already have a farm image and I accidentally click 'remove image' I should be able to undo it without having to refresh the page.

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/webapp/src/components/Form/FilePickerWrapper/index.tsx
navDhammu, line 1: Had to change this to typescript to avoid errors when using inside FarmImagePicker

packages/api/src/middleware/farm.js
```diff
@@ -0,0 +1,143 @@
...
+  await Promise.all([deleteImagePromsie, deleteThumbnailPromise]);
+}
+
+export async function getExistingImageKeys(farmId) {
+  const {
+    farm_image_url: imageUrl,
+    farm_image_thumbnail_url: thumbnailUrl,
+  } = await FarmModel.query().findById(farmId);
+
+  if (!imageUrl && !thumbnailUrl) return;
```
antsgar, line 130: This should probably be

```suggestion
  if (!imageUrl || !thumbnailUrl) return;
```

If either url for some weird reason is not defined, it will produce an error in the function below when attempting to parse that string

antsgar (approved):
This is working great and it's so nice to see live! Thank you so much for this awesome contribution. I left a small comment, but other than that it looks good to go 🙂

navDhammu:
> This is working great and it's so nice to see live! Thank you so much for this awesome contribution. I left a small comment, but other than that it looks good to go 🙂

Good catch, I updated it and thanks for reviewing!

PR #2983, opened 2023-11-14
LF-3814 The white screen is displayed once the user Updates the user info without actually selected the preferred LiteFarm Language

kathyavini:
**Description**

This one was a bit weird! The symptom is that the `language_preference` property is missing from the `userFarm` Redux entity when the `userFarmSelector` is called from `<Account />`, but ONLY if you go there directly after creating a new farm, rather than after logging in. In the Jira ticket there is one "symptom-level" fix for this issue.

The cause is that this GET request in `postFarmSaga` doesn't return the `language_preference` property:

``` 
call(axios.get, userUrl + '/' + user_id, header),`
```

This is the `getUserByID` controller. What I consider a bit weird is that it's actually not called anywhere else in app except within this Add Farm saga / flow (or at least, that's how it appears after a frontend code search) 🤷 

Jira link: https://lite-farm.atlassian.net/browse/LF-3814

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

I tested the flow by logging (not very high-tech) through the Add Farm > `postFarm` >  `postFarmSuccess` > Account. I found the lacking `language_preference` property to occur on both new and existing accounts after Add Farm. After login these sagas are not used and the `user` object is populated elsewhere.

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/userController.js
+++ b/packages/api/src/controllers/userController.js
@@ -321,6 +321,7 @@ const userController = {
           'user_id',
           'gender',
           'birth_year',
+          'language_preference',
         );
 
       if (!data) {


SayakaOno (approved):
Works perfect!

PR #2987, opened 2023-11-15
Lf 3215 a user creating a task at a location concurrently with another user retiring said location creates a task at a location that does not exist

gursimran-singh:
**Description**

When two users are logged on to the same farm and one user begins the task creation flow at a particular location while another user attempts to retire the same location, the tasks are created at a location that does not exist on the farm

Jira link:https://lite-farm.atlassian.net/browse/LF-3215

**Type of change**

**How Has This Been Tested?**

- Start to create a new task
- Before finishing the task creation flow, delete the location by setting the "deleted = true" in database to mimic the behaviour that another user has deleted the location at the same time.
- Try to save the task. Error will occur and a modal will be displayed giving info of the error.

- [x] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/taskController.js
+++ b/packages/api/src/controllers/taskController.js
@@ -13,6 +13,7 @@
  *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
  */
 
+import { transaction, Model } from 'objection';
 import TaskModel from '../models/taskModel.js';
 
 import UserFarmModel from '../models/userFarmModel.js';
@@ -28,7 +29,9 @@ import { typesOfTask } from './../middleware/validation/task.js';
 import IrrigationTypesModel from '../models/irrigationTypesModel.js';
 import FieldWorkTypeModel from '../models/fieldWorkTypeModel.js';
 import locationDefaultsModel from '../models/locationDefaultsModel.js';
+import Location from '../models/locationModel.js';
 import TaskTypeModel from '../models/taskTypeModel.js';
+import baseController from './baseController.js';
 const adminRoles = [1, 2, 5];
 // const isDateInPast = (date) => {
 //   const today = new Date();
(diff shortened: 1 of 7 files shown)


packages/webapp/public/locales/en/translation.json
```diff
@@ -1894,7 +1894,11 @@
     "TODO": "To do",
     "TRANSPLANT": "Transplant",
     "TRANSPLANT_LOCATIONS": "Select a transplant location",
-    "UNASSIGNED": "Unassigned"
+    "UNASSIGNED": "Unassigned",
+    "CREATE": {
+      "FAILED": "Can't create the task",
+      "LOCATION_DELETED": "It seems you are trying to create a task on a deleted location. Please try again afresh!"
```
antsgar, line 1900: I'd suggest

```suggestion
      "LOCATION_DELETED": "Another farmer has retired your selected location. Please refresh the page to get the most recent status."
```
so it's more consistent with a pre-existing string we have for attempting to delete a plan with completed tasks.

antsgar (approved):
It's working great, thank you so much for contributing! 🚀

packages/api/src/controllers/taskController.js
```diff
@@ -270,6 +273,20 @@ const taskController = {
     const nonModifiable = getNonModifiable(typeOfTask);
     return async (req, res, next) => {
       try {
+        // Only to satisfy "isDeleted" function requirement
```
kathyavini, line 276: If it's truly only for that and there is no other reason to open a transaction, I think you might be able to just pass `null` as the first argument to `baseController.isDeleted()` -- just tried that on another ticket and it seemed fine!

gursimran-singh: @kathyavini Thanks for the suggestion.

antsgar (approved):

PR #2989, opened 2023-11-15
LF-3423: disallow abandoning completed plan and viceversa

antsgar:
**Description**

Fixed concurrency issue where if a user completes a crop plan and the other one tries to abandon it without reloading the screen, or vice versa, we should prevent this from happening and show an error modal.

Jira link:

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

Tested locally by attempting the concurrent completing/abandoning with two different accounts.

**Checklist:**

### DIFF
--- a/packages/shared/constants/error.js
+++ b/packages/shared/constants/error.js
@@ -1,0 +1,2 @@
+export const CANNOT_COMPLETE_ABANDONED_PLAN = "Cannot complete management plan with an abandon date"
+export const CANNOT_ABANDON_COMPLETED_PLAN = "Cannot abandon management plan with a complete date"
--- a/packages/webapp/public/locales/en/translation.json
+++ b/packages/webapp/public/locales/en/translation.json
@@ -1036,7 +1036,7 @@
   "MANAGEMENT_PLAN": {
     "ABANDON": {
       "CANT_ABANDON_COMPLETED": "Can't abandon crop plan",
-      "CANT_ABANDON_CONCURRENT_USER": "Another farmer has completed this plan. Please refresh the page to get the most recent status."
+      "CANT_ABANDON_CONCURRENT_USER": "Another farmer has completed this plan. Please navigate back to get the most recent status for plans."
     },
     "ABANDON_MANAGEMENT_PLAN_CONTENT": "Abandoning this crop plan will abandon all the incomplete tasks associated with it and remove it from your farm map.",
     "ABANDON_MANAGEMENT_PLAN_TITLE": "Abandon crop plan?",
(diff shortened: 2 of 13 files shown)


packages/webapp/vite.config.ts
```diff
@@ -37,4 +38,9 @@ export default defineConfig({
   server: {
     port: 3000,
   },
+  resolve: {
+    alias: {
+      '@shared': path.resolve(__dirname, '../shared'),
```
antsgar, line 43: Added this alias so that it's easier to import from the shared folder in the webapp files.

packages/shared/constants/error.js
```diff
@@ -0,0 +1,2 @@
+export const CANNOT_COMPLETE_ABANDONED_PLAN = "Cannot complete management plan with an abandon date"
```
antsgar, line 1: I created this file for error strings returned by the API that need to be validated by the UI to display an error message. This way if on the BE side we decide to change the contents of the string for any reason, the UI will get the updated string to compare to.

packages/webapp/public/locales/en/translation.json
```diff
@@ -1046,6 +1050,8 @@
       "ABANDON_NOTES": "Abandonment notes",
       "ABANDON_PLAN": "Abandon plan",
       "ABANDON_REASON": "Reason for abandoning",
+      "CANT_COMPLETE_ABANDONED": "Can't complete crop plan",
+      "CANT_COMPLETE_CONCURRENT_USER": "Another farmer has abandoned this plan. Please refresh the page to get the most recent status.",
```
kathyavini, line 1054:
```suggestion
      "CANT_COMPLETE_CONCURRENT_USER": "Another farmer has abandoned this plan. Please return to the list of crop plans to get the most recent status.",
```

It's a minor point, but on the page where this modal appears (unlike the page where the user tries to delete a plan), there is no indication of status that would change if they refreshed the page.

From this view the user needs to go back to the list of crop plans (and actually that will be enough -- refreshing is not necessary) to get the status to update.

Screenshot 2023-11-15 at 12 18 17 PM

kathyavini (commented):
The logic looks great, and that's a neat solution for the API error message!

My only concern was on the modal wording -- for these two views you are taken to a different form page and so refreshing is not what the user needs to do (it will just clear their form and then they can submit again and get the same error). They need to navigate back in some way.

packages/webapp/public/locales/en/translation.json
antsgar: Thanks for pointing this out! I updated the copy 🙂

kathyavini (dismissed):
Looks good! Just has a merge conflict on the translation file.

kathyavini (approved):

PR #2990, opened 2023-11-15
LF-3766: Labour costs for harvest task is not counted if the task did not have an assignee when it was created

SayakaOno:
**Description**

1. fix the request body of the post harvest task API (we were not sending `override_hourly_wage` which is required later to add the appropriate wage when completing the task)
2. fix `completeHarvestTask` API to update `wage_at_moment` (we have two APIs to complete tasks, one is for harvest tasks, and the other is for other types of tasks. The chunk of code to add `wage_at_moment` was missing in the harvest task API)

NOTE: the existing data will not be corrected for now. (please see ticket's comments for more information)

Jira link: https://lite-farm.atlassian.net/browse/LF-3766

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/api/src/controllers/taskController.js
```diff
@@ -619,16 +619,30 @@ const taskController = {
...
+          .first();
+        wagePatchData.wage_at_moment = wage.amount;
+      }
+
       const result = await TaskModel.transaction(async (trx) => {
         const updated_task = await TaskModel.query(trx)
           .context({ user_id })
           .upsertGraph(
-            { task_id, ...req.body.task },
+            { task_id, ...req.body.task, ...wagePatchData },
```
SayakaOno, line 645: I referred these code:
https://github.com/LiteFarmOrg/LiteFarm/blob/361e95ac090fff4b6144b9f580142d6e1b9fe5f3/packages/api/src/controllers/taskController.js#L553-L578

packages/api/tests/task.test.js
```diff
@@ -1967,6 +1967,68 @@ describe('Task tests', () => {
       );
     });
 
+    test('wage_at_moment should be updated when completing a harvest task', async (done) => {
```
SayakaOno, line 1970: Almost the same as the one above...

packages/api/src/controllers/taskController.js
```diff
@@ -619,16 +619,30 @@ const taskController = {
...
+        assignee_role_id,
+        wage_at_moment,
+        override_hourly_wage,
+      } = await TaskModel.getTaskAssignee(task_id);
       const { role_id } = await UserFarmModel.getUserRoleId(user_id);
       if (!canCompleteTask(assignee_user_id, assignee_role_id, user_id, role_id)) {
         return res.status(403).send("Not authorized to complete other people's task");
       }
+
+      const wagePatchData = { wage_at_moment: override_hourly_wage ? wage_at_moment : null };
```
antsgar, line 633: There's a fair amount of code that is common between `completeHarvestTask` and `completeTask` -- could we abstract it to its own separate function/s so that when we make a change to this logic we can be sure that it applies to both? This way we would prevent bugs like this one where the wage logic was applied for regular tasks but not for a harvest task

packages/api/tests/task.test.js
antsgar, line 1970: Thanks for adding tests! 🙏

antsgar (commented):
I tested it and it works great! Left a comment but it's looking good 🙂

packages/api/src/controllers/taskController.js
SayakaOno: Thank you for the suggestion, I created two functions. It would be great if you could re-review!

packages/api/src/controllers/taskController.js
```diff
@@ -39,6 +39,51 @@ const adminRoles = [1, 2, 5];
...
+    assignee_role_id,
+    wage_at_moment,
+    override_hourly_wage,
+  } = await TaskModel.getTaskAssignee(task_id);
+  const { role_id } = await UserFarmModel.getUserRoleId(user_id);
+  if (!canCompleteTask(assignee_user_id, assignee_role_id, user_id, role_id)) {
+    throw new Error("Not authorized to complete other people's task");
+  }
+
+  const finalWage = { wage_at_moment: override_hourly_wage ? wage_at_moment : null };
```
antsgar, line 54: I think if we still called this one `wagePatchData` it'd be more clear to me that this is an object containing the patch data to be sent out rather than an actual wage amount

packages/api/src/controllers/taskController.js
```diff
@@ -39,6 +39,51 @@ const adminRoles = [1, 2, 5];
 //   return false;
 // };
 
+async function getTaskAssigneeAndFinalWage(farm_id, user_id, task_id) {
```
antsgar, line 42: I wonder if these two functions would actually be better suited in the model rather than the controller, since models are usually where data manipulation functions live. @kathyavini do you have any thoughts?

packages/api/src/controllers/taskController.js
```diff
@@ -39,6 +39,51 @@ const adminRoles = [1, 2, 5];
...
+  const {
+    assignee_user_id,
+    assignee_role_id,
+    wage_at_moment,
+    override_hourly_wage,
+  } = await TaskModel.getTaskAssignee(task_id);
+  const { role_id } = await UserFarmModel.getUserRoleId(user_id);
+  if (!canCompleteTask(assignee_user_id, assignee_role_id, user_id, role_id)) {
+    throw new Error("Not authorized to complete other people's task");
+  }
```
antsgar, line 52: If we moved this functions to the model, this bit could probably stay in the controller's endpoints even though it's repeated, since it handles sending out the unauthorized response.

antsgar (commented):
Left one small comment

packages/api/src/controllers/taskController.js
SayakaOno: I wanted to make it clear that this function is for completing task (I could have added a comment) and decided to name the function "...FinalWage" and had the variable name aligned with the function name.
I guess if I move the code to the model, that would not be a concern?
I'm not sure about the best practices, so I will wait for Joyce's response!

PR #2994, opened 2023-11-17
LF-3825 the white screen is displayed on the mobile phone device once the user switches the farms and opens the finance dashboard

kathyavini:
**Description**

Reorganizes the calling of sagas between switching farm and finances; puts finances behind a loading spinner until data fetching is compete; clears task data between farms.

Note: TypeError shown below may persist with a low reproduction rate.

Jira link: https://lite-farm.atlassian.net/browse/LF-3825

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

Mostly wrapped the TypeErrors in `try { } ... catch { }` and verified it was a mix of old + new farm data causing the crash. Only when switching farms, only on Slow 3G throttling.

**Checklist:**

### DIFF
--- a/packages/webapp/src/containers/Finances/index.jsx
+++ b/packages/webapp/src/containers/Finances/index.jsx
@@ -15,36 +15,30 @@
 
 import Moment from 'moment';
 import { extendMoment } from 'moment-range';
-import { useEffect, useRef } from 'react';
+import { useEffect, useRef, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { useDispatch, useSelector } from 'react-redux';
 import NoSearchResults from '../../components/Card/NoSearchResults';
-import { dateRangeOptions } from '../../components/DateRangeSelector/constants';
 import useDateRangeSelector from '../../components/DateRangeSelector/useDateRangeSelector';
 import AddTransactionButton from '../../components/Finances/AddTransactionButton';
 import DateRangeSelector from '../../components/Finances/DateRangeSelector';
 import FinancesCarrousel from '../../components/Finances/FinancesCarrousel';
 import PureTransactionList from '../../components/Finances/Transaction/Mobile/List';
 import PureCollapsibleSearch from '../../components/PopupFilter/PureCollapsibleSearch';
+import Spinner from '../../components/Spinner';
 import { Title } from '../../components/Typography';
 import { SUNDAY } from '../../util/dateRange';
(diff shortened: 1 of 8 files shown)


SayakaOno:
I tried twice on your branch, but the app crashed with the `planting_management_plan` error for the first time, and the second time was with `location_id`...
Screenshot 2023-11-16 at 10 54 18 PM

I waited until all APIs' responses were returned at least before switching to the finance page...

packages/webapp/src/containers/Finances/index.jsx
```diff
@@ -63,19 +57,26 @@ const Finances = ({ history }) => {
...
+
+  useEffect(() => {
+    if (!isFetchingData) {
+      setIsLoading(false);
+    }
+  }, [isFetchingData]);
+
+  useEffect(() => {
+    return () => {
+      setIsLoading(true);
```
SayakaOno, line 77: I don't understand why the state needs to be updated when unmounting...
@kathyavini do you have any ideas?

kathyavini: Hmmm I don't know! But with that cleanup you do get a loader when you go from Finances > Add Sale and then back button, don't you? 🤔 I don't like that! 😆

SayakaOno: You see the spinner when going back, after adding revenue/expense... whenever you go to the finance page. I don't like that either 😅 

I don't think unmounted component keeps its state, and I think it's always started with the initial value `true` passed to `useState`. I don't see any difference even if I remove `setIsLoading(true);`...

I'm reviewing your change now...

kathyavini: Oh yes sorry I used state too loosely there! The component state `isLoading` always resets to true. I just added the reducer bit to the other saga because that one actually does seem to control whether the spinner appears.

packages/webapp/src/containers/saga.js
```diff
@@ -621,6 +616,13 @@ export function* fetchAllSaga() {
   yield put(resetTasksFilter({ user_id, userFarms }));
 }
 
+export function* clearOldFarmStateSaga() {
+  yield put(resetTasks());
+
+  // Reset finance loading state
+  yield put(setIsFetchingData(true));
```
SayakaOno, line 623: UX is much better now!

SayakaOno (approved):

PR #3001, opened 2023-11-18
LF-3844 The user is not able to delete or abandon the crop plan once he deletes a task which is associated with this crop plan

kathyavini:
**Description**

Deleting or abandoning a plan with deleted tasks currently fails because the `getTaskType` method on `TaskModel` called here (to formulate the notification text):

```js
const { task_translation_key } = await TaskModel.getTaskType(task_id);
```

filters for deleted tasks, while the controllers for plan abandonment and delete themselves do not (presumably because there is no harm in re-deleting a deleted task or assigning an abandonment date to a deleted task). This means the method above gets called on deleted tasks, returns nothing, and then fails on the destructuring.

This PR:
- removes the check for deleted from that method (because logically speaking it should return the correct `task_type` regardless of the deletion status of the task in question
- Adds a check for task deletion in a more appropriate place in each of the four controllers that use `getTaskType`

Jira link: https://lite-farm.atlassian.net/browse/LF-3844

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/managementPlanController.js
+++ b/packages/api/src/controllers/managementPlanController.js
@@ -37,6 +37,7 @@ import {
 import knex from '../util/knex.js';
 import { getSortedDates } from '../util/util.js';
 import { TaskNotificationTypes, sendTaskNotification } from './taskController.js';
+import baseController from './baseController.js';
 const { transaction, Model } = objection;
 
 const managementPlanController = {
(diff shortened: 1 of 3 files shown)


SayakaOno (approved):
Works perfect!

PR #3019, opened 2023-11-30
LF-3892: migrate finances overview to rtk query

antsgar:
**Description**

Migrate Finances overview page to use RTK Query instead of sagas. This covers:
- Getting the necessary information for displaying the transactions.
- Adding a revenue or expense transaction.
- Getting the information to generate the report.

Jira link:

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/farmExpenseController.js
+++ b/packages/api/src/controllers/farmExpenseController.js
@@ -15,8 +15,8 @@
 
 import baseController from '../controllers/baseController.js';
 
+import { Model, transaction } from 'objection';
 import FarmExpenseModel from '../models/farmExpenseModel.js';
-import { transaction, Model } from 'objection';
 
 const farmExpenseController = {
   addFarmExpense() {
@@ -33,7 +33,7 @@ const farmExpenseController = {
           resultArray.push(result);
         }
         await trx.commit();
-        res.sendStatus(201);
+        res.status(201).send({ expenses: resultArray });
       } catch (error) {
         //handle more exceptions
         await trx.rollback();
--- a/packages/api/src/controllers/farmExpenseTypeController.js
+++ b/packages/api/src/controllers/farmExpenseTypeController.js
@@ -15,9 +15,9 @@
 
 import baseController from '../controllers/baseController.js';
 
+import { Model, transaction } from 'objection';
 import ExpenseTypeModel from '../models/expenseTypeModel.js';
 import FarmExpenseModel from '../models/farmExpenseModel.js';
-import { transaction, Model } from 'objection';
 
 const farmExpenseTypeController = {
   addFarmExpenseType() {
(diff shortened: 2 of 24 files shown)


kathyavini (commented):
Wow, it's truly mind-boggling to see how many fewer API calls are made this way! And I think I really like this API; it's so readable!

I just wanted to jot down my initial thoughts, but I will definitely spend some more time with this branch whenever I have some time this week:

1. Since finances are the most isolated part of the Redux store, they are a perfect place to begin the migration. But where we are touching on nested entities (here only crops and crop varieties) is still buggy, so I really want to see what fixing that up looks like. I'm mostly terrified of moving to the flat structure that this migration (not here in finances, but when we move it app-wide) will require 🙂 
2. I don't think it will necessarily be a lot of work, but it we'll need a new plan for fixing concurrency issues now. Right now we are addressing updates from other users with frequent GET requests to fetch the new data, and the new "refresh to see the most recent state" method we implemented last release is going to stop working since refreshing will now do nothing. I tried poking around in the RTK docs a little but nothing jumped out in terms of solutions for between-user updates other than polling or streaming updates. And it appears that tag invalidation can only be attached to a `build.mutation()`, is that your understanding too? If tags could be invalidated in a standalone way we could attach it to our current notification system, but if not, I'm not quite sure what to do -- it seems users could end up with very stale data if two users are logged onto the same farm.

But as long as we have a plan of attack for those two issues, I am soooo on board with starting this migration ASAP 😁

antsgar:
@kathyavini for the concurrency issues, it seems the cache for an endpoint could also be invalidated manually https://redux-toolkit.js.org/rtk-query/usage/manual-cache-updates

SayakaOno (commented):
I now have an overview of how to work with RTK query.
I really like the easy refetch by adding `invalidatesTags` and that everything is done in components so we don't need to go to saga files to see what's there.

Here are bugs that I came across:

https://github.com/LiteFarmOrg/LiteFarm/assets/33141219/348078a9-4fe8-45e6-b446-a9d954cdcdd5

Screenshot 2023-12-11 at 2 48 20 PM

Duncan-Brain (changes requested):
Looking forward to migrating to RTk query !

RTQ thoughts:
apiSlice.js:
- I am wondering if it might be better if we split up our apis and do many, as you said we will probably want to do more granular api calls (and by extension the tags) 
- We might then need to specify the reducerPath option to place it in the store?
- The name apiSlice is maybe confusing since there is a createSlice function similar to createApi the file structure I have seen names based on that function used.

What I am seeing on frontend is a broken finances
- the estimated value is not working on dashboard
- the other expense page breaks expenseTypes is not iterable console error 
- the revenues dont show up on revenues page

antsgar:
Closing, this is only an example of how to implement RTK Query moving forward

PR #3024, opened 2023-12-05
LF-3915 Clean up 'docker compose up' for new contributors

kathyavini:
Note: these are not high-priority changes! They are small quality of life improvements with `docker compose up` that were nonetheless going to bug me until fixed 🙂 

**Description**

We have an overarching task titled [Move to 'docker compose up' for new contributors](https://lite-farm.atlassian.net/browse/LF-3335) aimed at full containerization. This task is currently on pause; however, the modifications in this PR are smaller in scope and focus on making the current `docker compose up` command more user-friendly and efficient. These changes were partly inspired by feedback received during our August candidate search.

Changes:

- Improved the `minio_create_bucket` script: It now displays a success message instead of a confusing error if the bucket already exists. Additionally (as a side-effect, but retained because it's a bit useful), it lists now the bucket contents
- Moved the export server and its dependency `redis` to their own [Docker compose profile](https://docs.docker.com/compose/profiles/) so they don't start up by default. They are the slowest services to load and the least used, so don't belong in the default profile
- Change the export server serivce name to `export`, to make it easier to invoke manually with `docker compose up export`
- "Fixed" the export server startup crash. The underlying issue involved a script modifying a folder in `packages/api`, which triggered a nodemon reload, temporarily disabling the API and consequently crashing the export server. I just added a 10 second sleep to give nodemon time to do its reload.
- README updates: Changes related to the above changes, a new note on `test_farm` since it's not auto-created in Docker, and removing reference to the bug that was fixed in PR #3011.

Jira link: https://lite-farm.atlassian.net/browse/LF-3915

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

Please test by trying the new `docker compose up` in the root directory. And also `docker compose up export`.

**Checklist:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,6 +142,8 @@ Make sure that your `.env` variables are configured as outlined in the `.env.def
 
 Also see the section below [on using Docker](#docker).
 
+Note: the export server is not usually needed, and so is not started with the other development dependencies. To start it, make sure the the webapp and API are already running, and use the command `docker compose up export`.
+
 <details>
   <summary>Instructions for running image, document, and certification services natively</summary>
 
@@ -212,13 +214,15 @@ If you are unable to use Docker, please contact a core team member to get instru
 
 To run [ESLint](https://eslint.org/) checks execute `npm run lint`
 
-The [chai.js](https://www.chaijs.com/) and [jest](https://jestjs.io/) libraries automate tests that run real database operations using [JWT](https://jwt.io/introduction). The tests use a dedicated database named `test_farm`, distinct from the `pg-litefarm` database that the app normally uses .
+The [chai.js](https://www.chaijs.com/) and [jest](https://jestjs.io/) libraries automate tests that run real database operations using a dedicated database named `test_farm`, distinct from the `pg-litefarm` database that the app normally uses.
+
+You'll want to confirm that you have an empty `test_farm` database (otherwise use your preferred database client to create one) before continuing with the following:
 
 1. In a terminal, navigate to the `packages/api` folder.
 2. Execute `npm run migrate:testing:db` to set up the test database.
 3. Execute `npm test` to launch the tests. Or, to generate test coverage information, run `npm test -- --coverage .` and then see the `coverage/index.html` file.
 
-While the tests do attempt to clean up after themselves, it's a good idea to periodically use `psql` to `DROP` and `CREATE` the `test_farm` database, followed by the migrations from step 2 above.
+While the tests do attempt to clean up after themselves, it's a good idea to periodically use `psql` or your database client to `DROP` and `CREATE` the `test_farm` database, followed by the migrations from step 2 above.
 
 ## webapp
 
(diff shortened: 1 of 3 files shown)


kathyavini:
@Duncan-Brain

> My understanding of Ivans vision was that at some point docker compose up would add all configuration needed to run the entirety of LiteFarm. And then any developer specific preferences can be put in the docker override file?

Yes, that's right! But I think the spirit of what he was looking for was a very painless onboarding experience, which to me is actually sans export server.

The export server:
- is the slowest to start up (and even more so with the current code change to fix the crash). This would be a startup cost every time the dev boots up LiteFarm
- requires two docker services that don't have any application elsewhere
- obscures the API logs if you open it in the same terminal window
- frankly is not a very important part of the app for most contributors. If someone were to start working with the export server / job queue code, they would have to be pretty deep into LiteFarm development (and more than capable of starting up an individual service with `docker compose up export`

I use `docker compose up db minio imaginary` every day specifically for these reasons and it has covered all my daily dev needs. I'm guessing you probably never used the `docker compose up` as it is configured now (since you use native db + minio) but try it for a few days to see what I'm talking about... it's really quite annoying!

> I also noticed the Troubleshooting section of README disappeared?

This is covered in the PR description

kathyavini:
> should save you the hassle of typing those specific services:

@Duncan-Brain my point wasn't that I need a solution for my own environment, I was saying that in 6 months of work (and I think I was even the last person to touch the export server code anyway), those services have been more than adequate and the other two are just going to be bloat for devs. I truly do urge you to try `docker compose up` (without profiling redis and the export server) for a few days to understand what the issue is.

> Could be nice to keep the section and put something in it

Yeah I loved adding the troubleshooting section and I still think it's 👌 I'm all for adding in a different bug if you have a candidate! Maybe something with those annoying `.env` keys? (Otherwise we can always put it back when something appropriate comes up, that is pretty easy)

kathyavini:
Wow good memory @Duncan-Brain because I had zero recollection of writing that list 😂 It looks like most of those points still need to be addressed, but I don't think they belong in this PR thematically. I included the other update because it should have gone out with last week's PR and didn't, and that merge was fresh in my mind is all.

> but devs (for everyday devving) are they meant to be the only consumers of the command?

YES, big yes this is only for devs! Remember the idea was to have `docker compose up` -- the short notation -- be the dev environment since they are the ones using the while the server environments are the ones that require the full file notation:

`docker compose -f docker-compose.beta.yml up --build -d`

The dev docker compose file is not appropriate for a server environment, if that was one of the concerns.

README.md
```diff
@@ -278,33 +282,26 @@ Use cases in which we currently utilize docker at LiteFarm include:
...
+- To build the beta/prod Docker containers (generally used when troubleshooting the build process), you will also need to:
+  - create a `.env` file at the root directory of the project i.e. LiteFarm
+  - add key-value pairs in the `.env` by referring to the `docker-compose.[ENV].yml` that contains the docker env keys
 
 ## Commands
 
 These commands can be run from the root of the repo.
 
-- `docker compose up` to run all the local development services, OR
+- `docker compose up` to run the main development services, OR
```
antsgar, line 294: Should `docker compose up export` also be added to the commands list?

antsgar (dismissed):
Thank you for working on this and making it better! 🙌

antsgar (approved):

Duncan-Brain:
I don't want to hold this up and am happy to approve -- it just doesn't affect my workflow much. So I was curious about @SayakaOno opinion because maybe she uses docker up as well?

Happy to merge if the new configuration of making the export server its own extra command is good for her too.

SayakaOno (approved):
I rarely use docker - only when uploading files, so I don't have a strong opinion on this.
I confirmed that the both commands work, so I will merge :)

PR #3025, opened 2023-12-05
LF-3914 Post (some) handled errors to Sentry

kathyavini:
Update: we modified our Sentry permissions temporarily to allow troubleshooting from localhost. The [issue](https://litefarm.sentry.io/issues/4701829430/events/0096482202c7409b8d5311d028ab6d64/) format and stack trace look good and useable so I think it's fine to run on beta!

**Description**

Note: this relates to PR #3023 in that it piggy-backs on the Winston logger to get our errors to Sentry, but should be a standalone code change (i.e. doesn't depend on that PR's code changes). 

This PR adds a [custom Winston transport](https://github.com/winstonjs/winston#adding-custom-transports) to send Error objects passed to Winston to Sentry, using `Sentry.captureException(error)`, see [Sentry docs](https://docs.sentry.io/platforms/node/usage/#capturing-errors)). Because our needs were minimal, I wrote a basic transport, similar to [this StackOverflow example](https://stackoverflow.com/questions/75467152/why-is-sentry-capturemessage-not-working-inside-winston-transport-log-function).

I'm happy with the current output (although I wish it tracked users!). Just as a note, there is a more full-featured transport available here: [winston-transport-sentry-node](https://github.com/aandrewww/winston-transport-sentry-node), ([Source Code](https://github.com/aandrewww/winston-transport-sentry-node/blob/master/src/transport.ts)). I didn't use it because it has a bit more than we need (e.g. log levels), and I couldn't get it to work out-of-the-box, but it's a very popular package and always an option in the future.

Jira link: https://lite-farm.atlassian.net/browse/LF-3914

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

Tested by triggering the LF-3884 error from localhost with the environment code checks reversed (to run on `development`) and the beta sentry dsn key.

**Checklist:**

### DIFF
--- a/packages/api/package-lock.json
+++ b/packages/api/package-lock.json
@@ -57,6 +57,7 @@
         "uuid": "^8.3.0",
         "winston": "^3.8.2",
         "winston-daily-rotate-file": "^4.7.1",
+        "winston-transport": "4.6.0",
         "xlsx-populate": "^1.21.0"
       },
       "devDependencies": {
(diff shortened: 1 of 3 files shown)


SayakaOno (dismissed):
Let's test it!
The Jira ticket is linked instead of "Sentry docs" in the description 😉

SayakaOno (approved):

PR #3031, opened 2023-12-07
LF-3923a Add vite-pwa configuration 'registerSW({ immediate: true })'

kathyavini:
**Description**

This is the configuration recommended in the [Vite-PWA Documentation](https://vite-pwa-org.netlify.app/guide/auto-update.html#automatic-reload) to force **automatic page reload**. The goal in testing it in our application is to see if we can get better new code uptake in fewer refreshes 🙂 Of course we want to continue to not see any MIME type or other whitescreen errors!

I installed `workbox-window` as a dev dependency to address the Vite build error:

`Rollup failed to resolve import "workbox-window" from "/@vite-plugin-pwa/virtual:pwa-register".`

This was the recommended solution on the [vite-pwa GitHub issue tracker](https://github.com/vite-pwa/vite-plugin-pwa/issues?q=is%3Aissue+Rollup+failed+to+resolve+import+%22workbox-window%22+from+%22%2F%40vite-plugin-pwa%2Fvirtual%3Apwa-register%22.) for this error, and it resolved it in my local testing.

As before, this is a change that would be best tested with a subsequent frontend code merge to beta.

Jira link: https://lite-farm.atlassian.net/browse/LF-3923

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

I've continued to use the [local-docker](https://github.com/LiteFarmOrg/LiteFarm/tree/local-docker) branch, which contains an SSL-free Nginx configuration that can be run locally, to build and test the containers. However, the full implementation needs to be tested on beta itself.

**Checklist:**

### DIFF
--- a/packages/webapp/package.json
+++ b/packages/webapp/package.json
@@ -120,7 +120,8 @@
     "vite-plugin-istanbul": "^4.0.0",
     "vite-plugin-pwa": "^0.14.3",
     "vite-plugin-svgr": "^2.4.0",
-    "vitest": "^0.34.5"
+    "vitest": "^0.34.5",
+    "workbox-window": "^7.0.0"
   },
   "pnpm": {
     "overrides": {
--- a/packages/webapp/pnpm-lock.yaml
+++ b/packages/webapp/pnpm-lock.yaml
@@ -311,13 +311,16 @@ devDependencies:
     version: 4.0.0(vite@4.1.5)
   vite-plugin-pwa:
     specifier: ^0.14.3
-    version: 0.14.3(vite@4.1.5)(workbox-build@6.6.0)(workbox-window@6.6.0)
+    version: 0.14.3(vite@4.1.5)(workbox-build@6.6.0)(workbox-window@7.0.0)
   vite-plugin-svgr:
     specifier: ^2.4.0
     version: 2.4.0(rollup@3.25.3)(vite@4.1.5)
   vitest:
     specifier: ^0.34.5
     version: 0.34.5(happy-dom@12.9.1)(sass@1.56.1)
+  workbox-window:
+    specifier: ^7.0.0
+    version: 7.0.0
 
 packages:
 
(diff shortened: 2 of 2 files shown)


SayakaOno (commented):
Maybe I misunderstood it, but am I supposed to see a visual change in this PR?

>  I will first try adding the visual code change alongside the PR.

kathyavini:
@SayakaOno Ack no, sorry! Just an of data Jira ticket 😔

I was going to do it like that originally, but I'm starting to think these changes (nginx + vite-pwa) are a good patch candidate BEFORE our next minor release so in that case it's probably better to keep the PR pure to the vite-pwa config change.

SayakaOno (approved):

SayakaOno (approved):

Duncan-Brain:
Is this necessary? From the docs:
"The disadvantage of using this behavior is that the user can lose data in any browser windows/tabs in which the application is open and is filling in a form.

If your application has forms, we recommend you to change the behavior to use default prompt option to allow the user decide when to update the content of the application.

DANGER

Before you put your application into production, you need to be sure of the behavior you want for the service worker. Changing the behavior of the service worker from autoUpdate to prompt can be a pain."

kathyavini:
@Duncan-Brain wow, that's the second time I've linked the Jira ticket instead of PWA docs, thanks 😆 

> Is this necessary? 

Well, right now it takes logout + >2 page refreshes to get the new code to show, which... I don't know. If there are no crashes along the way, I think I would be happy with that as the final behaviour. But it's a bit weird to force logout only for the user to log back in to an unchanged app.

The goal here is to test it and see what the timing of the reload looks like (I'm truly not sure!) We can always revert. As to form behaviour, what we have now is to let the user finish filling in the form, and then log them out when they hit submit. Either way the form data is lost, and I would argue the second might be be a little more frustrating.

Regarding the danger, our application is already ON `autoUpdate` (see our vite config). So it's the moving off of it that would be a pain. As the docs mention, `autoUpdate` is not the same thing as automatic page reload.

Duncan-Brain:
Hmm it implies that there is a way to submit the form normally. But if its just a test then that is cool I guess.

Duncan-Brain (approved):

PR #3051, opened 2023-12-20
LF-3942 Fix release card update timing and test logout

kathyavini:
**Description**

Only code changes:
- Fixed handling of the translation on the New Release Card, so that it tracks the application version and is updated in tandem with client update, not before
- Changes to CSS of that card (decreasing padding to fit more version digits, moving inline CSS to module)

Existing functionality being tested:
- Included a version change in this PR to test the _existing_ logout functionality (please see Jira ticket for details)
- As far as I can tell, the webapp is already configured correctly to force logout when the app version constant is changed. As we don't change versions on integration, I have never seen this behaviour before on beta and would like to observe it

Jira link: https://lite-farm.atlassian.net/browse/LF-3942

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

Logout functionality can be tested in two ways, but both are only approximations of beta/prod:

_Simple test_

1. Change the constant `APP_VERSION` while running `pnpm dev`. You should immediately be logged out as soon as the application reloads.

_Closer (but imperfect) approximation to hosted environments_

1. Check out this branch. Run `pnpm vite build` and `pnpm vite preview`. Open the `localhost` preview
2. Change `APP_VERSION` and rebuild. Once build, reload the preview once to start the new service worker install (almost instantaneous on localhost), and reload a second time to update the application. You should be immediately logged out.

**NOTE:** This preview and build method is 100% susceptible to the TypeError on route change (the service worker does not behave like the hosted version in terms of the caching the app), so you cannot navigate _anywhere_ once the second vite build completes. Only reload.

Because neither of the above are perfect approximations for beta/prod, this needs to also be tested upon merge.

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/Card/NewReleaseCard/NewReleaseCard.jsx
+++ b/packages/webapp/src/components/Card/NewReleaseCard/NewReleaseCard.jsx
@@ -1,36 +1,27 @@
 import Card from '../index';
 import { ReactComponent as Star } from '../../../assets/images/signUp/new_feature.svg';
 import { Semibold, Text } from '../../Typography';
-import styles from '../../Typography/typography.module.scss';
+import typography from '../../Typography/typography.module.scss';
+import styles from './styles.module.css';
 import { useTranslation } from 'react-i18next';
-import { colors } from '../../../assets/theme';
+import { APP_VERSION } from '../../../util/constants';
 
 export function NewReleaseCard({ style }) {
   const { t } = useTranslation();
   return (
-    <Card
-      color={'info'}
-      style={{
-        display: 'flex',
-        flexDirection: 'row',
-        gap: '16px',
-        alignItems: 'center',
-        padding: '16px',
-        ...style,
-      }}
-    >
+    <Card color={'info'} style={{ ...style }} className={styles.card}>
       <Star />
       <div>
-        <Semibold style={{ color: colors.teal700, marginBottom: '12px', lineHeight: '20px' }}>
-          {t('SIGNUP.LITEFARM_UPDATED')}
+        <Semibold className={styles.updated}>
+          {t('RELEASE.LITEFARM_UPDATED', { version: APP_VERSION })}
         </Semibold>
-        <Text style={{ margin: 0, lineHeight: '18px' }}>
+        <Text className={styles.changes}>
           {t('SIGNUP.CHANGES')}{' '}
           <a
             href={'https://www.litefarm.org/post/an-investment-in-finances'}
             target="_blank"
             rel="noreferrer"
-            className={styles.underlined}
+            className={typography.underlined}
           >
             {t('common:HERE')}
           </a>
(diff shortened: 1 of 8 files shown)


packages/webapp/src/components/Card/NewReleaseCard/NewReleaseCard.jsx
```diff
@@ -1,36 +1,27 @@
...
-        alignItems: 'center',
-        padding: '16px',
-        ...style,
-      }}
-    >
+    <Card color={'info'} style={{ ...style }} className={styles.card}>
       <Star />
       <div>
-        <Semibold style={{ color: colors.teal700, marginBottom: '12px', lineHeight: '20px' }}>
-          {t('SIGNUP.LITEFARM_UPDATED')}
```
SayakaOno, line 25: Just to make sure, we shouldn't delete old translations, correct??

packages/webapp/src/components/Card/NewReleaseCard/styles.module.css
SayakaOno, line 1: It's not a big deal, but the usual indentation is two spaces 😉

packages/webapp/src/components/Card/NewReleaseCard/NewReleaseCard.jsx
kathyavini: Correct! That one should stay as long as we are "supporting" v3.5.2

packages/webapp/src/components/Card/NewReleaseCard/styles.module.css
kathyavini: Hmmm, `tabWidth` is right there in the `.prettierrc` and I thought VS code should be pulling from the repo setting foremost 🤔 But even if I configured my VS with an incorrect override, shouldn't it have been caught in the precommit hook? What's gone wrong?!

packages/webapp/src/util/constants.js
```diff
@@ -5,4 +5,6 @@ export const DO_ORIGIN_URL = `https://${
 export const DO_CDN_URL = `https://${
   import.meta.env.VITE_DO_BUCKET_NAME
 }.nyc3.cdn.digitaloceanspaces.com`;
-export const APP_VERSION = '3.5.1';
+
+// Changing this forces logout and updates the new release card
+export const APP_VERSION = '3.5.2.1';
```
antsgar, line 10: Will we be updating this version to 3.6 when we release the new navigation?

antsgar (approved):
Left a question but this is great! 🙂

SayakaOno (approved):

packages/webapp/src/util/constants.js
kathyavini: Is the IA release version 3.6? I wasn't sure if it would be that or patch (3.5.3). But I guess with full release notes a minor version bump makes more sense.

PR #3050, opened 2023-12-19
LF-3911: create side menu expanded version

antsgar:
**Description**

- Implemented new desktop and mobile navigation.
- Refactored finances routes so that they're all nested in /finances to reflect new IA.

Jira link:
https://lite-farm.atlassian.net/browse/LF-3911

**Type of change**

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

**How Has This Been Tested?**

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [ ] Passes test case
- [x] UI components visually reviewed on desktop view
- [ ] UI components visually reviewed on mobile view
- [ ] Other (please explain)

**Checklist:**

- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [x] The precommit and linting ran successfully
- [ ] I have added or updated language tags for text that's part of the UI
- [ ] I have added "MISSING" for all new language tags to languages I don't speak
- [x] I have added [the GNU General Public License](https://lite-farm.atlassian.net/l/cp/BT0Dd7WW) to all new files

PR #3053, opened 2023-12-20
LF-3912: compact desktop menu

antsgar:
**Description**

Added compact version of the menu and toggle to switch between the two versions.
Added animations.

Jira link:

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


SayakaOno (commented):
Looks great overall!

Here are my findings:
- modals and filters(= modals?) are in the centre of the browser not in the main content. Loïc says _"As a rule any content should hang centred within the main content area"_ (his [comment](https://lite-farm.atlassian.net/browse/LF-3932?focusedCommentId=16646) in a ticket)
  
Screenshot 2023-12-21 at 12 35 53 PM

Screenshot 2023-12-21 at 12 35 29 PM


- the toggle button moves along with main content scrolling, I assume it should stay.
   
Screenshot 2023-12-21 at 12 42 41 PM

- I don't think it needs to be addressed, but the crop icon is bolded when hovered over.

https://github.com/LiteFarmOrg/LiteFarm/assets/33141219/0d1f0554-e455-4a5e-8886-3b9d467af78b

antsgar:
@SayakaOno thank you for the feedback! Re the modals not being centered, I actually think in the case of the modals the navigation should be below the backdrop and not clickable, since the whole attention should be on the modal. I made this change in my last commit and addressed your other comments, let me know what you think

SayakaOno:
@antsgar The animation looks amazing! I love it!❤️❤️❤️ 

The updated modal looks great!
- I think filters need a separate fix...?
Screenshot 2023-12-21 at 12 35 29 PM

- The animation looks perfect after logging in, but after reload, the icon points the other way.
Screenshot 2023-12-21 at 4 59 05 PM

kathyavini:
Oh wow, ALL the animations look SO GOOD! The whole menu has become a place of delight 😁 And I'm strongly in favour of the animation every time, I think it's lovely and not at all too much 🙂 

Noticed two small things I wanted to check in on:
- If you open transactions and then collapse the menu, the side menu floats in space a bit (as if it were attached to the side of the expanded menu).

Screenshot 2023-12-21 at 9 35 11 PM

- Also, this change is not from this PR but rather from #3050, but because this is the last menu PR I wanted to make sure I checked: putting a nonsense URL used to clear the faulty path and load the homepage, but now it renders the menus but no content, and doesn't clear the path; is this intentional?

Screenshot 2023-12-21 at 9 56 38 PM

packages/webapp/src/components/Navigation/index.jsx
```diff
@@ -19,53 +21,85 @@ export default function PureNavigation({
...
-    !hidden && (
-      <>
+  return hidden ? (
+    children
+  ) : (
+    <>
+      {isFarmSelected && (
+        <>
+          {/* Sidebar Menu */}
+          {isMobile ? (
```
Duncan-Brain, line 48: I know I am not assigned to this PR but wanted to take a peek! The animation is really nice looking. 

I just saw you move this in (nice!) and thought the conditional was getting big here. The decision to use a drawer or not might be neatly handled in the SideMenu similarly to Top.

antsgar: Done! Can you re-test and see if everything still looks good? Worried I might have messed something up in the refactor 😅

antsgar:
@SayakaOno and @kathyavini thank you for the comments (that router bug was a big one! Thanks for finding that Joyce). I addressed your feedback, can you re-test please? 🙏

packages/webapp/src/components/Navigation/index.jsx
Duncan-Brain: Love it! Thats some nice stuff right there!

packages/webapp/src/styles.module.scss
```diff
@@ -31,22 +31,14 @@
...
-.sideMenu {
-  width: var(--global-side-menu-width);
-  height: 100vh;
-  position: sticky;
-  top: 0;
-  overflow: auto;
-  background-color: var(--bgInputListTile);
-}
-
 .mainColumn {
```
SayakaOno, line 42: I think this can be removed as well.

packages/webapp/src/components/Navigation/TopMenu/styles.module.scss
```diff
@@ -5,8 +5,9 @@
   height: var(--global-navbar-height);
 }
 
-.centerContent {
```
SayakaOno, line 8: This seems to be used in `TopMenu` (line 235). Probably that can be removed?

SayakaOno (commented):
I left small comments, but everything looks good!

kathyavini:
Hmmm, I just lost the mobile menu (like in a z-index sense; it's coming up blank now), is that reproducible for anyone else?


https://github.com/LiteFarmOrg/LiteFarm/assets/82857964/74002a58-e92d-48a2-9777-643cbae5b831

kathyavini:
Loïc would like the desktop view from the "lg tablet" breakpoint onwards. He thought that the discussion had landed on 600px for that, but wanted to double-check with @antsgar if that is correct 🙂

antsgar:
@kathyavini yes, the current breakpoint for the navbar switch is exactly 600px!

SayakaOno (approved):
I think it's working well!

kathyavini (approved):
Looks great!!

As discussed in Slack the two remaining related changes (map width and selector flow for the new flat menu structure) should be a separate PR 🙂

PR #3054, opened 2023-12-21
LF-3943 New release notification badge

kathyavini:
**Description**

This implements a small full-stack feature which shows a new release badge to returning users upon login.

- The implementation (sagas, routes, controllers) was most closely modelled after our spotlight feature, except slightly simpler. ReleaseBadge has no Redux store component, and there is no need to write additional migrations after the initial one.
- To greatly simplify the database setup, this won't show a badge to anyone on the first release after implementation: as no users have an entry in the `release_badge` table initially, everyone will be treated as a new user
- This means returning users will only see a badge on their first version upgrade AFTER they have visited LiteFarm on a previous version
- Please see Jira ticket below for details of the desired user experience for this badge
- Component [Storybook](http://localhost:6006/?path=/docs/components-releasebadge--docs)

Jira link: https://lite-farm.atlassian.net/browse/LF-3943

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

_Tested timing (including release-version logout) as follows:_
- Ran the database migrations
- Logged into LiteFarm (database entry created but no badge shown)
- Changed the `APP_VERSION` constant
- Observed: logout + badge shown after login, only once per user

**Checklist:**


packages/webapp/src/containers/ReleaseBadgeHandler/saga.js
```diff
@@ -0,0 +1,69 @@
...
+
+  const { user_id, farm_id } = yield select(loginSelector);
+  const header = getHeader(user_id, farm_id);
+
+  try {
+    const result = yield call(axios.get, releaseBadgeUrl, header);
+
+    const { app_version } = result.data;
+
+    if (app_version < currentVersion) {
```
Duncan-Brain, line 37: This is never true for me and seems to be preventing the functionality from working. Might need an else or a different way to compare versions.

packages/webapp/src/components/Navigation/index.jsx
```diff
@@ -51,6 +52,7 @@ export default function PureNavigation({
               open={showNotificationSpotlight}
               onFinish={resetSpotlight}
             />
+            <ReleaseBadgeHandler />
```
Duncan-Brain, line 55: Would this be more appropriately in App? The is Farm selected is handy here but it doesn't necessarily seem navigation related.

packages/api/src/routes/releaseBadgeRoute.js
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+import express from 'express';
+
+const router = express.Router();
+import releaseBadgeController from '../controllers/releaseBadgeController.js';
+
+router.get('/', releaseBadgeController.getReleaseBadgeVersion());
+
+router.patch('/', releaseBadgeController.updateReleaseBadgeVersion());
```
Duncan-Brain, line 23: Feels weird not to have any middleware haha. Would the isSelf middleware be appropriate? Makes me thing spotlight might be missing something.

packages/api/db/migration/20231219050204_create_release_badge_table.js
```diff
@@ -0,0 +1,27 @@
...
+ *  (at your option) any later version.
+ *
+ *  LiteFarm is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ */
+
+export const up = async function (knex) {
+  await knex.schema.createTable('release_badge', (table) => {
```
Duncan-Brain, line 17: Not against a new table per se -- but just curious what the value is to adding a new table versus keeping the same functionality on 
a) a new column in spotlights (since its kind of a version spotlight)
or
b) a new column on user

packages/webapp/src/components/ReleaseBadge/styles.module.scss
```diff
@@ -0,0 +1,103 @@
...
+  width: 400px;
+  height: fit-content;
+
+  border-radius: 8px;
+  background: #ffdad9;
+  box-shadow: 0px 4px 4px 0px rgba(0, 0, 0, 0.25);
+
+  position: fixed;
+  bottom: 32px;
+  z-index: 1400; // highest existing z-index is 1302
```
Duncan-Brain, line 11: z-index wars continue.. haha 

Makes me think there is probably a MUI component that handles all this functionality without making a custom component. Something like snackbar, alert, or popover.

packages/webapp/src/containers/ReleaseBadgeHandler/saga.js
kathyavini: Really? It was originally written `!==` (same as app version saga) but it works for me either way and this is the comparison we want (in case two devices have different versions loaded) 🤔 When you change your `APP_VERSION` constant in your testing, are you bumping it up alphabetically? Normal semantic versioning (bumping the minor / patch version) should work fine if you're not going into double-digits.

packages/webapp/src/components/Navigation/index.jsx
kathyavini: Can it be done from App and still maintain dependence on `isFarmSelected`? I wrote it in the old `<NavBar />` originally (back when there was a demarcation between farmSelected and not farmSelected there), not knowing that component was going to be refactored. I'm not at all particular about where it's placed, as long as it doesn't appear until the login is completed -- do you know of a better place with that logic?

packages/api/src/routes/releaseBadgeRoute.js
kathyavini: Yup, I went through that same thinking, initially adding and then removing `isSelf`. In the end I figured maybe the rationale (for spotlight, and so for here), is that it just doesn't matter that much? I guess that would be an argument against moving this to the `user` table where a patch could have some consequence.

Duncan-Brain:
I think the comparator function is all I can see thats actually a problem. But I can test by setting it to !=.

Left a couple other inline comments to think about.

It looks to me as scoped -- I might just ask if the Tablet version should be the same as mobile? Figma doesnt have it but to me it seems like our pattern is to keep tablet closer to desktop than mobile look?

packages/webapp/src/components/Navigation/index.jsx
Duncan-Brain: I don't know about App, but I think it would need to go in the ReleaseHandler or some container that holds the component.

packages/api/db/migration/20231219050204_create_release_badge_table.js
kathyavini: No particular reason, other than it was simpler to write the logic I wanted from scratch rather than complicate the Spotlight saga, controllers, and model with new behaviour (e.g. that one is boolean based but I needed the string comparison here, and of course the frontend handling is quite different). I never thought about the `user` table at all actually, but I think the thematic coupling is a lot tighter to Spotlight.

Having finished it, I think it IS easier to read and reason about this way, but I'm curious what @antsgar thinks.

packages/api/src/routes/releaseBadgeRoute.js
Duncan-Brain: Well it always matters to have good auth -- otherwise in this case any data can be written to that string column. Like a malicious html element. 

I forget if the id_token is verified against creds by this point so I am not sure its needed or anything but spotlights could just be wrong -- I also see it done on some sensor routes which also scares me haha.

packages/webapp/src/components/ReleaseBadge/styles.module.scss
kathyavini: We could also refactor it into a more sensible structure without MUI's help 😂  I think I've read a blog post (did you send it??!) against the random z-index situation we have gotten ourselves into.

packages/webapp/src/containers/ReleaseBadgeHandler/saga.js
Duncan-Brain: It works once I have a record (probably with a number) in the db but it doesn't create one until it works. So its probably stuck in a `null < null` situation. 

Do strings compare well this way? Are we sure future Litefarmers wont go high on patch versions or change version naming to some other approach?

Duncan-Brain: nvm this is what I am seeing: 
Screenshot 2023-12-21 at 4 49 30 PM

kathyavini: Ahhhhh okay thank you!! This is because I established my database records with the other method.

I'm not sure about the versioning; the fact that we're up to 3.6 potentially does make me nervous about double digits 🤔 Maybe I'll do a more appropriate semantic versioning compare then. I don't think we will change our whole version scheme (given that we've done so many releases on it already) but maybe a note for future LiteFarmers then!

packages/webapp/src/components/ReleaseBadge/styles.module.scss
Duncan-Brain: yes haha z-index bad, me no likey.  -- I think the 1302 is to override MUI styles too. This was the article: https://www.codegram.com/blog/stop-ab-using-z-index/

MUI is pretty nice though -- an info Alert makes a lot of sense to me -- the blue color is so nice too ;) https://mui.com/material-ui/react-alert/

kathyavini:
@Duncan-Brain just to confirm by tablet view, that would be up to the higher tablet breakpoint of `1024px` yes? I'll ask Loïc. I think it looks good, I would just want to check the distance between the text and the close button.

Screenshot 2023-12-21 at 1 55 30 PM



I will def fix the comparator logic thank you!!! ❤️

packages/webapp/src/components/Navigation/index.jsx
kathyavini: Yeah that makes sense. I don't think it's 100% _un_related to navigation -- as the whole logged-in UI just comes up together as a unit, including the badge -- but it REALLY doesn't belong in the Pure Component.

kathyavini:
Thanks @Duncan-Brain, that was really helpful! I added a more robust semantic versioning comparison (apparently not a trivial task? At least according to the [StackOverflow thread](https://stackoverflow.com/questions/6832596/how-can-i-compare-software-version-number-using-javascript-only-numbers) with the implemented solution). And the very important `null` handling!

I moved `<ReleaseBadgeHandler />` out of `<PureNavigation />` (yuck that was a bad choice!) to the container, but I couldn't get it working all the way up in `<App />` calling the farmSelected hook within itself. Maybe because the app version is populated in the Redux store after the farmSelection flow? I actually don't fully know but I didn't want to sink too much time on that particular refactor as it's now back where I wanted it originally. I hope it's okay that it's still infringing on Navigation a bit -- but less than before!

I also prefer the mobile version at tablet breakpoint, so I made that change, and I'll just revert/tweak if Loïc responds with a hard "no" 🙂

packages/webapp/src/components/ReleaseBadge/styles.module.scss
```diff
@@ -0,0 +1,103 @@
...
+  font-size: 19px;
+  font-weight: 700;
+
+  :visited {
+    color: #520f0d;
+  }
+}
+
+// Max tablet breakpoint
+@media (max-width: 1024px) {
```
Duncan-Brain, line 73: I think Antonella created some mixin breakpoints in the nav PR which would be good to start using.

Duncan-Brain:
Working well for me! 

I too tried moving the Release badge out of navigation and into app and putting the conditional for useIsFarmSelected directly into the Release handler and also found it didnt work well -- needing a refresh to see the popup.

Duncan-Brain (approved):

packages/webapp/src/components/ReleaseBadge/styles.module.scss
kathyavini: I'm only seeing the one for 600px (`sm-breakpoint`) in `/src/assets/mixin.scss` and 1024px is defined in each stylesheet. But we use it a lot -- should it be a mixin too @antsgar?

kathyavini:
Loïc has asked for the desktop view from the lg-tablet breakpoint onwards. He thought the discussion had landed on 600px for that, but suggested double-checking that value with you @antsgar 🙂

packages/api/db/migration/20231219050204_create_release_badge_table.js
antsgar: I originally thought this was going to be a new column in the spotlight table too, but since this isn't really a spotlight I like the idea of keeping it isolated to make the feature easier to understand -- otherwise you might think there should be a spotlight related to this when there isn't. As for the user table, if it was just the app version column we could have gone down that path, but if there's any usefulness to having an updated at / created at specifically for the app version entry I prefer to keep all of it in its own table rather than bloat the user table with 3 new columns

packages/api/src/controllers/releaseBadgeController.js
```diff
@@ -0,0 +1,66 @@
...
+            .findById(user_id)
+            .patch({ app_version })
+            .context({ user_id });
+        } else {
+          await ReleaseBadgeModel.query()
+            .insert({
+              user_id,
+              app_version,
+            })
+            .context({ user_id });
```
antsgar, line 54: Not super important right now, but it looks like knex has an `onConflict` method that we could use to make this kind of upsert logic more simple (not sure if it's available in the version that we have or a newer one, though). Could be worth looking into for future changes! https://knexjs.org/guide/query-builder.html#onconflict

packages/api/src/controllers/releaseBadgeController.js
```diff
@@ -0,0 +1,66 @@
...
+        }
+
+        res.sendStatus(200);
+      } catch (error) {
+        console.error(error);
+        res.status(400).json({ error });
```
antsgar, line 60: I think we do this a lot throughout the app, but since we don't have a reason to think here that the error is caused by something that was sent wrong by the client, I don't think "bad request" is the appropriate status code. For a catch all for generic errors like this, I think 500 makes more sense. Wdyt?

antsgar (dismissed):
Left minor comments, it looks awesome! This badge seems to have been a lot of work, nicely done!! 🙌

packages/api/src/controllers/releaseBadgeController.js
kathyavini: Ah this is something I'd love to establish guidelines for! I think we have always used 400 as our catch-all, with just a few 500s sprinkled in randomly. I don't know if it's intended, but as a side-effect of our habit now I associate 500 errors as being "severe" ones, since they are usually only sent by the general error handler in `server.js`. That said I think I threw one into a controller recently (finance report) so I'm not even internally consistent 😔

Duncan-Brain (approved):
Works well for me nice job!

Might be good to add the semantic versioning limitations somewhere like constants or release checklist page or something.

packages/api/src/controllers/releaseBadgeController.js
antsgar: The code shouldn't really be related to the severity, but to the cause of the error. 400 codes indicate something went wrong on the client side, so for example if you check for a required parameter and it's missing in the request, or the shape of the payload doesn't look the way it should be, or there's a missing header -- anything that indicates that the client sent out something you weren't expecting. Reading the definition for the 500 status code (https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/500), it says this is the "catch all" for a server error where you don't really know what went wrong, so I think that would be the fitting one in these generic catch blocks. Not a blocker for merging though! But something I think we should start doing throughout the app

antsgar (approved):

PR #3055, opened 2023-12-22
LF-3932: Fix broken styles due to navigation update

SayakaOno:
**Description**

This PR fixes three things that got broken by the navigation update.
1. adjust the snackbar position so that it is placed in the centre of the main content. (I moved `isCompactSideMenu` state up to `App` component which I didn't like. Let me know if you have suggestions)
2. fix farm selection buttons to fit in the main content. I had to pass styles, which we probably don't want to.
3. fix roles translations in people page
4. fix overflowing `WeatherBoard` on the home page
5. add a link to the homepage to the logo on mobile `TopMenu`

Jira link: https://lite-farm.atlassian.net/browse/LF-3932

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/App.jsx
+++ b/packages/webapp/src/App.jsx
@@ -13,7 +13,7 @@
  *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
  */
 
-import { Suspense } from 'react';
+import { Suspense, useState } from 'react';
 import Navigation from './containers/Navigation';
 import history from './history';
 import clsx from 'clsx';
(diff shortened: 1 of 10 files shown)


Duncan-Brain (approved):
Looks good to me.

1. I think that makes sense barring some way to get the width -- I thought I have seen you made a hook once like that or using the variables if they are dynamic.
2. Nice!
5. Thank you that was my bad

antsgar (approved):
Fixes look good, thanks for working on these!

Re moving up the state variable for `isCompact` -- unfortunately for absolutely positioned elements such as the snackbar, we need the component to somehow know the width of the sidebar if we want to use it for centering/sizing. I think overall there may be other components with these need (such as the release badge). In that sense we might be better off making this state variable part of the Redux store, although with our current persist configuration that would mean keeping the configuration for the side menu after reloading, that is, keeping it collapsed if it was collapsed and that seems incompatible with showing the expansion animation on load. Alternatively we could use React Context, which would usually be my preferred approach but might be a confusing addition since we already use Redux.

SayakaOno:
Thank you for reviewing and the feedback!

@Duncan-Brain 
The hook must be `useElementWidth`!? That could have been an option but I was obsessed with doing it with CSS for some reason...! 😮

@antsgar 
I haven't really used `Context`, but my preference would be 1. context 2. Redux, too. I agree that we should not just go for context since we have never used it in the project.
I have addressed some of your comments in https://github.com/LiteFarmOrg/LiteFarm/pull/3058. Thank you 🙏

PR #3056, opened 2023-12-22
Navigation fixes

antsgar:
**Description**

- Fixes missing entities errors in Finances subsections.
- Fixes tasks cards not sizing gracefully.
- Fixes buggy styling on crop detail page.

Jira link:

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/containers/Finances/ActualRevenue/index.jsx
+++ b/packages/webapp/src/containers/Finances/ActualRevenue/index.jsx
@@ -13,14 +13,13 @@ import { cropVarietiesSelector } from '../../cropVarietySlice';
 import { setPersistedPaths } from '../../hooks/useHookFormPersist/hookFormPersistSlice';
 import { allRevenueTypesSelector } from '../../revenueTypeSlice';
 import ActualRevenueItem from '../ActualRevenueItem';
-import { getRevenueTypes } from '../saga';
+import { getRevenueTypes, getSales } from '../saga';
 import { salesSelector } from '../selectors';
 import {
   calcActualRevenueFromRevenueItems,
   filterSalesByDateRange,
   mapSalesToRevenueItems,
 } from '../util';
-import { getSales } from '../actions';
 import { getCropVarieties } from '../../saga';
 
 export default function ActualRevenue({ history, match }) {
@@ -51,8 +50,8 @@ export default function ActualRevenue({ history, match }) {
   );
 
   useEffect(() => {
-    dispatch(getRevenueTypes());
     dispatch(getSales());
+    dispatch(getRevenueTypes());
     dispatch(getCropVarieties());
   }, []);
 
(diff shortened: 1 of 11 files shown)


kathyavini (commented):
These look like good fixes to me!

I'm still not seeing Actual Revenues on the first load though (visiting directly after login). Expenses is perfectly fixed for me.

packages/webapp/src/containers/Finances/ActualRevenue/index.jsx
```diff
@@ -49,9 +51,9 @@ export default function ActualRevenue({ history, match }) {
   );
 
   useEffect(() => {
```
kathyavini, line 53: An inelegant but working solution would be to just call `fetchAllData()` here 😬

packages/webapp/src/containers/Finances/ActualRevenue/index.jsx
```diff
@@ -49,9 +51,9 @@ export default function ActualRevenue({ history, match }) {
...
-    if (!allRevenueTypes?.length) {
-      dispatch(getRevenueTypes());
-    }
+    dispatch(getRevenueTypes());
```
kathyavini, line 54: An inelegant but working solution seems to be just calling `fetchAllData()` here instead 😬

antsgar: So I have no idea why, but the getSales action wasn't actually working -- setting it up with `createAction` as some of the others seems to have done the trick

kathyavini (commented):
Strange about `ADD_SALE` but it works well now!

Can you remind me where we landed regarding this error where the `sale_id` isn't being populated in the URL? It looks familiar (was it mentioned on one of the previous nav PRs?) but I can't fully remember. This is on beta/integration too, not just this branch:

https://github.com/LiteFarmOrg/LiteFarm/assets/82857964/dda5f32d-7186-4f7c-b1a4-6cbc86da13e8

Also, as a minor cosmetic point, I'm wondering if the scroll bars should be hidden in all cases on the new menu even when the content overflows due to a smaller screen height (it happens in the video above when Finances is expanded). The default scroll bar looks a little wonky to me here and breaks estimated revenue onto two lines 🤔

antsgar:
@kathyavini I (hopefully) fixed the redirects using this syntax https://github.com/remix-run/react-router/pull/5209, can you re-test please? 🙏 re the scrollbars, I think it's best not to override the browser's default behavior for accessibility/UX, since the bars are a visual cue that the content is scrolling. They look a bit different for me though in Chrome, which browser are you using?

kathyavini (approved):
Looks good!

@antsgar I am on Chrome as well. So I wonder if you have something different selected in MacOS itself -- @SayakaOno did some sleuthing once and found a system setting that controls scrollbar appearance in Chrome, but it's not a Chrome setting at all; kind of wild! Source: https://github.com/LiteFarmOrg/LiteFarm/pull/2794

Mine look pretty different depending on whether I have "Automatically based on mouse or trackpad" or "When scrolling" selected

SayakaOno (approved):
Looks good to me!

PR #3059, opened 2024-01-03
Nav menu styling fixes

antsgar:
**Description**

- Updated active state styling for menu subitems so they're different from items.
- Removed tabs from settings since they're no longer needed (we now have direct links navigating through each section in the navigation).
- Make map full width (approach extendable to future full width pages as well) and center Redraw / Confirm buttons.

Jira link:

https://lite-farm.atlassian.net/browse/LF-3948
https://lite-farm.atlassian.net/browse/LF-3950
https://lite-farm.atlassian.net/browse/LF-3952

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/routes/index.jsx
+++ b/packages/webapp/src/routes/index.jsx
@@ -279,7 +279,7 @@ const UnknownRecord = React.lazy(() =>
   import('../containers/ErrorHandler/UnknownRecord/UnknownRecord'),
 );
 
-const Routes = () => {
+const Routes = ({ isCompactSideMenu }) => {
   useScrollToTop();
   useReduxSnackbar();
   const userFarm = useSelector(
@@ -509,7 +509,13 @@ const Routes = () => {
             />
             <Route path="/tasks/:task_id/harvest_uses" exact component={HarvestUses} />
             <Route path="/tasks/:task_id/abandon" exact component={TaskAbandon} />
-            <Route path="/map" exact component={Map} />
+            <Route
+              path="/map"
+              exact
+              component={(history) => (
+                <Map isCompactSideMenu={isCompactSideMenu} history={history} />
+              )}
+            />
             <Route path="/map/videos" exact component={MapVideo} />
             <Route
               path="/create_location/farm_site_boundary"
@@ -780,7 +786,13 @@ const Routes = () => {
             <Route path="/documents/add_document" exact component={AddDocument} />
             <Route path="/documents/:document_id/edit_document" exact component={EditDocument} />
             <Route path="/documents/:document_id" exact component={MainDocument} />
-            <Route path="/map" exact component={Map} />
+            <Route
+              path="/map"
+              exact
+              component={(history) => (
+                <Map isCompactSideMenu={isCompactSideMenu} history={history} />
+              )}
+            />
             <Route path="/map/videos" exact component={MapVideo} />
             <Route
               path="/create_location/farm_site_boundary"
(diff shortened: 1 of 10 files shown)


SayakaOno (commented):
Looks good overall!

A few of my thoughts:
- I would like a title on the pages that had the tab (account, people and farm). Maybe it's just me, but I feel like I'm lost.
- You are still waiting for the decision on the small modal, but I think the longer modal like below looks nicer when it is on the map...
Screenshot 2024-01-03 at 2 36 47 PM

kathyavini (commented):
It looks good and I like the full-width solution!!

I checked the UX channel because I was curious about the buttons and modal going inside the map header now, but looks like no reply yet. I think specifically the modal might need an outline in this case, like the top and side menu bars have, because otherwise the contrast is disconcertingly low:

Screenshot 2024-01-03 at 2 49 25 PM

Other than that, I see a TypeError crash when drawing a line (posted in Slack), but locally I don't see it anymore after merging integration into this branch; could you confirm that from your end too? If the header styling gets the a-ok from UX and the crash is resolved with an integration merge, then this has my approval 🙂 (since I won't be here tomorrow to approve in actuality)

antsgar:
@SayakaOno I switched back to the buttons / modal being absolutely positioned so you should now see the modal over the map. For the pages titles, I'd like Loïc's input on that so I'd prefer we defer that to a future update, would you be okay with that?

As for the error in the map, I'm able to repro, it seems this is related to the Google Map React library + React 18 + Strict Mode (see here https://github.com/google-map-react/google-map-react/issues/1129). I'm not sure how to disable strict mode with Vite, but I'm also able to repro on integration without these changes so I don't think it's something new, and if it's related to Strict Mode it should just happen locally.

Duncan-Brain (changes requested):
I am seeing the error as well + "history.push is not a function" on areas and lines and everything. But I think it is because the history prop should be destructured from the arrow function -- that seems to work for me to fix all type errors. Will point it out inline. ... I think history is available through some parent component or something for the other routes but I am not 100% on that.

I am technically fine with the change to pass the isCompactSideMenu down as props in the route... but it is the only Route we do that for. So I am not a huge fan of the change of pattern. Just wondering if there is some sort of 100% width solution instead that will respond with it since it should just filling the content container. -- I am fine if thats not feasible for the release but ya seems like a new pattern to pass props this way.

packages/webapp/src/routes/index.jsx
```diff
@@ -509,7 +509,13 @@ const Routes = () => {
             />
             <Route path="/tasks/:task_id/harvest_uses" exact component={HarvestUses} />
             <Route path="/tasks/:task_id/abandon" exact component={TaskAbandon} />
-            <Route path="/map" exact component={Map} />
+            <Route
+              path="/map"
+              exact
+              component={(history) => (
```
Duncan-Brain, line 515: @antsgar I think history just needs to be destructured here is all to fix the errors. My only real change request

antsgar: Ahhh good catch! Will fix. Re passing isCompactSideMenu -- if the styling is done properly, we shouldn't even need to be aware of the width of the side menu. Sadly the absolute positioning of elements here in the map makes it hard to center stuff without relying on that number. Even then, instead of just drilling that down we should store it globally with React Context, but I figured this was okay for now.

antsgar: @Duncan-Brain should be fixed now, thank you for catching it!

Duncan-Brain (approved):
Works for me!

SayakaOno (approved):
The modal looks great!
As for the pages titles, I am okay with leaving it as a future improvement :)

Thank you for looking into the error, I will create a ticket!

PR #3060, opened 2024-01-03
LF-3944 Some translations become broken when the user refreshes the page and snackbar translations are always missing

kathyavini:
**Description**

*As tested in group*

Using `git bisect` commit 7f631d24bc5526cf7aaf7eefe63488f871fadd11 was identified as being responsible for the current translation issues. Also confirmed at that time that it is actually the namespacing in this hook (i.e. the inclusion of 'messages' in that array) which controls the availability of the translation elsewhere in app.

Note: we are NOT entirely sure why this should be, but we are restoring to the previous state/behaviour. We also, for consistency, added the reference to the one omitted translation file (`revenue`). We are not currently seeing errors on revenue translations, but did at one point [fairly recently](https://litefarm.slack.com/archives/C014931NRFY/p1701365564357259).

Jira link: https://lite-farm.atlassian.net/browse/LF-3944

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

Tested together and found:
- the namespacing itself is necessary (e.g. removing 'messages' from the namespace array means the snackbars will continue to break)
- (really inexplicably) the call should happen in the Pure Component; calling in the wrapping container didn't fix the error

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/Navigation/index.jsx
+++ b/packages/webapp/src/components/Navigation/index.jsx
@@ -10,6 +10,7 @@ import {
   NavBarNotificationSpotlightProvider,
 } from './NavbarSpotlightProvider';
 import styles from './styles.module.scss';
+import { useTranslation } from 'react-i18next';
 
 export default function PureNavigation({
   showNavigationSpotlight,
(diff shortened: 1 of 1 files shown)


antsgar (approved):

Duncan-Brain (approved):
Awesome find Joyce!

Oh the hours wasted on a nonsensical solution. Must remember git bisect haha.

PR #3061, opened 2024-01-04
LF-3956: Navigation bar is displayed on the login page

SayakaOno:
**Description**

After looking into `useHistoryLocation` hook, I found that `history.listen()` in which `setHistoryLocation` is called was not being called in `useEffect`.
So, we were referencing the old `historyLocation` here:

Screenshot 2024-01-03 at 4 21 13 PM

Replacing `useEffect` with `useLayoutEffect` seems to work fine.
I can't explain better than what ChatGPT did, so I will just copy and paste...

> The useEffect hook runs asynchronously after the render, which means it might not guarantee that the DOM has been fully updated or that certain events, like history changes, have been processed.
> 
> When you use useEffect with history.listen, there's a possibility that the initial render or reload might not immediately trigger the history change event, leading to the callback not being executed right away.
> 
> On the other hand, useLayoutEffect is similar to useEffect, but it fires synchronously after all DOM mutations. Using useLayoutEffect ensures that the DOM is updated before the effect runs, which can be crucial in scenarios where you want to perform actions that depend on the updated DOM state, like setting up event listeners for history changes.
> 
> In summary, if you need to ensure that certain actions, such as setting up listeners for external events like history changes, occur immediately after the DOM is updated, useLayoutEffect might be more appropriate than useEffect. However, be mindful of potential performance implications, as synchronous updates may impact perceived responsiveness. Use it when the synchronous behavior is necessary for your specific use case.

Jira link: https://lite-farm.atlassian.net/browse/LF-3956

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/containers/hooks/useHistoryLocation.js
+++ b/packages/webapp/src/containers/hooks/useHistoryLocation.js
@@ -1,10 +1,26 @@
-import { useEffect, useState } from 'react';
+/*
+ *  Copyright 2023 LiteFarm.org
+ *  This file is part of LiteFarm.
+ *
+ *  LiteFarm is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  LiteFarm is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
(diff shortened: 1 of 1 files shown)


antsgar:
Works great! 🎉

antsgar (approved):

antsgar (approved):
Works great, thank you for fixing! I think on the long run we might be better off reading from history directly instead of using this hook, but this works perfectly for the time being

Duncan-Brain:
Seems to be working for me! Thank you for fixing this, I think I messed up by making this hook.

Long long term I also wonder if we should move login page out of Routes and into App,  or Move Navbar out of app and into Routes.

PR #3074, opened 2024-01-11
LF-3972 Getting a list of available animal types

kathyavini:
**Description**

Here we go livestock! 😄 

This is the database table and GET endpoint for animal types. In looking through Jira, it seems the API endpoints have slightly re-designed (right @antsgar?); you can see it more clearly in the ticket for the GET endpoint for breeds, which has a query parameter. The only new element here is the removal of `farm_id` from the route and consequently a change to the route middleware pattern.

Jira link: https://lite-farm.atlassian.net/browse/LF-3972

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

Inspected the table and tested the new route (GET `/animal_type`) on Postman. Created new tests modelled after `revenue_type.test.js`

**Checklist:**

### DIFF
--- a/packages/api/db/migration/20240110204504_create_animal_type_table.js
+++ b/packages/api/db/migration/20240110204504_create_animal_type_table.js
@@ -17,14 +17,14 @@ export const up = async function (knex) {
   await knex.schema.createTable('animal_type', (table) => {
     table.increments('id').primary();
     table.uuid('farm_id').references('farm_id').inTable('farm').defaultTo(null);
-    table.string('type').defaultTo(null);
-    table.string('type_key').defaultTo(null);
+    table.string('custom_type_name').defaultTo(null);
+    table.string('default_type_key').defaultTo(null);
     table.boolean('deleted').notNullable().defaultTo(false);
     table.string('created_by_user_id').references('user_id').inTable('users');
     table.string('updated_by_user_id').references('user_id').inTable('users');
     table.dateTime('created_at').defaultTo(new Date('2000/1/1').toISOString()).notNullable();
     table.dateTime('updated_at').defaultTo(new Date('2000/1/1').toISOString()).notNullable();
-    table.check('?? is not null or ?? is not null', ['type', 'type_key']);
+    table.check('?? is not null or ?? is not null', ['custom_type_name', 'default_type_key']);
   });
 
   // Add initial default types
@@ -33,8 +33,8 @@ export const up = async function (knex) {
   for (const typeKey of defaultTypeKeys) {
     await knex('animal_type').insert({
       farm_id: null,
-      type: null, // only provide for user-created types (i.e. without translation keys)
-      type_key: typeKey,
+      custom_type_name: null, // only provide for user-created types (i.e. without translation keys)
+      default_type_key: typeKey,
       deleted: false,
       created_by_user_id: '1',
       updated_by_user_id: '1',
--- a/packages/api/db/migration/20240110204504_create_animal_type_table.js
+++ b/packages/api/db/migration/20240110204504_create_animal_type_table.js
@@ -22,8 +22,9 @@ export const up = async function (knex) {
     table.boolean('deleted').notNullable().defaultTo(false);
     table.string('created_by_user_id').references('user_id').inTable('users');
     table.string('updated_by_user_id').references('user_id').inTable('users');
-    table.dateTime('created_at').notNullable();
-    table.dateTime('updated_at').notNullable();
+    table.dateTime('created_at').defaultTo(new Date('2000/1/1').toISOString()).notNullable();
+    table.dateTime('updated_at').defaultTo(new Date('2000/1/1').toISOString()).notNullable();
+    table.check('?? is not null or ?? is not null', ['type', 'type_key']);
   });
 
   // Add initial default types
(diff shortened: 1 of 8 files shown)


packages/api/db/migration/20240110204504_create_animal_type_table.js
```diff
@@ -0,0 +1,77 @@
...
+  await knex.schema.createTable('animal_type', (table) => {
+    table.increments('id').primary();
+    table.uuid('farm_id').references('farm_id').inTable('farm').defaultTo(null);
+    table.string('type').defaultTo(null);
+    table.string('type_key').defaultTo(null);
+    table.boolean('deleted').notNullable().defaultTo(false);
+    table.string('created_by_user_id').references('user_id').inTable('users');
+    table.string('updated_by_user_id').references('user_id').inTable('users');
+    table.dateTime('created_at').notNullable();
+    table.dateTime('updated_at').notNullable();
```
antsgar, line 26: We could add a check constraint to this table to ensure that either `type` or `type_key` are filled in. I think it would be something like this:

`table.check('?? is not null or ?? is not null', ['type', 'type_key']);`

Knex docs for check constraints https://knexjs.org/guide/schema-builder.html#checks

antsgar, line 26: Should `created_at` and `updated_at` default to now?

antsgar (commented):
Left a few comments but it looks great 🎉 go first Livestock PR!

packages/api/db/migration/20240110204504_create_animal_type_table.js
```diff
@@ -0,0 +1,77 @@
...
+  for (const typeKey of defaultTypeKeys) {
+    await knex('animal_type').insert({
+      farm_id: null,
+      type: null, // only provide for user-created types (i.e. without translation keys)
+      type_key: typeKey,
+      deleted: false,
+      created_by_user_id: '1',
+      updated_by_user_id: '1',
+      created_at: new Date('2000/1/1').toISOString(),
+      updated_at: new Date('2000/1/1').toISOString(),
```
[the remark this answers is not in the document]
kathyavini: This is the timestamp Kevin has requested for all system-defined (vs user-created) entries.

packages/api/db/migration/20240110204504_create_animal_type_table.js
kathyavini: Okay awesome thank you, I will add that!

packages/api/db/migration/20240110204504_create_animal_type_table.js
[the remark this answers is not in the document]
antsgar: Cool, thank you!

packages/api/db/migration/20240110204504_create_animal_type_table.js
kathyavini: I'm not sure -- I see we have a mix in our existing migrations between the timestamp above and `now` and no default. Since the model provides timestamps for user inserts, I wonder if the system timestamp is actually the best default? But I don't love that we have no consistency between tables on this.

antsgar: I think system timestamps make sense come to think of it! I'll update my PR to match that

packages/api/src/controllers/animalTypeController.js
```diff
@@ -0,0 +1,39 @@
...
+ *  LiteFarm is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ */
+
+import AnimalTypeModel from '../models/animalTypeModel.js';
+
+const animalTypeController = {
+  getFarmAnimalType() {
```
antsgar, line 19: NIT, but I wonder if we should call this one `getFarmAnimalTypes` in plural since it fetches all of them?

packages/api/db/migration/20240110204504_create_animal_type_table.js
kathyavini: Okay I'll do that too 😄

antsgar (dismissed):

packages/api/tests/animal_type.test.js
```diff
@@ -0,0 +1,127 @@
...
+
+  afterAll(async (done) => {
+    await tableCleanup(knex);
+    await knex.destroy();
+    done();
+  });
+
+  // GET TESTS
+  describe('GET animal type tests', () => {
+    test('All users should get animal type by farm id (or null)', async () => {
```
antsgar, line 94: Working on tests super similar to these ones on my branch, I realized in this test we would be creating a custom animal type, but there aren't any default types necessarily, so we wouldn't be truly asserting that the types with farm ID null are actually being returned. In my PR I added something to populate a default type, maybe we could leverage that here too and ensure the response is the right length?

kathyavini: Thank you for writing those! I did run it locally with the defaults still present but of course yes, in CI there is no guarantee they are still there and I let it be only out of laziness 🙂 I do wish we had a better `tableCleanup` that didn't destroy the migration-created defaults.

kathyavini:
Forgot my comment when closing PR 🙂 

Closed in favour of the two-tables approach here https://github.com/LiteFarmOrg/LiteFarm/pull/3078

PR #2882, opened 2023-09-20
LF-3651 Revenue type table unique violation error on auto-incrementing primary key

kathyavini:
**Description**

This fixes the database error seen on first insert(s) into the new `revenue_type` table due to the auto-incrementing primary key which was ignored in favour of insertion.

This PR **amends the migration file**. Affected databases (beta, developer local environments) should be rolled back to before when this migration was run (likely 1 or two 2 batches -- please see your own database's `knex_migrations` table) and then run again once this PR is merged. You will lose your existing custom revenue types (on beta none exist yet as the API to create them has not been merged).

Jira link: https://lite-farm.atlassian.net/browse/LF-3651

**Type of change**

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] UI components visually reviewed on desktop view
- [ ] UI components visually reviewed on mobile view
- [x] Other (please explain)

Was written and tested on local database with all authors @Duncan-Brain and @SayakaOno and @kathyavini 

**Checklist:**

- [ ] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [x] The precommit and linting ran successfully
- [ ] I have added or updated language tags for text that's part of the UI
- [ ] I have added "MISSING" for all new language tags to languages I don't speak
- [ ] I have added [the GNU General Public License](https://lite-farm.atlassian.net/l/cp/BT0Dd7WW) to all new files

PR #3076, opened 2024-01-11
LF-3975 getting a list of available breeds for an animal type

antsgar:
**Description**

- Added `animal_breed` table and endpoint to fetch all animal breeds available to a farm.
- This PR points to the branch in #3074 since breeds are linked to animal types.

Jira link:
https://lite-farm.atlassian.net/browse/LF-3975

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

Run migration and checked local database and run unit tests.

**Checklist:**

### DIFF
--- a/packages/api/db/migration/20240111141747_create_animal_breed_table.js
+++ b/packages/api/db/migration/20240111141747_create_animal_breed_table.js
@@ -23,8 +23,8 @@ export const up = async function (knex) {
     table.boolean('deleted').notNullable().defaultTo(false);
     table.string('created_by_user_id').references('user_id').inTable('users');
     table.string('updated_by_user_id').references('user_id').inTable('users');
-    table.dateTime('created_at').notNullable().defaultTo(knex.fn.now());
-    table.dateTime('updated_at').notNullable().defaultTo(knex.fn.now());
+    table.dateTime('created_at').notNullable();
+    table.dateTime('updated_at').notNullable();
     table.check('?? is not null or ?? is not null', ['breed', 'breed_key']);
   });
 
--- a/packages/api/db/migration/20240111141747_create_animal_breed_table.js
+++ b/packages/api/db/migration/20240111141747_create_animal_breed_table.js
@@ -1,0 +1,103 @@
+/*
+ *  Copyright (c) 2024 LiteFarm.org
+ *  This file is part of LiteFarm.
+ *
+ *  LiteFarm is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  LiteFarm is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ */
+
+export const up = async function (knex) {
+  await knex.schema.createTable('animal_breed', (table) => {
+    table.increments('id').primary();
+    table.uuid('farm_id').references('farm_id').inTable('farm').defaultTo(null);
+    table.integer('type_id').references('id').inTable('animal_type').notNullable();
+    table.string('breed').defaultTo(null);
+    table.string('breed_key').defaultTo(null);
+    table.boolean('deleted').notNullable().defaultTo(false);
+    table.string('created_by_user_id').references('user_id').inTable('users');
+    table.string('updated_by_user_id').references('user_id').inTable('users');
+    table.dateTime('created_at').notNullable().defaultTo(knex.fn.now());
+    table.dateTime('updated_at').notNullable().defaultTo(knex.fn.now());
+    table.check('?? is not null or ?? is not null', ['breed', 'breed_key']);
+  });
+
+  // Add initial default breeds
+  const defaultBreedKeys = [
+    {
+      typeId: 1,
+      breedKeys: ['ANGUS', 'HEREFORD', 'CAROLAIS'],
+    },
+    {
+      typeId: 2,
+      breedKeys: ['YORKSHIRE_LARGE_WHITE', 'LANDRACE', 'DUROC'],
+    },
+    {
+      typeId: 3,
+      breedKeys: ['CORNISH_CROSS', 'ROSS_308', 'COBB_500'],
+    },
+    {
+      typeId: 4,
+      breedKeys: ['LEGHORN', 'RHODE_ISLAND_RED', 'PLYMOUTH_ROCK'],
+    },
+  ];
+
+  const rows = [];
+
+  defaultBreedKeys.forEach(({ typeId, breedKeys }) =>
+    breedKeys.forEach((breedKey) =>
+      rows.push({
+        farm_id: null,
+        type_id: typeId,
+        breed: null, // only provide for user-created types (i.e. without translation keys)
+        breed_key: breedKey,
+        deleted: false,
+        created_by_user_id: '1',
+        updated_by_user_id: '1',
+        created_at: new Date('2000/1/1').toISOString(),
+        updated_at: new Date('2000/1/1').toISOString(),
+      }),
+    ),
+  );
+
+  await knex('animal_breed').insert(rows);
+
+  // Add  permissions
+  await knex('permissions').insert([
+    { permission_id: 147, name: 'add:animal_breeds', description: 'add animal breeds' },
+    { permission_id: 148, name: 'delete:animal_breeds', description: 'delete animal breeds' },
+    { permission_id: 149, name: 'edit:animal_breeds', description: 'edit animal breeds' },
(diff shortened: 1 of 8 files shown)


packages/api/db/migration/20240111141747_create_animal_breed_table.js
```diff
@@ -0,0 +1,103 @@
...
+ *
+ *  LiteFarm is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ */
+
+export const up = async function (knex) {
+  await knex.schema.createTable('animal_breed', (table) => {
+    table.increments('id').primary();
```
Duncan-Brain, line 18: I know I suggested the increments as a primary key. Just wondering where we landed on the increments per farm. so making the primars key unique to the farm, and mayyyybe the breed as well . (ie many with id = 1, but only one id = 1 && farm_id = xxxx)

Duncan-Brain: Also wondering what you thought of renaming to `animal_breed_id`, id makes sense for the context of table but I am guessing the reason we are explicit is we usually work with multiple tables and if they all used 'id' in the context of the table it might get confusing.

packages/api/db/migration/20240111141747_create_animal_breed_table.js
```diff
@@ -0,0 +1,103 @@
...
+    table.string('updated_by_user_id').references('user_id').inTable('users');
+    table.dateTime('created_at').notNullable();
+    table.dateTime('updated_at').notNullable();
+    table.check('?? is not null or ?? is not null', ['breed', 'breed_key']);
+  });
+
+  // Add initial default breeds
+  const defaultBreedKeys = [
+    {
+      typeId: 1,
```
Duncan-Brain, line 34: I am concerned with using the index here versus searching for "CATTLE" key or something. Sometimes if a migration fails or if the entries are deleted, the table PK index count remains. 

If in the other PR the typeID is specifically added to make it the same as these type ids that is also fine. But right now it looks like they are added on a loop with the index taking whatever is next.

kathyavini: Yes they have to be added in a loop using the auto-increment and not explicitly because we learned last time that not doing so breaks the table and causes the key unique violation error -- remember those fun times? 😄 https://github.com/LiteFarmOrg/LiteFarm/pull/2882

Duncan-Brain: Oh I remember I am always pro letting the auto increment do its thing! So searching for 'CATTLE' and 'farm_id = null' here would be best practice then here I think just in case something goes wrong with exact numbers.

packages/api/src/models/animalBreedModel.js
```diff
@@ -0,0 +1,48 @@
...
+    return {
+      type: 'object',
+      required: ['farm_id', 'type_id', 'breed'],
+
+      properties: {
+        id: { type: 'integer' },
+        farm_id: { type: 'string' },
+        type_id: { type: 'integer' },
+        breed: { type: 'string' },
+        // breed_key will not be provided for user-generated types and should default to null
```
Duncan-Brain, line 40: Even if null should be added below as part of the schema.

packages/api/src/routes/animalBreedRoute.js
```diff
@@ -0,0 +1,24 @@
...
+ *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ */
+
+import express from 'express';
+
+const router = express.Router();
+import AnimalBreedController from '../controllers/animalBreedController.js';
+import checkScope from '../middleware/acl/checkScope.js';
+
+router.get('/', checkScope(['get:animal_breeds']), AnimalBreedController.getFarmAnimalBreeds());
```
Duncan-Brain, line 22: Usually we use the hasFarmAccess() to make sure the user can only get the breeds they have access to.

Duncan-Brain: Looking at it now the proper middleware is probably authFarmId instead of hasFarmAccess.

packages/api/src/controllers/animalBreedController.js
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+import AnimalBreedModel from '../models/animalBreedModel.js';
+
+const animalBreedController = {
+  getFarmAnimalBreeds() {
+    return async (req, res) => {
+      try {
+        const { farm_id } = req.headers;
+        const { type_id } = req.query;
```
Duncan-Brain, line 23: Curious about this decision to use query. We typically use param in url. 

Found this article that didn't find any best practices per se but had a point that query might be good for optional parameters and that path parameters were good for when they have subtrees.
https://stackoverflow.com/questions/4024271/rest-api-best-practices-where-to-put-parameters

Parent stack issue: https://stackoverflow.com/questions/25385559/rest-api-best-practices-args-in-query-string-vs-in-request-body

Duncan-Brain (changes requested):
Left a couple points about the db, route, and controller architecture.

Otherwise tests pass, and using postman I successfully return the object with a query string. If a required parameter is not sent should we send a 400 bad request? Without a query string I get an error which is expected but I am not sure desireable.
Screenshot 2024-01-11 at 5 59 45 PM

packages/api/db/migration/20240111141747_create_animal_breed_table.js
antsgar: The key set to autoincrement would still be unique per farm, so IMo I'd just let Postgres handle it, as long as they don't repeat each other it's all good. I'm not sure how we'd handle making them unique per farm, would we use the farm ID as some sort of seed? Seems a bit overcomplicated IMO.

For the ids, I was hesitant about it but that is the pattern that we follow on other tables and I can see how it could be useful when doing joins to avoid having to rename the columns, so I'll update it!

packages/api/db/migration/20240111141747_create_animal_breed_table.js
antsgar: It's a good point, it did cross my mind but I wasn't sure. I'll do that 🙂

packages/api/src/controllers/animalBreedController.js
antsgar: I want to chat about this a little bit more tomorrow in tech daily, but yes, essentially both would be correct. The reason I ended up leaning towards a query param is I'm not sure if we'll ever need to fetch all breeds for all types for a farm at some point. If our URL reflects the nested parameters, such as `animal_type/1/animal_breed`, there wouldn't be a way to fetch without filtering by animal type. I don't know if we'll ever have a use case for that, but with that in mind I think the query params provide a little bit more flexibility, allowing us to filter by the parent entity or not depending on what we need.

antsgar:
@Duncan-Brain oh great catch, a test should be added for this case too! No, in that case the endpoint shouldn't return an error, it should return all breeds for that farm regardless of the animal type they belong to. Will fix!

packages/api/src/models/animalBreedModel.js
kathyavini: That's my bad and I'll change it in the other one too! I thought setting it to `null` would force having to pass an explicit `null` from the controller, but that is not the case.

packages/api/db/migration/20240111141747_create_animal_breed_table.js
Duncan-Brain: Primary key: I guess I am stuck on the concept of farm specific increments like how Jira/ Github Issue starts off as 1 for every organization. Its true it doesn't apply to all situations -- short urls is nice tho. But I think its maybe easy to make a composite key https://stackoverflow.com/a/56744896 . If I think harder it is a half solution to ideal anyways so yah it doesn't matter... but maybe the perfect ideal result would be readable URLs like `/animal_breed/cow/ `vs `/animal_breed/22/ `

ID: Yah I like the way you have it better I am trying to understand why maybe it was done the way it is and if consistency is ok if there is a good enough reason

antsgar: Yeah I think the key here is there's a difference in what's shown to the user and what's used as a primary key for the table -- Jira has a company identifier (like "LF") and a ticket number which they extract from the URL and use to lookup the right ticket int their database. This doesn't mean though that they don't have an internal ID that is used as a primary key. We could definitely do a more user friendly URL if we want to, but I would add extra fields to do that rather than use our primary key ID.

Reusing the same id for different farms by using a composite key with the farm ID could end up being error prone as we'd need to look at the farm ID to distinguish one record from the other, and use both fields to lookup a record.

Duncan-Brain: Good point, yah unfortunately we do use primary key on FE url.

antsgar:
Closing this one in favor of two tables approach #3080

PR #3077, opened 2024-01-12
LF-3985 Introducing a new animal type to the catalogue

kathyavini:
**Description**

This adds a straightforward POST with conflict error check + tests, both most closely modelled after `revenue_type`. It only applies to custom types.

Jira link: https://lite-farm.atlassian.net/browse/LF-3985

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

Tested with Postman with a few different auth combinations. Tests again modelled after `revenue_type`.

**Checklist:**


packages/api/src/controllers/animalTypeController.js
```diff
@@ -34,6 +37,43 @@ const animalTypeController = {
...
+          return res.status(409).send();
+        } else {
+          const result = await baseController.postWithResponse(
+            AnimalTypeModel,
+            { type, farm_id },
+            req,
+            {
+              trx,
+            },
+          );
```
antsgar, line 65: To simplify this logic, maybe we could start giving knex's `onConflict` a try? https://knexjs.org/guide/query-builder.html#onconflict

Essentially we can tell it to insert except if certain attributes match in which case we want it to ignore and do nothing.

kathyavini: I think makes sense for when our goal is upsert (like the code you flagged in the notification badge PR https://github.com/LiteFarmOrg/LiteFarm/pull/3054) but don't we want the conflict response returned (rather than a silent fail) in this case?

I'm not sure because we also haven't really consumed this error status on the frontend -- usually we've aimed to give a validation error before we even hit the API in the case of existing records -- so this would be more about what you see when using the API directly.

antsgar: That's a good point! I do like that the API response is more explicit in this scenario and indicates the conflict, so I'm on board with keeping it this way. The rest looks great to me, I haven't tested it out yet but I'll do that in a bit!

antsgar (approved):

packages/api/src/controllers/customAnimalTypeController.js
```diff
@@ -30,6 +33,43 @@ const customAnimalTypeController = {
       }
     };
   },
+
+  addCustomAnimalType() {
+    return async (req, res) => {
+      const trx = await transaction.start(Model.knex());
+      try {
+        const { farm_id } = req.headers;
+        const { type } = req.body;
```
Duncan-Brain, line 42: Maybe we don't need this... but what do you think about trimming strings for trailing spaces? We could do it on the FE but we do have the API technically :p .

packages/api/src/controllers/customAnimalTypeController.js
```diff
@@ -30,6 +33,43 @@ const customAnimalTypeController = {
       }
     };
   },
+
+  addCustomAnimalType() {
```
Duncan-Brain, line 37: I saw in Anto's add animals PR that she was immediately allowing the possibility of adding an array of animals.

Should this allow an array of types? I am just thinking about how the designs show you can add multiple types at once.
Screenshot 2024-01-22 at 11 33 48 AM

Duncan-Brain (commented):
@kathyavini Sorry it took me a while to review, I did not see I was a reviewer for some reason.

Tests pass, endpoint works as designed. 

Added two comments: adding multiple types at once and string manipulation

Happy to approve and merge -- but I think there is a chance you might agree on the adding multiple types at once comment. So I will just wait until you reply yes or no on that so someone else doesn't merge it.

packages/api/src/controllers/customAnimalTypeController.js
antsgar: I don't think the UI will allow adding multiple types at once. You can definitely add multiple animals at once as reflected in this screenshot, but for types I think the way it's going to work is the "Animal type" dropdown is actually going to be an autocomplete, and you're going to be able to type in something that doesn't exist and click on "Add new" or something similar. So you'd add them one by one

antsgar:
@kathyavini I just realized that as of now, if someone creates a new type there'll be no associated breed to pick from. Depending on which PR gets merged first between this one and #3080, we should add something in the controller that adds a row for "Unknown" breed for the new type when it's created

packages/api/src/controllers/customAnimalTypeController.js
kathyavini: I like that as a rule but not necessarily as a one-off; I think translation strings were the only place we had been running `trim()` on user input so far, so I could see it as replacement for that in these tables that don't have the translation string column.

Were you thinking a new guideline or something more programatic like a change to `postWithResponse()`?

Duncan-Brain: Uhm, I think I was thinking like Gursimrans function formatTranslationKey() as some some of update to our new style. Up to you but it seems possibly quite useful.

Duncan-Brain (dismissed):

kathyavini:
@antsgar I thought I followed the requirement for `unknown`, but then writing the logic out I wasn't sure I did anymore 😂 

We could insert an entry into `custom_animal_breed` with the same `farm_id` and `custom_type_id` as the animal type being created here, but it's neither user-created nor a straight string (`UNKNOWN` should be a translation key) so it doesn't really fit within `breed`. I guess logically it more belongs in the `default_animal_breed` table since we are providing it to the user... but that one doesn't have an entry for `custom_type_id` only `default_type_id`!

We can always do a little check on `breed` on the frontend to turn `UNKNOWN` into a key, but then it seems weird to have anything in the db in the first place. The `custom_animal_breed` table would be a mix of true custom and auto-generated `UNKNOWN` records 🤔

antsgar (approved):

packages/api/src/controllers/customAnimalTypeController.js
```diff
@@ -39,7 +39,8 @@ const customAnimalTypeController = {
       const trx = await transaction.start(Model.knex());
       try {
         const { farm_id } = req.headers;
-        const { type } = req.body;
+        let { type } = req.body;
+        type = type.trim();
```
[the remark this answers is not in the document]
kathyavini: Oh you also want to enforce it's not empty? Is that what you're doing for the other POST controller?

antsgar (dismissed):

packages/api/src/controllers/customAnimalTypeController.js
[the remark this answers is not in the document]
kathyavini: @Duncan-Brain do you prefer it abstracted? Would you want to use it your controller?

```js
  /**
   * Checks if the input is a string and not just white space.
   * Returns a trimmed version of the input if it's a valid string, or null
   *
   * @param {string} input - The input string to validate and trim.
   * @return {string | null} - The trimmed string if valid, otherwise null.
   */
  checkAndTrimString(input) {
    if (typeof input !== 'string' || !input.trim()) {
      return null;
    }
    return input.trim();
  },
  ```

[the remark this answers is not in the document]
Duncan-Brain: I am good with any way really haha! Just asking that we probably don't want empty strings in the db is all I am thinking. I havent gotten to that part in the post yet I decided to review others PRs as first priority.

If I were to really really abstract it haha... it would probably take an array of objects to check the null status of and handle the the rollback and error. But thats probably overkill -- the null or empty string check is quite fine and very clear.

Duncan-Brain (approved):
Nice looks great!

antsgar (approved):

PR #3078, opened 2024-01-13
LF-3972b Getting a list of available animal types (two tables)

kathyavini:
**Description**

As discussed today in tech daily, here is a re-write of PR https://github.com/LiteFarmOrg/LiteFarm/pull/3074 but separating the default and custom animal type functionalities for evaluation 🙂 

Jira link: https://lite-farm.atlassian.net/browse/LF-3972

**Type of change**

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] UI components visually reviewed on desktop view
- [ ] UI components visually reviewed on mobile view
- [x] Other (please explain)

As before: Postman and new API tests. Note the `custom_animal_type` tests are basically unchanged from the previous tests.

**Checklist:**

- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] The precommit and linting ran successfully
- [ ] I have added or updated language tags for text that's part of the UI
- [ ] I have added "MISSING" for all new language tags to languages I don't speak
- [x] I have added [the GNU General Public License](https://lite-farm.atlassian.net/l/cp/BT0Dd7WW) to all new files

PR #3080, opened 2024-01-15
LF-3975 getting a list of available breeds for an animal type (two tables)

antsgar:
**Description**

V2 of #3076 with separate tables for default and custom types. This is based off of Joyce's PR with the same approach for types #3078.

Jira link:

**Type of change**

**How Has This Been Tested?**

- [x] Passes test case
- [] Other (please explain)

Run migration locally and run tests.

**Note: If you run any of the "one table approach" PRs locally, before checking out to this branch rollback the migrations previously applied (the files were deleted in the new PRs).**

**Checklist:**


packages/api/tests/custom_animal_breed.test.js
```diff
@@ -0,0 +1,184 @@
...
+        const secondBreed = await makeCustomAnimalBreed(mainFarm);
+
+        const res = await getRequestAsPromise({
+          user_id: user.user_id,
+          farm_id: mainFarm.farm_id,
+          query_params_string: `custom_type_id=${secondBreed.custom_type_id}`,
+        });
+
+        expect(res.status).toBe(200);
+        // Should return breed with default type only
```
SayakaOno, line 142: "custom" type only?

SayakaOno (commented):
Looks good! I left some questions :)

packages/api/src/controllers/customAnimalBreedController.js
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+import CustomAnimalBreedModel from '../models/customAnimalBreedModel.js';
+
+const customAnimalBreedController = {
+  getCustomAnimalBreeds() {
+    return async (req, res) => {
+      try {
+        const { farm_id } = req.headers;
+        const { default_type_id, custom_type_id } = req.query;
```
Duncan-Brain, line 23: Should this check that at least 1 is null?

packages/api/db/migration/20240115182150_create_animal_breed_tables.js
```diff
@@ -0,0 +1,109 @@
...
+    },
+    {
+      typeKey: 'CHICKEN_LAYERS',
+      breedKeys: ['LEGHORN', 'RHODE_ISLAND_RED', 'PLYMOUTH_ROCK'],
+    },
+  ];
+
+  const rows = [];
+
+  for await (const entry of defaultBreedKeys) {
```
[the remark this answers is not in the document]
antsgar: Actually I don't think it's needed, ty! Removing

packages/api/src/controllers/defaultAnimalBreedController.js
```diff
@@ -0,0 +1,39 @@
...
+  getDefaultAnimalBreeds() {
+    return async (req, res) => {
+      try {
+        const { default_type_id } = req.query;
+        const rows = await DefaultAnimalBreedModel.query().modify((queryBuilder) => {
+          if (default_type_id) {
+            queryBuilder.where('default_type_id', default_type_id);
+          }
+        });
+        return res.status(200).send(rows);
```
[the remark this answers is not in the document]
antsgar: To me a 404 is more "the resource couldn't be found" -- in this case default animal breeds are an existing resource, it's just with that specific filter applied there aren't any matches. If we were sending the type ID through query params then I'd be more inclined towards a 404, but with the route specified as is I think the empty array is better

packages/api/tests/custom_animal_breed.test.js
antsgar: Yes, thank you! Copy paste 😅

packages/api/src/controllers/customAnimalBreedController.js
antsgar: Yup, good call!

packages/api/db/migration/20240115182150_create_animal_breed_tables.js
```diff
@@ -0,0 +1,109 @@
...
+
+  await knex.schema.createTable('custom_animal_breed', (table) => {
+    table.increments('id').primary();
+    table.uuid('farm_id').notNullable().references('farm_id').inTable('farm');
+    table.integer('default_type_id').references('id').inTable('default_animal_type');
+    table.integer('custom_type_id').references('id').inTable('custom_animal_type');
+    table.string('breed').notNullable();
+    table.boolean('deleted').notNullable().defaultTo(false);
+    table.string('created_by_user_id').references('user_id').inTable('users');
+    table.string('updated_by_user_id').references('user_id').inTable('users');
```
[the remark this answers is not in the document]
antsgar: Maybe! @kathyavini I was kinda mimicking your PR here, is there any reason you made them nullable?

packages/api/db/migration/20240115182150_create_animal_breed_tables.js
```diff
@@ -0,0 +1,109 @@
...
+export const up = async function (knex) {
+  await knex.schema.createTable('default_animal_breed', (table) => {
+    table.increments('id').primary();
+    table.integer('default_type_id').notNullable().references('id').inTable('default_animal_type');
+    table.string('key').notNullable();
+  });
+
+  await knex.schema.createTable('custom_animal_breed', (table) => {
+    table.increments('id').primary();
+    table.uuid('farm_id').notNullable().references('farm_id').inTable('farm');
```
Duncan-Brain, line 25: You mentioned wanting to check this against the types table custom types?

packages/api/db/migration/20240115182150_create_animal_breed_tables.js
```diff
@@ -0,0 +1,109 @@
...
+    table.integer('default_type_id').references('id').inTable('default_animal_type');
+    table.integer('custom_type_id').references('id').inTable('custom_animal_type');
+    table.string('breed').notNullable();
+    table.boolean('deleted').notNullable().defaultTo(false);
+    table.string('created_by_user_id').references('user_id').inTable('users');
+    table.string('updated_by_user_id').references('user_id').inTable('users');
+    table.dateTime('created_at').notNullable().defaultTo(new Date('2000/1/1').toISOString());
+    table.dateTime('updated_at').notNullable().defaultTo(new Date('2000/1/1').toISOString());
+    table.check(
```
Duncan-Brain, line 34: I think this check still allows both to be filled which also shouldnt be allowed.

packages/api/src/controllers/defaultAnimalBreedController.js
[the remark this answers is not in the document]
SayakaOno: I think I understand.
"If we were sending the type ID through query params" - how would this be different from the current one `/default_animal_breeds?default_type_id=4`?

packages/api/db/migration/20240115182150_create_animal_breed_tables.js
antsgar: I think when I mentioned that I hadn't realized that custom breeds would need a farm ID because they can belong either to a default or custom type (if they always belonged to default they would just take on the farm ID of the type). Unfortunately I don't think there's any way to enforce this with constraints, the only way would be to add a trigger after inserting the record and deleting it if it doesn't fulfill the criteria.

packages/api/db/migration/20240115182150_create_animal_breed_tables.js
antsgar: Added an extra constraint!

packages/api/db/migration/20240115182150_create_animal_breed_tables.js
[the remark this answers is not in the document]
kathyavini: @antsgar I absolutely just copied them from another migration where they were defined like this!

Looks like have the above version, 

`table.string('created_by_user_id').references('user_id').inTable('users').defaultTo('1');`

and

`table.string('created_by_user_id').references('user_id').inTable('users').notNullable();`

to pick from as codebase precedents. The version used above is the most common, but my personal preference is actually `.defaultTo('1');` 🤷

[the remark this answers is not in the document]
antsgar: We could set it as not nullable and also default to 1, right? I'm good with that!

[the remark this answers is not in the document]
Duncan-Brain: @antsgar and @kathyavini should the type table be altered to reflect this too on this pr?

Duncan-Brain (approved):
Everything looks good to me. The new checks are working for me. 

I approved but two small comments:
- the created_by user should the type table be updated here too to reflect the new default to 1 strategy.

- for whatever reason my task.test.js file is no longer passing -- wondering if anyone else is seeing it or just me are we opening too many trx connections in the mocks creations or something? 
Screenshot 2024-01-18 at 2 29 25 PM

packages/api/db/migration/20240115182150_create_animal_breed_tables.js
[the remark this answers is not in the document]
kathyavini: Personally I don't think that's worth a migration, as there are plenty of tables that use the same definition for those two values that `animal_type` currently has. I think we are just agreeing that we prefer this schema moving forward?

packages/api/src/controllers/defaultAnimalBreedController.js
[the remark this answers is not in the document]
antsgar: Sorry I meant to say if we had the type ID as a regular param in the URL, instead of a query param!

antsgar:
@Duncan-Brain yes, good call, I'll update the type tables too! Not sure what you mean by this one? 

> Edit deleted tests comment

Duncan-Brain:
Ah I just originally posted an issue with tests, but it was just some local bug with postgres having too many clients. You can see in tech daily.

SayakaOno (dismissed):

packages/api/db/migration/20240115182150_create_animal_breed_tables.js
```diff
@@ -0,0 +1,130 @@
...
+      typeKey: 'CHICKEN_LAYERS',
+      breedKeys: ['LEGHORN', 'RHODE_ISLAND_RED', 'PLYMOUTH_ROCK'],
+    },
+  ];
+
+  const rows = [];
+
+  for (const entry of defaultBreedKeys) {
+    const { typeKey, breedKeys } = entry;
+    const { id: typeId } = await knex('default_animal_type').where('key', typeKey).first();
```
Duncan-Brain, line 86: I think this has been changed to just id as I am getting a migration error: 
Screenshot 2024-01-22 at 11 55 39 AM

packages/api/tests/testEnvironment.js
```diff
@@ -115,6 +115,8 @@ async function tableCleanup(knex) {
     DELETE FROM "pesticide";
     DELETE FROM "task_type";
     DELETE FROM "farmDataSchedule";
+    DELETE FROM "default_animal_breed";
```
Duncan-Brain, line 118: So I just ran into an Error created from #3078 . Because the defaults are deleted from the table at the end the testing db migration fails when referencing the types. 

I think either we do not want to be deleting the defaults after the test is complete or we want to populate the db afterward similar to (but not beforeAll) populateTaskTypes. 

Screenshot 2024-01-22 at 11 55 39 AM

To reproduce this would be Run the tests after running LF-3972b migration. Then try to run this migration on testing db after the table has been cleaned up.

packages/api/db/migration/20240115182150_create_animal_breed_tables.js
Duncan-Brain, line 86: This line in the migration is failing for me on testing:db I wrote the reason for it below in the testEnvironment file.

packages/api/tests/testEnvironment.js
antsgar: The way I structured the tests I actually don't want any defaults populated since the tests themselves take care of populating default data as they need, so I think I'll just adjust the migration so that it won't insert the breed if it doesn't find the type

Duncan-Brain: So yeah I don't really mind what ever you want to do in tests. I just don't think I like the idea of skipping if it doesn't exist. It is a useful error if for some reason you misspelled the custom types here compared to Joyces.  I would want to see a failed migration.

I don't see my scenario as one off but a consequence of testing in order. I am a little surprised more of us didn't run into it. So personally if it is between leaving it as is and having someone like me who tested it in order rollback their db or adding the special case -- I think I would prefer to leave it as is and I just rollback my db and not run tests again on each branch until it is all merged into integration as one.

Duncan-Brain (dismissed):
Approving to leave the migration as is!

Question about adding in the UNKNOWN --- should you also add in MIXED then? lol 😂 They may know the parent breeds but not have a name for the hybrid

packages/api/src/models/customAnimalBreedModel.js
```diff
@@ -0,0 +1,54 @@
...
+
+  // Optional JSON schema. This is not the database schema! Nothing is generated
+  // based on this. This is only used for validation. Whenever a model instance
+  // is created it is checked against this schema. http://json-schema.org/.
+  static get jsonSchema() {
+    return {
+      type: 'object',
+      oneOf: [
+        {
+          required: ['farm_id', 'default_type_id', 'key'],
```
Duncan-Brain, line 35: sorry just noticed this working on LF-3986 -- should key be 'breed' instead ?

Duncan-Brain (dismissed):

SayakaOno (approved):

PR #3083, opened 2024-01-17
LF-3978 creating one or more individual animals

antsgar:
**Description**

- Created table for individual animals as well as supporting tables with enum.
- Added POST endpoint to create a new animal.
- Added GET endpoint to fetch the list of animals for a farm.

Jira link:
https://lite-farm.atlassian.net/browse/LF-3978
https://lite-farm.atlassian.net/browse/LF-3993

**Type of change**

**How Has This Been Tested?**

- [x] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/api/db/migration/20240116175706_create_animal_tables.js
```diff
@@ -0,0 +1,189 @@
...
+    table.string('name');
+    table.date('birth_date');
+    table.string('identifier');
+    table.integer('identifier_color_id').references('id').inTable('animal_identifier_color');
+    table
+      .integer('identifier_placement_id')
+      .references('id')
+      .inTable('animal_identifier_placement');
+    table.integer('origin_id').references('id').inTable('animal_origin');
+    table.string('dam');
```
Duncan-Brain, line 64: Should this self reference an animal id?

packages/api/db/migration/20240116175706_create_animal_tables.js
```diff
@@ -0,0 +1,189 @@
...
+  });
+
+  await knex.schema.createTable('animal', (table) => {
+    table.increments('id').primary();
+    table.uuid('farm_id').notNullable().references('farm_id').inTable('farm');
+    table.integer('default_breed_id').references('id').inTable('default_animal_breed');
+    table.integer('custom_breed_id').references('id').inTable('custom_animal_breed');
+    table.integer('sex_id').references('id').inTable('animal_sex');
+    table.string('name');
+    table.date('birth_date');
```
Duncan-Brain, line 56: I am wondering if we should really avoid dates (vs datetimetz). Not necessarily because it is wrong but because we don't use them consistently in our app and the timezone info is probably relevant if we manipulate anything in the controller. 

We have some more funky functions in server.js for dealing with dates so if we go that route and keep dates - I think we would want to avoid the existing date pattern. (issue somewhat related - https://lite-farm.atlassian.net/browse/LF-3396)

packages/api/src/controllers/animalController.js
```diff
@@ -0,0 +1,78 @@
...
+        const { farm_id } = req.headers;
+        const result = [];
+
+        if (!Array.isArray(req.body)) {
+          return res.status(400).send('Request body should be an array');
+        }
+
+        for (const animal of req.body) {
+          if (!animal.identifier && !animal.name) {
+            return res.status(400).send('Should send either animal name or identifier');
```
Duncan-Brain, line 49: Going to need a trx.rollback() here and anywhere else you return unsuccessfully.

packages/api/src/controllers/animalController.js
```diff
@@ -0,0 +1,78 @@
...
+      } catch (error) {
+        console.error(error);
+        return res.status(500).json({
+          error,
+        });
+      }
+    };
+  },
+
+  addAnimal() {
```
Duncan-Brain, line 36: should this be plural?

packages/api/src/controllers/animalController.js
```diff
@@ -0,0 +1,78 @@
...
+import { Model, transaction } from 'objection';
+import AnimalModel from '../models/animalModel.js';
+import baseController from './baseController.js';
+
+const animalController = {
+  getFarmAnimals() {
+    return async (req, res) => {
+      try {
+        const { farm_id } = req.headers;
+        const rows = await AnimalModel.query().where({ farm_id });
```
Duncan-Brain, line 25: Are we waiting to add query parameters?

packages/api/src/controllers/animalController.js
```diff
@@ -0,0 +1,78 @@
...
+        if (!Array.isArray(req.body)) {
+          return res.status(400).send('Request body should be an array');
+        }
+
+        for (const animal of req.body) {
+          if (!animal.identifier && !animal.name) {
+            return res.status(400).send('Should send either animal name or identifier');
+          }
+
+          if (!animal.default_breed_id && !animal.custom_breed_id) {
```
Duncan-Brain, line 52: Should we also check that the custom_breed_id belongs to that farm?

packages/api/src/controllers/animalController.js
```diff
@@ -0,0 +1,78 @@
...
+            return res.status(400).send('Should send either default_breed_id or custom_breed_id');
+          }
+
+          const individualAnimalResult = await baseController.postWithResponse(
+            AnimalModel,
+            { ...animal, farm_id },
```
Duncan-Brain, line 58: You did this correctly -- but if for some reason farm_id came before ...animal then it would be possible to override farm_id and I could add animals to another persons farm accidentally or on purpose. 

There is some risk that some developer who copies this pattern in the future does not know/remember that -- it could be prudent to either only spread specific animal properties or check the body against farm_id.

Duncan-Brain (changes requested):
Added a few comments inline about 
- trx's,
- dates
- nullables 
- extra checks

Couple other things I was wondering:
1. Normally we create models at the same time as we make the table and it seems a few Models are missing for the attributes key tables that we will probably need to run get's on in the future.
2. My tests are still failing for too many clients error when I run the whole test suite including animals.test.js -- on their own animals.test.js pass -- I really wonder whether something has changed for backend tests lately or if its my local db that is struggling for some reason.

packages/api/db/migration/20240116175706_create_animal_tables.js
antsgar: Nope, I believe after discussing in the design file it was decided that since the parents of the animal could be animals not registered in the inventory of the farm, for now we'll keep these two fields free text to make things simpler

packages/api/db/migration/20240116175706_create_animal_tables.js
```diff
@@ -0,0 +1,189 @@
...
+    table.string('key').notNullable();
+  });
+
+  await knex.schema.createTable('animal', (table) => {
+    table.increments('id').primary();
+    table.uuid('farm_id').notNullable().references('farm_id').inTable('farm');
+    table.integer('default_breed_id').references('id').inTable('default_animal_breed');
+    table.integer('custom_breed_id').references('id').inTable('custom_animal_breed');
+    table.integer('sex_id').references('id').inTable('animal_sex');
+    table.string('name');
```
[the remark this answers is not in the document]
antsgar: I think the result is pretty much the same and since the default is nullable I assume that's the case if I don't see a .notNullable, but I don't mind adding .nullable if it makes readability easier

packages/api/db/migration/20240116175706_create_animal_tables.js
antsgar: I was unsure about that too and had actually started with datetime and then changed it. It's quite possible the UI might mess up the dates because of timezones, so I'll update this

packages/api/src/controllers/animalController.js
antsgar: True, thank you for catching this!

packages/api/src/controllers/animalController.js
antsgar: I don't think we'll need query options for this one for now, the use case I could think of for query options in this scenario would be filtering, but since we need to pre-fetch everything upfront anyway for offline purposes we'll end up doing the filtering on the client side. For creation there's no offline support so we can fetch only breeds for the type they've selected in the form. That being said, we can add query options for completeness if we want to have that tool at hand in case we end up needing it.

packages/api/src/controllers/animalController.js
antsgar: Yes, good call!

packages/api/src/controllers/animalController.js
antsgar: Yeah, I could just strip farm_id off of the animal object before spreading to make extra sure

antsgar:
@Duncan-Brain I missed the fact that we're going to have to query these enum tables from the UI to show the available options in the dropdowns, so I'll actually have to add routes, controllers and models for all of them

packages/api/db/migration/20240116175706_create_animal_tables.js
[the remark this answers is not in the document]
Duncan-Brain: Its definitely the same -- I think we have just kept it explicit in the past so thats definitely my preference. Helpful for thinking each one through and which (null or not) it should be.

packages/api/src/controllers/animalController.js
Duncan-Brain: Whatever you think is best! Just something I noticed.

packages/api/src/controllers/animalController.js
Duncan-Brain: It might be easier to specify what we want. There could be other table columns we don't want the user to specify in their body req to addAnimals. And instead let the logic do its thing.

Not sure this is possible but user_id, deleted, id etc.  <<- I havent tested that yet though

packages/api/src/controllers/animalIdentifierPlacementController.js
```diff
@@ -0,0 +1,39 @@
...
+ *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ */
+
+import AnimalIdentifierPlacement from '../models/animalIdentifierPlacementModel.js';
+
+const animalIdentifierPlacementController = {
+  getAnimalIdentifierPlacements() {
+    return async (req, res) => {
+      try {
+        const { default_type_id } = req.query;
```
Duncan-Brain, line 22: Just clarifying it won't be possible to assign an identifier placement to custom_type_id

antsgar: Yeah, it's not too clear yet how we will handle identifier placement for custom types (there's an ongoing Slack convo) so for now I've left that out here entirely

packages/api/src/controllers/animalController.js
Duncan-Brain: We might also want the check the other constraint if both id types exist on one animal :( -- otherwise you get a model validation error which is much less user friendly. 

I wish we had TS on our backend because these checks can get out of hand on how many are needed haha

Duncan-Brain (changes requested):
Tests still pass, and new models and controllers look good to me!

Really just 
- one trx.rollback missing that I can see,
-  the overwrite risk of spreading animals
- (less important as try catch should work but) the unhandled error from the other constraint check for if (default_id && custom_id)

antsgar:
@Duncan-Brain I've added the check in the last item, and for the second item I had already added `delete animal.farm_id` so there shouldn't be any risk with spreading anymore. Can you point out where the missing trx.rollback would need to be?

packages/api/src/controllers/animalController.js
```diff
@@ -0,0 +1,78 @@
...
+          );
+
+          result.push(individualAnimalResult);
+        }
+
+        await trx.commit();
+        return res.status(201).send(result);
+      } catch (error) {
+        console.error(error);
+        return res.status(500).json({
```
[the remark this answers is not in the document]
Duncan-Brain: @antsgar I think it is right here for the rollback

packages/api/src/controllers/animalController.js
Duncan-Brain: @antsgar Yes you switched the farm_id so that solves cross farm issues. 

But the principle applies to the baseProperties too. So I could add a created_by_user_id="someones_id_I_want_to_frame_bad_data" or deleted=true on creation. 

I might be wrong on the importance here but seems like something we want the context and defaults to handle.

SayakaOno (commented):
That was a lot of tables, great job!

For breed unknown, I am a little confused about what was discussed [here](https://litefarm.slack.com/archives/C05B9KJLJ82/p1706030964515469).
`breed` was nullable but is not now?

packages/api/db/migration/20240116175706_create_animal_tables.js
```diff
@@ -0,0 +1,206 @@
...
+    table
+      .integer('identifier_placement_id')
+      .references('id')
+      .inTable('animal_identifier_placement');
+    table.integer('origin_id').references('id').inTable('animal_origin');
+    table.string('dam');
+    table.string('sire');
+    table.dateTime('brought_in_date');
+    table.dateTime('weaning_date');
+    table.text('notes');
```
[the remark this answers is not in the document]
antsgar: I used `text` because it doesn't have a set character limit, we could also go with `string` but I think we'd probably need to increase the 256 character limit in that case depending on what we want that notes field to look like in the UI

packages/api/src/controllers/animalController.js
```diff
@@ -0,0 +1,115 @@
...
+import baseController from './baseController.js';
+import DefaultAnimalBreedModel from '../models/defaultAnimalBreedModel.js';
+import CustomAnimalBreedModel from '../models/customAnimalBreedModel.js';
+
+const animalController = {
+  getFarmAnimals() {
+    return async (req, res) => {
+      try {
+        const { farm_id } = req.headers;
+        const rows = await AnimalModel.query().where({ farm_id });
```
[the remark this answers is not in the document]
antsgar: Probably not, but I wasn't sure since we're not filtering deleted records out for types and breeds. @kathyavini did you have any thoughts on this, or should we actually add that condition on all controllers?

packages/api/src/controllers/animalController.js
```diff
@@ -0,0 +1,115 @@
...
+  },
+
+  addAnimals() {
+    return async (req, res) => {
+      try {
+        const trx = await transaction.start(Model.knex());
+        const { farm_id } = req.headers;
+        const result = [];
+
+        if (!Array.isArray(req.body)) {
```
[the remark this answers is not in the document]
antsgar: If array is empty the code in line 51 won't execute since the loop in line 50 will iterate over the empty array

packages/api/src/controllers/animalController.js
```diff
@@ -0,0 +1,115 @@
...
+            const customBreed = await CustomAnimalBreedModel.query()
+              .whereNotDeleted()
+              .findById(animal.custom_breed_id);
+
+            if (!customBreed || customBreed.farm_id !== farm_id) {
+              trx.rollback();
+              return res.status(400).send('custom_breed_id has invalid value');
+            }
+          }
+
```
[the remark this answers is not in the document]
antsgar: Yeah, we should probably add it but I'm not entirely sure yet how we will handle identifier placement for custom breeds, so I'd rather keep this a little bit more flexible for now

packages/api/src/controllers/animalController.js
```diff
@@ -0,0 +1,115 @@
...
+      } catch (error) {
+        console.error(error);
+        return res.status(500).json({
+          error,
+        });
+      }
+    };
+  },
+
+  addAnimals() {
```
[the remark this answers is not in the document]
antsgar: That's true, we might even want that for creating them from the UI so that not all work is lost if there's an error on one animal, although I imagine that'd be less likely to happen from the UI. We can definitely improve this logic when we work on the CSV upload, it's a great observation!

packages/api/tests/animal.test.js
```diff
@@ -0,0 +1,281 @@
...
+      {
+        promisedUser: [user],
+        promisedFarm: [mainFarm],
+      },
+      fakeUserFarm(role),
+    );
+    return { mainFarm, user };
+  }
+
+  async function makeAnimal(mainFarm, properties) {
```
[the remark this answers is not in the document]
antsgar: I inverted the logic so that the default will be a default 😅 but also added a comment

antsgar:
@SayakaOno re unknown breeds -- we landed on the breed actually being nullable since it's an optional field, I had forgotten to remove the constraint from the migration! Let me know if it makes sense now. I've also addressed your other comments, too

SayakaOno:
@antsgar Maybe I'm missing something, but without breed_ids, how do we know what animal types they are?

packages/api/src/controllers/animalController.js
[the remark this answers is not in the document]
SayakaOno: I need to re-learn coding 😇 
Would it be okay to return `[]` with 201 in that case?

packages/api/src/controllers/animalController.js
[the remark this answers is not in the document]
kathyavini: Sorry I read this in the morning and then forgot! Yes, all controllers should probably have it 🙂 

I think we still haven't definitively decided whether we are excluding deleted entities here, or again overriding the hidden properties to filter on the frontend -- and maybe RTK Query will impact on which of those we should do? -- but `.whereNotDeleted()` is a good default to start with and we can always adjust later.

antsgar:
@SayakaOno thank you so much for catching that! I think I ended up confusing myself and this was the reason I had originally gone for non nullable breeds -- in my original design, there's no type, we're inferring the type from the breed and we'd have an "Unknown" breed option, but since we're making breeds nullable we need a type column in animals too. I've updated the PR, let me know your thoughts

packages/api/src/controllers/animalController.js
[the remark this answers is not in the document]
antsgar: Yeah, I think a 201 is okay in that case since the request is the right shape, but it contains no actual new data

packages/api/src/controllers/animalController.js
```diff
@@ -65,12 +74,32 @@ const animalController = {
...
+            );
+
+            if (!defaultType) {
+              trx.rollback();
+              return res.status(400).send('default_type_id has invalid value');
+            }
+          }
+
+          if (animal.custom_type_id) {
+            const customType = await CustomAnimalTypeModel.query().findById(animal.default_type_id);
```
SayakaOno, line 90: This should be `.findById(animal.custom_type_id);`, right? I wonder why this doesn't affect any tests...

At a quick glance, the changes look good overall! I will spend some more time later on!

packages/api/src/controllers/animalController.js
```diff
@@ -0,0 +1,151 @@
...
+          if (animal.custom_breed_id) {
+            const customBreed = await CustomAnimalBreedModel.query()
+              .whereNotDeleted()
+              .findById(animal.custom_breed_id);
+
+            if (
+              !customBreed ||
+              customBreed.farm_id !== farm_id ||
+              customBreed.default_type_id !== animal.default_type_id ||
+              customBreed.custom_type_id !== animal.custom_type_id
```
SayakaOno, line 118: These could compare `null` and `undefined`, and an error is returned unexpectedly.
I was trying to create an animal with a custom type and a custom breed and didn't send `default_type_id`. `customBreed.default_type_id` was `null`, but `animal.default_type_id` was `undefined`, so I got a 404. `default_type_id` (or `custom_type_id`) null should be included in the request body?

antsgar: It shouldn't! Great catch, I've updated the logic and modified the test to assert it

Duncan-Brain (approved):

packages/api/tests/animal.test.js
```diff
@@ -0,0 +1,334 @@
...
+
+        const res = await postRequestAsPromise(
+          {
+            user_id: user.user_id,
+            farm_id: mainFarm.farm_id,
+          },
+          [firstAnimal, secondAnimal, thirdAnimal],
+        );
+
+        console.log(res.error);
```
SayakaOno, line 186: One more console.log here :)

antsgar: 😅

packages/api/src/controllers/animalController.js
antsgar: Ohhh this is interesting, so apparently this is caused by the animal having both a name and an identifier and that breaking the "oneOf" in the model. I changed it to "anyOf" now and it seems to be working fine, can you test again please?

SayakaOno (approved):
It worked! 🎊

Duncan-Brain (approved):
woo!

PR #3089, opened 2024-01-23
LF-3798: Clicking the back button on the SSO account creation page takes you to the welcome screen

SayakaOno:
**Description**

**What was happening?**
Users are expected to go through these paths:
`/` (loading status `true`) -> `/sso_signup_information` -> `/farm_selection` (loading status `true` -> `false`) -- redirect --> `/welcome` -> `/add_farm`
An account is created when leaving `/`, and the user is considered "logged in" at this point. If the user goes back to `/` by clicking the browser button, they are redirected to `/welcome` (because they are logged in but have no farms). The next page is `/add_farm`, and since the user didn't go to `/farm_selection` where the loading status is updated to `false`, the loading status is still `true` in the `/add_farm` page, and the button reads "Loading..."

============================================================================

Redirect the user that created an account with Google to the welcome page instead of the user creation page, and remove `onLoadingUserFarmsStart()` that starts loading that is not stopped or used properly. Also, the userFarm API is expected to not return any farms, but it returns a 404 for that case, so I fixed it as well.

I tested as many cases as I could think of including invitation flows, but everything seemed working fine. Please let me know if I am missing something!

- update `userFarmController`'s `getUserFarmByUserID` not to return `404` when there's no farm to return.
- fix wrong error logs in ChooseFarm saga
- remove `onLoadingUserFarmsStart()` that is called when the user selects their Google account. (the loading status is not being set properly to `false`, and also the loading status is not referred)
- redirect the user that are currently redirected to `/sso_signup_information` to `/welcome`
- delete the route `/sso_signup_information`, component and saga

Jira link: https://lite-farm.atlassian.net/browse/LF-3798

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/webapp/src/containers/GoogleLoginButton/saga.js
```diff
@@ -17,7 +17,6 @@ export const loginWithGoogle = createAction(`loginWithGoogleSaga`);
 
 export function* loginWithGoogleSaga({ payload: google_id_token }) {
   try {
-    yield put(onLoadingUserFarmsStart());
```
kathyavini, line 20: Can the fail reducer (and both imports) also be removed then? Are these maybe even just a copy-paste relic since they are called anyway in the farm selection flow?

And I guess I never really understood the relationship between this issue and the Continue button bug because I never even noticed this line here, and getting rid of that intermediate page still fixed the button bug! Did you figure out how they relate in the end?

kathyavini (commented):
Looks good to me and thank you for the other cleanups!!  🎉 

So I'm not sure why, but I had the incorrect idea (maybe because this is how the invite flow works) that the push change would also fix getting stuck on the Welcome Screen when you use the back button (i.e. so that it returns you to the login screen).

Do you think @SayakaOno (and also @antsgar given your current navigation investigations) that adding this behaviour to the Welcome Screen would be a bad idea?

```js
  useEffect(() => {
    const unlisten = history.listen(() => {
      if (history.action === 'POP') {
        logout();
        history.push('/');
      }
    });

    return () => {
      unlisten();
    };
  }, [history]);
```

I think that's the UX I most like, and I don't _think_ it has repercussions, but obviously I'm not too sure 😬

SayakaOno (commented):
Thank you Joyce!
Being able to go back to the login page feels like an expected behaviour, but I am not sure about logging out the user without them clicking a login button... (there is a low priority bug https://lite-farm.atlassian.net/browse/LF-4030 and they might be able to be tackled altogether!?)
Getting "stuck" at `/farm_selection` after entering a password is kind of the same experience, and I feel like I want a consistent solution there, too.
But I'm not sure either! I would like Loïc's inputs 😁

packages/webapp/src/containers/GoogleLoginButton/saga.js
kathyavini: Ohhhhh, that's exactly right!! I did indeed push to `/farm_selection` and not `/welcome` so I unintentionally piggy backed on `getUserFarms` to clean up that loading state! Makes sense.

kathyavini (approved):

antsgar:
Looks good to me, thanks for fixing it! It does feel weird to me to be stuck in the Welcome screen, although I think all of these situations would likely be resolved by adding the logout button to the navbar for those pages (Welcome and Farm selection).

@SayakaOno were you planning on working on adding the birth date / gender fields to the profile page in a follow-up PR?

SayakaOno:
Thank you @antsgar!
Yes, I will create another PR for those fields!

antsgar (approved):

PR #3092, opened 2024-01-25
LF-4037: add default to last name

antsgar:
**Description**

Add default of `''` to `last_name` colum in `users` table.
We're currently not setting a default at the db level and have a fallback of '' on the client when we send out the user information from an invitation. To avoid having to set this fallback in each place where we add a user, I defaulted the column and made the field not required at the model level.

Jira link:
https://lite-farm.atlassian.net/browse/LF-4037

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

Tested bug by creating a Google account without a last name and verifying that I could use it to log into Litefarm.

**Checklist:**

### DIFF
--- a/packages/api/db/migration/20240125135642_add_user_last_name_default.js
+++ b/packages/api/db/migration/20240125135642_add_user_last_name_default.js
@@ -1,0 +1,26 @@
+/*
+ *  Copyright (c) 2024 LiteFarm.org
+ *  This file is part of LiteFarm.
+ *
+ *  LiteFarm is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  LiteFarm is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ */
+
+export const up = async function (knex) {
+  await knex.schema.alterTable('users', (table) => {
+    table.string('last_name').notNullable().defaultTo('').alter();
+  });
+};
+
+export const down = async function (knex) {
+  await knex.schema.alterTable('users', (table) => {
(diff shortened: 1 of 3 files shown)


packages/api/db/migration/20240125135642_add_user_last_name_default.js
```diff
@@ -0,0 +1,26 @@
...
+ *
+ *  LiteFarm is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ */
+
+export const up = async function (knex) {
+  await knex.schema.alterTable('users', (table) => {
+    table.string('last_name').notNullable().defaultTo('').alter();
```
Duncan-Brain, line 18: What existed before was really weird... `notNullable().defaultTo(null)` 😂 .  Almost makes me think it was a typo.

My question is why default to empty string instead of just allowing null with `.nullable()`? It might make it easier to deal with on the frontend but it seems weird to me https://stackoverflow.com/a/44890240

antsgar: Well the not nullable with a null default is not that weird, essentially null is the default if there's no default if that makes sense. Essentially if you set a column to not nullable and you don't set a default, you're expecting that whoever's inserting a new record will always specify that field, which is totally fine if the field is required and that's what you want to enforce.

To the second point, I'd totally prefer to make this nullable, but that seems higher risk especially in the context of a hotfix since there might be places where the client is expecting a string and is not prepared with nullish coalescing to handle the null scenario, so this seems like the safest choice.

kathyavini (approved):
Yikes so the frontend was doing it 😂 

I agree with the comprehensive approach of putting it on the db schema 👍

packages/api/db/migration/20240125135642_add_user_last_name_default.js
Duncan-Brain: Ok I did not realize it was hotfix 👍

Duncan-Brain (approved):
I am just now realizing that we are inconsistent with full name and first last name between our signup and our profile page 😂 ... so bad.

PR #3097, opened 2024-01-28
LF-3872-not-able-to-edit-expense-value

navDhammu:
**Description**

This appears to be only reproducible in certain situations. 

**Editing an expense value that is between 1-9**

To reproduce:
1. Add an expense between $1 - $9 and save.
2. Try to edit the expense by pressing backspace, then entering a new value. The save button will remain disabled not allowing you to submit form. However, if you blur the input field after clearing it before entering a new value then it still works.

**Editing any expense value by first clearing the input at once**

To reproduce:
1. Add an expense with more than 1 digit and save.
2. Clear the input by highlighting the entire number at once and delete. 
3. Try to edit right away without blurring, you will find the save button remains disabled.

What seems to be happening is that when the value becomes empty, it is being set as NaN and for some reason that doesn't trigger the dirty or valid state of the form which doesn't trigger a rerender and the button remains disabled. It works in other cases because the dirty state is changed first before the value becomes empty (or NaN). 

Jira link: https://lite-farm.atlassian.net/browse/LF-3872

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/Finances/PureExpenseDetail/index.jsx
+++ b/packages/webapp/src/components/Finances/PureExpenseDetail/index.jsx
@@ -115,7 +115,7 @@ const PureExpenseDetail = ({
         type={'number'}
         hookFormRegister={register(VALUE, {
           required: true,
-          valueAsNumber: true,
+          setValueAs: (v) => (v === '' ? null : +v),
           min: { value: 0 },
         })}
         currency={useCurrencySymbol()}


packages/webapp/src/components/Finances/PureExpenseDetail/index.jsx
navDhammu, line 1: Just replaced valueAsNumber with setValueAs and ensuring it never becomes NaN by setting empty value to null

Duncan-Brain (approved):
Looks good and works for me! 

It is also triggering validation which is great. I might be interested in making this a reusable function and fixing throughout the app. As I assume it is happening everywhere where that exists.

Slightly wish it could be fixed with valueAsNumber since they have it built in but I cannot find any solutions there that make sense.

navDhammu:
@Duncan-Brain Thanks for the review! 

Yes I think this might be happening in other places where numbers are meant to be edited. But oddly this issue doesn't happen when there are no validation rules on the input. So If you remove the required and the min value and only leave the valueAsNumber in the register options then it works as expected. 

As a side note - I tested the NumberInput I worked on in the other ticket and it doesn't have this issue. So if we're planning on migrating to that then it's safe to say this problem would also be fixed along with it.

antsgar:
@navDhammu thank you for fixing this!

antsgar (approved):

PR #3098, opened 2024-01-29
LF-4038: The app crashes on the home page once the user saves the user profile without preferred language

SayakaOno:
**Description**

When the user with a non-LF language changes their profile without setting a language, `language_preference` becomes `undefined` ~~in the Redux store~~ and the `litefarm_lang` key in local storage is set to undefined in saga, which causes a crash in the home page. (Thank you Joyce!)

https://github.com/LiteFarmOrg/LiteFarm/blob/37d34dfd0fcc623567583f9c1ba48fef9ea84641/packages/webapp/src/containers/WeatherBoard/utils/index.jsx#L28-L34
(⬆️ `lang` being `undefined` causes a crash)

**Changes**
- update `useLanguageOptions` hook to keep `language_preference` as is. Currently the selected option is becoming `null` for a non-LF language.
- add `option` tag to the last name input label.

NOTE:
I realized that the selected gender does not reflect the newly saved language preference, so I created a [ticket](https://lite-farm.atlassian.net/browse/LF-4040)

Jira link: https://lite-farm.atlassian.net/browse/LF-4040

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/Profile/Account/index.jsx
+++ b/packages/webapp/src/components/Profile/Account/index.jsx
@@ -82,6 +82,7 @@ export default function PureAccount({ userFarm, onSubmit, history, isAdmin }) {
           maxLength: { value: 255, message: t('PROFILE.ERROR.LAST_NAME_LENGTH') },
         })}
         errors={errors[userFarmEnum.last_name] && errors[userFarmEnum.last_name].message}
+        optional
       />
       <Input
         label={t('PROFILE.ACCOUNT.EMAIL')}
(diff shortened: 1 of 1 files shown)


kathyavini (approved):
I like this solution a lot, particularly how it stays within the pure component ☺️ 

Just to clarify on the PR description... it's the Redux entity becoming `undefined` here in this component that causes the crash, but only because the `litefarm_lang` key in local storage is set from it in the submit saga, right? I tried to search up if language was ever pulled from Redux itself and I couldn't find any cases outside of this component.

SayakaOno:
@kathyavini Thank you!! I had not checked where the `lang` was coming from. I updated the description :)

antsgar (approved):
Looks good, thank you for fixing it!

PR #3100, opened 2024-01-30
LF-4026 Get Cypress tests passing on current state of integration

kathyavini:
TODO:
- [x] Opening to test running in CI -- passed!!! 🎉 
- [x] Refactor the beforeEach of each spec into a reusable function

**Description**

This is an update of PR #2931, addressing comments in that PR and getting the tests passing on the current state of integration by updating selectors and points of failure/flakiness. It has been running smoothly locally, no flakiness, for several days now so hopefully the same on GitHub!! 🤞 

Notes on running locally within `packages/end-to-end` with `npm run cypress`:
- Because of the spotlight interactions, the tests need to be run on a clean database, the way that they will be run in CI (I have been using a script that drops and re-migrates my db for the purposes of my own testing)
- Similarly, there is a temporal dependency such that the `crops` test needs to be run before the `tasks` test. This will always be the case in CI (they are run sequentially) but I'll make a note in the code because eventually it would be much better if the specs could be run in total isolation
- The root cause for both of those issues is that I couldn't successfully get conditional "interact-if-present" logic working. This would be first on my list for the next iteration of the tests, but for now I let it be because it only is relevant in local use, not CI
- The specs are setup to run through multiple language users (there a variable that controls this and it's well-integrated into the tests) but then the GitHub workflow was hardcoded for the English user. Discussed this with @antsgar and English-only is preferred for now (it will also keep the tests short)

Jira link: https://lite-farm.atlassian.net/browse/LF-4026

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

Run locally from within `packages/end-to/end` with refreshed database using `npm run cypress`

**Checklist:**


packages/end-to-end/cypress/e2e/tasks.js
```diff
@@ -51,88 +68,59 @@ describe('Tasks', () => {
...
     cy.contains('First Field').should('be.visible');
     // eslint-disable-next-line cypress/no-unnecessary-waiting
     cy.wait(500, { log: false });
-    cy.get('[data-cy=map-selectLocation]').click({ force: false });
+    cy.get(Selectors.MAP_SELECT_LOCATION).click({ force: false });
+
+    cy.get(Selectors.ADD_TASK_LOCATION_CONTINUE).should('exist').and('not.be.disabled').click();
+
+    // This screen only present when this test file is run after crops.js
+    cy.get(Selectors.ADD_TASK_CROPS_CONTINUE).should('exist').and('not.be.disabled').click();
```
kathyavini, line 103: This is the cause of the temporal dependency between `crops.js` and `tasks.js`. There is an additional page in this multi-page form when a crop plan exists on the farm.

antsgar: In the future to make this more robust and ensure the tests are independent, we'll likely need to bake in the creation of a crop in the tasks test itself as a part of the test setup before assertions. We'll probably end up abstracting the code that does the crop creation on its own reusable command

packages/end-to-end/cypress/support/utilities.js
```diff
@@ -0,0 +1,45 @@
+/*
```
kathyavini, line 1: What got timeboxed here was writing this function in ts and accepting an array instead of a single additional file -- after wrangling with the `cy.fixture()` promises for a while I left it be. It would be good for those improvements to be made someday 🙂

kathyavini:
@Duncan-Brain hmmm I don't know if I too much extra info there! Mostly it's the exact same as how Mika set it up and I've come to love the pattern. I think the only things I changed were:

- There was a sudden flakiness (or maybe more like sudden failure) on the Farm Selection page so I tried intercepting the network request I was seeing in dev tools and I haven't seen a failed test there since 👍
- There was one hardcoded `cy.wait(500)` on a farm map interaction, so I intercepted and waited instead (not a move but a duplication of what was used in another farm map test)
- There was one intercept of `@markerJsRequest` that I truly don't know the purpose of; by the time I was running the tests this was causing a failure (the `cy.wait` never concluded) so I removed it

packages/end-to-end/cypress/e2e/tasks.js
kathyavini: Ah that's a good idea! I was thinking the spec needed to be more flexible here, but instead the environment could have been made more predictable.

antsgar (approved):
This is really great work! 🙌 we might need to figure out a way to get rid of the explicit waiting at some point if it ends up in flakiness, but for now we move on 🙂 (relatedly, I found this article that I really enjoyed about cy.wait https://filiphric.com/waiting-in-cypress-and-how-to-avoid-it)

Duncan-Brain (approved):
Thanks @kathyavini for the explaining those little changes! 

Might be worth asking Mika about the marker thing or others for the future..... I agree I cant find any use of it in our frontend.

packages/end-to-end/.gitignore
```diff
@@ -0,0 +1 @@
+.nyc_output
```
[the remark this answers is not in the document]
kathyavini: Yes exactly! And running the tests locally regenerates this folder, with the contents a single JSON file containing this 😆 

```json
{}
```

PR #3101, opened 2024-01-30
LF-4002, LF-3998, LF-4028: create animal group tables and GET/POST requests

antsgar:
**Description**

Co-authored with @kathyavini
- Created animal group table and tables relating group to animals and batches.
- Created POST request to add new groups.
- Created GET request to get groups for a farm along with their related batches and animals.

Jira link:
https://lite-farm.atlassian.net/browse/LF-4002
https://lite-farm.atlassian.net/browse/LF-3998
https://lite-farm.atlassian.net/browse/LF-4028

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

Run migration locally.

**Checklist:**

### DIFF
--- a/packages/api/db/migration/20240130190029_create_animal_group_tables.js
+++ b/packages/api/db/migration/20240130190029_create_animal_group_tables.js
@@ -1,0 +1,70 @@
+/*
+ *  Copyright (c) 2024 LiteFarm.org
+ *  This file is part of LiteFarm.
+ *
+ *  LiteFarm is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  LiteFarm is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ */
+
+export const up = async function (knex) {
+  await knex.schema.createTable('animal_group', (table) => {
+    table.increments('id').primary();
+    table.string('farm_id').notNullable();
+    table.string('name').notNullable();
+    table.string('notes');
+  });
+
+  await knex.schema.createTable('animal_group_relationship', (table) => {
+    table.increments('id').primary();
+    table.integer('animal_id').references('id').inTable('animal').notNullable();
+    table.integer('animal_group_id').references('id').inTable('animal_group').notNullable();
+  });
+
+  await knex.schema.createTable('animal_batch_group_relationship', (table) => {
+    table.increments('id').primary();
+    table.integer('animal_batch_id').references('id').inTable('animal_batch').notNullable();
+    table.integer('animal_group_id').references('id').inTable('animal_group').notNullable();
+  });
+
+  // Add  permissions
+  await knex('permissions').insert([
+    { permission_id: 163, name: 'add:animal_groups', description: 'add animal groups' },
+    { permission_id: 164, name: 'delete:animal_groups', description: 'delete animal groups' },
+    { permission_id: 165, name: 'edit:animal_groups', description: 'edit animal groups' },
+    { permission_id: 166, name: 'get:animal_groups', description: 'get animal groups' },
+  ]);
+
+  await knex('rolePermissions').insert([
+    { role_id: 1, permission_id: 163 },
+    { role_id: 2, permission_id: 163 },
+    { role_id: 5, permission_id: 163 },
+    { role_id: 1, permission_id: 164 },
+    { role_id: 2, permission_id: 164 },
+    { role_id: 5, permission_id: 164 },
+    { role_id: 1, permission_id: 165 },
+    { role_id: 2, permission_id: 165 },
+    { role_id: 5, permission_id: 165 },
+    { role_id: 1, permission_id: 166 },
+    { role_id: 2, permission_id: 166 },
+    { role_id: 3, permission_id: 166 },
+    { role_id: 5, permission_id: 166 },
+  ]);
+};
+
+export const down = async function (knex) {
+  const permissions = [163, 164, 165, 166];
+
+  await knex('rolePermissions').whereIn('permission_id', permissions).del();
+  await knex('permissions').whereIn('permission_id', permissions).del();
+
+  await knex.schema.dropTable('animal_group');
+  await knex.schema.dropTable('animal_group_relationship');
+  await knex.schema.dropTable('animal_batch_group_relationship');
+};
--- a/packages/api/src/controllers/animalGroupController.js
+++ b/packages/api/src/controllers/animalGroupController.js
@@ -25,25 +25,17 @@ const animalGroupController = {
     return async (req, res) => {
       try {
         const { farm_id } = req.headers;
-        const groups = await AnimalGroupModel.query().where({ farm_id }).whereNotDeleted();
-        const groupsWithRelatedIds = await Promise.all(
-          groups.map(async (group) => {
-            const animalRelationships = await AnimalGroupRelationshipModel.query().where({
-              animal_group_id: group.id,
-            });
-            const batchRelationships = await AnimalBatchGroupRelationshipModel.query().where({
-              animal_group_id: group.id,
-            });
-            return {
-              ...group,
-              related_animal_ids: animalRelationships.map((relationship) => relationship.animal_id),
-              related_batch_ids: batchRelationships.map(
-                (relationship) => relationship.animal_batch_id,
-              ),
-            };
-          }),
+        const rows = await AnimalGroupModel.query()
+          .where({ farm_id })
+          .whereNotDeleted()
+          .withGraphFetched({ animal_relationships: true, batch_relationships: true });
+        return res.status(200).json(
+          rows.map(({ animal_relationships, batch_relationships, ...rest }) => ({
+            ...rest,
+            related_animal_ids: animal_relationships.map(({ animal_id }) => animal_id),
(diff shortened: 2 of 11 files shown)


packages/api/db/migration/20240130190029_create_animal_group_tables.js
```diff
@@ -0,0 +1,70 @@
...
+  // Add  permissions
+  await knex('permissions').insert([
+    { permission_id: 163, name: 'add:animal_groups', description: 'add animal groups' },
+    { permission_id: 164, name: 'delete:animal_groups', description: 'delete animal groups' },
+    { permission_id: 165, name: 'edit:animal_groups', description: 'edit animal groups' },
+    { permission_id: 166, name: 'get:animal_groups', description: 'get animal groups' },
+  ]);
+
+  await knex('rolePermissions').insert([
+    { role_id: 1, permission_id: 163 },
```
antsgar, line 45: Current last permission number is 158, but I think @Duncan-Brain's PR for batches will need to add 159, 160, 161 & 162

SayakaOno (commented):
Could you remind me how we decided to go with two join tables? I was expecting one table for some reason 😅

kathyavini:
Okay so we're treating it a bit like a join table where we'll delete the whole record when the group is removed? Will that be okay with the historical tracking?

antsgar:
@kathyavini we could have two tables or one, but I think having two keeps things simpler with null constraints and it makes sense since we're already having batches and individual animals stored separately.

Actually I missed the base properties in the `animal_group` table, we shouldn't remove the whole record when the group is removed, just soft delete!

antsgar:
@SayakaOno the difference is the `animal_batch_id` vs the `animal_id` columns which are foreign keys to different tables, so even if we merged the two into one we'd still need to keep two columns to relate to the two separate tables

SayakaOno:
@antsgar My bad, I meant the same columns except for the animal/animal batch id. I was thinking about a similar design like `custom_animal_breed` having to have default_type_id or custom_type_id.

antsgar:
@SayakaOno yeah it'd be totally valid to do that too, although what I like with the separate tables is we don't get all of those records with `null` ids and we don't have to add constraints. It should also be more performant if for example we want to query for all groups an animal belongs to, since there'll be less volume of data to query through. With animals and breeds it's a bit different because an animal can only belong to one breed, so it didn't make much sense to split that relationship up into its own separate table.

SayakaOno:
@antsgar That's a good point. My focus was to query animals and animal batches using group id, so when thiking about the other way, I agree that the two table approach seems better.

packages/api/src/controllers/customAnimalTypeController.js
```diff
@@ -23,7 +23,7 @@ const customAnimalTypeController = {
     return async (req, res) => {
```
kathyavini, line 23: These two changes to custom animal type are unrelated; I just noticed these points in discussions of other animal-related PRs and realized they were missing here.

packages/api/src/controllers/animalGroupController.js
```diff
@@ -0,0 +1,150 @@
...
+          name,
+          farm_id,
+          deleted: false,
+        });
+
+        if (record) {
+          await trx.rollback();
+
+          return res.status(409).send(); // Enforcing uniqueness on group name
+        } else {
```
Duncan-Brain, line 111: Do we need this else block?

Duncan-Brain (dismissed):
Looks good to me tests are passing.

Left a couple small comments.

packages/api/src/controllers/animalGroupController.js
```diff
@@ -0,0 +1,150 @@
...
+        let { name, notes } = req.body;
+        name = baseController.checkAndTrimString(name);
+        notes = baseController.checkAndTrimString(notes);
+
+        if (!name) {
+          await trx.rollback();
+          return res.status(400).send('Group name is required');
+        }
+
+        if (!Array.isArray(related_animal_ids) || !Array.isArray(related_batch_ids)) {
```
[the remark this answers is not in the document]
kathyavini: Correct

packages/api/src/models/animalGroupModel.js
```diff
@@ -0,0 +1,46 @@
...
+      properties: {
+        id: { type: 'integer' },
+        farm_id: { type: 'string' },
+        name: { type: 'string' },
+        notes: { type: ['string', 'null'] },
+        ...this.baseProperties,
+      },
+      additionalProperties: false,
+    };
+  }
```
Duncan-Brain, line 43: Could possibly add the relationMappings here for the relationship tables to avoid inserting multiple models and just upsert using the graph instead. I think both are acceptable just wondering the preference.

kathyavini: Hmmm, I'm interested in that! It seems weird to have models with no associated controllers + routes, but I hate the (un-) readability of the `.withGraphFetched()` syntax. Do you necessarily have to construct one of those graph strings in order to take advantage of `relationMappings`? Do you have an existing example you like?

Duncan-Brain: My understanding is that whole point of the relationMapping is to use graph-like queries/insertions. So you dont have to use it at all but if you want to receive or insert data in one call you can. I will try to use it in the batch requests and see if it is liked pattern.

packages/api/tests/animal_group.test.js
```diff
@@ -0,0 +1,419 @@
...
+  afterAll(async (done) => {
+    await tableCleanup(knex);
+    await knex.destroy();
+    done();
+  });
+
+  // GET TESTS
+  describe('Get animal groups tests', () => {
+    test('All users should get animal groups for their farm', async () => {
+      const roles = [1];
```
SayakaOno, line 137: Why 1 only...?

antsgar: Oops! I was testing this out and wanted it to run faster and forgot to switch it back

packages/api/src/controllers/animalGroupController.js
```diff
@@ -0,0 +1,149 @@
...
+  getFarmAnimalGroups() {
+    return async (req, res) => {
+      try {
+        const { farm_id } = req.headers;
+        const groups = await AnimalGroupModel.query().where({ farm_id }).whereNotDeleted();
+        const groupsWithRelatedIds = await Promise.all(
+          groups.map(async (group) => {
+            const animalRelationships = await AnimalGroupRelationshipModel.query().where({
+              animal_group_id: group.id,
+            });
```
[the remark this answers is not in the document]
antsgar: I added relation mappings to the model now so I'm only fetching animal_id, and added this mapping to the controller 🙂

packages/api/src/models/animalGroupModel.js
antsgar: I added the relationMappings and it made the logic for the GET method a lot easier, thanks for that! Not sure how it works from the insert aspect though

packages/api/tests/animal_group.test.js
```diff
@@ -0,0 +1,419 @@
...
+  afterAll(async (done) => {
+    await tableCleanup(knex);
+    await knex.destroy();
+    done();
+  });
+
+  // GET TESTS
+  describe('Get animal groups tests', () => {
+    test('All users should get animal groups for their farm', async () => {
+      const roles = [1, 2, 5];
```
SayakaOno, line 137: All users, so I think we need 3 as well.

packages/api/src/controllers/animalGroupController.js
```diff
@@ -0,0 +1,141 @@
...
+  getFarmAnimalGroups() {
+    return async (req, res) => {
+      try {
+        const { farm_id } = req.headers;
+        const rows = await AnimalGroupModel.query()
+          .where({ farm_id })
+          .whereNotDeleted()
+          .withGraphFetched({ animal_relationships: true, batch_relationships: true });
+        return res.status(200).json(
+          rows.map(({ animal_relationships, batch_relationships, ...rest }) => ({
```
Duncan-Brain, line 33: I would just plug the idea that when you keep the structure of the graph and related table it can make reading the data as well as the code manipulation much easier.

packages/api/src/models/animalGroupModel.js
```diff
@@ -0,0 +1,72 @@
...
+        filter: (query) => query.select('animal_id'),
+      },
+      batch_relationships: {
+        modelClass: AnimalBatchGroupRelationshipModel,
+        relation: Model.HasManyRelation,
+        join: {
+          from: 'animal_group.id',
+          to: 'animal_batch_group_relationship.animal_group_id',
+        },
+        filter: (query) => query.select('animal_batch_id'),
```
Duncan-Brain, line 66: It seems like filter is an alias for modify -- wondering if we could use modify keyword instead.  I was trying to find examples for this and was having trouble in the docs.

Duncan-Brain (approved):

packages/api/src/controllers/animalGroupController.js
antsgar: Yeah here I still had to do this mapping because the related graph was getting me something like `related_batches: [{ id: 5 }, { id: 7}]` which is fine if I wanted more properties than just an ID but seemed overkill for that

packages/api/src/models/animalGroupModel.js
antsgar: Updated! Yup they're the same thing

SayakaOno (approved):

packages/api/src/controllers/animalGroupController.js
Duncan-Brain: Yeah as long as its a "join like" table thats probably fine. If we put base properties on it in the future though then it makes more sense. 

Eg with deleted and updated and created property we can track when an animal entered the group and left the group.

Duncan-Brain (approved):

PR #3102, opened 2024-01-31
LF-3998 Create POST endpoint to create a new animal group

kathyavini:
TODO:
- [ ] Clarify question noted in controller code: (behaviour when some animal/batch ids are invalid)
- [ ] Requires update where noted once batch model is completed
- [ ] Didn't write any tests yet

**Description**

Jira link: https://lite-farm.atlassian.net/browse/LF-3998
(shape of endpoint described here: https://lite-farm.atlassian.net/browse/LF-3989)

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

HAVE TESTED NOTHING

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/animalGroupController.js
+++ b/packages/api/src/controllers/animalGroupController.js
@@ -27,8 +27,9 @@ const animalGroupController = {
       try {
         const { farm_id } = req.headers;
         const { related_animal_ids, related_batch_ids } = req.body;
-        let { name } = req.body;
+        let { name, notes } = req.body;
         name = baseController.checkAndTrimString(name);
+        notes = baseController.checkAndTrimString(notes);
 
         if (!name) {
           await trx.rollback();
@@ -59,6 +60,7 @@ const animalGroupController = {
         const record = await baseController.existsInTable(trx, AnimalGroupModel, {
           name,
           farm_id,
+          deleted: false,
         });
 
         if (record) {
@@ -68,7 +70,7 @@ const animalGroupController = {
         } else {
           const result = await baseController.postWithResponse(
             AnimalGroupModel,
-            { name, farm_id },
+            { name, notes, farm_id },
             req,
             {
               trx,
--- a/packages/api/src/controllers/animalGroupController.js
+++ b/packages/api/src/controllers/animalGroupController.js
@@ -1,0 +1,112 @@
+/*
+ *  Copyright 2024 LiteFarm.org
+ *  This file is part of LiteFarm.
+ *
+ *  LiteFarm is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  LiteFarm is distributed in the hope that it will be useful,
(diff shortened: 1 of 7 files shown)


packages/api/src/controllers/animalGroupController.js
```diff
@@ -0,0 +1,112 @@
...
+
+        if (!Array.isArray(related_animal_ids) || !Array.isArray(related_batch_ids)) {
+          await trx.rollback();
+          return res.status(400).send('Animal ids and batch ids must be arrays');
+        }
+
+        if (!related_animal_ids.length && !related_batch_ids.length) {
+          await trx.rollback();
+          return res.status(400).send('Either animal ids or batch ids are required');
+        }
```
antsgar, line 46: We might want to allow creation of an empty group depending on how the UI works -- I could imagine a scenario where you can create the group with its name and leave it empty in a sort of "draft" state to add animals and batches later

packages/api/src/controllers/animalGroupController.js
```diff
@@ -0,0 +1,112 @@
...
+
+        // Check ownership of animals
+        for (const animalId of related_animal_ids) {
+          const animal = await AnimalModel.query(trx)
+            .findById(animalId)
+            .where({ farm_id })
+            .whereNotDeleted();
+
+          // QUESTION: do we prefer the whole transaction to fail and an error response if any of the animals in the provided array are invalid (as written here)? Or should they be skipped and the others still inserted?
```
antsgar, line 55: In the individual animal controller, I did the first (everything fails if there's something wrong with one animal). But I think we discussed it with @SayakaOno in my PR and the second option would actually be better, create (in this case add to the group) what we can, and return in the error which entitites/records failed. I didn't pursue that with individual animals because I wanted to move on for the time being, but I think if possible and if there's not a lot of extra effort involved the second option would be preferrable

antsgar (commented):
Looks good so far! 🎉

packages/api/src/controllers/animalGroupController.js
kathyavini: I was about to clarify whether we would to be sending an error code or a success code in this case (I would have thought still sending a 201 for the inserted records would make more sense) but then I Googled it and realized there is an actual response status for this case, mixed error and success! Cool! [207 Multi-Status](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/207).

SayakaOno: I actually thought the all-or-nothing approach was good @antsgar, but when it fails, the client may want all errors depending on what's it for (eg. file upload). Partial success requires more complexed UI, so I prefer to disscuss with Loïc and Kevin before finalizing it!

kathyavini: > when it fails, the client may want all errors depending on what's it for (eg. file upload)

This is a good point and I can definitely make that change (returning the array of all failures instead of the id of the first failure).

packages/api/src/controllers/animalGroupController.js
kathyavini: Done, it will accept both arrays as empty now.

PR #3110, opened 2024-02-05
LF-3946 Show a fallback screen instead of white screen

kathyavini:
**Description**

- Installs the [react-error-boundary](https://github.com/bvaughn/react-error-boundary) package
- Wraps the App + routes with the Error Boundary
- Adds a fallback container which handles the reload + logout behaviour, including sending a tagged error to Sentry (tag is `fallback_user_action`; see [example event here](https://litefarm.sentry.io/issues/4947685358/events/14991a1b5da145b69c24d47c27bf37a3/))
- Adds the pure component + [Storybook](http://localhost:6006/?path=/docs/components-purereacterrorfallback--docs)
- Begins adding the new design system colours to `src/assets/colors.scss`

*Note about the `Failed to import dynamic module` error*:

Sadly I think the "TypeError: Failed to fetch dynamically imported module" error will not trigger the error boundary. I thought this would be considered an error in the `<Routes />` component, but you can see looking [at this Sentry error](https://litefarm.sentry.io/issues/4946614753/?project=5777759) that it is instead a system exception. If you scroll to the bottom and expand the Event Grouping Information, the `By In-App` section (which contains the Component Stack for an error within the React component tree), is empty.

Jira link: https://lite-farm.atlassian.net/browse/LF-3946

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

I triggered what would normally be a whitescreen type error crash in two ways:
- clear the `litefarm_lang` local storage value and click the LF logo (inspired by our recent bug bash!)
- from transaction list, create an expense > view details > edit + save > view details > delete > back button until crash

I also tried to recreate the `Failed to fetch dynamically imported module` error, and as mentioned above, could not trigger the boundary that way. If you can think of any other kinds of known TypeError to test on, that would be great 🙏 

**Checklist:**


packages/webapp/src/stories/ReactErrorFallback/ReactErrorFallback.stories.tsx
```diff
@@ -0,0 +1,62 @@
...
+ */
+
+import { ReactNode } from 'react';
+import { Meta, StoryObj } from '@storybook/react';
+import { componentDecorators } from '../Pages/config/Decorators';
+import { PureReactErrorFallback } from '../../components/ErrorHandler/PureReactErrorFallback';
+import { Main } from '../../components/Typography';
+import styles from './styles.module.scss';
+
+// https://storybook.js.org/docs/writing-stories/typescript
```
SayakaOno, line 23: ❤️

packages/webapp/src/components/ErrorHandler/PureReactErrorFallback/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+            <RefreshIcon className={styles.icon} />
+            {t('ERROR_FALLBACK.RELOAD')}
+          </TextButton>
+          <Main className={styles.or}>{t('common:OR')}</Main>
+          <TextButton className={styles.iconLink} onClick={handleLogout}>
+            <LogoutIcon className={styles.icon} />
+            {t('PROFILE_FLOATER.LOG_OUT')}
+          </TextButton>
+        </div>
+        <Main className={styles.supportText}>
```
SayakaOno, line 66: Should we make the email clickable?

SayakaOno (commented):
Looks great!

I made a `TypeError: Cannot read properties of undefined (reading 'value')` happen by adding `console.log(undefined.value);` and saw the fall back view :)

packages/webapp/src/components/ErrorHandler/PureReactErrorFallback/index.tsx
kathyavini: Yes! Good idea.

SayakaOno (dismissed):

antsgar:
@loicsans tagging you in this one for French translations

packages/webapp/src/components/ErrorHandler/PureReactErrorFallback/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+  handleLogout: () => void;
+}
+
+export const PureReactErrorFallback = ({
+  handleReload,
+  handleLogout,
+}: PureReactErrorFallbackProps) => {
+  const { t } = useTranslation();
+
+  const isDesktop = useIsAboveBreakpoint('(min-width: 600px)');
```
antsgar, line 41: Lately we've been using MUI's breakpoints and their hook, in this case like this

`const isDesktop = useMediaQuery(theme.breakpoints.down('sm'));`

That way we always use the same breakpoints instead of typing in the number in the components. I think there's only one remaining component using `useIsAboveBreakpoint` which could be refactored as well and then we could remove the custom hook

packages/webapp/src/components/ErrorHandler/PureReactErrorFallback/styles.module.scss
```diff
@@ -0,0 +1,149 @@
...
+
+    p {
+      font-size: 14px;
+      letter-spacing: -0.308px;
+    }
+  }
+}
+
+.title {
+  color: #16423d; /* Figma's teal 900 on this page is not is not the app's teal 900 -- this should be checked */
```
antsgar, line 80: The app's var is close enough that we could probably use that one?

antsgar (commented):
Left a minor comment but this looks great! Love the new screens 🙂 and the use of Typescript!

packages/webapp/src/components/ErrorHandler/PureReactErrorFallback/index.tsx
kathyavini: No issues with this component, but for refactoring the last component that uses `useIsAboveBreakpoint`... the breakpoint is `856px`. Is it safe to round up to 900px (MUI breakpoint `md`)? @SayakaOno it is your component so I want to defer to you! This is the floating button on finances.

I think it looks okay at 900 but I want to make sure I'm not missing something 🙂

packages/webapp/src/components/ErrorHandler/PureReactErrorFallback/styles.module.scss
kathyavini: (Discussed today in tech daily but just noting here for GitHub too)...

This colour is part of the new design system; Loïc provided this really helpful [Figma link](https://www.figma.com/file/eMffC0h3JPlWtzNttxJpil/Litefarm-Updated?type=design&node-id=1-60)). Since we're going to move towards using CSS variables according to that convention, we'll eventually move the other way and migrate the old teal over when ready.

packages/webapp/src/components/ErrorHandler/PureReactErrorFallback/index.tsx
SayakaOno: 900px looks fine to me too. It looks like it was originally 768px, and I thought I just had followed Loïc's design, but I couldn't find any breakpoints in the mockups... also, it looks like Anto updated it to 856px [here](https://github.com/LiteFarmOrg/LiteFarm/pull/3053/files).
Setting a higher breakpoint seems safe...

SayakaOno (approved):

PR #3111, opened 2024-02-06
Lf 3792 the sale will no longer be visible once the crop plan will be deleted completed or abandoned

navDhammu:
**Description**

This PR fixes the issue where revenue for a crop sale does not show the crop when the crop management plan for that crop is not active or planned. This can happen when the user deletes, abandons or completes a plan after generating a revenue.

Changes:
- Added a selector to select all management plans 
- Modified useSelector inside useCropSaleInputs to include management plans for crop(s) associated with that particular sale irrespective of their management plan status. 

Jira link: https://lite-farm.atlassian.net/browse/LF-3792

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/containers/Finances/useCropSaleInputs.jsx
+++ b/packages/webapp/src/containers/Finances/useCropSaleInputs.jsx
@@ -28,8 +28,12 @@ import { STATUS } from '../../components/Forms/GeneralRevenue/constants';
 import { useTranslation } from 'react-i18next';
 import styles from '../../components/Forms/GeneralRevenue/styles.module.scss';
 import { useSelector } from 'react-redux';
-import { currentAndPlannedManagementPlansSelector } from '../managementPlanSlice';
 import { measurementSelector } from '../userFarmSlice';
+import {
+  allManagementPlansSelector,
+  currentAndPlannedManagementPlansSelector,
+} from '../managementPlanSlice';
+import { createSelector } from 'reselect';
 
 /**
  * Reformat the sale data for ease of use with react-hook-forms.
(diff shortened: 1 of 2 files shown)


antsgar:
@navDhammu thank you for working on a fix for this one too! The fix works great, the only issue I see is that since `useCropSaleInputs` is used both for viewing/editing an existing revenue but also creating one, this change will cause a crop that has no current plans to show up when creating a new revenue, which I think isn't intended behavior. I'm thinking we could probably send some parameter to the hook so that it knows whether to include crops from pre-existing sales or not. Alternatively, we could have `useSaleInputs` receive a parameter for the current sale (which would only be passed when editing/viewing) and extract the extra management plan needed from that specific sale.

navDhammu:
@antsgar Thank you reviewing this. As far as I understand this doesn't allow crops that don't have a management plan to show up when creating a revenue. I tested this by marking a crop plan as complete and deleted and checking if it shows up when creating a revenue. 

`useCropSaleInputs` already accepts a sale parameter which is what I used to filter the management plans:

```
const managementPlans = useSelector((state) =>
    selectManagementPlans(state, sale?.crop_variety_sale),
  );
```

When editing, the crops for existing sale would show up, along with crops that have current and planned plans. When adding, only the current and planned ones would show up. I think that's the desired behavior?

Duncan-Brain:
I think it looks correct to me adding the existing sale crop to the options, if none exists (new sale) none is added.

My only comment might be that it could be good to keep the creation of selectors in the slice.

navDhammu:
@Duncan-Brain We can put the selector in the slice but I just kept in the file because it was very specfic to the crop sale and not really reusable. Also I couldn't come up with a name that doesn't sound like a sentence - currentAndPlannedManagementPlansAndPlansWithCropSale?

Duncan-Brain (approved):

navDhammu:
@Duncan-Brain `managementPlanByCropIdSelector` uses managementPlanSelector which filters out deleted plans so it wouldn't work if the user were to delete a crop plan associated with a historical sale. And I think a crop sale can have more than one crop as part of that sale if I'm not wrong?

antsgar:
@navDhammu yup, that's the correct behavior! You're right, I was using a crop for testing that I thought no longer had active plans but it turns out that it did, sorry about that

antsgar (approved):

PR #3115, opened 2024-02-06
LF-4079 Create reusable group pill component

kathyavini:
TODO:
- [x] Confirm the status of the new dark blue with Loïc
  - All the new colours are part of the new design system

**Description**
- This creates the `+ {{count}} others` Hover Pill for use on the Group column of the Animal table
- The hover pill text adapts to count (@antsgar I am leaving gender as a problem to solve the first time we encounter an inconsistency!)
- [Storybook](http://localhost:6006/?path=/story/components-hoverpill--plural) There is a property "position" (property of the Storybook story, not of the component) which allows you to place the hover pill in different corners of the Storybook iFrame to watch the Popper/MUI tooltip adapt to the viewpoint edges. Which isn't even our code anyway so it was a bit pointless to test, but I really wanted to see it working 😂 

Small note: on mobile the distance from anchor to tooltip is further than on Desktop. This is true for the other MUI Tooltip in app (the `<Infoi />` component) as well and I am not sure the reason.

Jira link: https://lite-farm.atlassian.net/browse/LF-4079

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

Only viewed in Storybook UI as not yet present in App.

**Checklist:**


SayakaOno (commented):
Super quick, and looks great!

I wonder how flexible this component should be.
Screenshot 2024-02-06 at 1 26 16 PM
If there is a pill at the bottom of the window and the number is 5 for example, the tooltip will be cut off.

I'm not sure what Loïc's solution will be, but I think it would be nice if the placement changes depending on where in the window the pill is. (it might be easy to implement it with a Mui component https://mui.com/material-ui/react-tooltip/)

kathyavini:
@SayakaOno Oh, that is the kind of thing I would never have considered! 😂 

Thank you for the MUI component link. I'll give it a try and also see if I can fix the story to be better about positioning.

antsgar:
@loicsans some strings to translate here too

packages/webapp/public/locales/fr/translation.json
```diff
@@ -882,6 +882,10 @@
   "HOME": {
     "GREETING": "Bonjour, "
   },
+  "HOVERPILL": {
+    "PLUS_OTHERS_COUNT_one": "autre",
+    "PLUS_OTHERS_COUNT_other": "autres"
```
kathyavini, line 887:
```suggestion
    "PLUS_OTHERS_COUNT_one": "+ {{count}} autre",
    "PLUS_OTHERS_COUNT_other": "+ {{count}} autres"
```

SayakaOno (commented):
Looks great!
The options for the position in the story are very nice! Thank you ❤️ 
I'll approve once the changes are finalized.

packages/webapp/public/locales/fr/translation.json
loicsans: Yep, sorry about that. I hadn't check the English file

packages/webapp/src/components/HoverPill/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+    </>
+  );
+
+  // https://mui.com/material-ui/react-tooltip/#distance-from-anchor
+  const PopperProps = {
+    modifiers: [
+      {
+        name: 'offset',
+        options: {
+          offset: [0, -8],
```
antsgar, line 44: There's something super small that's bothering me 😅 which I think is actually buggy behavior from MUI, but hopefully we can fix it? When you hover over the pill and the tooltip opens, if you then hover back to the pill there's a space within it where hovering doesn't work and the styling turns back to the "unhovered" style. I recorded a video which hopefully shows it better.

https://github.com/LiteFarmOrg/LiteFarm/assets/20230771/357fb51f-ff6b-494a-b683-929fb2d3061f

I couldn't figure out why this was happening, but it turns out that the tooltip has a bottom margin and that bit of it is overlapped with the pill.

Captura de pantalla 2024-02-08 a la(s) 10 05 28


The solution seems to be to apply a `margin: 0` style to the tooltip to override MUI's margin, and then adjust the offset to be a positive number so that there's still some space between the tooltip and the pill. The only problem is that the MUI selector is more specific than the `.hoverDetails` one, so for it to work the custom selector needs to be more specific. One way that I found to do this without using `!important` is to apply the `hoverDetails` styling to MUI's styling for specific tooltip placements, like this:

```
<Tooltip
      classes={{
        tooltip: styles.hoverDetails,
        tooltipPlacementTop: styles.hoverDetails,
      }}
      ...
>

```

kathyavini: Ah, so that's why the offset had to be negative in the first place! These default component styles always get me!! 🤣 

I couldn't get it to work just like you described though (with override classes but without `!important`); could you check again? 😅 And once the overall offset is changed to positive I think we'll need to override classes for more placements (I would think at least `tooltipPlacementBottom`)? If so I think it might be okay (and at least more succinct) to go the `!important!` route...

packages/webapp/src/components/HoverPill/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+  return (
+    <Tooltip
+      title={hoverContent}
+      placement="bottom-end"
+      classes={{
+        tooltip: styles.hoverDetails,
+      }}
+      PopperProps={PopperProps}
+      enterTouchDelay={0}
+      leaveTouchDelay={900000}
```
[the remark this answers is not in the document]
kathyavini: These values come from the other MUI tooltip we have in app (the Infoi icon tooltip). Personally my reaction to the MUI default (that you have to long press and not tap) was to think the hover pill was broken on mobile! I think probably that's how the Infoi got its settings too. But is the long press interaction more standard?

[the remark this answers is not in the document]
antsgar: Ah, I have no idea which one would be more common tbh! But that makes sense to me

packages/webapp/src/components/HoverPill/index.tsx
antsgar: Ah no, you're right, it doesn't work, I had probably done a different change and thought it worked when it didn't. At some point I just repeated the same selector to increase the specificity, like `.hoverDetails.hoverDetails.hoverDetails` but I think anyone reading that would wonder what's going on there probably 😅 so I'm alright with keeping the important. For the record this is one of the things I strongly dislike about MUI, which is how hard it is to override their styling

antsgar (approved):

PR #3122, opened 2024-02-09
LF-4082 RTK Query setup and fetching inventory

kathyavini:
**Description**

This creates the RTK Query `apiSlice.ts` and associated changes. You can see that once that part was done, fetching the inventory into the new container was a piece of cake! ☺️ Right now the inventory displays as a wall of text. Please comment out the tag invalidation line to observe caching.

Reference was @antsgar's wonderful PR #3019 and the only differences were due to the migration to TypeScript.

NOTE: I did not migrate the [`useQueries` custom wrapper](https://github.com/LiteFarmOrg/LiteFarm/pull/3019/files#diff-a5fa4ddca7a025f39751bab0473f53135a043ceb94fb6e3295c222fee59021c3), as I could not figure out how to do so without losing the type safety of the built-in RTK-Query hooks -- that is a TS conversion well beyond my skills (and might even be impossible). See code note below.

Jira link: https://lite-farm.atlassian.net/browse/LF-4082

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/webapp/src/containers/Animals/Inventory/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+function AnimalInventory() {
+  const dispatch = useDispatch();
+
+  // For demonstration only! (remove to observe caching); force refresh via:
+  dispatch(api.util.invalidateTags(['Animals', 'AnimalBatches', 'AnimalGroups']));
+
+  const { data: animals } = useGetAnimalsQuery();
+  const { data: animalBatches } = useGetAnimalBatchesQuery();
+  const { data: animalGroups } = useGetAnimalGroupsQuery();
+
```
kathyavini, line 35: This is the point at which type-checking will be enforced on the returned data based on the interfaces that were provided in `apiSlice.ts`-- it's fun to see by trying to access existent + non-existent properties on each of those data objects.

Although these can be successfully batched with the wrapper function `useQueries` linked above:

```ts
const { data, isLoading } = useQueries([
  { hook: useGetAnimalsQuery, label: 'animals' },
  { hook: useGetAnimalBatchesQuery, label: 'animalBatches' },
  { hook: useGetAnimalGroupsQuery, label: 'animalGroups' },
]);

const { animals, animalBatches, animalGroups } = data;
```

I never figured out how to modify the wrapper to still get that type checking.

packages/webapp/src/containers/userFarmSlice.ts
```diff
@@ -1,46 +1,75 @@
...
+}
+
+interface UserFarm {
+  farm_id: string;
+  user_id: string;
+  units: Units;
+
+  // https://www.typescriptlang.org/glossary#index-signatures
+  [key: string]: any;
+}
```
kathyavini, line 19: I looked a bit more into this and I think the index signature is still the best way to avoid having to write out the whole interface to `UserFarm` (a Partial would be correct as well, but would still require the base interface which I think we want to avoid in these situations where it's not relevant and we aren't trying to type check on it).

packages/webapp/src/apiConfig.js
```diff
@@ -1,5 +1,5 @@
 /*
- *  Copyright 2019-2022 LiteFarm.org
+ *  Copyright 2019-2024 LiteFarm.org
```
Duncan-Brain, line 2: Does the new copyright standard just use the one current year now? https://lite-farm.atlassian.net/wiki/spaces/LITEFARM/pages/1269202970/Copyright+GNU+General+Public+License

SayakaOno (commented):
@kathyavini are the APIs being called over and over again in `/inventory` for you?

packages/webapp/src/apiConfig.js
Duncan-Brain: Well I have been doing it wrong then lol -- I remember Ivan sharing some kind of blog saying 1 year was enough. I think I have actually deleted these intervals in favour of 1 year approach.

packages/webapp/src/store/api/apiSlice.ts
```diff
@@ -0,0 +1,55 @@
+/*
+ *  Copyright 2023-2024 LiteFarm.org
```
SayakaOno, line 2: I think this can be `2024`.

packages/webapp/src/store/api/types/index.ts
```diff
@@ -0,0 +1,60 @@
...
+ *  LiteFarm is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ */
+
+// If we don't necessarily want to type an endpoint
+export type Result = Array<{ [key: string]: any }>;
+
+export interface Animal {
```
SayakaOno, line 19: We need to add `internal_identifier` and `photo_url` to Animal and AnimalBatch ([PR](https://github.com/LiteFarmOrg/LiteFarm/pull/3125)), can it be done now??

packages/webapp/src/store/api/apiSlice.ts
kathyavini: I wanted to give credit to @antsgar for the code she wrote in 2023 (upon which this was based!) 😄 But I guess since it never got merged, this is a bit confusing.

packages/webapp/src/apiConfig.js
kathyavini: Oh no 😂  The original block we were copy-pasting had all the years that LiteFarm was active, and I think Ivan was just asking us to get rid of all the other dates if the file itself hadn't existed back then. I think this method with the range of dates (from creation up to last modification) is pretty standard.

packages/webapp/src/store/api/apiSlice.ts
```diff
@@ -0,0 +1,55 @@
...
+    },
+    responseHandler: 'content-type',
+  }),
+  tagTypes: ['Animals', 'AnimalBatches', 'AnimalGroups'],
+  endpoints: (build) => ({
+    // redux-toolkit.js.org/rtk-query/usage-with-typescript#typing-query-and-mutation-endpoints
+    // <ResultType, QueryArg>
+    getAnimals: build.query<Animal[], void>({
+      query: () => `${animalsUrl}`,
+      providesTags: ['Animals'],
```
Duncan-Brain, line 42: Just wondering are we waiting until we have more endpoints to do specific animal tags? 

To get both individual animal tags and we could do : 
```
(result, error, userFarm) => {
  const tags= result.map((animal) => return {type: 'Animal', id: animal.id});
  tags.push {return type: 'FarmAnimals', id: userFarm.farm_id}
}
```

Just have to pass in useFarm on the query selector.

packages/webapp/src/store/api/apiSlice.ts
```diff
@@ -0,0 +1,55 @@
...
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ */
+
+import { createApi, fetchBaseQuery } from '@reduxjs/toolkit/query/react';
+import { RootState } from '../store';
+import { animalsUrl, animalBatchesUrl, animalGroupsUrl, url } from '../../apiConfig';
+import type { Animal, AnimalBatch, AnimalGroup } from './types';
+
+export const api = createApi({
```
Duncan-Brain, line 21: As mentioned on Antos PR I think this file is going to get huge. I know the redux docs have this pattern for their examples but in the class I took they did it like this:

```
store 
> apis 
    > animalBatchesApi 
    > animalApi
> slices
    > whateverSlice
> thunks
    > whateverThunk
```

This has the benefit of setting the baseQuery to match the endpoints, we can start using the new pattern where we don't have to pass in user_id to the headers for the other routes. And of course the nice pattern of making an redux api for each new table.

packages/webapp/src/store/api/types/index.ts
```diff
@@ -0,0 +1,64 @@
+/*
```
Duncan-Brain, line 1: Just wondering if the redux store is the best location for types. As we start converting our codebase to .ts I get the feeling we will be using types in our containers or components or even backend too.

packages/webapp/src/containers/userFarmSlice.ts
Duncan-Brain: I think this interface could go in your new types file as we scope our data hierarchy. I think its good to write out the whole interface but I also am good with the idea of leaving it for now and maybe using whatever we touch as candidates for the next .ts update. Like a contagion we can create a ticket and say we need to update userFarm to .ts next.

packages/webapp/src/store/api/apiSlice.ts
kathyavini: Oh shoot I meant to ask about this in meeting this morning because I don't quite follow.

The idea is that you can create a new tag and entity (the singular animal) based upon the `/animals` endpoint, without ever having to have an actual `GET` singular animal endpoint? I didn't know that was a pattern, but it sounds interesting.

What I don't follow is the role of `userFarm` though, since the base endpoint was already only for one farm. I would think the nested tag would want to be associated with `animal.id`, not `farm_id`.

packages/webapp/src/containers/userFarmSlice.ts
kathyavini: Yeah, let's talk about that with @antsgar when she returns in terms of how we want to manage the transition over to TS. I would prefer to not write any interfaces that we aren't actively using in code (i.e. where we aren't actually type checking on the properties) because it feels like bloat. So I like the "whatever we touch" approach personally (like the full interface would be written the day we use the `userFarm` selector in a TS file and need the properties), but it might be easier to establish a system now rather than do piecemeal 🤔

packages/webapp/src/store/api/apiSlice.ts
Duncan-Brain: The idea is you can invalidate tags for an individual animal (update, remove). The backend endpoint/query param will still be needed. 

For 'FarmAnimals' tag -- its essentially the same as ['Animals'] you have currently as its goal would be to refresh all animals.  I just chose a structure which is probably not necessary.... The example I have from my class uses users.id to go with something like 'UsersMovies' tag. I was just transposing.

packages/webapp/src/store/api/apiSlice.ts
kathyavini: I think Anto said we will start with one file but almost certainlly move to some sort of chunking. But looking at this closer, I'm not sure about one file-one endpoint (is that what the diagram is indicating?) I guess I would have thought of chunking all the animal endpoints together, although now I realize our API wasn't really set up like that with a common base URL 🤔

Duncan-Brain (approved):
It seems to be working well for me ! I did not see the multiple requests at first but navigating around added it. If you want to keep the demonstration you could make the dispatch into a button 😄 .

I just added a few architecture questions mostly carried over from the template PR. 

My main approval question was about if breed name, and type name should be retrieved here or if there is another PR supposed to do those RTK queries. I think we decided it can go in the "implement table" PR.

packages/webapp/src/store/api/types/index.ts
kathyavini: Discussed in tech daily but want to leave a ping to discuss with @antsgar too.

Yeah I think your idea of having an animal type that we use throughout the frontend (and then extend to get the `apiSlice` interface) makes sense -- and I agree, must avoid interface duplication 👍 👍 👍 Would love to discuss types organization generally as even within the RTK Query docs alone there were several patterns.

packages/webapp/src/store/api/apiSlice.ts
Duncan-Brain: Even all the animal endpoints together could get chunky -- image addOne, removeOne, addMany, fetchOne, fetchMany etc etc. for each of animal, animal batches, animal groups, animal identifiers, animal sexes, breeds, types etc. etc.

With one per endpoint it would mirror our routes/models/tables with probably one clearly defined get, post, and delete and typings for most.

kathyavini: True, true. And yes it does feel particularly appealing to mirror the routes file exactly...!

antsgar (approved):
Looks and works great, so excited to see this one come to life!

SayakaOno:
@kathyavini I was looking at `dispatch(api.util.invalidateTags(['Animals', 'AnimalBatches', 'AnimalGroups']))`, it doesn't seem to refetch when reloading...?
This is off topic, but if a query needs refetching every time on mount, passing `refetchOnMountOrArgChange`(https://redux-toolkit.js.org/rtk-query/api/createApi#refetchonmountorargchange) `true` to the hook may be what we need to do.

kathyavini:
@SayakaOno wait are you saying that dispatching `invalidateTags` in the component creates an infinite loop OR does nothing that all then??! 😓

I like the sound of `refetchOnMountOrArgChange`!! I thought it was something we would need especially while testing, but I think it sounds like a good default always to refetch on page reload.

PR #3125, opened 2024-02-10
LF-4081: Support animal internal identifier

SayakaOno:
**Description**

- migration for `animal` and `animal_batch` tables
  - add `internal_identifier` and `photo_url` columns 
  - remove `name_identifier_check` in `animal` and make `name` nullable in `animal_batch`
  - add unique constraint on `internal_identifier` within the same farm_id
  - add `internal_identifier` to existing records (if it's not necessary this can be removed)
  - create function and triggers to add `internal_identifier` on insertion
- update `baseController` `insertGraph` method to fetch the item before returning (`internal_identifier` etc. would not be returned otherwise)
  - https://vincit.github.io/objection.js/api/query-builder/mutate-methods.html#insertandfetch
  - https://vincit.github.io/objection.js/recipes/returning-tricks.html
- update models and controllers
- update `userFarmFactory` to take roleId (was there a way to create `userFarm` with a desired `role_id`?)
- add/update tests for animal and batch internal_identifier

Jira link: https://lite-farm.atlassian.net/browse/LF-4081

**Type of change**

**How Has This Been Tested?**

- [x] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/api/tests/animal_batch.test.js
```diff
@@ -439,4 +462,28 @@ describe('Animal Batch Tests', () => {
...
   });
+
+  describe('Animal batch model tests', () => {
+    test('Unique internal_identifier should be added within the same farm_id between animal and animal_batch tables', async () => {
+      const [firstFarm] = await mocks.farmFactory();
+      const [secondFarm] = await mocks.farmFactory();
+      const [thirdFarm] = await mocks.farmFactory();
+
+      const testCases = [
+        { farm: firstFarm, isAnimal: false, expectedInternalIdentifier: 1 },
```
kathyavini, line 473: Just a minor thing, but since the test cases are defined separately from the function call `makeAnimalOrBatchForFarm`, maybe a string format would be more immediately readable than an `isAnimal` boolean?

```js
{ farm: firstFarm, kind: 'animal', expectedInternalIdentifier: 1 },
{ farm: firstFarm, kind: 'batch', expectedInternalIdentifier: 2 },

File: packages/api/src/controllers/baseController.js [hunk 2]
```diff
@@ -195,7 +195,8 @@ export default {
     return await model
       .query(trx)
       .context({ user_id: req?.auth?.user_id, ...context })
-      .insertGraph(removeAdditionalPropertiesWithRelations(model, data));
+      .insertGraph(removeAdditionalPropertiesWithRelations(model, data))
+      .returning('*');
```
kathyavini, line 199: Oh! I'm surprised we never had to do this before to get the auto-incrementing integers back.

kathyavini (dismissed):
WOW what a migration!!! 🎉 And the triggers work perfectly, I love it!!!

packages/api/tests/animal_batch.test.js
SayakaOno: I thought about it, but in that function, I would only do `if (kind === 'animal')`, and I would not use 'batch', so I went with `isAnimal`.
But I don't have a strong preference, so I will update it :)
Thank you for reviewing!

packages/api/src/controllers/baseController.js
SayakaOno: To be honest, I have no idea why `id` is returned...😅

kathyavini (approved):

packages/api/tests/animal_batch.test.js
kathyavini: You're right, it's a little more weird in the ternary, but I do like it, thank you!! 😂

packages/api/tests/animal.test.js
```diff
@@ -227,11 +230,12 @@ describe('Animal Tests', () => {
       expect(res.status).toBe(400);
     });
 
-    test('Should not be able to create an animal without name or identifier', async () => {
+    test('Should be able to create an animal without name and identifier', async () => {
```
Duncan-Brain, line 233: Maybe we don't need this test anymore...

packages/api/tests/animal_batch.test.js
```diff
@@ -281,7 +284,7 @@ describe('Animal Batch Tests', () => {
       expect(res.status).toBe(400);
     });
 
-    test('Should not be able to create an animal batch without a name', async () => {
+    test('Should be able to create an animal batch without a name', async () => {
```
Duncan-Brain, line 287: Maybe we don't need this one either..

packages/api/src/controllers/baseController.js
Duncan-Brain, line 199: I am seeing that internal_identifier is returned without this. So I am not sure I understand.

SayakaOno: Could you try sending a post request and checking what's returned?

Duncan-Brain: Yup my bad! I did on /animals but its only used for /animal_batches.

So only comment is that maybe the baseController function should be renamed to `insertGraphWithResponse()` to match the `postWithResponse()`.

Duncan-Brain:
Nice job! This definitely looks a lot tougher than I originally thought it might be. 

As we continue having to compare across batches and animal it definitely makes me wish we found a way to put them in the same table originally. But I really like that the identifier removes some of those cluttered constraints! 

Everything works really great for me and I am happy to approve -- but I figured Anto might want to check it out since it is a new and robust backend pattern! So I thought I might ask some questions about the pattern if we have time.

--

So I was looking through how others solve this issue and it seems like the approach suggested in the Jira ticket is similar to:
https://dba.stackexchange.com/q/161080. In that exchange they mention drawbacks where there can be race conditions if concurrent users try to update the tables. That makes me wonder if we have this same concurrent user risk? Or if the situations are very different.

This exchange mentions using views as a way to overcome that.
https://dba.stackexchange.com/a/135446

Do you know anything more about the concurrency risk? I did not quite understand where it comes from but it says it is "subtle" lol

packages/api/src/controllers/baseController.js
antsgar: Sorry I'm not following! Why do we need all the properties returned here, is this mainly for tests?

antsgar:
@Duncan-Brain that's a great point! So yes, with the current approach and with triggers in general there's a risk of race conditions -- if someone tried to create an animal or batch at the same time as another user within the same farm, the row insertion might fail because the identifier isn't unique. With the scale of users that we have per farm I'm not super worried about that right now, but the concept of using views to fix that is interesting, and it would also be an interesting mechanism to reconcile the info from individual animals and batches in the future for display in the UI (like in the dashboard table)

antsgar (dismissed):
This is great, awesome job going into more complex BE stuff like triggers!! 🙌

packages/api/src/controllers/baseController.js
SayakaOno: I will rename it, thank you!

SayakaOno: I don't think we need all the properties, but we at least need `internal_identifier` which wouldn't be returned without  `.returning('*')` to update the list of batches that we already have on the frontend.
If there's a way to not include unnecessary properties like created_by..., updated_by... etc. maybe we can go with that, but I'm not sure if we are doing it already or we should start doing it...

SayakaOno:
Thank you @Duncan-Brain  and @antsgar !

For concurrency and race condition, if I am not mistaken, if you try to add an animal and a batch within the same farm at the same time, they might get the same `internal_dentifier`. There's a constraint in each table, but we can't add a constraint to check uniqueness between the two tables. (I believe the risk is "subtle" though)
Should we try using `VIEW` now? (thank you for the links Duncan)

Duncan-Brain:
@SayakaOno you are probably asking.. not me..haha but I can approve as-is! It sounds like a very small risk.

But, my personal preference is to try the `VIEW` since 
- we have an existing solution of name or identifier = required which is good enough in my opinion until this is decided
- this pattern will likely be reused / copied and now would be a good time to choose between patterns
- I dislike the possibility of bad prod data even if the risk is very very small -- so if this is the db architecture community recommended solution then I like that

I wanted to see what it looked like so this is not correct but it is a starting place should should someone decide to try: 
```
await knex.schema.createView('v_farm_animal_identifier', function (view) {
      view.columns(['farm_id', 'id', 'identifier_id'])
      view.as(knex('animal_batch').select('farm_id', 'id').rowNumber('identifier_id', 'id', 'farm_id').unionAll(function() {
        this.select('farm_id', 'id').from('animal');
      }));
    });
```

SayakaOno:
@duncan I was asking everyone 😂 I would like to give it a try too.

packages/api/db/migration/20240207214919_create_animal_union_batch_view_modify_animal_batch_tables.js
```diff
@@ -0,0 +1,106 @@
...
+ * @returns { Promise<void> }
+ */
+export const down = async (knex) => {
+  await knex.schema.dropView('animal_union_batch');
+
+  await knex.schema.alterTable('animal', (table) => {
+    table.dropColumn('photo_url');
+  });
+
+  // table.check does not work in alterTable
```
kathyavini, line 95: Do you know why this is by chance? You can get `table.check()` to work -- but only with an empty binding array (something I saw suggested on an open issue on the knex repo):

```js
table.check('name IS NOT NULL OR identifier IS NOT NULL', [], 'name_identifier_check');
 ```
 
This was driving me nuts just now on a migration I was trying to write. Do you know why the ?? interpolation works in `createTable` but not `alterTable`?

**Comment by SayakaOno** [comment]
I created `animal_union_batch` VIEW with `internal_identifier`.
- In addition to rows Anto included in the VIEW, I added `identifier`, `name` and `deleted` rows so that it could be used for the inventory table.
- `created_at` is used to sort the records.
- To join the view, `relationMappings` was added/updated in the models.
- We need to manually attach internal_identifiers after creating animals/batches. So I created a function.

I love `VIEW` so far!!

File: packages/api/db/migration/20240207214919_create_animal_union_batch_view_modify_animal_batch_tables.js (hunk 5 above)
Comment (SayakaOno) [reply packages/api/db/migration/20240207214919_create_animal_union_batch_view_modify_animal_batch_tables.js:95]: Ah! So table.check still works!
I couldn't find any info other than that `alterTable` may have some limitations  😢

**Comment by SayakaOno** [comment]
Now the VIEW only has necessary columns to get `internal_identifier`.
I wonder if creating a table dedicated to `internal_identifier` is the best solution to avoid potential issues...

File: packages/api/db/migration/20240207214919_create_animal_union_batch_internal_id_view_modify_animal_batch_tables.js [hunk 6]
```diff
@@ -0,0 +1,83 @@
...
+      table.string('photo_url');
+      table.dropChecks(['name_identifier_check']);
+    });
+
+    await knex.schema.alterTable('animal_batch', (table) => {
+      table.string('photo_url');
+      table.string('name').nullable().alter();
+    });
+
+    // Create function to assign internal_identifier to a record (NEW) to be inserted
```
kathyavini, line 32: I think I still prefer to see this SQL code timestamped in an `sql/` directory (or `views/` directory) in case we do have to touch it in the future in another migration.

If we start doing that, I don't know -- I'm curious if you have a preference? -- if it would be better to save the raw SQL as JS string to make import trivial, or creating SQL files and using the full `fs.readFileSync(path.join(__dirname`... as in this [knex GitHub thread](https://github.com/knex/knex/issues?q=fs.readFileSync%28path.join%28__dirname%2C+in%3Acomments) but then having SQL in an .sql file, which gives syntax highlighting and maybe some other benefit I'm not considering.

packages/api/db/migration/20240207214919_create_animal_union_batch_internal_id_view_modify_animal_batch_tables.js
```diff
@@ -0,0 +1,83 @@
...
+    await knex.raw(`
+      CREATE VIEW animal_union_batch_internal_identifier AS
```
kathyavini, line 34: Copilot suggested (but it makes sense to me!) that the view might be named with `_view` to make it clearer in the future that this is a view and not a table; e.g. `animal_union_batch_id_view`.

I was thinking of that in particular looking at the model, which has no indication that this is not full table.

packages/api/src/models/animalUnionBatchInternalIdentifierModel.js
```diff
@@ -0,0 +1,24 @@
...
+ *  the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  LiteFarm is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ */
+
+import Model from './baseFormatModel.js';
```
kathyavini, line 16: As mentioned above, it would be nice if there was some way (and I think changing the view name is probably best, but at least a code comment here) to make it clear that this is not a table.

packages/api/src/util/animal.js
```diff
@@ -0,0 +1,34 @@
...
+ *  LiteFarm is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ */
+
+import knex from './knex.js';
+
+/**
+ * Assigns internal identifiers to records based on the provided record IDs and kind.
```
kathyavini, line 19: Maybe nit, but should this be called `getInternalIdentifiers` (and the JSDoc comment revised) as identifiers are not being assigned to records at all -- just generated when the view is called up?

kathyavini (commented):
I like the view approach -- I just left a few minor comments about it below -- but I am curious about your dedicated table idea! I guess that's different from the original approach of columns in each table respectively? So it would be like just making this view concrete?

packages/api/src/util/animal.js
SayakaOno: Oh sorry, I modified this function and forgot to update the JSDoc.

Identifiers are assigned on line 31: `record.internal_identifier = internalIdentifier`, so after this function is run, the records that were passed will have identifiers.
Initial records:
```
[{id: 1, name: "name"}, {id: 2, name: "name2"}]
```
Updated records:
```
[{id: 1, name: "name", internal_identifier: 1}, {id: 2, name: "name2", internal_identifier: 2}]
```

SayakaOno (commented):
@kathyavini thank you for your suggestions, I will do some research and improve the VIEW approach.

For the table idea, you are correct. I'm picturing something like this:
- a new table with columns ("id" (foreign keys: animal.id, animal_batch.id), "batch" (bool), internal_identifier)
- make `internal_identifier` unique within `farm_id` (oh no, would we need to duplicate `farm_id`? 😅 I just realized that this approach may not be a good idea) 
- add `internal_identifier` from controllers or using triggers

I believe the only risk would be that the request will fail when attempting to add the same id at the same time, and it's better than having duplicated keys (the first trigger approach) or the potential inconsistency of `internal_identifier` (this approach).

packages/api/src/util/animal.js
kathyavini: Oh okay, I think we just had different definitions of `record` in mind. I thought a record was specifically a database term (an entry in a table), so a record update makes me think that a database update operation was performed.

But it sounds like here, a record is just a data object (or in context, the result object about to be sent to the client). Which I should have noticed, since it is the exact variable name you are using! 🙂 If that's clear to everyone else, that's fine with me too.

SayakaOno:
@kathyavini I think I've addressed all of your comments. It would be great if you could re-review when you have the time, thank you!

packages/api/src/util/animal.js
```diff
@@ -16,14 +16,15 @@
 import knex from './knex.js';
 
 /**
- * Assigns internal identifiers to records based on the provided record IDs and kind.
- * @param {Array<Object>} records - The array of records to which internal identifiers will be assigned. Each record is expected to contain an 'id' property.
- * @param {string} kind - The kind of records being processed ('batch' or other).
+ * Assigns internal identifiers to records.
+ * @param {Array<Object>} records - The array of animals or animal batches to which internal identifiers will be assigned.
+ *                                  Each record is expected to contain an 'id' property.
```
kathyavini, line 21: Extra whitespace?

packages/api/db/sql/20240207214919_animal_union_batch_id_view.sql
```diff
@@ -0,0 +1,25 @@
+CREATE VIEW animal_union_batch_id_view AS
```
kathyavini, line 1: Probably needs license as well

kathyavini (approved):
It looks good! I like the separate SQL and that we will be able to diff it now ☺️ 

We never discussed this, but I imagine we will add the license to SQL code as well?

packages/api/db/sql/20240207214919_animal_union_batch_id_view.sql
SayakaOno: ChatGPT said "Adding the GNU General Public License (GPL) or any other software license directly to SQL files isn't a common practice", so I didn't add it. (but I didn't do any further research)
Do you think we can just add it?

kathyavini: I'll defer to @antsgar to decide for us! 😄

PR #3126, opened 2024-02-12
LF-3566 add common name, varietal, and cultivar to edit varietal screen

navDhammu:
**Description**

- Added missing varietal and cultivar input fields and changed label for variety name to common name.
- Did minor refactoring by moving field names out of the component and into an object.

Jira link: https://lite-farm.atlassian.net/browse/LF-3566

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/EditCropVariety/index.jsx
+++ b/packages/webapp/src/components/EditCropVariety/index.jsx
@@ -3,7 +3,7 @@ import React, { useState } from 'react';
 import PropTypes from 'prop-types';
 import { useTranslation } from 'react-i18next';
 import { Label } from '../Typography';
-import Input, { integerOnKeyDown } from '../Form/Input';
+import Input, { getInputErrors, integerOnKeyDown } from '../Form/Input';
 import styles from './styles.module.scss';
 import Radio from '../Form/Radio';
 import Form from '../Form';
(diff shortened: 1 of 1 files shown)


Duncan-Brain (approved):
Nice thanks! Looks good to me! 

I would normally say that we might put the constants into a constants.js file so it can be shared between add and edit views ... but these folders are weirdly high level and not grouped so this works for me!

packages/webapp/src/components/EditCropVariety/index.jsx
```diff
@@ -143,12 +141,30 @@ export default function PureEditCropVariety({
 
       <Input
         style={{ marginBottom: '40px' }}
-        label={t('translation:FIELDS.EDIT_FIELD.VARIETY')}
-        type="text"
-        hookFormRegister={varietyRegister}
-        hasLeaf={true}
+        label={t('CROP.VARIETY_COMMON_NAME')}
+        hookFormRegister={commonNameRegister}
+        errors={getInputErrors(errors, FIELD_NAMES.COMMON_NAME)}
```
antsgar, line 146: Is the missing `hasLeaf` property here intentional?

antsgar (commented):
Left one small question but looks great, thank you for fixing!

packages/webapp/src/components/EditCropVariety/index.jsx
navDhammu: Yes because when adding a crop there is no leaf, so I assumed that's how it should be for both views.

antsgar (approved):

PR #3127, opened 2024-02-13
LF-4090 (backend): Display KPIs in dashboard 

SayakaOno:
**Description**

This PR is for the APIs for LF-4090. Don't worry about reviewing for now if you're busy with your tasks! I will determine whether to include frontend code in this PR or open another PR depending on the team's availability.

- Update get default and custom animal types APIs to take `count` query parameter
- Tests
   - Update `getRequest` functions to allow query params. It looks like we are trying to make `getRequest` function compatible with async/await syntax using `util.promisify`, but `Chai-HTTP` supports promises natively, so I got rid of it.
   - Add `afterEach` in default animal type test file so that the table is cleaned up after each test. (without this, the new test will fail)

Jira link: https://lite-farm.atlassian.net/browse/LF-4090

**Type of change**

**How Has This Been Tested?**

- [x] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/api/src/util/animal.js
```diff
@@ -0,0 +1,76 @@
...
+    .select(typeIdColumnName, 'count')
+    .where('farm_id', farmId)
+    .whereNotDeleted();
+
+  const map = {};
+  updateMapWithTypeCounts(map, animals, typeIdColumnName);
+  updateMapWithTypeCounts(map, animalBatches, typeIdColumnName);
+
+  return map;
+};
```
antsgar, line 76: You can simplify this by using `UNION` to get results from both tables, group them by your column name and return the counts

```sql
select default_type_id as type, count(*) from (
  (select default_type_id from animal a)
  union all
  (select default_type_id from animal_batch ab)
) as animal_batch_union
group by default_type_id
```

I think if you do it this way you'll get the counts for each type directly from the query and won't need any `updateMap` extra functions.

SayakaOno: Oh true, I forgot that what were returned from the two tables were the same structure.
Thank you!

packages/api/src/controllers/customAnimalTypeController.js
```diff
@@ -17,13 +17,25 @@ import { transaction, Model } from 'objection';
...
+import { getAnimalTypeIdCountMap } from '../util/animal.js';
 
 const customAnimalTypeController = {
   getCustomAnimalTypes() {
     return async (req, res) => {
       try {
         const { farm_id } = req.headers;
         const rows = await CustomAnimalTypeModel.query().where({ farm_id }).whereNotDeleted();
+
+        if (rows.length && req.query.count === 'true') {
```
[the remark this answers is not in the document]
SayakaOno: That was intentional, I am just attaching `count` info to the rows 😂
So you are saying it is possible to query types with an extra column `count`, right?
I will try it.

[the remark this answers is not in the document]
antsgar: Ah no, never mind, you're right! I got confused here and thought we were querying animals when we're actually querying types. My bad!

packages/api/src/util/animal.js
kathyavini: I might have missed some context here -- maybe from the animal identifier discussion? -- but why are we now using raw SQL in the controllers if it's more expedient?

I've always remembered this bit of technical guidance from our [Pre-integration peer review document](https://lite-farm.atlassian.net/wiki/spaces/LITEFARM/pages/1191575576/Pre-integration+peer+review) because it made a lot of intuitive sense to me

Screenshot 2024-02-15 at 3 15 36 PM

Are we revising these guidelines?

SayakaOno: You haven't missed anything, I just don't remember those rules, so I will read and follow the guidelines.
I wrote new queries in controllers. Are they supposed to be in models?

kathyavini: Well it was more a question for @antsgar in suggesting an SQL solution (I thought the original map read pretty well)

SayakaOno: Sorry 😂 

Understood. I did those two approaches, and with the raw SQL approach, I was able to do everything in one query, but there were full of errors at first. I forgot to check `deleted` and `farm_id` and ended up adding more tests.

antsgar: Oh yeah, you're right the controller should call the model instead of executing the query and it's the model's responsability to query the database, but if we wanted a more complex query we could just add that to the model and call it from the controller -- my code suggestion didn't necessarily imply using that code directly in this file, my bad! That being said I understand why we went with this approach here because of the counts for batches, so it makes sense now

antsgar: Oh @SayakaOno I just saw you implemented the more complex query! Query looks great, nice job 🙌 but I think @kathyavini is right that it belongs in the model. I'd create a new method in the model and call it from the controller. Wdyt?

SayakaOno: Thank you Anto, sounds good. I will move the queries to the models.
But as Joyce was asking, I'm curious if we should avoid writing raw SQL and use ORM...

antsgar: If you can write the same query with Knex then sure, since we're already using Knex that should give us a more readable output! Sometimes Knex has some limitations and some queries need to be written raw, but if both approaches are doable I'd agree the ORM should be preferred if we do have an ORM like in this case

SayakaOno: I understand.
I believe those queries are some of those that need to be written raw.
I have moved the queries to the models. It would be appreciated if you could re-review when you have the time. Thank you!

packages/api/tests/custom_animal_type.test.js
```diff
@@ -48,17 +48,14 @@ describe('Custom Animal Type Tests', () => {
...
       .request(server)
-      .get('/custom_animal_types')
+      .get(`/custom_animal_types${query}`)
       .set('user_id', user_id)
-      .set('farm_id', farm_id)
-      .end(callback);
+      .set('farm_id', farm_id);
   }
 
-  const getRequestAsPromise = util.promisify(getRequest);
```
kathyavini, line 60: Thank you for looking into this! I had always copied it as boilerplate without questioning whether it was necessary.

packages/api/src/models/defaultAnimalTypeModel.js
```diff
@@ -39,6 +40,30 @@ class DefaultAnimalTypeModel extends Model {
       additionalProperties: false,
     };
   }
+
```
kathyavini, line 43: Oh gosh, the query is completely identical except for the table + column name? I think we should declare our experimental two-table approach a failure 😂

I like having this as a codebase example of how to do SQL in a model though. Even for an SQL-phobe like me, I think it reads very clearly side by side with the model definition ❤️

kathyavini (approved):
I love it!

I guess it will just be on me to brush up on my SQL now to keep up with these fancy queries with coalesce and stuff 😂

To me it seemed like the two table approach just ended up creating massive repetition for you here (especially in the tests) -- did you find that to be true? Or was there was still some conceptual value to the separation?

packages/api/tests/custom_animal_type.test.js
SayakaOno: I had errors in my tests and blamed this, but they were just due to my mistakes...
However, ChatGPT says this is not used how it is supposed to be 😂 I think we can stop copying this in the future!

packages/api/src/models/defaultAnimalTypeModel.js
SayakaOno: For the custom type, we also need to check `farm_id`, but we could do `farm_id = ? OR farm_id is NULL` if we had one table.

SayakaOno (commented):
@kathyavini Thank you for reviewing! I couldn't have done it without ChatGPT... I need to learn more about SQL too.

I want to emphasize that the two-table approach becomes burdensome especially when redoing things 😂 I hadn't considered it...
I haven't found any good reasons why using two tables is better yet.

antsgar (approved):
Looks great! 🙂

PR #3128, opened 2024-02-14
LF-4089 Create KPI component

kathyavini:
**Description**

Overall KPI Component structure:

```plaintext
Container (handles selection state)
  └── Pure Tile Dashboard / Main Component (handles counts + collapsing more logic)
    ├── SummaryTiles (presentational)
    ├── Dashboard Tile (presentational)
    └── MoreComponent (handles its collapsed, active, etc. state)
```

Because the container keeps the selected state, the main [Storybook story](http://localhost:6006/?path=/story/components-tiledashboard--many-types) includes a wrapper with state handling ([code](https://github.com/LiteFarmOrg/LiteFarm/blob/314e41406b0c2ae2a3429e5b23dedc7039aa8247/packages/webapp/src/stories/TileDashboard/TileDashboard.stories.tsx)) which models the role of the future in-app container and allows the selected interaction to be tested.

The subcomponents are as follows and have minor Storybook stories:
- [Summary Tiles](http://localhost:6006/?path=/docs/components-tiledashboard-summarytiles--docs)
- [Dashboard Tile](http://localhost:6006/?path=/docs/components-tiledashboard-dashboardtile--docs)
- [More Component](http://localhost:6006/?path=/docs/components-tiledashboard-morecomponent--docs)

subcomponents

The component organization is something I struggled with so I would love some suggestions for improvement! As written, I feel like they are chunked arbitrarily rather than with any particular patterns of reusability.

Jira link: https://lite-farm.atlassian.net/browse/LF-4089

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/webapp/src/stories/TileDashboard/TileDashboard.stories.tsx
```diff
@@ -0,0 +1,179 @@
...
+import { ReactComponent as CattleIcon } from '../../assets/images/animals/cattle-icon-btn-list.svg';
+import { ReactComponent as ChickenIcon } from '../../assets/images/animals/chicken-icon-btn-list.svg';
+import { ReactComponent as GoatIcon } from '../../assets/images/animals/goat-icon-btn-list.svg';
+import { ReactComponent as PigIcon } from '../../assets/images/animals/pig-icon-btn-list.svg';
+import { ReactComponent as RabbitIcon } from '../../assets/images/animals/rabbit-icon-btn-list.svg';
+import { ReactComponent as SheepIcon } from '../../assets/images/animals/sheep-icon-btn-list.svg';
+
+// https://storybook.js.org/docs/writing-stories/typescript
+const meta: Meta<typeof PureTileDashboard> = {
+  title: 'Components/PureTileDashboard',
```
antsgar, line 32: Could we call this just `TileDashboard`? I think the "Pure" makes it harder for me to search for what I'm looking for

packages/webapp/src/components/TileDashboard/DashboardTile/styles.module.scss
```diff
@@ -0,0 +1,111 @@
...
+  display: flex;
+
+  justify-content: center;
+  align-items: center;
+  flex-shrink: 0;
+  gap: 8px;
+  padding: 0px 8px;
+
+  border-radius: 4px;
+  background: var(--Colors-Accent---singles-Icon-action, #2d5550);
```
antsgar, line 32: When inspecting the tiles in dev tools I see that this variable is not defined -- and although the fallback color matches Figma, in Figma there's an opacity of 25% applied. For reference this is what I'm seeing in Storybook (basically the tiles are blending with the background)

Captura de pantalla 2024-02-15 a la(s) 10 13 26

antsgar (commented):
Left a couple small comments but otherwise looks perfect, love it! 🎉

packages/webapp/src/components/TileDashboard/DashboardTile/styles.module.scss
kathyavini: Wait, sorry could you explain this one further?  I would see this over and over in Storybook, and every time it would be resolved by a restart of the Storybook server -- then the colours would return to normal! This is from an undefined CSS variable and the fallback only sometimes being applied? How does the opacity figure in?

packages/webapp/src/stories/TileDashboard/TileDashboard.stories.tsx
kathyavini: Sure -- rename just the Storybook story you mean (as opposed to the component + Story?)

I already get thrown off by how the component folder name (doesn't include Pure) and the export name (includes Pure) don't match, but I thought there was a reason behind that pattern so I kept following it. Definitely seems pointless in Storybook though where they're supposed to all be pure components.

packages/webapp/src/components/TileDashboard/DashboardTile/styles.module.scss
kathyavini: @antsgar after adding the new CSS colour this still happens for me all the time in Storybook 😞 If not CSS variable, I don't know what could be causing this!

packages/webapp/src/stories/TileDashboard/TileDashboard.stories.tsx
antsgar: Yeah that's what I meant! Just the story 🙂

antsgar (dismissed):
Looks awesome! 🎉

packages/webapp/src/components/TileDashboard/MoreComponent/index.tsx
SayakaOno, line 1: License is missing ;)

packages/webapp/src/components/TileDashboard/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+            </div>
+          ))}
+          {hiddenIconTiles.length ? (
+            <MoreComponent
+              moreIconTiles={hiddenIconTiles}
+              selectedFilterIds={selectedFilterIds}
+              className={styles.flexItem}
+            />
+          ) : (
+            <></>
```
SayakaOno, line 84: I'm just curious, is this the same thing as `null`??

packages/webapp/src/stories/TileDashboard/mockTiles.tsx
SayakaOno, line 1: License is missing ;)

packages/webapp/src/components/TileDashboard/useDynamicTileVisibility.ts
```diff
@@ -0,0 +1,89 @@
...
+  gap = 4,
+  tileWidth = 90,
+  moreButtonWidth = 90,
+  minWidthDesktop = 600,
+  rowsPerView = { desktop: 1, mobile: 2 },
+}: useDynamicTileVisibilityParams) => {
+  const [visibleIconTiles, setVisibleIconTiles] = useState<TypeCountTile[]>([]);
+  const [hiddenIconTiles, setHiddenIconTiles] = useState<TypeCountTile[]>([]);
+
+  const totalTiles = typeCountTiles.length;
```
SayakaOno, line 51: I think this can be moved to `useLayoutEffect`, and `isDesktopView`can be `const`.

packages/webapp/src/components/TileDashboard/useDynamicTileVisibility.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  useLayoutEffect(() => {
+    let resizeObserver: ResizeObserver;
+
+    const updateLayout = (container: HTMLElement) => {
+      const containerWidth = container.getBoundingClientRect().width;
+      let isDesktopView = containerWidth > minWidthDesktop;
```
SayakaOno, line 58: It sounds like the desktop view should start at `600px`? `>= minWidthDesktop;`

packages/webapp/src/components/TileDashboard/useDynamicTileVisibility.ts
```diff
@@ -0,0 +1,89 @@
...
+
+      const rowMultiplier = isDesktopView ? rowsPerView.desktop : rowsPerView.mobile;
+
+      const availableWidth = containerWidth - moreButtonWidth;
+      const tilesPerRow = Math.floor(availableWidth / (tileWidth + gap));
+      const totalFittableTiles = tilesPerRow * rowMultiplier;
+
+      let tilesToHide = totalTiles - totalFittableTiles;
+      if (tilesToHide < 0) tilesToHide = 0;
```
SayakaOno, line 67: Could be `const tilesToHide = Math.max(totalTiles - totalFittableTiles, 0)`

SayakaOno (commented):
Looks great! Stories will make my task much easier, thank you!!! ❤️ 

I wanted to confirm two things:

- Are we supposed to see "+1 more"?
Screenshot 2024-02-16 at 4 14 19 PM

- The icon may be pushed unintentionally?
Screenshot 2024-02-16 at 4 22 02 PM

packages/webapp/src/components/TileDashboard/SummaryTiles/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+interface SummaryTilesProps {
+  totalCount: number;
+  categoryCount: number;
+  categoryLabel: string;
+}
+
+export const SummaryTiles = ({ totalCount, categoryCount, categoryLabel }: SummaryTilesProps) => (
+  <div className={styles.summaryTiles}>
+    <div className={styles.totalCount}>
+      <Main className={styles.countLabel}>total</Main>
```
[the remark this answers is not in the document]
kathyavini: Thank you!!

packages/webapp/src/components/TileDashboard/index.tsx
kathyavini: Yes, in terms of the DOM structure! I'll change to `null` as it is nicer looking in the code 🙂

packages/webapp/src/components/TileDashboard/useDynamicTileVisibility.ts
kathyavini: Is there an advantage to calculating `totalTiles` from within the `useLayoutEffect`?

SayakaOno: If it's in the `useLayoutEffect`, I think it will be calculated only once when the component is rendered.

kathyavini:
Thank you @SayakaOno for your careful review eyes!!! 🙏

The `moreTileWidth` was a holdover from a previous iteration and was not used correctly.  I think the small correction added now should be sufficient to fix, but this hook was my nemesis last week so I would appreciate if you could take a look at the new logic!!

SayakaOno:
@kathyavini I haven't reviewed the logic yet, but I'm wondering if you'd like to make improvements to these.
I don't think they are wrong, so please let me know if we're okay with leaving them as they are!

Screenshot 2024-02-20 at 2 10 28 PM
Screenshot 2024-02-20 at 2 10 57 PM

kathyavini:
@SayakaOno those look fine to me, is there something wrong with them? 😅 I remember thinking when I looked at the original [Figma](https://www.figma.com/file/eMffC0h3JPlWtzNttxJpil/Litefarm-Updated?node-id=3089%3A43641&mode=dev) that I didn't entirely understand the logic for how the second row tiles were collapsed, so I've been happy with whatever the flexbox default suggests.

SayakaOno:
@kathyavini For the first one, I expect the first row to have three types and the second row to have two types. For the second one, I expect types are shown instead of "+ 2 more" on the second row. Actually, I was expecting both to look like the screenshot below: 
Screenshot 2024-02-20 at 3 42 35 PM

But I just saw the mockup that "+ 6 more" occupying the second row 😅
The tiles were being displayed as I expected depending on the browser width, so I thought it was due to a calculation error, but now I understand it's the result of flex-box. Thank you for clarifying.

packages/webapp/src/components/TileDashboard/SummaryTiles/styles.module.scss
```diff
@@ -0,0 +1,81 @@
...
+
+.totalCount:hover,
+.typeCount:hover {
+  background: var(--Colors-Neutral-Neutral-900, #2b303a);
+  box-shadow: 0px 1px 0px 0px rgba(0, 0, 0, 0.25);
+
+  .countText {
+    color: var(--Btn-primary-pristine, #ffcf54);
+  }
+}
```
SayakaOno, line 81: Just FYI, I asked Loïc why the colour changes, and it looks like it was leftover from his previous idea, so I will remove these in my PR later.
https://www.figma.com/file/eMffC0h3JPlWtzNttxJpil?node-id=3089:43641&mode=design#710095392

packages/webapp/src/components/TileDashboard/useDynamicTileVisibility.ts
```diff
@@ -0,0 +1,89 @@
...
+    let resizeObserver: ResizeObserver;
+    const totalTiles = typeCountTiles.length;
+
+    const updateLayout = (container: HTMLElement) => {
+      const containerWidth = container.getBoundingClientRect().width;
+      const isDesktopView = containerWidth >= minWidthDesktop;
+
+      const rowMultiplier = isDesktopView ? rowsPerView.desktop : rowsPerView.mobile;
+
+      const tilesPerRow = Math.floor(containerWidth / (tileWidth + gap));
```
SayakaOno, line 59: One of the tiles doesn't need gap, so this could be improved by:
`const tilesPerRow = Math.floor(containerWidth + gap / (tileWidth + gap));`

Currently `tilesPerRow` and the actual number of tiles displayed don't match sometimes, but adding "+ gap" will fix it and it will work as designed.

SayakaOno (approved):

antsgar (approved):
It looks awesome! 🔥

PR #3129, opened 2024-02-17
LF-4120 Add fields to animal and animal batch tables

kathyavini:
**Description**

This begins setting up the backend for the animal & animal batch removal functionality by:
- Creating the animal_removal_reason enum table
- Adding removed, animal_removal_reason_id, and removal_explanation properties to animal and animal_batch
- Creating a GET endpoint for fetching the values in the enum table (I used animal_sex as a template)

Jira link: https://lite-farm.atlassian.net/browse/LF-4120

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

Test for the new GET endpoint is animal_removal_reason.test.js. I tested the model validation and database schema check (i.e. the requirement for providing an `animal_removal_reason_id` using Postman on the existing POST endpoints for animal and batch.

**Checklist:**


antsgar (dismissed):
It all looks good to me, nice job! 🙌

packages/api/src/models/animalBatchModel.js
```diff
@@ -51,9 +51,21 @@ class AnimalBatchModel extends baseModel {
...
       additionalProperties: false,
+      if: {
+        properties: {
+          removed: { const: true },
+        },
+        required: ['removed'],
+      },
+      then: {
+        required: ['animal_removal_reason_id'],
+      },
```
kathyavini, line 68: This if-then on the schema is only triggering on POST (which was the only controller that existed at the time I wrote this) and not on PATCH. Therefore it's not a useful approach (and not my favourite syntax either) soI'll remove it in the PATCH PR, in favour of either some other custom model function or just sticking with the database validation.

packages/api/src/models/animalBatchModel.js
```diff
@@ -51,9 +51,21 @@ class AnimalBatchModel extends baseModel {
         id: { type: 'integer' },
         name: { type: 'string' },
         notes: { type: ['string', 'null'] },
+        removed: { type: 'boolean' },
```
Duncan-Brain, line 54: I know its in the ticket but is this necessary? Could we not just:

`animal_removal_reason_id = null === removed= false`
and
`animal_removal_reason_id != null === removed= true`

This has the added benefit of not needing constraints or checks

Duncan-Brain (changes requested):
Tests run successfully !

I really don't think it makes sense to have the `removed` boolean property since we can infer it from the id's existence. It is causing the need to add complex checks.

If I am wrong and we do go that route the reverse check is maybe needed since the following still possible: To have removed == false, and animal_removal_id + reason set to some value.

packages/api/src/models/animalBatchModel.js
kathyavini: Oh my gosh I think I love that idea 😂  Let me run through the logic in the PATCH controllers + tests to see if holds up cleanly throughout the whole flow.

kathyavini: @antsgar what do think? It's super easy to implement on the backend (particularly if we don't care about keeping `animal_removal_reason_id` and the optional string `removal_explanation` in sync, which the current code doesn't do anyway.

The only drawback I can think of is that throughout the app we're going to be checking/filtering on `animal.animal_removal_reason_id` which is awkward compared to `animal.removed`. We could generate a `removed` property on the frontend, but I feel like making the RTK Query-managed entity start to diverge from the backend like that would be a mistake that would just confuse us later... if we want a `removed` property I think it should be in the database, and @Duncan-Brain is right, we would need the opposite check on it too.

So I could go either way with this...!

Duncan-Brain (approved):
Very nice!

antsgar (approved):

PR #3141, opened 2024-02-27
LF-4118 create DELETE endpoints for animals and batches

kathyavini:
**Description**

Adds DELETE endpoints for animals and animal batches.

- Since deleted records are not included in the historical tracking, they shouldn't have any need to store deleted date
- As part of this PR, deleted records are filtered out from the related animal / animal batch id arrays on animal group GET (as discussed as a group during pointing). Animal group POST already disallowed using deleted records

Jira link: https://lite-farm.atlassian.net/browse/LF-4118

**Type of change**

**How Has This Been Tested?**

- [x] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/api/src/controllers/animalBatchController.js
```diff
@@ -149,6 +149,63 @@ const animalBatchController = {
       }
     };
   },
+
+  deleteAnimalBatches() {
+    return async (req, res) => {
```
antsgar, line 154: This PR's working great! As a potential improvement, I was thinking that in this particular case the code for deletion for both animals and batches seems to be almost identical, except for the different model they use, so this would be a good case to abstract away into a middleware. I tried it out in a separate branch, `middleware-for-common-code` -- essentially the middleware receives the model and handles the deletion logic. I did have to make the error messages a bit more generic, they could be customized with a parameter too but it didn't seem worth the trouble. Can you take a look and let me know your thoughts?

kathyavini: Cool, I love it!!

My only hesitation is with the concept of doing CRUD in middleware itself, because I would have thought that is so completely a controller action. Logically I would lean towards doing the loop that checks on IDs (which is the bulk of the middleware code anyway) in this function (and maybe even extend it to handle the other routes that do the same check on the `req.body` array?) but still let the controller do the actual deletion 🤔 

But if handling the whole CRUD in middleware is a legitimate pattern that I just haven't heard of yet, then I'm fine with fine with giving it a go; it reads very well.

antsgar: Yup that's a really good point, I kinda thought about that too! It is a bit weird to have a middleware without any associated controller method being called after it. I like your idea of abstracting everything except for the actual deletion!

antsgar (dismissed):
Love the final outcome!

packages/api/src/middleware/checkAnimalEntities.js
```diff
@@ -0,0 +1,70 @@
+import { Model, transaction } from 'objection';
+import { handleObjectionError } from '../util/errorCodes';
```
Duncan-Brain, line 2: My backend was crying that this isn't errorCodes.js

packages/api/src/middleware/checkAnimalEntities.js
```diff
@@ -0,0 +1,70 @@
...
+      const invalidIds = [];
+
+      for (const id of idsSet) {
+        // For query syntax like ids=,,, which will pass the above check
+        if (!id || isNaN(Number(id))) {
+          await trx.rollback();
+          return res.status(400).send('Must send valid ids');
+        }
+
+        const existingRecord = await model
```
Duncan-Brain, line 44: I can't help but wonder if there is a way to do this in one query so that a trx is not needed. While unlikely... having two trx (here and then in delete) means the existing record status could (extremely unlikely) change between the time between actions.

packages/api/src/middleware/checkAnimalEntities.js
```diff
@@ -0,0 +1,70 @@
...
+      if (invalidIds.length) {
+        await trx.rollback();
+        return res.status(400).json({
+          error: 'Invalid ids',
+          invalidIds,
+          message:
+            'Some entities do not exist, are already deleted, or are not associated with the given farm.',
+        });
+      }
+
```
Duncan-Brain, line 64: I think you need a trx.commit() here so that the trx does not stay open. (or pass it to the next() even? I dunno if thats a possible thing so that 2 trx are not made)

Duncan-Brain (changes requested):
Tests and postman testing works well!

Only blocker for me is the `'.js'` otherwise my backend crashes and adding `trx.commit()`.  I am surprised it does not crash for you both? Seems like all the other middleware uses `'.js'`. 

Some notes I have generally - non blockers:

1. The delete with query params surprised me -- so I looked up other patterns here https://stackoverflow.com/questions/21863326/delete-multiple-records-using-rest. What I took from it was:
  - So there is a possibility (very unlikely) of reaching a URI char limit with this pattern I think they sometimes limited to 2000 chars.
  - I think the reason we are not using a POST or PATCH is that our existing soft-delete pattern is DELETE right?
  - If we did use post or patch then we could use a regular array without splitting the string on comma -- small but satisfying.

2. We've already talked about this in the daily - But for me one really nice thing about following the crop plan delete pattern of hard deleting the animal from the group join, (or even adding a deleted property to the group join) is more strightforward (vs builder) modifiers in the model relations. It seems reasonable to me that the deleted animal leaves the group if the animal never existed, or is not allowed to be deleted if in group (must be removed (hard-deleted) manually before that can be done). It could always be added back to the group manually if for some reason we wanted to un-delete it.

kathyavini:
Thanks @Duncan-Brain!! Huh, I do see the API crash now. I wonder why the test suites behave differently re:imports and don't care.

Neat, I did not know about the [char limit](https://stackoverflow.com/a/26809439/15876096)! I think we could even write a chai test for this, but I'm not sure if that will be accurate to how the real app behaves (e.g. we might hit an [nginx limit](https://stackoverflow.com/questions/1067334/how-to-set-the-allowed-url-length-for-a-nginx-request-error-code-414-uri-too) before the Node limit). I think I'd just to keep it in mind for the time being but make sure we test this on beta. It won't even require creating hundreds of animals since the failure will happen before that point.

I actually find the modifier approach kind of elegant 😉 But after our discussion yesterday I was assuming we are more likely to go with the `deleted` property on the join table approach (when we can start removing animals from groups), so we aren't in some hybrid state of soft delete + irrevocable data loss -- I was quite convinced by that argument, actually. I mean, we tell the client "animal never existed" but that's just the client; for us we are making the effort to create a soft delete, so why throw away some but not all data at the same time?

packages/api/src/middleware/checkAnimalEntities.js
kathyavini: _Can_ you pass it in `next()`??? I'll commit it, thanks for catching 😄

packages/api/src/middleware/checkAnimalEntities.js
kathyavini: I was a little worried by the separate transactions too; actually even opening a transaction in middleware seems a little strange!

In this particular case, what you describe should be safe because the only status that could change like that would be `deleted` itself (record farm ownership isn't going to change) and a re-DELETE on a deleted record won't do anything other than bump the updated properties.

packages/api/src/middleware/checkAnimalEntities.js
kathyavini: Thank you! 🙏

packages/api/src/middleware/checkAnimalEntities.js
Duncan-Brain: I dno haha -- seems reasonable to me!

Duncan-Brain:
Yeah I think I am good with the modifier in general. But query builder queries specifically get more confusing for me. Not terrible just makes me think a little harder. lol

Duncan-Brain (approved):

PR #3145, opened 2024-02-29
LF-4138 Convert current filter implementation to TypeScript

kathyavini:
**Description**

This was originally intended to assist with the new filter component; however, I'm hoping it will be equally if not more useful in terms of understanding the filter dataflow from Redux to the component for use in Animals.

Any feedback on readability (particularly on the placement, naming, and structure of interfaces -- which was the main point of doing this conversion -- would be so appreciated!! 🙏 

Notes:
- Where indicated in the comments, there is one aspect of data handling in the current FilterMultiSelect that would make it incompatible with use in the Finance Report page. I have not fixed that issue, just highlighted it.
- Other than the types and some JSX changes that were necessary to please the type checker, there are no functional changes in this code, except:
  - Changing, as discussed in S82 Sprint Planning today, the placeholder for a FilterMulltiSelect from "Select..." to "Showing all..."
  - The filter multi-select with the filter type `SEARCHABLE_FILTER_MULTI_SELECT` was not, in fact, searchable because it was missing its `isSearchable` prop... that has been added (or more likely, restored).

Jira link: https://lite-farm.atlassian.net/browse/LF-4138

**Type of change**

The placeholder and the searchability of the multi-select filter have been fixed; otherwise all changes should be runtime silent as they are just in the typescript types.

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

Note: I would have liked to add licenses to some of the files I touched, but GitHub's ability to detect a JSX -> TSX conversion as the same file (vs a delete + new file) is brittle, and seems to require changing as few lines of code as possible.


packages/webapp/src/components/FilterPage/index.tsx
```diff
@@ -6,6 +6,24 @@ import { Underlined } from '../Typography';
 import { useTranslation } from 'react-i18next';
 import Button from '../Form/Button';
 import FilterGroup from '../Filter/FilterGroup';
```
kathyavini, line 8: The data flow for the non-finance filters is:

1. Filter Container (Filter/Task, Filter/CropVariety, Filter/Documents etc.)
2. PureFilterPage
3. FilterGroup

The container defines the filterRef and the relationship with the Redux store. The PureFilterPage component is a layout component with buttons that control the overall clearing and applying (saving) of filters. The FilterGroup component displays each individual filter. Each controls their own update of the filterRef, and, through the passed onChange, dirties the parent filter state to allowing saving. The onApply handler passed from the container to the PureFilterPage saves the updated filterRef to the Redux store.

packages/webapp/src/containers/Filter/Transactions/index.tsx
```diff
@@ -1,5 +1,5 @@
 /*
```
kathyavini, line 1: The flow for transactions filter is

1. Finances/Report -AND- Finances/TransactionFilter
2. (This component) TransactionsFilterContent
3. FilterGroup + subfilters

Dataflow differs in that:
- Filters are only coupled bi-directionally to the Redux state in Finances/TransactionFilter. The report filters set local component state (via their passed onChange) and have a read-only relationship with the Redux state
- The Redux filter state shape is transformed into the component-used shape here in a common component (this is a good pattern I think)
- The Report Filter state is controlled via the passed onChange, which required modifying the callback passed from FilterGroup to the individual filters in #2942.
  - This is a very clever approach but was hard to type 😄 
  - It means that depending on the calling container, the parameter passed to the onChange by the component is either completely ignored, or is used to the set application state and needs to be a particular shape. It happens to be the wrong shape in the MultiSelect as written, but that combination of container + component is not yet seen in app (it will be)

packages/webapp/src/containers/Filter/types.ts
```diff
@@ -0,0 +1,37 @@
...
+  };
+}
+
+// (filterKey, filterState) => setTypesFilter({ ...typesFilter, [filterKey]: filterState })
+export type FinanceReportOnChangeCallback = (filterKey: string, filterState: FilterState) => void;
+
+// setDirty: () => !isDirty && setIsDirty(true)
+export type FilterPageOnChangeCallback = () => void;
+
+// type ContainerOnChangeCallback = FinanceReportOnChangeCallback | FilterPageOnChangeCallback;
```
kathyavini, line 34: I would love to define this as a union, but then the types shown on hover are the names of the union elements and not the full type.

packages/webapp/src/containers/Filter/Transactions/index.tsx
antsgar: Thank you for these notes, very helpful! I think the structure itself is not entirely different (although the file names change), in the sense that the `TransactionFilter` container defines the ref and relationship with the Redux store, similarly to what the other filter containers do, and the `TransactionsFilterContent` component replaces `FilterPage` in defining the layout and calling `FilterGroup`. `FilterPage` could not be reused because in Finances (and in general from now on) we're not doing separate pages for filters, so I think the Finances pattern will likely need to be repeated in that sense for Animals too. The difference with Animals I think will be that we'll need to filter from the filters modal itself and from the KPIs, but both will need bidirectional flow to the Redux store. The KPIs, unlike the Finances report, will need to dispatch the actual action to Redux to modify what's being displayed. If we do at any point need a report for Animals too though, we'd be well setup to implement the same pattern as here in Finances. 

Re this bit

> It happens to be the wrong shape in the MultiSelect as written, but that combination of container + component is not yet seen in app (it will be)

I think I knew this at the time and chose to ignore any other filter types since like you mention they weren't used in Finances or any similar combination that required this `onChange` function to work properly. Will need to be adjusted, although that entire `MultiSelect` component would now be replaced by the new one anyway.

packages/webapp/src/containers/Filter/types.ts
```diff
@@ -0,0 +1,37 @@
...
+export type ContainerOnChangeCallback =
+  | ((filterKey: string, filterState: FilterState) => void) // Finance Report
+  | (() => void); // PureFilterPage and TransactionFilter
```
antsgar, line 37: I would suggest defining this more generically as 

`export type ContainerOnChangeCallback = (filterKey?: string, filterState?: FilterState) => void`

There needs to be some uniformity to the callback's shape since the `FilterGroup` component will always be calling it with the same parameters, the only difference is whether the container cares about them or not

packages/webapp/src/containers/Filter/types.ts
```diff
@@ -0,0 +1,37 @@
...
+ *
+ *  LiteFarm is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ */
+
+// In tempStateReducer > filterReducer
+export interface ReduxFilterEntity {
+  [filterKey: /* e.g. STATUS, LOCATION */ string]: FilterState;
```
antsgar, line 18: It might be nice for the type for `filterKey` to be passed as a generic so that if we need to call this type from more specific filter components we can specify the enum

```ts
export type ReduxFilterEntity<FilterKey extends string = string> = Record<FilterKey, FilterState>;
```

kathyavini: I can't read generics yet, but I'll give this a close study, thank you 😄

packages/webapp/src/containers/Filter/types.ts
kathyavini: I think my main goal -- because it was the hardest thing for me to understand reading these files -- was trying to convey that in some cases `onChange` is just calling `setDirty` and that passed argument -- which particularly reading the `FilterGroup` defined `onChange` would make me assume it's always doing something significant -- is completely dropped; completely irrelevant.

But maybe my original types didn't do that very clearly either... or maybe it's not even something that can/should be expressed in Types since it's not exactly about the function signature? Or the optional parameters makes it clear enough? 🤔 I'm not really sure...!

packages/webapp/src/components/Filter/FilterMultiSelect/index.tsx
```diff
@@ -38,7 +49,7 @@ export const FilterMultiSelect = ({
   }, [shouldReset]);
 
   useEffect(() => {
-    filterRef.current[filterKey] = produce(defaultFilterState, (defaultFilterState) => {
+    filterRef.current![filterKey] = produce(defaultFilterState, (defaultFilterState) => {
```
Duncan-Brain, line 52: Does it break if you wrap this in `if(filterRef.current) {}`?

This _could_ hide errors otherwise

packages/webapp/src/containers/Filter/types.ts
```diff
@@ -0,0 +1,37 @@
...
+    active: boolean;
+    label: string /* e.g. 'Abandoned', 'Completed */;
+  };
+}
+
+// (filterKey, filterState) => setTypesFilter({ ...typesFilter, [filterKey]: filterState })
+export type FinanceReportOnChangeCallback = (filterKey: string, filterState: FilterState) => void;
+
+// setDirty: () => !isDirty && setIsDirty(true)
+export type FilterPageOnChangeCallback = () => void;
```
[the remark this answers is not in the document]
kathyavini: No, I'll do as @antsgar suggested above, but I'm still looking for the 100% pleasing solution to convey that sometimes this is just `setDirty()`!  Any thoughts? Does the chain of onChange read okay to you overall and in the component file (as a previous consumer of the filter pattern when building the sales hook)?

I know I didn't actually type the component you used (since we're not touching it anymore this sprint).

packages/webapp/src/components/Filter/FilterMultiSelect/index.tsx
kathyavini: I doubt it would break, but I wanted to avoid changing the JS logic the way a new conditional would (while the non-null assertion is just for the type checker). Since the code works, I'm assuming we don't need the check for functionality? Sounds like @antsgar will be re-writing this one though anyway.

packages/webapp/src/components/Filter/FilterGroup/index.tsx
```diff
@@ -21,9 +21,26 @@ import { FilterMultiSelect } from '../FilterMultiSelect';
...
-const FilterItem = ({ filter, showIndividualFilterControls, ...props }) => {
-  if ((filter.type === PILL_SELECT || !filter.type) && filter.options.length > 0) {
+// i.e (filterState) => ContainerOnChangeCallback(filter.filterKey, filterState)
+export type ComponentOnChangeCallback = (filterState: FilterState) => void;
+
+interface FilterItemProps {
+  filter: ComponentFilter;
+  filterRef: React.RefObject<ReduxFilterEntity>;
+  onChange: ComponentOnChangeCallback;
+  shouldReset?: number;
```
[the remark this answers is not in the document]
kathyavini: I know 😭

packages/webapp/src/containers/Filter/Transactions/index.tsx
kathyavini: > TransactionsFilterContent component replaces FilterPage in defining the layout and calling FilterGroup

Ah sorry, I should have been more clear -- by layout I meant not of the filter components, but of the overall structure (specifically the title, apply button and the state that determines when that button can be clicked), which in finances were lifted up one level to the container and in the others, sent down one level to the Page.

> FilterPage could not be reused because in Finances (and in general from now on) we're not doing separate pages for filters

Okay now I'm realizing the two situations are even more confusing than I initially thought because... yes, this is exactly right! Aaaaand... all those other filters ARE already modals, not pages! Complete with Drawer behaviour. And they do it by situating the filter container within a Drawer, to which they don't pass buttons, because they let that interior page handle the buttons 😅 

So I think both patterns are still technically viable, but I like the non-nesting approach better for sure; `<Layout />` doesn't look that attractive within a `<Drawer />` and there is less flexibility to style or even change the apply button behaviour.

packages/webapp/src/components/Filter/FilterMultiSelect/index.tsx
```diff
@@ -47,22 +58,25 @@ export const FilterMultiSelect = ({
 
   return (
     <ReactSelect
-      style={style}
-      placeholder={`${t('common:SELECT')}...`}
+      //@ts-ignore
+      style={style} // I suspect the forwardRef is at fault for the type error?
```
Duncan-Brain, line 62: May as well type that one out too? lol 

https://react-typescript-cheatsheet.netlify.app/docs/basic/getting-started/forward_and_create_ref/

packages/webapp/src/containers/Filter/types.ts
kathyavini: Ah okay, so we can call it like

```ts
  animalsFilter: ReduxFilterEntity<AnimalsFilterKeys>
 ```
 
Nice! 👍

packages/webapp/src/components/Filter/FilterMultiSelect/index.tsx
kathyavini: Noooooo not the React Select 🤣 I really didn't want to touch that, but I guess it's imported a lot; Nav had issues on his component too. Thanks for link.

Duncan-Brain (dismissed):
I think its all fine -- I am slacking on the ts course so it might be good to get another set of eyes from someone who knows more.

Only thing I am wary of is `!` and `ts-ignore`.

packages/webapp/src/components/Filter/FilterMultiSelect/index.tsx
Duncan-Brain: Without knowing anything -- even though the code works there might just be an edge case we haven't reached yet but its possible the type checker missed a conditional somewhere else. 🤷

packages/webapp/src/containers/Filter/types.ts
antsgar: I see your point! But we could probably throw in a comment instead to clarify that

packages/webapp/src/components/Filter/FilterMultiSelect/index.tsx
antsgar: The filterRef current value _could_ technically be null so it'd be correct to add the conditional here I think, but this component will all get re-written anyway!

antsgar (approved):

Duncan-Brain (approved):

PR #3147, opened 2024-02-29
LF-4090(2): Display KPIs in dashboard

SayakaOno:
**Description**

- Update queries not to count removed animals and batches. For custom types, select only necessary fields.
- Add KPI component to Inventory (deal with it as "full width" component)
- Update z-index (hoverpill's z-index is high, so I updated associated components)
- Update `TileDashboard` components. I realized that `isSelected` in each tile data was being ignored, so I updated it. (the property is not going to be used in KPI)
- Add `useQueries` hook Anto created
- Create functions (`generateUniqueAnimalTypeId`, `uppercaseTheFirstLetter`)
- Convert sort functions to TypeScript

Jira link: https://lite-farm.atlassian.net/browse/LF-4090

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/webapp/src/containers/Animals/Inventory/KPI/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+    const { defaultAnimalTypes, customAnimalTypes } = data;
+    const types = formatAnimalTypes([...defaultAnimalTypes, ...customAnimalTypes], onTypeClick, t);
+    types.sort(getComparator('asc', 'label'));
+
+    return types;
+  }, [data, isLoading, onTypeClick]);
+
+  return (
+    <div style={{ height: kpiHeight || 0 }}>
```
SayakaOno, line 111: I don't like the inline CSS, but KPI's height changes unpredictably, so I went with the easiest way. Since the child component's position is `fixed`, we need to maintain the height of the KPI component so that the table remains visible and is not hidden beneath the KPI component.

packages/webapp/src/App.jsx
```diff
@@ -27,7 +27,7 @@ import Routes from './routes';
 
 function App() {
   const [isCompactSideMenu, setIsCompactSideMenu] = useState(false);
-  const FULL_WIDTH_ROUTES = ['/map'];
+  const FULL_WIDTH_ROUTES = ['/map', '/animals/inventory'];
```
kathyavini, line 30: I love this functionality; it works so well!! I tried it just like this once though and @antsgar reminded me to use the new path constants 😁

packages/webapp/src/components/TileDashboard/MoreComponent/index.tsx
```diff
@@ -62,7 +64,7 @@ export const MoreComponent = ({
                 <DashboardTile
                   key={index}
                   {...item}
-                  isSelected={selectedFilterIds?.includes(item.id)}
+                  isSelected={item.isSelected ?? selectedFilterIds?.includes(item.id)}
```
kathyavini, line 67: I'm so sorry 🙇 This was my sheer laziness -- `isSelected` was never meant to be passed within the item data array, I only (and this is really bad, I am a little surprised I did this) put it on the TypeCountTile interface... in order to recycle that interface on the DashboardTile 😬🙇 

What I should have done is remove `isSelected` from `TypeCountTile` and done this:

```ts
interface DashboardTileProps extends TypeCountTile {
  isSelected: boolean;
}
```

I don't think there is a way to get the interaction to work with a data-passed `isSelected` -- but is that perhaps the purpose with the `??` -- to override the interactivity? If so it's a neat feature now!

I'm sorry about the misleading TypeScript 😞

packages/webapp/src/containers/Animals/Inventory/KPI/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+) => {
+  const formattedTypes: TypeCountTile[] = [];
+  types.forEach((type) => {
+    if (type.count) {
+      const label = 'type' in type ? type.type : t(`animalTypes:${type.key}`);
+      const iconKey = ('key' in type ? type.key : '') as keyof typeof icons;
+      const id = generateUniqueAnimalTypeId(type);
+
+      formattedTypes.push({
+        label: uppercaseTheFirstLetter(label),
```
kathyavini, line 56: It's a good utility function that we should for sure keep available, but I wonder if `text-transform: capitalize` might look better on tiles?

Screenshot 2024-02-29 at 5 13 47 PM


Did Loïc express a preference? I think Figma was all one-word types so it's hard to judge from there.

packages/webapp/src/containers/Animals/Inventory/KPI/index.tsx
kathyavini: Clean solution! Will it make a difference when the table itself has some fixed property (like the header)? I see the table is properly placed with the height correction, but the headers do go under the KPI on scroll -- not sure if this will still be the case with @Duncan-Brain's update?

Screenshot 2024-02-29 at 6 08 31 PM

packages/webapp/src/containers/Animals/Inventory/KPI/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+  return (
+    <div style={{ height: kpiHeight || 0 }}>
+      <div
+        ref={wrapperRef}
+        className={clsx(styles.wrapper, isCompactSideMenu && styles.withCompactMenu)}
+      >
+        <PureTileDashboard
+          typeCountTiles={types}
+          dashboardTitle={t('SECTION_HEADER.ANIMALS_INVENTORY')}
+          categoryLabel={t('TYPES').toUpperCase()}
```
kathyavini, line 119: The CSS already has `text-transform: uppercase` so it shouldn't be necessary.

packages/webapp/src/util/animal.ts
```diff
@@ -0,0 +1,21 @@
...
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ */
+
+import { CustomAnimalType, DefaultAnimalType } from '../store/api/types';
+import { ANIMAL_TYPE_ID_PREFIX } from '../containers/Animals/types';
+
+export const generateUniqueAnimalTypeId = (type: DefaultAnimalType | CustomAnimalType): string => {
+  return `${ANIMAL_TYPE_ID_PREFIX['key' in type ? 'DEFAULT' : 'CUSTOM']}_${type.id}`;
```
kathyavini, line 20: Ack our two table approach!!! 😅

kathyavini (commented):
It looks great!! Thank you for handling the removal (very nice), and you worked around all the mess we made by having two type tables! (And you kept the hover effect 🤭)

All of my inline comments are very minor, but I did have this issue, which is not unique to your branch (I've seen it playing around with this component too):

Screenshot 2024-02-29 at 4 52 49 PM


The first time I go to the component after logging in, the tiles don't render. The counts are populated, so the data must be in the component, but it's not until after I reload that I see the tiles.

Do you ever see that behaviour?

SayakaOno (commented):
Thank you @kathyavini for reviewing!

I was able to reproduce the empty titles so I looked into it.
It looks like the data can't make it to `useLayoutEffect` in `useDynamicTileVisibility`, so I updated the dependency. Could you double check?
(I see an empty table too...)

packages/webapp/src/containers/Animals/Inventory/KPI/index.tsx
kathyavini: Ohhh! I've gotten so used to `toLocaleCompare()` I totally forgot that part of native JS comparison on strings! Got it 👍

kathyavini (dismissed):
Thank you so much for tracking down and fixing the rendering bug!! I would not have known to fix with the dependency array; very clean fix 🙏

packages/webapp/src/containers/Animals/Inventory/KPI/index.tsx
Duncan-Brain: @kathyavini I am not handling any stickyness in the inventory implementation -- does it say somewhere that table header should be sticky?  I can't find it anywhere.

kathyavini: @Duncan-Brain I thought I remember Loïc saying it verbally, something like the "these are fixed as obviously all table headers are" 🤔 Might be misremembering though!

packages/webapp/src/store/api/apiSlice.ts
```diff
@@ -49,7 +62,21 @@ export const api = createApi({
       query: () => `${animalGroupsUrl}`,
       providesTags: ['AnimalGroups'],
     }),
+    getDefaultAnimalTypes: build.query<DefaultAnimalType[], string>({
```
kathyavini, line 65: One small thing I noticed when combining this with @Duncan-Brain's inventory branch is that `getDefaultAnimalTypes` will also be called without the `count` param at all, so the `QueryArg` type of this one should be `string | void` to cover both cases -- not sure which branch that is easier reflected on.

SayakaOno: Thank you, I will update my PR.
I'm just wondering, does it give you an error??

kathyavini: Thank you! 🙏  Yes the type checker error was in Duncan's `useAnimalInventory`:

Screenshot 2024-03-05 at 9 44 03 AM

SayakaOno: Ah! I am using `useQueries` so it didn't give me an error!
Thank you Joyce!!

kathyavini (approved):

antsgar (approved):
Looks great, love it! ❤️

PR #3148, opened 2024-03-01
LF-4092 - update animal controllers to use handle objection error

Duncan-Brain:
**Description**

On creation of the objection error handler it was mentioned by @SayakaOno that we should update all animal controllers to be consistent. (and a modest -50 lines of code ;) )

Now every endpoint in animals with a trx and excess checks should be covered.

Also added in a test that was missing from batch and controller -- should not be able to add a default breed with a custom type.

To test: For exhaustive testing remove one check at a time and check that it is still handled.

Jira link: [LF-4092](https://lite-farm.atlassian.net/jira/software/c/projects/LF/issues/LF-4092)

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

[LF-4092]: https://lite-farm.atlassian.net/browse/LF-4092?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/api/src/controllers/animalController.js
+++ b/packages/api/src/controllers/animalController.js
@@ -20,7 +20,7 @@ import DefaultAnimalBreedModel from '../models/defaultAnimalBreedModel.js';
 import CustomAnimalBreedModel from '../models/customAnimalBreedModel.js';
 import CustomAnimalTypeModel from '../models/customAnimalTypeModel.js';
 import { assignInternalIdentifiers } from '../util/animal.js';
-import { handleDBError } from '../util/errorCodes.js';
+import { handleObjectionError } from '../util/errorCodes.js';
 
 const animalController = {
   getFarmAnimals() {
@@ -121,7 +121,7 @@ const animalController = {
         await assignInternalIdentifiers(result, 'animal');
         return res.status(201).send(result);
       } catch (error) {
-        await handleDBError(error, res, trx);
+        await handleObjectionError(error, res, trx);
       }
     };
   },
@@ -186,7 +186,7 @@ const animalController = {
         await trx.commit();
         return res.status(204).send();
       } catch (error) {
-        handleDBError(error, res, trx);
+        handleObjectionError(error, res, trx);
       }
     };
   },
--- a/packages/api/src/controllers/animalController.js
+++ b/packages/api/src/controllers/animalController.js
@@ -18,7 +18,6 @@ import AnimalModel from '../models/animalModel.js';
 import baseController from './baseController.js';
 import DefaultAnimalBreedModel from '../models/defaultAnimalBreedModel.js';
 import CustomAnimalBreedModel from '../models/customAnimalBreedModel.js';
-import DefaultAnimalTypeModel from '../models/defaultAnimalTypeModel.js';
 import CustomAnimalTypeModel from '../models/customAnimalTypeModel.js';
 import { assignInternalIdentifiers } from '../util/animal.js';
 import { handleObjectionError } from '../util/errorCodes.js';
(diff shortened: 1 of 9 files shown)


packages/api/src/controllers/animalController.js
```diff
@@ -51,86 +50,56 @@ const animalController = {
         const { farm_id } = req.headers;
         const result = [];
 
-        if (!Array.isArray(req.body)) {
-          await trx.rollback();
-          return res.status(400).send('Request body should be an array');
-        }
```
SayakaOno, line 57: In my humble opinion, this is nicer than an Unknown internal server error. You haven't removed the one from animalBatchController!

packages/api/src/controllers/animalGroupController.js
```diff
@@ -57,16 +58,6 @@ const animalGroupController = {
...
 
-        if (!name) {
-          await trx.rollback();
-          return res.status(400).send('Group name is required');
-        }
-
-        if (!Array.isArray(related_animal_ids) || !Array.isArray(related_batch_ids)) {
-          await trx.rollback();
-          return res.status(400).send('Animal ids and batch ids must be arrays');
-        }
```
SayakaOno, line 68: I prefer keeping this too!

SayakaOno (commented):
I prefer receiving 400 rather than 500 when possible. I would like to hear what others think!

packages/api/src/controllers/customAnimalBreedController.js
```diff
@@ -59,37 +59,30 @@ const customAnimalBreedController = {
...
+        //   await trx.rollback();
+        //   return res.status(400).send('Only default_type_id or custom_type_id should be specified');
+        // }
+
+        // if (!default_type_id && !custom_type_id) {
+        //   await trx.rollback();
+        //   return res
+        //     .status(400)
+        //     .send('One of default_type_id or custom_type_id should be specified');
+        // }
```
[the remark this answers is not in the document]
Duncan-Brain: yes thank you!

packages/api/src/controllers/animalController.js
Duncan-Brain: Its more explicit yeah, for me I prefer to not feel like I have to handle every case where there should be an array provided it could add up to a lot of code.

The message provided by objections error "req.body is not iterable" makes enough sense to me. But I also think if you want to make a new Error Type in the future that makes this generic error more readable -- that is nice too!

Duncan-Brain:
I agree it is nice to receive 400 where possible. But to me the solution is to create handled error code types. And continue to allow "unknowns" to be 500.

Let the bad 500 be incentive to make more error codes haha

antsgar:
@Duncan-Brain is there a way we could add more cases to the handle error function so that what is currently returned as a 400 would still be returned as a 400?

Duncan-Brain:
@SayakaOno @antsgar  I added a custom test for the Array check .. I don't really love the solution I came up since it seems a little messy, and the error message is not the nicest of things... but check it out maybe you think it is ok. 

One thing that came up from doing this: 

In our controllers we use the Models to write to the DB, and in our tests we use plain Knex. By not testing our controller during tests this creates two different error messages. Handled in this case but reasonable that we will cause some problems down the line... I think we should be testing our API through our controllers and not knex directly, expecially since we will be putting more constraint and response logic.

Also there was a test in animal_groups being flakey (the order it was returned was not the same as it was added in test) so I made a little band-aid there.

packages/api/tests/animal_group.test.js
```diff
@@ -172,15 +172,19 @@ describe('Animal Group Tests', () => {
         expect(res.status).toBe(200);
         // Should return first two animal groups
         expect(res.body.length).toBe(2);
+        // Sometimes response is returned in different order than created
+        const idMap = res.body.map((group) => groups.findIndex(({ name }) => name === group.name));
```
SayakaOno, line 176: I think naming an array `idMap` is misleading.

SayakaOno (commented):
I'm open to approval, but I'm still in favour of finding a better way of validating the request body that Duncan will feel comfortable with rather than letting the handleCustomError function handle internal errors :)

Duncan-Brain:
I can put all the Array checking back in if that is what is preferred by others too. Maybe we can discuss in tech daily... I definitely agree it is not as nice looking as is.

I think when we add typescript to the backend we could enforce types that way a lot easier and the returned errors would be more parseable or automatic. We just don't have that right now. 

For me the reason why I do not personally want to make it nicer now is that: 
- Error is not translated so it is not really user facing -- and so it is only useful for us the developers. 
- The burden (lines of code) of enforcing every type, for every variable, on every route seems high and the value of letting a generic error handle it seems useful to me. 
- I guess I see it that creating a custom error message for `related_animal_ids = ARRAY` is no more important than creating a custom error message for `breed = STRING` (putting "breed": 5 in your body will throw an objection error that unless you know what it means it might not make sense either).

packages/api/tests/animal_group.test.js
Duncan-Brain: I added a todo since it is just a patch for flaky test. Is the name any better?

SayakaOno: Yes! The new name sounds perfect ❤️ Thank you!
I think your solution is good. Another solution maybe to find a corresponding group from `groups` in `forEach`.

SayakaOno (approved):

packages/api/src/util/errorCodes.js
```diff
@@ -146,6 +166,33 @@ export async function handleObjectionError(err, res, trx) {
...
       data: {},
     });
+  }
+}
+
+// Objection response
+const isNotIterable = 'is not iterable';
+// Plain knex response for tests
+const isNotIterableTests =
+  'Invalid attempt to iterate non-iterable instance.\nIn order to be iterable, non-array objects must have a [Symbol.iterator]() method.';
```
antsgar, line 176: I'm not a fan of this parsing of specific error messages tbh, it seems brittle -- if they change the message in the slightest this will break. Out of all the options, I think I'd rather just keep the 500 or keep the explicit checks for 400s. Thoughts?

Duncan-Brain: Last approach it is! lol 

@antsgar should I make a tech debt ticket (spike, rfc) about specifying our backend request shape. There has been a few suggestions in the past about improving our backend in this regard and we have OAS stuff but to my knowledge we haven't chosen a direction for this improvement yet.

SayakaOno (approved):

antsgar (approved):

PR #3149, opened 2024-03-01
LF-4084: Update table/2 to include checkboxes and row selection behaviour

SayakaOno:
**Description**

- Update V2 table to support checkbox and sticky header and style selected rows
- Add [stories](http://localhost:6006/?path=/story/components-tables-v2--with-checkboxes) and cleanup

NOTE: I didn't apply the new checkbox style as it seems like it would require a good amount of effort.

Jira link: https://lite-farm.atlassian.net/browse/LF-4084

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/webapp/src/components/Table/styles.module.scss
```diff
@@ -14,25 +14,57 @@
  */
 
 .table {
-  padding: 0.5px;
```
SayakaOno, line 17: @Duncan-Brain I had to remove this not to show the table content from the space above the table header when scrolling. Please let me know if this change will break anything.

https://github.com/LiteFarmOrg/LiteFarm/assets/33141219/889fec04-0dbd-4f48-8598-233da58230e0

Duncan-Brain: Yeah no worries from me, I just copied it over from previous version!

Duncan-Brain (dismissed):
It looks good to me! 

I think the only thing I would consider adding is a clear all on the story when the select all is checked.

packages/webapp/src/components/Table/TableV2.jsx
```diff
@@ -92,6 +93,11 @@ export default function TableV2(props) {
     shouldFixTableLayout,
     defaultOrderBy,
     alternatingRowColor,
+    onCheck,
+    handleSelectAllClick,
```
antsgar, line 97: I wonder if these two could be combined and simplified by passing a `setSelectedIds` function prop instead, and letting the table handle the different scenarios where the selected IDs need to change

SayakaOno: It's possible, but I'm hesitant to do it.
A few of my concerns:
- flexibility - what if we want to do something else when checkbox is clicked?
- duplicated `handleSelectAllClick` - we will have a function for this in its parent component anyways because we have "clear selection" in the action menu

Are they not a big deal??

packages/webapp/src/stories/Table/TableV2.stories.jsx
```diff
@@ -266,3 +217,64 @@ export const TasksLabourWithPagination = {
...
+    };
+    const handleSelectAllClick = () => setSelectedIds(data.map(({ id }) => id));
+
+    return (
+      <>
+        <TextButton
+          color="secondary"
+          sm
+          disabled={!selectedIds.length}
+          onClick={() => setSelectedIds([])}
```
Duncan-Brain, line 250: Ah sorry for not being more clear I think I was expecting `handleSelectAllClick()` to be able to handle both selecting and deselecting all as well .

SayakaOno: Oh! I guess that's how that checkbox works 😂 I will fix it, thank you!

antsgar (approved):

packages/webapp/src/stories/Table/TableV2.stories.jsx
Duncan-Brain: Looks great very nice!

Duncan-Brain (commented):
@SayakaOno happy to approve but I think this morning the other table pr was merged so there are now conflicts

PR #3150, opened 2024-03-07
LF-4093 Create animal inventory filter modal

kathyavini:
- [x] Data flow may have to be adjusted depending on where the `<AnimalsFilter />` is eventually placed within the inventory component structure
  - This will be pretty trivial to move by calling `useFilteredInventory` from a different component

**Description**

This is the filter functionality on animals -- filterSlice (LF-4094) and modal container (LF-4094). Set up mostly identically to transactions (particularly the component structure), except:
- Returned to the typical state pattern of using `{}` not `undefined` as the initial state for multi-option filter keys
- Transactions are filtered within the same hook where they are built; in inventory the inventory is built first and filtered afterwards; thank you @SayakaOno for this idea -- I like it!!

*VISUAL ISSUES I DID NOT ADDRESS*
  - As shown in tech daily the modal overlay is not interacting correctly with the top menu bar; this applies to the `<Drawer />` component in all contexts and @antsgar will be looking into it
  - The drawer state of `<Drawer />` could made into a MUI `<SwipeableDrawer />` but I think we decided not to scope that at this point
  - The modal state backdrop of `<Drawer />` is not as dark as in Figma. To allow the correct prop (which I couldn't identify anyway) to be passed through `<Drawer />` down to the MUI `<Dialogue />` was a refactor I did not want to undertake. Also this backdrop should arguably be darkened consistently throughout the app if at all
  - There is still the inconsistent (= sometimes appears, sometimes doesn't!) behaviour for the `[+ Load more]` button as @Duncan-Brain showed us in tech daily

Jira link: https://lite-farm.atlassian.net/browse/LF-4093 | https://lite-farm.atlassian.net/browse/LF-4094

Note: Apologies for the inconsistent branch + commit names; I worked on LF-4093 and LF-4094 as a unit and named the commits as LF-4094, but then opened under LF-4093 as I figured the modal name was more clear.

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/webapp/src/containers/Animals/Inventory/KPI/index.tsx
```diff
@@ -115,7 +115,7 @@ function KPI({ isCompactSideMenu, selectedTypeIds, onTypeClick }: KPIProps) {
         <PureTileDashboard
           typeCountTiles={types}
           dashboardTitle={t('SECTION_HEADER.ANIMALS_INVENTORY')}
-          categoryLabel={t('TYPES')}
+          categoryLabel={t('common:TYPES')}
```
kathyavini, line 118: Passing the filename to `useTranslation` works in-app, but the `pnpm i18n` script did not detect it and generated new keys within `translation.json`

packages/webapp/src/components/Drawer/index.tsx
```diff
@@ -29,63 +47,69 @@ const Drawer = ({
   classes = {
     modal: '',
     drawer: '',
-    backdrop: '',
```
kathyavini, line 32: I wanted to be explicit (because I was using them assuming something different -- i.e. that I could use the backdrop class to style both the drawer + the modal) that these props only apply to the drawer state; there actually was zero way to style the modal state using this classes object as written.

packages/webapp/src/components/Drawer/index.tsx
```diff
@@ -29,63 +47,69 @@ const Drawer = ({
...
+      titleClassName={styles.title}
+      dismissModal={onClose}
+      buttonGroup={buttonGroup}
+    >
+      <div className={styles.modalContent}>{children}</div>
+    </ModalComponent>
   ) : (
     <>
       <div
-        className={clsx(styles.backdrop, isOpen ? styles.openC : '', classes.drawerBackdrop)}
```
kathyavini, line 57: The click away and filtered overlay behaviour of the drawer were both not working because the correct classnames had become jumbled and the stylesheet styles were no longer applied.

packages/webapp/src/components/Filter/FilterMultiSelect/index.tsx
```diff
@@ -54,6 +54,8 @@ export const FilterMultiSelect = ({
         defaultFilterState[option.value].active = true;
       }
     });
+
+    onChange?.(filterRef.current![filterKey]);
   }, [value]);
```
kathyavini, line 59: Changing the value passed to this `onChange` should also now make the filter multi-select compatible with Report Filters; however for the purpose of using in animals (`useVisibleBreeds` makes use of the current `filterRef`) the change in the timing is the only thing that matters.

packages/webapp/src/components/Modals/ModalComponent/v2/index.jsx
```diff
@@ -23,6 +23,7 @@ export default function ModalComponent({
   warning,
   error,
   tooltipContent,
+  className = '',
```
kathyavini, line 26: `className` was passed by `<Drawer />` but never applied.

packages/webapp/src/containers/Animals/Inventory/useAnimalInventory.tsx
```diff
@@ -229,6 +353,7 @@ const useAnimalInventory = () => {
     { label: 'customAnimalTypes', hook: useGetCustomAnimalTypesQuery },
     { label: 'defaultAnimalBreeds', hook: useGetDefaultAnimalBreedsQuery },
     { label: 'defaultAnimalTypes', hook: useGetDefaultAnimalTypesQuery },
+    { label: 'animalSexes', hook: useGetAnimalSexesQuery },
```
kathyavini, line 356: Data not used by inventory, but I wanted it to get into the aggregate `isLoading` state.

SayakaOno (commented):
I know you followed what's done for finance transaction, but is it possible to filter the original `inventory`? I'm thinking about having `inventory` and `filteredInventory` in `AnimalInventory` component. Let me try to explain...

Current Process:
=== Initial Load ===
- Animal: Filter -> Map
- Batch: Filter -> Map
- Sort Animal & Batch

=== Changes to Filter ===
- Animal: Filter -> Map
- Batch: Filter -> Map
- Sort Animal & Batch

=== Changes to Filter ===
- Animal: Filter -> Map
- Batch: Filter -> Map
- Sort Animal & Batch

Proposed Approach:
=== Initial Load ===
- Animal: Filter -> Map
- Batch: Filter -> Map
- Sort Animal & Batch

=== Changes to Filter ===
- Filter `inventory` for Updates

=== Changes to Filter ===
- Filter `inventory` for Updates

Does it make sense? 😅 It would be great if we could avoid redundant operations.

kathyavini:
@SayakaOno I think so! That was actually how I assumed filters worked before I looked into transactions. But just to make sure I understand, the proposed solution would be like this, right (so we have a base unfiltered inventory to work with)?

Proposed Approach:
=== Initial Load ===

Animal -> Map
Batch -> Map
Sort Animal & Batch
=== Changes to Filter ===

Filter inventory for Updates

=== Changes to Filter ===

Filter inventory for Updates

SayakaOno:
@kathyavini Ah yes sorry 🙈 You got it right!!

kathyavini:
@SayakaOno DONE! I think I like it a lot better this way. I needed to keep some filtering IDs which the inventory array hadn't been tracking, but this has to be more efficient than rebuilding the inventory on each filter change.

And I love encapsulating logic into custom hooks so there is that too 😂

SayakaOno (commented):
Working great! And I love the extracted filter logic, thank you!!

- I am not sure when it started, but doesn't it look like the pill has got an extra padding??
Screenshot 2024-03-08 at 12 50 25 PM

- Should we ask Anto to look into the issue that the drop down of the Location select is cut off? (and the component is moving up in the background...)

https://github.com/LiteFarmOrg/LiteFarm/assets/33141219/b6ca7993-37ff-4b45-b025-ab9e17ced9ca

- There's one more thing I would like to bring up... I hope the animation becomes seamless when opening the filter modal for the first time after loading the app. (the action menu moves, so I think something is happening)

https://github.com/LiteFarmOrg/LiteFarm/assets/33141219/22fff4f0-75b2-40af-8d4c-4c29ce0a70d6

packages/webapp/public/locales/en/translation.json
```diff
@@ -161,8 +161,19 @@
     "ANIMAL_ID": "ID",
     "ANIMAL_IDENTIFICATION": "Identification",
     "ANIMAL_LOCATIONS": "Location",
+    "ANIMAL_SEXES": "Sex",
     "ANIMAL_TYPE": "Type"
   },
+  "ANIMALS": {
```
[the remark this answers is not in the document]
kathyavini: Gosh I don't know; I think I had some sort of weird impulse to not interpolate into the existing keys as they seemed to be a unit??! 😅 But of course we should nest everything under `ANIMAL`

packages/webapp/src/util/animal.ts
```diff
@@ -13,9 +13,21 @@
...
  */
 
-import { CustomAnimalType, DefaultAnimalType } from '../store/api/types';
+import {
+  CustomAnimalType,
+  DefaultAnimalType,
+  CustomAnimalBreed,
+  DefaultAnimalBreed,
+} from '../store/api/types';
 import { ANIMAL_TYPE_ID_PREFIX } from '../containers/Animals/types';
```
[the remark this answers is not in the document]
kathyavini: Then it could be shorter too! 🙏

packages/webapp/src/containers/Filter/Animals/index.tsx
```diff
@@ -0,0 +1,183 @@
...
+  filterRef,
+  filterContainerClassName,
+  onChange,
+}: AnimalsFilterContentProps) => {
+  const { t } = useTranslation(['translation', 'filter']);
+
+  const locations: Location[] = useSelector(locationsSelector);
+
+  const { data: defaultTypes = [] } = useGetDefaultAnimalTypesQuery();
+  const { data: customTypes = [] } = useGetCustomAnimalTypesQuery();
```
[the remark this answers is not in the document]
kathyavini: I was also unsure about this, because that's exactly how I initially interpreted Loïc's `ON YOUR FARM` Figma note. But I brought this question to tech daily earlier this week and we decided it actually just mean types _available_ to your farm (default + your own custom) rather than types _active_ on your farm since that is more true to the existing filter pattern.

And then as  a bonus it's much easier to implement since the note is also applied to breed, and we have no `count` setup on breed at all!

kathyavini:
@SayakaOno thank you for the review and suggestions!!

I'm afraid I don't know about those visual things, other than the multi-select filter pill... it looks via `git blame` that the padding has probably _always_ been that way! 😄 (I think we have tended to use `FilterPillSelect` a lot more so maybe that detail has flown under the radar?) The multi-select is getting re-written, but might as well fix it in the meantime!

Yes, I'm going to defer to @antsgar (= I don't know how to fix! 😅) the modal behaviour re:cutoff, and I'm not quite sure what that should look like in the end. If we could discuss the issue in tech daily/as a group coding exercise, I'd love that though, as the working with the MUI modal/`<Dialogue />` really frustrates me!

I'm not quite sure what the correct animation behaviour would be so could we discuss that in tech daily as well? 🙏 Do you think the action bar animation should be triggered in that instance? I would think no... right?

packages/webapp/src/containers/Filter/Animals/index.tsx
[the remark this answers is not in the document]
SayakaOno: Sorry I can't remember the discussion...
I thought it would be nice if breeds of types that I don't have in my farm were hidden, but I understand!

SayakaOno:
@kathyavini Ah I see! I was comparing the pill with removable `FilterPill`!
The action menu bar animation is annoying so I will get rid of it, but it happens only once and it is not supposed to happen. I'm curious what triggers it, but I'm okay with just ignoring it since it's not a big deal :)

SayakaOno (dismissed):

packages/webapp/src/components/Filter/FilterMultiSelect/index.tsx
antsgar: Not 100% sure but I think this might be solvable by getting rid of this `useEffect` altogether and calling things in the right order in the `onChange` function:

```jsx
const handleChange = (value: ComponentFilterOption[]): void => {
    setValue(value);
    filterRef.current![filterKey] = produce(defaultFilterState, (defaultFilterState) => {
      for (const option of value) {
        defaultFilterState[option.value].active = true;
      }
    });
    onChange?.(value as unknown as FilterState);
    /* NOTE: What is being passed here is in ComponentFilterOption format, and would not work if the state-setting onChange had been passed. However, the finance report filters -- the only container using a state-setting onChange -- don't use multi-select filters. */
  };

  return (
    <ReactSelect
      //@ts-ignore
      style={style} // I suspect the forwardRef is at fault for the type error?
      placeholder={`${t('common:SHOWING_ALL')}...`}
      options={options}
      label={subject}
      value={value}
      onChange={handleChange}
      isMulti
      isSearchable
      className={className}
    />
  );
```

Long term, I feel pretty strongly that we shouldn't be using refs but a state variable to store the filter's state (I'm not sure why we ended up with a ref), so the less we have to rely on that ref the easier it'll be to get rid of it.

packages/webapp/src/containers/Filter/Animals/index.tsx
```diff
@@ -0,0 +1,183 @@
...
+      options: [
+        ...filteredCustomBreeds.map((animalBreed) => ({
+          value: generateUniqueAnimalId(animalBreed),
+          default:
+            animalsFilter[AnimalsFilterKeys.BREED][generateUniqueAnimalId(animalBreed)]?.active ??
+            false,
+          label: animalBreed.breed,
+        })),
+        ...filteredDefaultBreeds.map((animalBreed) => ({
+          value: generateUniqueAnimalId(animalBreed),
```
antsgar, line 123: I'm not sure where we landed on this one, but would using the `key` for default breeds and the `name` for custom breeds be an option here for value so that we don't have to prefix?

antsgar (commented):
Looks and works great! 👏  Left a few minor comments, I think the only one worth looking into before merging is the one related to the change to `onChange`

packages/webapp/src/containers/Filter/Animals/index.tsx
kathyavini: I remember discussing, but I'm not sure we decided. I think we also said that to go that route we need a check on the backend to make sure that a user doesn't create a custom that duplicates an existing default (although I guess it would have to be in English and in all caps to actually create a conflict?) and since we didn't get into details of what and when to enforce that, I left it!

I think we will need that check before animals is released anyway though, so we can add the frontend change now. Should it apply to types (and the KPI) as well?

packages/webapp/src/components/Filter/FilterMultiSelect/index.tsx
kathyavini: Thanks!!! Although in testing that change I realized I was missing a `useEffect()` in `useVisibleBreeds`...that needs the `filterRef` 😅 

That's interesting because personally I love the `filterRef` -- I think it's tied to how I dislike the React render cycle, but the idea a value that _just is_ independent of the component's render cycle is such a relief 😂  But it's use here is probably tied how the component uses a different data format than the Redux state (source of so much mess and complexity) -- and I bet they could both be eliminated together.

packages/webapp/src/components/Drawer/index.tsx
```diff
@@ -29,63 +47,69 @@ const Drawer = ({
...
   },
   fullHeight,
   responsiveModal = true,
-}) => {
+}: DrawerProps) => {
   const theme = useTheme();
   const isDesktop = useMediaQuery(theme.breakpoints.up('sm'));
 
+  if (!isOpen) {
+    return null;
```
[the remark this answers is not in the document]
kathyavini: This change was 100% to please the type checker; I think it was complaining that some logical combinations did not return JSX or `null`? But if there is another way to solve that error I would be very interested; I don't love this!

packages/webapp/src/containers/Filter/Animals/index.tsx
antsgar: I'm good with keeping as is for now!

packages/webapp/src/components/Drawer/index.tsx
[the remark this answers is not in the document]
antsgar: Ah, not sure! Might be the condition on line 43 previously that should be `isOpen ? <ModalComponent ... > : null`? Anyway not a big deal!

antsgar (approved):

PR #3152, opened 2024-03-07
Lf 4083 add search to animal list

Duncan-Brain:
**Description**

Adds search bar to Inventory page. 

Does not: 
- add stickyness as discussed on ticket and now handled by [LF-4142 Adjust inventory layout](https://lite-farm.atlassian.net/browse/LF-4142)
- Change where the Layout component is from component to container and the search bar utilities. @kathyavini is it right I thought I heard you mention you wanted it elsewhere but told me not to change it yet -- are you wanting to tackle that in [LF-4104](https://lite-farm.atlassian.net/browse/LF-4104)?

Jira link: [LF-4083](https://lite-farm.atlassian.net/browse/LF-4083)

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

[LF-4104]: https://lite-farm.atlassian.net/browse/LF-4104?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LF-4083]: https://lite-farm.atlassian.net/browse/LF-4083?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kathyavini:
@Duncan-Brain no I think the idea of moving it was nixxed altogether since it was about maintaining the same structure as transactions (and keeping the container/component distinction) which were both deemed arbitrary. If we aren't worried about either of those two points there is no need to move where the data gets manipulated.

kathyavini:
So did you and Loïc decide not to implement the [modal effect](https://www.figma.com/file/eMffC0h3JPlWtzNttxJpil/Litefarm-Updated?type=design&node-id=2918-63566&mode=dev) when clicking into the searchbar on mobile? That's what all that resize observer code and `searchOverlayOpen` code in `<PureCollapsibleSearch />` is about.

I actually don't mind not having the effect (and wouldn't mind removing it for transactions, as we should be consistent and the need for resize observer is kind of annoying -- you can see how all that logic bulked up that component file), but if you aren't going to for the modal effect then is no need to use `<PureCollapsibleSearch />` as your base component -- you can just use regular `<PureSearchbarAndFilter />`. The way you wrapped all the additional logic in `collapse` you kind of re-generated the base searchbar anyway 😉 

Personally I vote for removing the modal effect in both places, but I'm curious what the team thinks.

Duncan-Brain:
@kathyavini Good question, I think I missed that screen haha. 

I agree I don't need it personally but I will make it the same for now!

kathyavini:
@Duncan-Brain I dunno; it's kind of a mess as written, isn't it? It's my component so I take responsibility 😅 I'm thinking of trying to re-write it so the modal functionality and the collapsing functionality reside in separate components; I assumed when I made it that this was going to be a new pattern for all searches so I made those two things too dependent on each other.

packages/webapp/src/components/Animals/Inventory/index.tsx
```diff
@@ -12,46 +12,101 @@
...
+        <div
+          className={clsx(isMobile ? styles.searchBar : styles.searchBarDesktop)}
+          ref={searchContainerRef}
+        >
+          <PureCollapsibleSearch
+            value={searchString}
+            onChange={(e) => setSearchString(e.target.value)}
+            isSearchActive={!!searchString}
+            containerRef={searchContainerRef}
+            placeholderText="Search your inventory"
```
SayakaOno, line 83: Hard-coded! 😆

packages/webapp/src/components/Animals/Inventory/index.tsx
```diff
@@ -12,46 +12,101 @@
...
+            collapse={false}
+            isDesktop={isDesktop}
+          />
+        </div>
+        <div
+          className={clsx(isMobile ? styles.resultsCount : styles.resultsCountDesktop)}
+        >{`Showing ${filteredInventory.length} results`}</div>
```
SayakaOno, line 90: This should also go in translation files eventually.

packages/webapp/src/components/Animals/Inventory/index.tsx
Duncan-Brain: Hmm I have been missing a lot of little things lately haha

Duncan-Brain:
Took @kathyavini advice and just copied and cut the non desired pieces for refactoring another day!

packages/webapp/public/locales/en/translation.json
```diff
@@ -161,7 +161,9 @@
     "ANIMAL_ID": "ID",
     "ANIMAL_IDENTIFICATION": "Identification",
     "ANIMAL_LOCATIONS": "Location",
-    "ANIMAL_TYPE": "Type"
+    "ANIMAL_TYPE": "Type",
+    "SEARCH_INVENTORY_PLACEHOLDER": "Search your inventory",
+    "SEARCH_RESULTS_TEXT": "Showing {{count}} results"
```
SayakaOno, line 166: I wonder if we need to do something like this:
```
"SPECIES_COUNT_one": "{{count}} species",
"SPECIES_COUNT_other": "{{count}} species",
```

packages/webapp/src/components/PopupFilter/PureCollapsibleSearch/styles.module.scss
```diff
@@ -96,24 +96,35 @@
...
+//   .circleContainer,
+//   .circle {
+//     display: none;
+//   }
+
+//   // Only relevant if resizing screen beyond breakpoint with modal open
+//   .modalContent {
+//     display: none;
+//   }
+// }
```
SayakaOno, line 130: Can these be removed?

packages/webapp/src/components/PopupFilter/PureSearchWithBackdrop/index.jsx
```diff
@@ -0,0 +1,75 @@
+/*
+ *  Copyright 2023 LiteFarm.org
```
SayakaOno, line 2: `2024`?

packages/webapp/src/components/PopupFilter/PureSearchWithBackdrop/styles.module.scss
```diff
@@ -0,0 +1,26 @@
+/*
+ *  Copyright 2023 LiteFarm.org
```
SayakaOno, line 2: `2024`?

packages/webapp/src/components/PopupFilter/PureSearchWithBackdrop/index.jsx
SayakaOno, line 1: Could this component be used for the finance search? I wasn't expecting a new component, so I'm just making sure.

packages/webapp/src/components/PopupFilter/PureCollapsibleSearch/index.jsx
```diff
@@ -44,45 +46,54 @@ export default function PureCollapsibleSearch({
 
   // Position modal searchbar
   useLayoutEffect(() => {
-    let resizeObserver;
```
kathyavini, line 47: I think you can revert the changes in this component + its styles + its Storybook since you are not using it (and your changes didn't exactly address collapse).

But I'm also interested in the answer to @SayakaOno's question! Can the backdrop solution be used for finances? I am loving `<Backdrop />` -- it looks way the heck simpler!

kathyavini (commented):
Wait by using `<Backdrop />` you were able to skip needing a resize observer???! 😍 Okay I want that in the old component now so much!!!

There is just one range of screen widths (tablet range) where everything goes insane -- is this a byproduct of `<Backdrop />` or unrelated?

Screenshot 2024-03-11 at 5 09 07 PM

packages/webapp/src/components/PopupFilter/PureSearchWithBackdrop/index.jsx
Duncan-Brain: No it does not collapse or make an overlay dialog component. So it cannot be used for finance.

packages/webapp/public/locales/en/translation.json
Duncan-Brain: Cool added that! Thanks

packages/webapp/src/components/PopupFilter/PureCollapsibleSearch/index.jsx
Duncan-Brain: No I think the existing Dialog is necessary because it overlays a new search bar "as a dialog". Backdrop as I am using it just hides everything and I am lifting up the input higher than the backdrop using the zIndex and themes (saw the` (theme) => theme.zIndex + 1  ` trick in a MUI example somewhere. Hopefully we can use that to fix our other layout issues too with the menus.

This reminds me I need to make the zIndex conditional to not mess with drawer.

Duncan-Brain:
@kathyavini Yeah thanks I def missed that weird behaviour. It's because of the Table I didn't update that to the mobile behaviour in line with the designs -- fixed now.  

I will remove the collapse and update stories but then rename it from Collapsible to Collapsing so as not to be misleading.

But it raised some interesting issues there that I will ignore for now haha, but collapsible icon does not seem to be responsive to the menu, and layout should probably overflow right creating a scrollbar instead of forcing the flex state of the menu.

packages/webapp/src/components/PopupFilter/PureCollapsibleSearch/index.jsx
Duncan-Brain: Well maybe it could be used by creating a hidden input that displays conditionally...

packages/webapp/public/locales/en/translation.json
SayakaOno: Sorry Duncan, the library handles this. I should've shared this 🙏 
https://www.i18next.com/translation-function/plurals#singular-plural

packages/webapp/src/stories/PureSearchBarWithBackdrop/PureSearchBarWithBackdrop.stories.jsx
```diff
@@ -0,0 +1,60 @@
+/*
+ *  Copyright 2023 LiteFarm.org
```
SayakaOno, line 2: `2024` lol

packages/webapp/public/locales/en/translation.json
Duncan-Brain: ooo purdy!

packages/webapp/src/stories/PureSearchBarWithBackdrop/PureSearchBarWithBackdrop.stories.jsx
Duncan-Brain: 😢 lol

SayakaOno (commented):
It's working well!
I'm wondering:
1. whether we should use the `isDesktop` prop rather than breakpoints in scss?
2. switching with or without the table header at 1200px seems a bit wide to me. (but I understand < 1200px is considered as tablet size) Can we find the requirement in Figma?  (this is a general question...)

packages/webapp/src/stories/PureSearchBarWithBackdrop/PureSearchBarWithBackdrop.stories.jsx
```diff
@@ -0,0 +1,60 @@
...
+  return (
+    <Suspense fallback="loading">
+      <div
+        style={{
+          outline: '1px dashed',
+          display: 'flex',
+          justifyContent: 'space-between',
+          gap: '20px',
+          padding: '16px',
+        }}
```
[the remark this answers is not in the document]
Duncan-Brain: .. I just copied the collapsing story and took some extras with it. I personally don't care for stories if its used but I can change/delete.

packages/webapp/src/components/PopupFilter/PureCollapsingSearch/index.jsx
```diff
@@ -22,13 +22,14 @@ import Input from '../../Form/Input';
...
 
-export default function PureCollapsibleSearch({
+export default function PureCollapsingSearch({
   value,
   isSearchActive,
   onChange,
   placeholderText,
   className,
   containerRef,
+  isDesktop,
```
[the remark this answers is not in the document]
Duncan-Brain: It was a media query -- no we should not use media query I don't think. 

- `mixin.scss` has breakpoints that are OK since we can change the whole app at the same time. If this is the preferred choice I can change 🤷 .

- isDesktop is nice in my personal opinion because:
  - we copied MUI breakpoints to make the `mixin.scss` so now we have 2 copies of source of truth
  - Using good CSS class naming like [BEM](https://medium.com/trabe/a-comprehensive-guide-to-using-bem-with-react-14d00c199e0d) I think javascript files can be slightly more readable versus hidden in CSS (I thought this was blocker so I rushed naming but fixed to BEM) Screenshot 2024-03-12 at 4 30 27 PM
  -  sometimes we want to use isDesktop for in page content or design decisions too so I enjoy letting MUI control our design system as a whole (since they are the experts)
Screenshot 2024-03-12 at 4 31 20 PM

packages/webapp/src/components/Animals/Inventory/index.tsx
```diff
@@ -14,44 +14,90 @@
  */
 import Table from '../../../components/Table';
 import Layout from '../../../components/Layout';
+import PureSearchBarWithBackdrop from '../../PopupFilter/PureSearchWithBackdrop';
+import NoSearchResults from '../../../components/Card/NoSearchResults';
 import type { AnimalInventory } from '../../../containers/Animals/Inventory/useAnimalInventory';
 import { TableV2Column, TableKind } from '../../Table/types';
-import type { DefaultTheme } from '@mui/styles';
+import type { Dispatch, SetStateAction, MutableRefObject } from 'react';
```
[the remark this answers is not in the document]
Duncan-Brain: removed!

packages/webapp/src/components/Animals/Inventory/index.tsx
```diff
@@ -14,44 +14,90 @@
...
-        columns={animalsColumns}
-        data={tableData}
-        shouldFixTableLayout={!isMobile}
-        minRows={tableData.length}
-        dense={false}
-        showHeader={!isMobile}
-      />
+      <div
+        className={clsx(
+          isDesktop ? styles.searchAndFilterContainerDesktop : styles.searchAndFilterContainer,
```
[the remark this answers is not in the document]
Duncan-Brain: We can if thats what is preferred -- linking other discussion thread:  https://github.com/LiteFarmOrg/LiteFarm/pull/3152#discussion_r1522083032

packages/webapp/src/stories/PureSearchBarWithBackdrop/PureSearchBarWithBackdrop.stories.jsx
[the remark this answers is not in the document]
Duncan-Brain: deleted!

packages/webapp/src/components/PopupFilter/PureCollapsingSearch/index.jsx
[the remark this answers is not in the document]
kathyavini: I really like having the breakpoints being controlled by JS/TS and no longer the mixins... those multiple methods and sources of truth (when finances was created we weren't yet unified around the MUI breakpoints) made it really hard to reason about e.g. the CollapsibleSearch.

But GOSH I hate those BEM underscores... do we have to?? 🤣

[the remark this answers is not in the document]
Duncan-Brain: lol no not necessary -- but its a good architecture vs "willy nilly" haha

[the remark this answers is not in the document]
SayakaOno: Okay, I was in favour of using CSS. I don't think we have discussed which strategy to stick with? I added to the tech daily agenda!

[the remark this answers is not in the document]
Duncan-Brain: @SayakaOno sounds awesome! I don't have a strong opinion between mixins and inline breakpoints just no more media queries haha

packages/webapp/src/stories/PureSearchBarWithBackdrop/PureSearchBarWithBackdrop.stories.jsx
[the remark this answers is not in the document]
SayakaOno: I was kind of hoping you wouldn't delete them 😂 I didn't really want to create a SCSS file just for one style... but I guess I will...

packages/webapp/src/stories/PureSearchBarWithBackdrop/PureSearchBarWithBackdrop.stories.jsx
SayakaOno: It looks like you updated another file? 😂

Duncan-Brain: lol dangit! 😂

packages/webapp/src/components/PopupFilter/PureCollapsingSearch/index.jsx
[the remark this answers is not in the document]
Duncan-Brain: I don't think we made a decision today. @SayakaOno can I make a tech debt ticket to choose ? That way search bar can go through.

packages/webapp/src/stories/PureSearchBarWithBackdrop/PureSearchBarWithBackdrop.stories.jsx
Duncan-Brain: Okay! I think I got it this time! hahaha

packages/webapp/src/components/PopupFilter/PureCollapsingSearch/index.jsx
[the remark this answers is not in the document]
SayakaOno: I thought we wouldn't need any further discussion 😅 It's up to you!

packages/webapp/src/components/Animals/Inventory/index.tsx
```diff
@@ -14,44 +14,95 @@
...
-        columns={animalsColumns}
-        data={tableData}
-        shouldFixTableLayout={!isMobile}
-        minRows={tableData.length}
-        dense={false}
-        showHeader={!isMobile}
-      />
+      <div
+        className={clsx(
+          isDesktop ? styles.SearchAndFilter_container___desktop : styles.SearchAndFilter_container,
```
antsgar, line 66: I like BEM overall, but since we haven't made a decision to go for that throughout the app, I'd advocate for consistency here and naming the styles more similarly to how classes are currently named in other files. In general I'm all for BEM but I think these will be out of place unless there's overall consensus to go with BEM from now on. If we do make a decision to go with BEM from now on, I think the convention would actually be something like `search-and-filter__container--desktop`. Unless this is a different one and not BEM?

packages/webapp/src/components/Animals/Inventory/index.tsx
```diff
@@ -14,44 +14,95 @@
...
+          onChange={(e: any) => setSearchString(e.target.value)}
+          isSearchActive={!!searchString}
+          placeholderText={placeHolderText}
+          zIndexBase={zIndexBase}
+          isDesktop={isDesktop}
+          className={clsx(isDesktop ? styles.SearchBar___widthFixed : styles.SearchBar___widthFull)}
+        />
+        <div
+          className={clsx(
+            isDesktop ? styles.SearchResults___marginLeft8 : styles.SearchResults___textAlignCenter,
```
antsgar, line 80: I'd try to give these classes more generic names rather than names that specifically point out the styling applied. The reason for that is, it leaves little flexibility to update the styling and reuse that selector without having to change the name of the class. Even if we went for BEM, the "modifiers" used are usually things like "compact", "disabled", "active", "big" rather than calling out specific styles

antsgar (commented):
Left a couple of comments re the CSS naming conventions! Other than that it works great, excellent work 🙌

packages/webapp/src/components/Animals/Inventory/index.tsx
Duncan-Brain: I think there are some react problems with the - or something. I linked the article I used to try to apply it but it may be out of date. 

I just added it to discuss the isDesktop inline breakpoints readability.

packages/webapp/src/components/Animals/Inventory/index.tsx
Duncan-Brain: Yeah. I would only call out specific styles if it is the only style on the property. I don't see value in making the modifier less readable (in JS) when there is just one style. 

But yeah anyone wanting to add another style would probably rename it to be more generic. Most just showcasing the inline breakpoints readability -- I haven't mastered BEM by any means ahha but it is nice

Duncan-Brain:
@antsgar BEM is gone!

packages/webapp/src/containers/Animals/Inventory/index.tsx
```diff
@@ -107,14 +110,32 @@ function AnimalInventory({ isCompactSideMenu }: AnimalInventoryProps) {
...
+  const [searchAndFilteredInventory, searchString, setSearchString] = useSearchFilter(
+    filteredInventory,
+    makeAnimalsSearchableString,
+  );
+
+  const searchProps: SearchProps = {
+    searchString,
+    setSearchString,
+    placeHolderText: t('ANIMAL.SEARCH_INVENTORY_PLACEHOLDER'),
+    searchResultsText: t('ANIMAL.SHOWING_RESULTS_WITH_COUNT', { count: filteredInventory?.length }),
```
SayakaOno, line 136: Since you renamed, `count` should be `searchAndFilteredInventory.length`

Duncan-Brain: fixed ! 🙏

packages/webapp/src/components/PopupFilter/PureCollapsingSearch/index.jsx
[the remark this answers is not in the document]
Duncan-Brain: I think it would be nice to choose a direction haha: https://lite-farm.atlassian.net/browse/LF-4143

packages/webapp/src/components/Animals/Inventory/index.tsx
```diff
@@ -14,44 +14,85 @@
...
+          dense={false}
+          showHeader={isDesktop}
+        />
+      ) : (
+        <NoSearchResults
+          className={clsx(isDesktop ? styles.noSearchResultsDesktop : styles.noSearchResults)}
+          searchTerm={searchString}
+          includeFiltersInClearSuggestion
+        />
+      )}
```
SayakaOno, line 95: I looked into "+ Load more" button being shown sometimes.
In `Table`, `minRows` is the initial value of `data.length` and won't change.

In `Inventory` component, if you have 10 animals and filter out 6 animals and refresh the page, `minRows` is `4` and when you clear filters, "Load more" button shows up.

The same thing happens after `hasSearchResults` is changed to `false` to `true` here since `Table` is re-rendered.

(I hope it makes sense)

I will make a change to pass `inventory.length` as `minRows` from the container in my PR later.

SayakaOno (approved):
🙌

packages/webapp/src/components/Animals/Inventory/index.tsx
Duncan-Brain: Nice! Right that makes sense I think it doesn't need to be dynamic unless data is refetched.

antsgar (approved):

PR #3153, opened 2024-03-08
LF-4142: Adjust inventory layout

SayakaOno:
**Description**

- create `FixedHeaderContainer` component (the calculation that is currently done in `KPI` was moved to this component). There seem to be several variations of the [fixed header + body] pattern 😅 We can update this component as necessary. I hope this makes the review easier...
  
Screenshot 2024-03-18 at 1 13 28 PM

- update `Table` to: 
   -  take `spacerRowHeight` so that the last row of the table will not to be hidden by action menu (it looks better with this solution than adding some space after the table)
   -  take `headerBackgroundColor` to override grey background that Mui adds to sticky header by default
   - fix `fullColSpan` to take `checkbox` into account
- make search & filter sticky on tablet/mobile view
- make the table header sticky
- revert `z-index`s that I added in https://github.com/LiteFarmOrg/LiteFarm/pull/3147
- display table instead of `<NoSearchResults />` when there's no inventory data

Jira link: https://lite-farm.atlassian.net/browse/LF-4142

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/webapp/src/components/Animals/FixedHeaderContainer/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+ *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ */
+
+import React, { ReactElement, ReactNode, useLayoutEffect, useMemo, useRef, useState } from 'react';
+import { Paper, useMediaQuery } from '@mui/material';
+import { useTheme } from '@mui/styles';
+import { sumObjectValues } from '../../../util';
+import styles from './styles.module.scss';
+
+const globalNavbarHeight = {
```
kathyavini, line 22: These aren't dynamic, right? They could be uppercase (or enum)?

kathyavini (commented):
Wow, the magic that makes FixedHeaderContainer work goes completely over my head! Thank you for figuring out all that crazy stuff!!!

Small and quickly fixed thing: the `.moreContent` class of the KPI `<MoreComponent />` needs a z-index boost so it renders on top of the searchbar + filter area.

Screenshot 2024-03-19 at 10 27 31 AM

And thing which I'm hoping is minor, but honestly I have zero idea what might cause this... on this branch I can only input one character at a time into the searchbar, and then  focus is removed from the input and I have to click in again to continue typing (same for backspace). Do you see that as well? Input isn't cleared either, so it doesn't seem at first glance like a re-render... a bit mysterious to me!

packages/webapp/src/components/Animals/FixedHeaderContainer/index.tsx
SayakaOno: I changed constants to uppercase! Personally, I've been using camelCase for object and array constants since I found information that uppercase is not always used for non-primitive types. It's nice to make it consistent, but it looks like uppercase, camelCase and snake_case are mixed in our project... 😅

SayakaOno:
@kathyavini Thank you so much for finding those issues!!!

- I didn't want to add z-index directly to MoreComponent because it's a reusable component, but I did... Please let me know if there's a better solution!
- It was kind of minor, but I spent a lot of time figuring out what was going on with the searchbar 🤮  It was really bad, thank you for catching it!!

This was causing unmounting and remounting, so I got rid of it!
https://github.com/LiteFarmOrg/LiteFarm/blob/76a42c13787cbe54d9289fc4f00e3e1de516c56c/packages/webapp/src/containers/Animals/Inventory/index.tsx#L218-L254

packages/webapp/src/components/Animals/FixedHeaderContainer/index.tsx
kathyavini: Interesting about non-primitive types. I had not thought about it before and it does truly seem to be up to developer  convention. In reading into it, I am horrified to learn that Airbnb style guide recommends not capitalizing even primitive constants if they are not exported 🤣  But the way you did the object after edit seems like a good convention to me (I would have been guilty of even capitalizing even the keys); maybe we can start a mini-LF style guide for this!

kathyavini (approved):
I would not have suspected the `<Body />` code at all, that part seemed so innocuous!!!

It looks great. It did not occur to me to bothered by the z-index on the reusable component, but now I am thinking wherever it ends up in app, we need that part to be on top so it's all good 😉

packages/webapp/src/components/Animals/FixedHeaderContainer/index.tsx
SayakaOno: Oh, I didn't know that! Sounds good!!

antsgar (commented):
This works great, and I'm very impressed by how you handled the logic for all the layout calculations! However, I was thinking there may be a way of letting the layout flow without having to do the math for the height of the contents by using flexbox and the overflow property, which should hopefully be a little bit easier to maintain. We'd still have to account for the top menu height because that's outside of the layout we're working on, but we wouldn't need to do calculations to account for the header. It's a bit hard to explain in writing so I pushed a branch with what I'm thinking -- however it's totally possible that I missed something and this solution wouldn't work, so let me know if that's the case! The branch is `example-flexbox-layout`

SayakaOno:
@antsgar thank you so much for the feedback and the code example! That solution was eye-opening for me.
We still need the exact height for the table to make its header sticky, so I left some calculations, but please let me know if you see any further improvement.
I will finalize the space after the table after Loïc confirms.

antsgar (approved):
@SayakaOno glad it was helpful! ❤️ I had forgotten that the table header needed to be sticky too, your solution looks great 🙌 awesome work!

kathyavini (approved):
Very nice!

PR #3154, opened 2024-03-11
LF-4087: Create or modify containers to associate action bar and table

SayakaOno:
Adjustments are required after the search & filter PRs are merged.

**Description**

- Add unique ids to animal and batches
- Add checkboxes to Inventory table
- Implement selection functionality in ActionMenu
- Remove ActionMenu animation
- Fix routes
   I found that this causes re-render when collapsing/expanding the side menu, so I fixed it. (`component` -> `render`)
   ```
   <Route
      path={ANIMALS_INVENTORY_URL}
      component={(props) => <Inventory isCompactSideMenu={isCompactSideMenu} {...props} />}
   />
   ```

~~NOTE:~~
- ~~After selecting a table row, if you collapse/expand the side menu, the selection is cleared. This happens because the whole component is re-rendered. I will come back and check the behaviour after https://github.com/LiteFarmOrg/LiteFarm/pull/3153 is merged. (I'm hoping that layout change fixes the issue 🙏 )~~

Jira link: https://lite-farm.atlassian.net/browse/LF-4087

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/webapp/src/containers/Animals/Inventory/index.tsx
```diff
@@ -115,33 +112,96 @@ function AnimalInventory({ isCompactSideMenu }: AnimalInventoryProps) {
     [t, isMobile],
   );
 
+  // TODO: Adjust once LF-4093 & 4094 are merged
+  const filteredInventory = inventory.slice();
```
antsgar, line 116: I'm getting an error that `filteredInventory` is alredy declared so I think this line needs to be removed now that the filters PR got merged

antsgar (commented):
There's one line that needs to be removed, but the checkboxes and final action menu look 🔥!

packages/webapp/src/containers/Animals/Inventory/index.tsx
SayakaOno: I'm sorry, I totally forgot about this... It has been removed, thank you!!

antsgar (dismissed):

packages/webapp/public/locales/en/common.json
```diff
@@ -55,6 +59,8 @@
   "SAVE_CHANGES": "Save Changes",
   "SEARCH": "Search",
   "SELECT": "Select",
+  "SELECT_ALL_COUNT": "Select all {{count}}",
```
Duncan-Brain, line 62: Is "Select all 1" okay or does this need one of those special _one, _other modifiers?

packages/webapp/src/components/Table/TableV2.jsx
```diff
@@ -184,7 +184,6 @@ export default function TableV2(props) {
                 <TableRow
                   key={row.id || index}
                   onClick={(event) => handleRowClick(event, row)}
-                  isItemSelected={isItemSelected}
```
Duncan-Brain, line 187: Should this not use the MUI prop `selected={isItemSelected}` ?

https://mui.com/material-ui/api/table-row/

packages/webapp/public/locales/en/common.json
SayakaOno: What should it be? Should I ask Loïc?

packages/webapp/src/util/animal.ts
```diff
@@ -31,3 +33,8 @@ export const generateUniqueAnimalId = (
 ): string => {
   return `${ANIMAL_ID_PREFIX['key' in entity ? 'DEFAULT' : 'CUSTOM']}_${entity.id}`;
 };
+
+export const generateInventoryId = (animalOrBatch: Animal | AnimalBatch): string => {
+  const key = 'count' in animalOrBatch ? AnimalOrBatchKeys.BATCH : AnimalOrBatchKeys.ANIMAL;
```
Duncan-Brain, line 38: Instead of relying on 'count', I wonder if it might be okay to just pass the key from useAnimalInventory(). 

Also is `internal_identifier` not still passed that we could use?

packages/webapp/public/locales/en/common.json
Duncan-Brain: I don't have a good idea. "Select 1" doesn't make sense to me either...

packages/webapp/src/util/animal.ts
SayakaOno: That's what I initially did, and I don't mind changing it back.

We could use `internal_identifier`; I just thought the new key would make the actions easier.
For example, when removing animals and batches, we can just separate animal and batch keys and use their ids.
- animal keys [,,,,] => call remove animals API
- animal batch keys [,,,,] => call remove animal batches API

If selectedIds are `internal_identifier`, we need to find animals and batches from `inventory` again to get animal/batch and their ids. If that's better (it feels more reliable?), we can use `internal_identifier`.

Duncan-Brain:
This looks like it was a tough one to me! Nice job!

There are no blockers for me but I left a few inline comments:
- translation string - ( just a small thing )
- Mui styles - I dont think you need to change but it is good to get used to using their design system if they are our main component library
- internal identifier - I do wonder why we cant use that
- ActionBar seems maybe almost "Pure" it might be nice to see it in the component vs container

Duncan-Brain (approved):

packages/webapp/src/util/animal.ts
Duncan-Brain: Ohh I understand now. It is up to you. We also have `batch` property on the row too: 
Screenshot 2024-03-14 at 11 42 14 AM

antsgar (approved):

packages/webapp/src/util/animal.ts
SayakaOno: I am not sure how I can use that `batch` effectively...
I will leave it for now!

PR #3155, opened 2024-03-12
LF-4141 The animal inventory of the previous farm is shown to the user once he changes the farm without logging out

kathyavini:
**Description**

This adds RTK Query tag invalidation to `clearOldFarmStateSaga()` which is called during farm switching. For rationale and alternative approaches considered, please see Jira ticket.

Jira link: https://lite-farm.atlassian.net/browse/LF-4141

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/containers/saga.js
+++ b/packages/webapp/src/containers/saga.js
@@ -151,6 +151,7 @@ import {
   onLoadingWatercourseFail,
   onLoadingWatercourseStart,
 } from './watercourseSlice';
+import { api } from '../store/api/apiSlice';
 
 const logUserInfoUrl = () => `${url}/userLog`;
 const getCropsByFarmIdUrl = (farm_id) => `${url}/crop/farm/${farm_id}`;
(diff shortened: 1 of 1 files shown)


packages/webapp/src/containers/saga.js
```diff
@@ -619,6 +620,17 @@ export function* fetchAllSaga() {
...
+  yield put(
+    api.util.invalidateTags([
+      'Animals',
+      'AnimalBatches',
+      'AnimalGroups',
+      'CustomAnimalBreeds',
+      'CustomAnimalTypes',
+      'DefaultAnimalTypes', // needs to be cleared for KPI count
+    ]),
+  );
```
antsgar, line 632: Since almost all of the data in the app is dependent on the farm, I wonder if we should consider calling `api.util.resetApiState()` instead?

kathyavini: Well it would result in re-fetching the enum keys and the default breeds unnecessarily. But the advantage is, it would be a set it and forget it type situation 🤔

antsgar: That's a good point!

antsgar (approved):

PR #3159, opened 2024-03-14
LF-4122 handle additional removal reasons in removal flow

kathyavini:
**Description**

Integration of removal functionality with inventory.

Deletion (the `CREATED_IN_ERROR` option) is a separate ticket and selecting it will do nothing other than return a network error.

Jira link: https://lite-farm.atlassian.net/browse/LF-4122

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/webapp/src/containers/Animals/Inventory/index.tsx
```diff
@@ -231,6 +240,12 @@ function AnimalInventory({ isCompactSideMenu }: AnimalInventoryProps) {
               }}
             />
           ) : null}
+          <RemoveAnimalsModal
+            isOpen={removalModalOpen}
+            onClose={() => setRemovalModalOpen(false)}
+            onConfirm={handleAnimalOrBatchRemoval}
+            showSuccessMessage={false}
+          />
```
kathyavini, line 248: The success message text mentions historical inventory, so I don't think it needs to be displayed until that has been built.

packages/webapp/src/containers/Animals/Inventory/index.tsx
```diff
@@ -231,6 +240,12 @@ function AnimalInventory({ isCompactSideMenu }: AnimalInventoryProps) {
               }}
             />
           ) : null}
+          <RemoveAnimalsModal
+            isOpen={removalModalOpen}
+            onClose={() => setRemovalModalOpen(false)}
+            onConfirm={handleAnimalOrBatchRemoval}
+            showSuccessMessage={false}
```
kathyavini, line 247: Success message mentions historical inventory, so I am assuming it should not be shown before that has been built out.

packages/webapp/src/containers/Animals/Inventory/useAnimalInventory.tsx
```diff
@@ -153,26 +153,32 @@ const formatAnimalsData = (
...
-      // preserve some untransformed data for filtering
-      group_ids: animal.group_ids,
-      sex_id: animal.sex_id,
-      custom_type_id: animal.custom_type_id,
-      default_type_id: animal.default_type_id,
-      custom_breed_id: animal.custom_breed_id,
-      default_breed_id: animal.default_breed_id,
-    };
-  });
+  return animals
```
kathyavini, line 156: Ultimately this filter would probably be put somewhere else (once we have a view for removed animals), but for now they are simply removed from the table + KPI

packages/webapp/src/containers/Animals/Inventory/useAnimalOrBatchRemoval.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  const handleAnimalOrBatchRemoval = (formData: FormFields) => {
+    const timestampedDate = toLocalISOString(formData.date);
+
+    const animalRemovalArray = [];
+    const animalBatchRemovalArray = [];
+
+    for (const id of selectedInventoryIds) {
+      const { kind, id: entity_id } = parseInventoryId(id);
+      if (kind === 'ANIMAL') {
```
SayakaOno, line 41: I think it would be nice if `AnimalOrBatchKeys` is used here.

packages/webapp/src/util/animal.ts
```diff
@@ -38,3 +38,8 @@ export const generateInventoryId = (animalOrBatch: Animal | AnimalBatch): string
   const key = 'count' in animalOrBatch ? AnimalOrBatchKeys.BATCH : AnimalOrBatchKeys.ANIMAL;
   return `${key}_${animalOrBatch.id}`;
 };
+
+export const parseInventoryId = (inventoryId: string): { kind: string; id: number } => {
```
SayakaOno, line 42: Can `AnimalOrBatchKeys` be used instead of `string`?

SayakaOno (commented):
Working great! I like that the hook keeps `Inventory` so clean.
I left small comments :)

packages/webapp/src/components/Animals/RemoveAnimalsModal/index.tsx
```diff
@@ -70,39 +72,58 @@ export default function RemoveAnimalsModal(props: RemoveAnimalsModalProps) {
...
+  // Reset form when a new set of animals/batches is selected
+  useEffect(() => {
+    if (props.isOpen) {
+      reset({ [DATE]: getLocalDateInYYYYDDMM() });
+    }
+  }, [props.isOpen]);
+
+  const { data: apiAnimalRemovalReasons = [], isLoading } = useGetAnimalRemovalReasonsQuery();
+
+  const RemovalReasons = useMemo(() => {
```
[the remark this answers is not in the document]
kathyavini: I think I might have pulled it only for the `useMemo`, where it's completely redundant with `data` 😬

packages/webapp/src/containers/Animals/Inventory/useAnimalOrBatchRemoval.ts
kathyavini: Thank you!!

packages/webapp/src/util/animal.ts
kathyavini: I did have to type cast what's returned from the split; I don't know if that's unavoidable or not.

SayakaOno: I tried myself and did the same too 😅

kathyavini:
Thank you @SayakaOno!

Honestly I had completely forgotten that snackbar until I saw @Duncan-Brain's daily standup report today. If Duncan has work underway on that, I'll defer to his solution as I'm pretty sure it's one snackbar for both removal and deletion on Figma and I'm assuming they can be implemented pretty easily together -- @Duncan-Brain is that right?

Duncan-Brain:
@kathyavini Its a bit of a challenge lol. Thinking of doing 2 snackbars or rewriting api because non atomic between animals and batch. Mostly have solution but need to troubleshoot over rendering monday.

SayakaOno (dismissed):

packages/webapp/src/components/Animals/RemoveAnimalsModal/index.tsx
```diff
@@ -70,39 +72,58 @@ export default function RemoveAnimalsModal(props: RemoveAnimalsModalProps) {
   const theme = useTheme();
   const isMobile = useMediaQuery(theme.breakpoints.down('sm'));
 
+  // Reset form when a new set of animals/batches is selected
+  useEffect(() => {
+    if (props.isOpen) {
+      reset({ [DATE]: getLocalDateInYYYYDDMM() });
+    }
+  }, [props.isOpen]);
```
antsgar, line 80: If we want to reset the form to default every time the modal's open, I think this can be achieved by passing `shouldUnregister: true` to `useForm`

antsgar (approved):
Works great! 🙌 I noticed two UI things that aren't really related to this PR so no need to handle them here, but just noting them. The removal reasons dropdown would look better without the padding around the options, and on the Action Menu, it always says "Remove animal" even if there's more than one selected (could be generically "Remove animals" in plural or change depending on the number selected)

packages/webapp/src/components/Animals/RemoveAnimalsModal/index.tsx
kathyavini: Oh nice, thank you! I didn't know that reset to default value!

kathyavini:
@antsgar let's do "Remove animals" in all cases because the singular entity selected might be a batch... and it's also easier to implement 😉 

Could you show me which padding you mean? I actually specifically asked Nav to increase the padding on this dropdown beyond the default React Select 😅 -- but I was basing it on Loïc's [Figma design here](https://www.figma.com/file/eMffC0h3JPlWtzNttxJpil/Litefarm-Updated?node-id=5278%3A86914&mode=dev).

antsgar:
@kathyavini this spacing here!

Captura de pantalla 2024-03-20 a la(s) 16 53 55

I do see it in the designs too but I find it odd that the options don't go to the border of the dropdown. The top one's fine actually, it's the left right/margins that look weird to me, we could add in more spacing to the option's padding itself instead of around it?

kathyavini:
@antsgar ah so that the hover effect extends like this? That can definitely be done

Screenshot 2024-03-20 at 1 24 20 PM


Alas just noticed that the extra character in Remove Animals doesn't fit 😉

Screenshot 2024-03-20 at 1 24 26 PM

Will adjust, but I imagine will we want to revisit this after all the translations are in?

packages/webapp/src/components/ActionMenu/styles.module.scss
```diff
@@ -31,7 +31,7 @@
     left: 50%;
 
     &.iconCount_4 {
-      width: 408px;
+      width: 424px;
```
kathyavini, line 34: @SayakaOno this was my first thought for fitting the longer label, but now actually I think I slightly prefer the look of decreasing the gap from 8px to 4px; curious which you think is better?

SayakaOno: I think I set the gap to 8px because the space looked too narrow with 4px.
Screenshot 2024-03-20 at 3 20 52 PM

To me, an 8 pixel gap still seems better when thinking about other languages...

SayakaOno (approved):

antsgar (approved):

PR #3164, opened 2024-03-18
LF-4104 Associate filter actions with animal KPIs filtering on types

kathyavini:
**Description**

Jira link: https://lite-farm.atlassian.net/browse/LF-4104

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/containers/Animals/Inventory/KPI/useAnimalsFilterReduxState.ts
+++ b/packages/webapp/src/containers/Animals/Inventory/KPI/useAnimalsFilterReduxState.ts
@@ -1,0 +1,50 @@
+/*
+ *  Copyright 2024 LiteFarm.org
+ *  This file is part of LiteFarm.
+ *
+ *  LiteFarm is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  LiteFarm is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ */
+
+import { useDispatch, useSelector } from 'react-redux';
+import produce from 'immer';
+import { AnimalsFilterKeys } from '../../../Filter/Animals/types';
+import { animalsFilterSelector, setAnimalsFilter } from '../../../filterSlice';
+import { getActiveTypeIds } from '../../../Filter/Animals/utils';
+
+export const useAnimalsFilterReduxState = () => {
+  const dispatch = useDispatch();
+  const animalsFilter = useSelector(animalsFilterSelector);
+
+  const selectedTypeIds = getActiveTypeIds(animalsFilter);
+
+  const updateSelectedTypeIds = (typeIds: string[]) => {
+    let newFilterState = produce(
+      animalsFilter[AnimalsFilterKeys.TYPE],
+      (filterState: { [key: string]: { label?: string; active: boolean } }) => {
+        for (const key in filterState) {
+          filterState[key].active = false;
+        }
+
+        for (const key of typeIds) {
+          if (!filterState[key]) {
+            filterState[key] = { active: true };
+          } else {
+            filterState[key].active = true;
+          }
+        }
+      },
+    );
+
+    dispatch(setAnimalsFilter({ ...animalsFilter, [AnimalsFilterKeys.TYPE]: newFilterState }));
+  };
+
+  return { selectedTypeIds, updateSelectedTypeIds };
+};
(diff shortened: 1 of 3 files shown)


Duncan-Brain:
Looks great and really cool to use immer! The use cases for that are still pretty unknown to me so I would love to hear more about it sometime in tech daily. 

My main question what is the expected behaviour of clicking a KPI when a filter on another animals breed is already being filtered for. I am not sure what I expected but I think maybe it would select the KPI breeds in the filter too? (sorry the video is long I forgot how to find the stop button)


https://github.com/LiteFarmOrg/LiteFarm/assets/30075973/05686e97-6b3d-4194-afde-cbbac31cc8c8

kathyavini:
@Duncan-Brain I think the immer code was @antsgar's, I just copy-pasted from one of the filter components 😉

Re: the video though, that is some seriously buggy behaviour @0:18 that I did not catch -- if the types are not modified in the filter, they should not be selected or deselected from saving a change to breeds like that. Thank you I'm gonna look into that now!!

I'm not sure if that was your question though.... I didn't quite follow this part

> maybe it would select the KPI breeds in the filter too

since the KPI only containers types, not breeds. But hold on a minute because I think the bug made it look like more things were associated than should be (only the type selection should be associated between filter + KPI)

Duncan-Brain:
We might be seeing different things haha -- happy to discuss if needed. I will try to explain further

So my understanding is types and breeds are "connected" (maybe not in this PR?) . What I am saying is that I might expect that if I select the "Cattle" type -- then all cattle breeds are selected in filter until I remove some of them. When I remove the last breed of a type then the type also gets removed from the filter.

Why I said maybe I expect this (but maybe not). Is that maybe the expectation is that the filter should always be working on the visible results. So it would indeed be expected that selecting cattle does not increase the selected animals in the filter but in fact "further reduces" the results (in this existing case no effect because there are no cattle types of cornish cross chickens haha ).

kathyavini:
@Duncan-Brain got it! Okay, this is worth following up with Loïc on I think.

I agree with the idea that selecting more types should increase, not narrow the selection. But the relationship between type + breed described on Figma was that the types should narrow the possible breed selection (in the dropdown), not also auto-select the breeds 🤔

I guess another way to fix the narrowing issue would be to deselect all breeds once a new type is picked. But I'm not sure I like either approach because we're manipulating breed selection (either selecting or deselecting) independent from user interaction.

packages/webapp/src/containers/Animals/Inventory/KPI/useAnimalsFilterReduxState.ts
```diff
@@ -0,0 +1,50 @@
...
+            filterState[key] = { active: true };
+          } else {
+            filterState[key].active = true;
+          }
+        }
+      },
+    );
+
+    dispatch(setAnimalsFilter({ ...animalsFilter, [AnimalsFilterKeys.TYPE]: newFilterState }));
+  };
```
antsgar, line 47: This works great, but could be simplified by having this function receive the type that was clicked and its state rather than the whole state object from the component that's calling it. If the Inventory component, the onClick function could do

```js
const onTypeClick = useCallback(
    (typeId: string) => {
      const isSelected = selectedTypeIds.includes(typeId);
      updateSelectedTypeIds(typeId, isSelected);
    },
    [updateSelectedTypeIds],
  );
```

And then on the hook

```js
const updateSelectedTypeIds = (typeId: string, isSelected: boolean) => {
    const newFilterState = {
      ...animalsFilter[AnimalsFilterKeys.TYPE],
      [typeId]: { active: !isSelected },
    };
    dispatch(setAnimalsFilter({ ...animalsFilter, [AnimalsFilterKeys.TYPE]: newFilterState }));
  };
```

Basically you only keep track of the entire state object on the Redux store, and let the component only pass what's changed.

antsgar (commented):
Glad you could sort out the ref issue (hopefully we can get rid of that one soon), and love how this is working! I left a comment on a portion of the code that could be simplified

packages/webapp/src/containers/Animals/Inventory/KPI/useAnimalsFilterReduxState.ts
antsgar: Come to think of it we probably don't need to pass `isSelected` from the component either, since the hook already has access to `selectedTypeIds`, so it could just do

```js
const updateSelectedTypeIds = (typeId: string) => {
    const isSelected = selectedTypeIds.includes(typeId);
    const newFilterState = {
      ...animalsFilter[AnimalsFilterKeys.TYPE],
      [typeId]: { active: !isSelected },
    };
    dispatch(setAnimalsFilter({ ...animalsFilter, [AnimalsFilterKeys.TYPE]: newFilterState }));
};
```

Duncan-Brain (approved):
I no longer see that issue you pointed out in my video and the other thing is undecided by the sounds of it!

packages/webapp/src/containers/Animals/Inventory/KPI/useAnimalsFilterReduxState.ts
kathyavini: Gosh that's SO much cleaner 😍 I don't feel like I could have come up with that solution though, no matter how long I thought about it 😅 The `active: !isSelected` is such a good way to do the toggle!!

antsgar: It's easier when you already have someone else's code to read through and think through without being too involved 😅 glad it helps!

antsgar (approved):

Duncan-Brain (approved):
Still loogs good!

PR #3165, opened 2024-03-18
Lf 4119 display component for animal removal and handle delete reason

Duncan-Brain:
**Description**

Thanks @kathyavini for completing most of the points on this Jira ticket in your pr #3159 !

This PR creates a deletion handler and a choice for choosing between removal and deletion.

EDIT Now includes removal handler.

If differs from the scope in that it creates a snackbar for success and failure of each of batches and animals. If instead we grouped the success + failure it could result in a situation where the user deletes something and then we return a success message (or the opposite: the user successfully deletes something and we return an error even though the animals disappear from table).

Sorry it took a bit longer since I had a small amount of trouble copying the architecture of the solution from the RFC (keeping the snackbars in the container) [#3892](https://github.com/LiteFarmOrg/LiteFarm/pull/3019/files#diff-8db0fa9c368e05f4b570474ec9842334335cd873756d537a4bf97be90cf5ce10)   -- and I am still not super familiar with RTK mutations.

Because we use hooks here returning the lifecycle status of the mutations in combination with changing the props the hook using the handler resulted in some infinite snackbar problems when placing the snackbar handling in container. Maybe I was doing it wrong.. but thats what I saw.

So another reason I looked into the listeners and the [recommended side effects](https://redux.js.org/usage/side-effects-approaches#recommendations) from redux was that we might see some of our more complex logic from sagas creep into our containers. I don't love that because as we possibly move towards removing some distinctions between our existing containers and components these might get to be very heavy code files. Anyways we might want to look further into those!

Jira link:[ LF-4119](https://lite-farm.atlassian.net/browse/LF-4119)

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain):
  - comment out the batches or animals controller and return 400 response to see a combination error and success.

**Checklist:**


kathyavini:
@Duncan-Brain #3159 had some minor updates last week -- can you merge those changes too ? And it would be helpful to diff to that branch to start, with so it's clearer what code belongs to delete.

packages/webapp/src/containers/Animals/Inventory/useAnimalOrBatchRemoval.ts
```diff
@@ -0,0 +1,143 @@
...
+    const newArray = selectedInventoryIds;
+    removeTheseIds.forEach((removalId: string): void => {
+      const index = newArray.indexOf(removalId);
+      if (index > -1) {
+        // only splice array when item is found
+        newArray.splice(index, 1); // 2nd parameter means remove one item only
+      }
+    });
+    return newArray;
+  };
```
kathyavini, line 56: Could you please use filter instead of splice loop

```suggestion
  const removeFromSelectedIds = (removeTheseIds: string[]) => {
    return selectedInventoryIds.filter((id) => !removeTheseIds.includes(id));
  };
```

Duncan-Brain: Oh yeah thats better haha 😂  .. removed this function and converted to a one liner so no more goofy names either like removeTheseIds haha.

Duncan-Brain:
@kathyavini Thanks yeah I spaced out I guess and forgot to mark this as draft until yours was merged, and simultaneously merged integration instead of your branch. Woopsie!

packages/webapp/src/containers/Animals/Inventory/useAnimalOrBatchRemoval.ts
```diff
@@ -13,29 +13,43 @@
...
 
 const useAnimalOrBatchRemoval = (
   selectedInventoryIds: string[],
   setSelectedInventoryIds: Dispatch<SetStateAction<string[]>>,
+  dispatch: Dispatch<any>,
+  t: TFunction,
 ) => {
-  const [removeAnimals] = useRemoveAnimalsMutation();
-  const [removeBatches] = useRemoveAnimalBatchesMutation();
+  const { mutations } = useMutations([
```
SayakaOno, line 37: We sometimes need to combine the results of hooks for GET APIs using `useQueries`, but I don't see any benefit in combining mutations here... do we want to wait for two APIs to return the results and show snackbars at the same time?

packages/webapp/src/containers/Animals/Inventory/useAnimalOrBatchRemoval.ts
```diff
@@ -46,24 +60,90 @@ const useAnimalOrBatchRemoval = (
...
+
+  const handleAnimalOrBatchDeletion = async () => {
+    const animalIds: number[] = [];
+    const selectedAnimalIds: string[] = [];
+    const animalBatchIds: number[] = [];
+    const selectedBatchIds: string[] = [];
+
+    for (const id of selectedInventoryIds) {
+      const { kind, id: entity_id } = parseInventoryId(id);
+      if (kind === 'ANIMAL') {
```
SayakaOno, line 106: I asked Joyce to replace this with the enum `AnimalOrBatchKeys`! Could you please do the same?

packages/webapp/src/containers/Animals/Inventory/useAnimalOrBatchRemoval.ts
```diff
@@ -13,29 +13,43 @@
...
+import useMutations from '../../../hooks/api/useMutations';
+import { enqueueErrorSnackbar, enqueueSuccessSnackbar } from '../../Snackbar/snackbarSlice';
+import { TFunction } from 'i18next';
 import { AnimalOrBatchKeys } from '../types';
 
 const useAnimalOrBatchRemoval = (
   selectedInventoryIds: string[],
   setSelectedInventoryIds: Dispatch<SetStateAction<string[]>>,
+  dispatch: Dispatch<any>,
+  t: TFunction,
```
SayakaOno, line 35: Let me ask a question I've always had... (sorry Duncan)
@antsgar I'm not sure if I should pass `dispatch`/`t` or call `useDispatch()` /`useTranslation()` in this hook, or if I should pass `onAddAnimalsSuccess`, `onAddAnimalsFail`, `onAddBatchesSuccess` etc.
Do you have any advice on a good approach to take?

SayakaOno (commented):
It seems to be working great! I left small comments.

I have a question about how to use mutation hooks in general.
I think we also can do something like this:
```
const [removeXXX, { isError, isSuccess }] = useXXXMutation();

const handleRemoval = () => {
  removeXXX(); // without awaiting
}

useEffect(() => {
  if (isError) {
    // snackbar for failure
  }
}, [isError]);

useEffect(() => {
  if (isSuccess) {
    // snackbar for success
  }
}, [isSuccess]);
```
@antsgar Can we go either way? Is there anything to keep in mind?

packages/webapp/src/containers/Animals/Inventory/useAnimalOrBatchRemoval.ts
Duncan-Brain: There might be some value in the isSuccess (all succeed) and isLoading (still waiting) pieces... since I think those can be genericized. 

But the main usefulness I think is renaming 😂 . Having to rename all the isXXXLoading, isXXXError gets cumbersome in the same component. So having one object hold them all with a named label is useful to me.

packages/webapp/src/containers/Animals/Inventory/useAnimalOrBatchRemoval.ts
Duncan-Brain: Yeah for sure!

packages/webapp/src/containers/Animals/Inventory/useAnimalOrBatchRemoval.ts
Duncan-Brain: And I thiiiink we want to show them at the same time, it might be weird if they are staggered. 

It sounds like we are going to put effort towards the partial success in the near future so this is temporary

SayakaOno: Renaming - true, I've experienced the same thing 😅 

I don't mind subtle difference in timing, but I think all the `dispatch(...Snackbar())` should be written after those two try-catch blocks.

Oh it's temporary so you don't mind it too 😂

Duncan-Brain:
@SayakaOno regarding the useEffect route. I think that is possible and could work. But I am not sure is suggested way to do it. Here is the VSCode function docs I see when I use the function without await:
Screenshot 2024-03-27 at 3 31 15 PM

You can see the example returns a promise, and also has the suggested syntax I copied from @antsgar RFC.

Specifically for this case I actually tried adding the useEffect() to the inventory container with success and failure events (snackbar and` setSelectedIds()` ) and it caused an infinite loop on snackbar because the `setSelectedIds()` is a prop for the custom hook that calls the mutation. It is probably possible to fix that but I stuck with what was easy for now haha.

Happy to consider doing it that way but I don't have a clear idea how right now.

What do you think about RTK listeners for effects too? https://redux.js.org/usage/side-effects-approaches#recommendations

SayakaOno:
@Duncan-Brain 
I agree with the current approach, I just wanted to address my question before Anto leaves!
I think it would be easier to go with the other approach for a case where we have an API that takes a while to get a response and we want to display a spinner while waiting...

And thank you for trying! I can't fully understand without code example, but I would like to explore at the right time!

As for RTK listeners, I would like to try it! While I really like the current approach where the entire flow (remove animal -> snackbar) is written in one component, there should be the case we would like to use listeners...?

packages/webapp/src/hooks/api/useMutations.js
```diff
@@ -0,0 +1,43 @@
...
+ *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ */
+
+const useMutations = (namedMutations) => {
+  const results = namedMutations.map((mutation) => mutation.hook());
+  const data = results.map((result) => result[1]);
+
+  const isError = data.some((result) => result.isError);
+  const isLoading = data.some((result) => result.isLoading) && !isError;
+  const isSuccess = data.every((result) => result.isSuccess);
```
SayakaOno, line 22: I just realized that when only some of the mutations are actually called, even if all of them are successful, `isSuccess` won't be true.
For instance, we pass removeAnimals, removeBatches, deleteAnimals, and deleteBatches, but we will call only two of them at most (either removeAnimals and removeBatches, or deleteAnimals and deleteBatches) at the same time. I expect `isSuccess` to be true if all the APIs I called were successful, but it's not the case because the other APIs aren't called and `isSuccess` of their results won't be `true`.

Is there a solution?

packages/webapp/src/containers/Animals/Inventory/useAnimalOrBatchRemoval.ts
```diff
@@ -46,24 +60,90 @@ const useAnimalOrBatchRemoval = (
...
-    animalBatchRemovalArray.length && removeBatches(animalBatchRemovalArray);
+    try {
+      if (animalBatchIds.length) {
+        await mutations['deleteBatches'].trigger(animalBatchIds).unwrap();
+        setSelectedInventoryIds(selectedInventoryIds.filter((i) => !selectedBatchIds.includes(i)));
+        dispatch(enqueueSuccessSnackbar(t('ANIMALS.SUCCESS_REMOVE_BATCHES', { ns: 'message' })));
+      }
+    } catch (err) {
+      dispatch(enqueueErrorSnackbar(t('ANIMALS.FAILED_REMOVE_BATCHES', { ns: 'message' })));
+    }
```
antsgar: Since you've already grouped the mutations states in the `useMutations` hook (nice!), it'd be great to take advantage of that here to already start showing a single a single snackbar. I think the only change needed would be to get rid of the `dispatch` lines in this function and add a `useEffect` within this hook that reacts to changes in `inError` and `isSuccess` returned by `useMutations`

```js
 const { mutations, isError, isSuccess } = useMutations([
    { label: 'deleteAnimals', hook: useDeleteAnimalsMutation },
    { label: 'deleteBatches', hook: useDeleteAnimalBatchesMutation },
  ]);
useEffect(() => {
    if (isSuccess) {
      dispatch(enqueueSuccessSnackbar(t('ANIMALS.SUCCESS_REMOVE_BATCHES', { ns: 'message' })));
    }
    if (isError) {
      dispatch(enqueueErrorSnackbar(t('ANIMALS.FAILED_REMOVE_BATCHES', { ns: 'message' })));
    }
  }, [isError, isSuccess]);
```

We'd still have to implement the changes in the BE to allow partial success for each individual endpoint, but this way if either of the endpoints failed we'd display the single error snackbar so we'd be halfway there.

antsgar (commented):
Left a couple of suggestions but it's working great! ❤️

packages/webapp/src/hooks/api/useMutations.js
antsgar: Ah this is a really good point! I think to be able to leverage `isSuccess` and `isError` correctly we'd have to call `useMutations` twice, once for deletion and once for removal, and handle the states separately

antsgar:
I just saw @SayakaOno had mentioned a similar approach to my `useEffect` comment, I hadn't reloaded the PR so hadn't seen that! My comment is very similar except that we'd use the result from the `useMutations` hook, and from what I've tried out if we do this in the hook and setup the dependencies in the `useEffect` correctly, there aren't any infinite re-renders happening, but @Duncan-Brain let me know if I'm missing something

packages/webapp/src/containers/Animals/Inventory/useAnimalOrBatchRemoval.ts
antsgar: If the effect after removing animals is always going to be the same regardless of where we reuse this logic (which won't be reused at the moment I think so it's hard to tell), I think it's easier to encapsulate those side effects within the hook itself and let the component not have to worry about it. If it's dependent on where in the UI the hook is called, I'd pass the handlers as props.

The only minor caveat is I'd call `useDispatch` and `useTranslation` in the hook itself as well so that each component doesn't have the extra work of calling that and passing that through.

packages/webapp/src/hooks/api/useMutations.js
Duncan-Brain: Thanks for spotting! I added the remove part after un-drafting and did not think that through!

Duncan-Brain: I think I will remove those from the useMutations since as you mention the nature of mutations makes the likelihood of simultaneous triggers of all unlikely. Just use it for the naming convenience for now 😂

Duncan-Brain: For more info -- I think I would have to call it four times actually in this case haha -- since we conditionally call the animals or batches if there are any selected. So it would probably defeat the purpose of an organizing hook.

packages/webapp/src/hooks/api/useMutations.js
```diff
@@ -0,0 +1,43 @@
...
+ *
+ *  LiteFarm is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ */
+
+const useMutations = (namedMutations) => {
+  const results = namedMutations.map((mutation) => mutation.hook());
+  const data = results.map((result) => result[1]);
```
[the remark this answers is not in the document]
Duncan-Brain: Yeah good call, I am thinking that the scenario @SayakaOno mentioned about isSuccess and isError is probably quite common.  Wondering if the main usefulness of this hook is the structured/named response to avoid scenarios like `deleteAnimalsTrigger` ,`isDeletingAnimalsSuccess`, `isDeletingAnimalsError` etc in favour of `mutations['delAnimals'].lifecycle.isSuccess`. While not necessarily shorter it's more iterable I think.

Are you good if I keep only results and mutations and delete the rest? I could also possibly scrap this utility function.

packages/webapp/src/containers/Animals/Inventory/useAnimalOrBatchRemoval.ts
```diff
@@ -46,24 +60,90 @@ const useAnimalOrBatchRemoval = (
...
         });
+        selectedBatchIds.push(id);
       }
     }
 
-    animalRemovalArray.length && removeAnimals(animalRemovalArray);
-    animalBatchRemovalArray.length && removeBatches(animalBatchRemovalArray);
+    try {
+      if (animalRemovalArray.length) {
+        await mutations['removeAnimals'].trigger(animalRemovalArray).unwrap();
```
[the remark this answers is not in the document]
Duncan-Brain: I guess it was an easy way to use the implicit isSuccess, isError for the particular endpoint to send the snackbars.

I think you used this same way in your RFC
Screenshot 2024-03-28 at 9 24 32 AM

Duncan-Brain:
@antsgar Ah yes, I think the difference was that my instinct was to put the useEffect in the Inventory container. The idea being that other bulk actions might also want to use the snackbar functions. Putting it in the `useAnimalOrBatchRemoval()` would indeed prevent that infinite re-render.

But I am having trouble seeing a succinct solution to the using the useEffects instead of the try catch block without a decent sized refactor of how this hook works. Because of the 4 endpoints and varying messages.

Added:
- state fix on `selectedIds `- occassionally would see the count not update correctly on action bar because close setstates utilizing current value.
- console log
- moved dispatch and translation hooks into hook. good call there 👍

packages/webapp/src/containers/Animals/Inventory/useAnimalOrBatchRemoval.ts
antsgar: Ah I see! In that case yeah, I'd refactor the `useMutations` hook so that it only takes care of the "re-naming" aspect of things and doesn't infer anything from the bulk of the mutations, or maybe even get rid of the `useMutations` hook altogether? I think it might lose value once the whole awaiting for both mutations with `Promise.all` to display a single snackbar enters the picture.

Are you thinking that step of using `Promise.all` to await both mutations and show a single snackbar would be part of the "partial success" ticket's scope?

packages/webapp/src/containers/Animals/Inventory/index.tsx
```diff
@@ -12,7 +12,7 @@
  *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
  *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
  */
-import { useCallback, useMemo, useState, ChangeEvent } from 'react';
+import { useCallback, useMemo, useState, ChangeEvent, useEffect } from 'react';
```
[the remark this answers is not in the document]
Duncan-Brain: Yah no problem!

SayakaOno (approved):

kathyavini (approved):
Looks good @Duncan-Brain! This half of the flow (with the snackbars) was way harder than the first half!!

PR #3167, opened 2024-03-21
LF-4124 new multi select filter component

antsgar:
**Description**

- Created new multi select filter component.
- Replaced current instances of both multiselect and pill select filters in the app with new component.
- Deleted some unnecessary components after replacement.
- Altered/added stories.
- Refactored filters to get rid of `filterRef` and use a state variable instead. **Note: Because of this change, I'll need some help testing this thoroughly throughout the app and making sure no filters are broken due to the change!**

Sorry about the giant PR! The ref refactor really blew it out of proportion and would have been better split up into a separate PR, but too late for that.

Jira link:
https://lite-farm.atlassian.net/browse/LF-4124

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/webapp/src/assets/colors.scss
```diff
@@ -80,13 +80,17 @@
   --Colors-Secondary-Secondary-green-100: #d5eee1;
   --Colors-Secondary-Secondary-green-200: #c1e6d2;
   --Colors-Secondary-Secondary-green-400: #93d4b1;
+  --Colors-Secondary-Secondary-green-500: #F2FAF5;
```
Duncan-Brain, line 83: I think @kathyavini knows an extension we use to keep these all lowercase

antsgar:
@Duncan-Brain thank you so much for the feedback! I think the way it's working for Animals is the way that's intended per the design spec -- if "Showing all" is selected, we don't show any individual pills and the "Showing All (Default)" placeholder should be visible (points 1 and 3 would be alright in that case according to spec) -- but the Transactions one should also behave like this and it isn't doing that, I'll check that out!

Duncan-Brain:
@antsgar I find it surprising that for 1 it is expected that Showing all checkbox does nothing, I expect to be able to toggle the checkbox and have different behaviour happen. I think the designs show that.

Maybe I did not describe it right. Here is a video of what I see.

https://github.com/LiteFarmOrg/LiteFarm/assets/30075973/558f5155-8696-4579-8d82-7b9f95894bc7

packages/webapp/src/components/Filter/FilterMultiSelectV2/index.tsx
```diff
@@ -0,0 +1,251 @@
...
+    <XCircle />
+  </components.MultiValueRemove>
+);
+
+const ClearIndicator = (props: ClearIndicatorProps<ComponentFilterOption>) => {
+  const { getValue, clearValue } = props;
+  const value = getValue();
+  const { t } = useTranslation();
+
+  const onClick = (e: MouseEvent) => {
```
kathyavini, line 52: Hmm, a bit surprised that the stopPropagation + preventDefault aren't enough to keep the filterSelect from re-opening when you hit clear? Ideally it would stay closed in this interaction, right?


https://github.com/LiteFarmOrg/LiteFarm/assets/82857964/dba791ab-ce82-4a8d-8125-c619917df891

packages/webapp/src/components/Filter/FilterMultiSelectV2/index.tsx
```diff
@@ -0,0 +1,251 @@
...
+
+  useEffect(() => {
+    if (shouldReset) {
+      setValue(options.filter((option) => option.default));
+    }
+  }, [shouldReset]);
+
+  const sortedOptions = useMemo(() => {
+    const sorted = [...options].sort((a, b) => {
+      const isSelected = (option: ComponentFilterOption) => value.includes(option);
```
kathyavini, line 129: Sorting doesn't work for me; was it working previously? The value array is always `[]` here for some reason.

packages/webapp/src/components/Filter/FilterMultiSelectV2/index.tsx
```diff
@@ -0,0 +1,251 @@
...
+  return (
+    <components.Option {...props}>
+      <TextButton className={styles.optionButton} onClick={onClick}>
+        <Checkbox label={label} checked={isSelected} />
+      </TextButton>
+    </components.Option>
+  );
+};
+
+const MenuList = (props: MenuListProps<ComponentFilterOption>) => {
```
kathyavini, line 86: Very clever use of MenuList 😄

packages/webapp/src/components/Filter/FilterMultiSelectV2/index.tsx
```diff
@@ -0,0 +1,251 @@
...
+
+  const partiallyChecked = getValue().length !== options.length;
+
+  return (
+    <components.MenuList {...props}>
+      <div className={styles.option}>
+        <TextButton className={styles.optionButton} onClick={clearValue}>
+          <Checkbox
+            label={t('FILTER.SHOWING_ALL')}
+            checked={hasValue}
```
kathyavini, line 98: Default state should also be checked

packages/webapp/src/components/Filter/FilterMultiSelectV2/index.tsx
```diff
@@ -0,0 +1,251 @@
...
+
+const MenuList = (props: MenuListProps<ComponentFilterOption>) => {
+  const { children, getValue, hasValue, options, clearValue } = props;
+  const { t } = useTranslation();
+
+  const partiallyChecked = getValue().length !== options.length;
+
+  return (
+    <components.MenuList {...props}>
+      <div className={styles.option}>
```
kathyavini, line 94: Both the checked and partially checked states need to have the [selected option styling](https://www.figma.com/file/eMffC0h3JPlWtzNttxJpil/Litefarm-Updated?type=design&node-id=2805-37627&mode=design&t=mloVzFJhA0P2XRXV-0) (the grey background).

packages/webapp/src/components/Filter/FilterMultiSelectV2/index.tsx
```diff
@@ -0,0 +1,251 @@
...
+        <TextButton className={styles.optionButton} onClick={clearValue}>
```
kathyavini, line 95: And I _think_ -- this is what I understand from [this conversation]( https://www.figma.com/file/eMffC0h3JPlWtzNttxJpil/Litefarm-Updated?type=design&node-id=2805-37627&mode=design&t=mloVzFJhA0P2XRXV-0) between Loïc and Sayaka -- that it can just be plain non-interactable; doesn't need an onClick handler and can probably get `pointer-events: none;'

packages/webapp/src/components/Filter/FilterMultiSelectV2/styles.module.scss
```diff
@@ -0,0 +1,131 @@
...
+ *  the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  LiteFarm is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ */
+
+@import '../../../assets/mixin.scss';
```
kathyavini, line 16: This is gorgeous CSS 😉 ❤️

kathyavini (commented):
@antsgar it's beautiful!!! And I'm so happy (or actually maybe even astonished) that the filterRef could be replaced with component state so cleanly. I don't think it's caused any bugs in the existing filters!

I know that you don't have a lot of time to keep working on this, and I think _most_ of the things I mentioned below in-line we can handle (well the sorting bug has me little stumped -- if you have some time for insight on that, it would be appreciated!! 😅).

But there is one other thing that relates to something mentioned in-line... do you know the source of this blip? I see either a flash of pills or a flash of selection (depending on whether the dropdown is currently open/closed) when using the `clear` functionality:

https://github.com/LiteFarmOrg/LiteFarm/assets/82857964/aca219e1-15c8-4d80-825b-0a806b274670

I wonder if, were the dropdown not open-closed toggled by the clear button, would this blip not appear?

packages/webapp/src/components/Filter/FilterMultiSelectV2/index.tsx
kathyavini: Actually let me also tag @Duncan-Brain on that since it was relevant to what he said above (and the interaction he shows in the video where clicking the "All selected" row does nothing but close the dropdown was also the first thing I noticed and was bothered by 😉).

I think full spec according to the linked Figma is no interaction on that indicator row (clearing or otherwise) and should open with default state checked and selected.

packages/webapp/src/components/Filter/FilterMultiSelectV2/index.tsx
antsgar: Ah I thought I had got rid of this 😭 so annoying!

packages/webapp/src/components/Filter/FilterMultiSelectV2/index.tsx
antsgar: I thought we had mentioned keeping it interaction free at some point too but based on this in Figma

Captura de pantalla 2024-03-27 a la(s) 21 41 58

I thought it does need to clear the filters if not in the default state? Might be that's just outdated though

packages/webapp/src/components/Filter/FilterMultiSelectV2/index.tsx
antsgar: So I just need to use the default indicator component as a wrapper for mine, not sure why but it makes sense to use it anyway! Solving this also gets rid of the pills flicker you mentioned above

packages/webapp/src/components/Filter/FilterMultiSelectV2/index.tsx
antsgar: Sad! It did work at some point. Checking it

packages/webapp/src/assets/colors.scss
antsgar: Thank you!

packages/webapp/src/components/Filter/FilterMultiSelectV2/index.tsx
antsgar: So the problem was I was using `includes` but since the element I'm checking for is an object, it's not actually the same element. Should be fixed now!

packages/webapp/src/components/Filter/FilterMultiSelectV2/index.tsx
antsgar: Sorry, could you clarify this one please? Not sure I understand!

antsgar: Never mind, I got it now!

packages/webapp/src/components/Filter/FilterMultiSelectV2/index.tsx
antsgar: I've gone ahead and made this non-interactive for now to avoid confusion, should be easy to adjust down the road if needed!

antsgar:
@Duncan-Brain @kathyavini thank you for the thoughtful feedback! ❤️ I think I've addressed most if not all comments in my last commit

kathyavini (commented):
Lovely! ❤️‍🔥 

@Duncan-Brain Only thing I want to check with with Loïc after the holidays is this deselection when the last item is checked -- it's true to what he wrote (or maybe he wanted it to stay partial selection?) but I would prefer that it did not deselect itself when all are selected, that seems backwards ☺️ 

https://github.com/LiteFarmOrg/LiteFarm/assets/82857964/73e23f9f-f9b9-456b-a00f-62e5b3c9a1bd

packages/webapp/src/stories/Filters/FilterMultiSelectV2.stories.jsx
```diff
@@ -0,0 +1,69 @@
...
+ *  (at your option) any later version.
+ *
+ *  LiteFarm is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ */
+
+import { componentDecorators } from '../Pages/config/Decorators';
+import { FilterMultiSelect } from '../../components/Filter/FilterMultiSelectV2';
```
SayakaOno, line 17: `import{ FilterMultiSelectV2 }`

packages/webapp/src/stories/Filters/FilterMultiSelectV2.stories.jsx
```diff
@@ -0,0 +1,69 @@
...
+  ],
+  onChange: () => {
+    console.log('onChange fired!');
+  },
+  shouldReset: 0,
+};
+
+export default {
+  title: 'Components/Filters/FilterMultiSelect',
+  component: FilterMultiSelect,
```
SayakaOno, line 56: `component: FilterMultiSelectV2`

SayakaOno (commented):
A few things I noticed:
- "Continue" button is not enabled even if I check the checkbox in this view.
Screenshot 2024-04-10 at 1 16 08 PM
- The dropdown closes after selecting an option on mobile.

https://github.com/LiteFarmOrg/LiteFarm/assets/33141219/72c2ea4a-3e06-4115-9698-43e704910066

- "Showing All (Default)" is always displayed  when the dropdown is opened

packages/webapp/src/stories/Filters/FilterMultiSelectV2.stories.jsx
kathyavini: @SayakaOno do you think we should create a story for the original FilterMultiSelect as well since it wasn't deleted?

SayakaOno: I don't think it's a must, but now that you mentioned, I think it would be nice to have one!

kathyavini:
@SayakaOno thank you for spotting those things. That checkbox bug especially flew under my radar and it would have been a pain to track down later!

I think those three are fixed now, but I'm curious if you also disliked the way the keyboard menu pops up on mobile and hides the options in the dropdown? I did not like that!

packages/webapp/src/components/Form/Checkbox/checkbox.module.scss
```diff
@@ -27,63 +60,39 @@
...
-  width: 7px;
-  height: 13px;
-  border: solid white;
-  border-width: 0 2px 2px 0;
-  -webkit-transform: rotate(45deg);
-  -ms-transform: rotate(45deg);
-  transform: rotate(45deg);
+  ~ .label {
+    color: var(--Colors-Neutral-Neutral-900);
+    font-weight: 600;
```
SayakaOno, line 78: @kathyavini Not to always bold the label, I think we could pass a prop and apply a className conditionally like this.

```
...
  ~ .label {
    color: var(--Colors-Neutral-Neutral-900);
    // font-weight: 600;
  }
}

.container.shouldBoldSelected input:checked {
  ~ .label {
    font-weight: 600;
  }
}
```

SayakaOno (commented):
Thank you for addressing those! Looks great!

> I'm curious if you also disliked the way the keyboard menu pops up on mobile and hides the options in the dropdown? I did not like that!

It's annoying, but do you have an ideal UX in mind??

packages/webapp/src/components/Form/Checkbox/checkbox.module.scss
kathyavini: I love it, thank you!!! I was banging my head against the Select's styling forever instead of just fixing the new Checkbox 😄

kathyavini:
Regarding the mobile behaviour, the other option would be making it non-searchable on mobile, which was the solution the GitHub issue thread (the one cited in the closing on mobile commit) went with. I would prefer that, but I'll leave a note for Loïc to please evaluate on mobile too when he validates so he can decide then.

SayakaOno (approved):
I didn't review much of the code, but I think it is working well!

PR #3170, opened 2024-03-26
LF-3360 Add "interested" validation to organic certifier survey

thxv3n0lvl:
**Description**

When requesting for a certification through API, the user is able to get it even the farm is not pursuing it (interested = false). 
This PR adds a validation to make sure the farm is interested on certifications and returns `202` status code
[Accepted](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/204) 

Also updates the test cases that are not mocking this property from the DB therefore they were expecting a 200 status code. 

Tests related to validate status code 202 can be added but it's a bit daunting the amount of test cases and code in organicCertifierSurvey.test.js file so I decided to skip that for now.

Jira link:
https://lite-farm.atlassian.net/browse/LF-3360

**Type of change**

**How Has This Been Tested?**

- [x] Passes test case
- [x] Other - Making the API call with a farm that is not pursuing certification vs a farm that is

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/organicCertifierSurveyController.js
+++ b/packages/api/src/controllers/organicCertifierSurveyController.js
@@ -161,8 +161,17 @@ const organicCertifierSurveyController = {
         });
       }
       const organicCertifierSurvey = await knex('organicCertifierSurvey')
-        .where({ farm_id })
+        .where({ farm_id, interested: true })
         .first();
+
+      // Skip the whole flow in case this Farm is not pursuing any cert.
+      if (organicCertifierSurvey === undefined) {
+        // Status code 202 Accepted https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/202
+        return res
+          .status(202)
+          .json({ message: 'You are not currently pursuing any certifications.' });
+      }
+
       const certification = organicCertifierSurvey.certification_id
         ? await knex('certifications')
             .where({ certification_id: organicCertifierSurvey.certification_id })
--- a/packages/api/src/controllers/organicCertifierSurveyController.js
+++ b/packages/api/src/controllers/organicCertifierSurveyController.js
@@ -166,9 +166,8 @@ const organicCertifierSurveyController = {
 
       // Skip the whole flow in case this Farm is not pursuing any cert.
       if (organicCertifierSurvey === undefined) {
-        // Status code 202 Accepted https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/202
         return res
-          .status(202)
+          .status(400)
           .json({ message: 'You are not currently pursuing any certifications.' });
       }
 
--- a/packages/api/tests/organicCertifierSurvey.test.js
+++ b/packages/api/tests/organicCertifierSurvey.test.js
@@ -817,7 +817,7 @@ describe('organic certification Tests', () => {
 
         await mocks.organicCertifierSurveyFactory(
           { promisedUserFarm: [{ farm_id, user_id }] },
-          mocks.fakeOrganicCertifierSurvey(farm_id, { certifier_id: 1 }),
+          mocks.fakeOrganicCertifierSurvey(farm_id, { certifier_id: 1, interested: true }),
         );
         [{ crop_id }] = await mocks.cropFactory(
           { promisedFarm: [{ farm_id }], createdUser: [{ user_id }] },


thxv3n0lvl:
As there was no suggestion on the ticket how the response should look like I decided to take the same string from the web when there's no certification pursued. Please let me know if this is ok or needs an update, also I left a couple comments in the code that I can get rid of but wanted you to look first at it.
The new response will look like:

```
{
    "message": "You are not currently pursuing any certifications."
}
```

packages/api/src/controllers/organicCertifierSurveyController.js
```diff
@@ -161,8 +161,17 @@ const organicCertifierSurveyController = {
...
       const organicCertifierSurvey = await knex('organicCertifierSurvey')
-        .where({ farm_id })
+        .where({ farm_id, interested: true })
         .first();
+
+      // Skip the whole flow in case this Farm is not pursuing any cert.
+      if (organicCertifierSurvey === undefined) {
+        // Status code 202 Accepted https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/202
+        return res
+          .status(202)
```
SayakaOno, line 171: The doc says "`202 Accepted` response status code indicates that the request has been accepted for processing, **but the processing has not been completed**", but the processing has been completed here. `204` seems suitable to me.

@Duncan-Brain @kathyavini what do you think?

SayakaOno (commented):
Thank you @thxv3n0lvl for working on this!
I believe the status code should be something other than `202`, but I'm not so sure, so I tagged Duncan and Joyce!

packages/api/src/controllers/organicCertifierSurveyController.js
Duncan-Brain: For the error code:

I almost expect an error code in the 400's tbh haha. 

I agree with Sayakas point about processing not being completed. And 202 is non-committal to the outcome of the processing described in the Mozilla description. But we do know exactly why it fails to return an export document.

For the premise of the ticket:

I wonder if you actually should still maybe be able to export documents even if you are not interested. My understanding is we are capable of returning something for any farm -- for example creating an "other" certification type.

Duncan-Brain:
@thxv3n0lvl Were you also able to confirm or deny the farm name mix up on the document export also described in the PR?

packages/api/src/controllers/organicCertifierSurveyController.js
thxv3n0lvl: @SayakaOno @Duncan-Brain thanks for the feedback, would you mind sharing what status code should be used? 
204 or in the 400s range.
To me is pretty difficult to say this is an error because it looks like the user gave the consent of not being interested on certifications so the backend should inform that through the frontend, not as an error but some sort of goodmoodhappy reminder (204?). 
Also thinking it depends on the team taste on how to deal with this situation so I'll update the PR according your expectations. 

About this:
> I wonder if you actually should still maybe be able to export documents even if you are not interested. My understanding is we are capable of returning something for any farm -- for example creating an "other" certification type.

Is this something that needs some designing within the core team ? Not sure about the discussions around this topic.

thxv3n0lvl:
> @thxv3n0lvl Were you also able to confirm or deny the farm name mix up on the document export also described in the PR?

Yes, I left a comment in the ticket about this.

packages/api/src/controllers/organicCertifierSurveyController.js
kathyavini: Ah sorry, I was slow to notice this tag. I would say it's a `400` response code. There is no intentional frontend flow that hits this API endpoint with `interested` set to false. If it were to happen via the frontend, it would be an error from the client for making that request.

> My understanding is we are capable of returning something for any farm -- for example creating an "other" certification type.

@Duncan-Brain yeah there is a working flow for no supported certifier / "generic forms" -- is that the situation you mean? That flow also sets `interested` to true though, so it should still be compatible with adding this API check.

SayakaOno: @thxv3n0lvl 
> To me is pretty difficult to say this is an error because it looks like the user gave the consent of not being interested on certifications so the backend should inform that through the frontend, not as an error but some sort of goodmoodhappy reminder (204?).

I agree with you. But let's go with `400` this time. Thank you again!

SayakaOno (approved):

Duncan-Brain (approved):
Thanks Alan!

PR #3171, opened 2024-03-26
LF-4178 Update button colours and variants throughout app

kathyavini:
**Description**

This updates the `<Button />` component used app-wide with the new colours and variant names found in [here in Figma](https://www.figma.com/file/eMffC0h3JPlWtzNttxJpil/Litefarm-Updated?node-id=3080%3A37232&mode=dev).

- Added
  - Secondary-CTA button variant
  - Focus states for all buttons
  - Styling to control the color of an embedded icon (as show in Figma)
- Updated
  - Primary, Secondary, and disabled variants throughout app
  - 'Success' variant (used twice in app); now called 'Secondary-2' per Figma
- Removed:
  - ~Warning and error style variants; they are not used in App and not present on the component chart~ Thank you @Duncan-Brain for spotting they are indeed both used in `<DeleteBox />`! So they will be preserved, although they don't have styles defined for all states like the other variants
- Not yet implemented:
  - Button link variant (I don't think we use the `<Button />` component for this purpose in app, but rather `<TextButton />` which is style neutral; we can revisit this, but it's not yet pressing)
  - Per-variant disabled states; ~this I want to check with Loïc when we next meet~ -- discussed; not necessary for now as only primary will ever be disabled
  - Any new component API (e.g. icon-related props, new sizes beyond what is already used in app). This is a conservative update keeping fully compatible with our current use of `<Button />` only.

Updates have been reflected in the `<Button />` [Storybook story](http://localhost:6006/?path=/docs/components-button--docs)

Jira link: https://lite-farm.atlassian.net/browse/LF-4178

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/assets/colors.scss
+++ b/packages/webapp/src/assets/colors.scss
@@ -69,6 +69,8 @@
   --tableSortPill: #f4f5f7;
 
   // New design system colours
+  --White: #fff;
+
   --Colors-Primary-Primary-teal-50: #ebf5f4;
   --Colors-Primary-Primary-teal-100: #c0e1dd;
   --Colors-Primary-Primary-teal-300: #78bdb6;
@@ -87,16 +89,25 @@
   --Colors-Secondary-Secondary-green-900: #325442;
 
   --Colors-Neutral-Neutral-50: #f0f1f3;
+  --Colors-Neutral-Neutral-200: #b9bfc9;
+  --Colors-Neutral-Neutral-300: #98a1b1;
   --Colors-Neutral-Neutral-400: #858fa1;
+  --Colors-Neutral-Neutral-500: #66738a;
   --Colors-Neutral-Neutral-600: #5d697e;
   --Colors-Neutral-Neutral-700: #485262;
   --Colors-Neutral-Neutral-900: #2b303a;
 
   --Btn-primary-pristine: #ffcf54;
   --Btn-primary-hover: #e8a700;
+  --Btn-primary-disabled: #e7ebf2;
 
   --Colors-Accent-Accent-yellow-50: #fff8e6;
+  --Colors-Accent-Accent-yellow-100: #ffe9b0;
   --Colors-Accent-Accent-yellow-400: #ffc633;
+  --Colors-Accent-Accent-yellow-500: #ffb800;
+  --Colors-Accent-Accent-yellow-600: #e8a700;
+  --Colors-Accent-Accent-yellow-700: #b58300;
+  --Colors-Accent-Accent-yellow-900: #6b4d00;
   --Colors-Accent---singles-Red-light: #ffdad9;
   --Colors-Accent---singles-Red-full: #d02620;
   --Colors-Accent---singles-Red-dark: #520f0d;
--- a/packages/webapp/src/components/Crop/Detail.jsx
+++ b/packages/webapp/src/components/Crop/Detail.jsx
@@ -75,11 +75,6 @@ function PureCropDetail({
               },
             ]}
           />
-
-          {/* <Button style={{ marginBottom: '32px', width: '100%' }} color={'success'}>
-            <Document style={{ marginRight: '1px' }} /> {t('CROP_DETAIL.COMPLIANCE_DOC')}
-            <Leaf style={{ marginLeft: '14px' }} />
-          </Button> */}
         </>
       )}
       {isEditing && (
--- a/packages/webapp/src/components/Form/Button/index.tsx
+++ b/packages/webapp/src/components/Form/Button/index.tsx
@@ -1,9 +1,24 @@
+/*
+ *  Copyright 2022-2024 LiteFarm.org
+ *  This file is part of LiteFarm.
+ *
+ *  LiteFarm is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  LiteFarm is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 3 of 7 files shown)


packages/webapp/src/components/Crop/Detail.jsx
```diff
@@ -75,11 +75,6 @@ function PureCropDetail({
               },
             ]}
           />
-
-          {/* <Button style={{ marginBottom: '32px', width: '100%' }} color={'success'}>
```
Duncan-Brain, line 79: ❤️

packages/webapp/src/assets/colors.scss
```diff
@@ -69,6 +69,8 @@
   --tableSortPill: #f4f5f7;
 
   // New design system colours
+  --White: #fff;
```
Duncan-Brain, line 72: Do we need this one? lol . Its a very minor thing but I wonder if instead of `var(--White)` we can use `white`

Duncan-Brain (approved):
Wow so little code with a big impact!

Everything looks great to me! 

Added one small inline comment. 

If it's of interest to you I made a mixin to color custom icons.  (in the icon fix pr that I have not posted yet). 

Right now the svg stroke color styles may not work as is with all our custom icons but should be fine for any coming from icon libraries:
Screenshot 2024-03-27 at 4 39 45 PM

(notice the green in there^)

Duncan-Brain:
Woops noticed this one late: Delete box button should be red on read only revenues
Screenshot 2024-03-27 at 4 51 44 PM

Duncan-Brain (changes requested):
Red button

packages/webapp/src/stories/Button/Button.stories.jsx
```diff
@@ -66,13 +75,38 @@ DisabledSmall.args = {
...
 
-const style = {
-  background: 'linear-gradient(to right, orange , yellow, green, cyan, blue, violet)',
+export const WithIconSmall = Template.bind({});
+WithIconSmall.args = {
+  color: 'primary',
+  sm: true,
+  children: (
+    <>
+      <div>Edit</div>
```
SayakaOno, line 84: It's a story so it's fine, but `div` should be `span` 😄

packages/webapp/src/components/Form/Button/index.tsx
```diff
@@ -1,9 +1,24 @@
...
+ *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ */
+
 import React, { ReactNode } from 'react';
 import styles from './button.module.scss';
 import clsx from 'clsx';
 
 type ButtonProps = {
-  color?: 'primary' | 'secondary' | 'success' | 'error' | 'warning' | 'none';
+  color?: 'primary' | 'secondary' | 'secondary-2' | 'secondary-cta' | 'none';
```
SayakaOno, line 21: You might have noticed, but `warning` is being used too.

Screenshot 2024-03-27 at 3 35 26 PM

SayakaOno (commented):
Looks great!!!!!

packages/webapp/src/components/Form/Button/index.tsx
kathyavini: I did indeed! 🙂

packages/webapp/src/stories/Button/Button.stories.jsx
kathyavini: I actually thought the flexbox `gap` would only apply to a block-level element!! I'm kind of surprised to learn it works on span!!

packages/webapp/src/assets/colors.scss
kathyavini: I think the only advantage is being able to re-define it all at once throughout app if we want, right? If Loïc ever did want to make this an off-white or something? (Although I'll be we have so much hard-coded through app anyway 😢)

kathyavini:
@Duncan-Brain and @SayakaOno thank you for the `<DeleteBox />` spot!!! 🙏 🙏 🙏 I was foiled by the (quite nice, actually) dynamic `color = {color}` on that component, since I was searching for the hard-coded strings.

Fixed and I gave Loïc a heads up too about their existence so he can add some interactivity styles if he wishes (right now they don't have hover or active states).

@Duncan-Brain hey nice, yes! I want to see the icon mixins when ready! Yes manipulating the stroke like this won't work for all custom icons... has worked so far for those Loïc gave us that belong in buttons, but we'll have to keep checking as we add new ones. I would enjoy a more robust solution.

SayakaOno (approved):

packages/webapp/src/assets/colors.scss
Duncan-Brain: Yeah for sure and that makes sense to me.  Maybe then it is worth a specific name (or names) that would make it more re-definable? Like `--secondary-btn-background: #fff` . I don't think I would redefine white haha

Duncan-Brain:
Here is what I have got that seems to work for me regarding the custom icon color filling (it is not exhaustive with svg path types) if you wanted it. It works by checking if the attribute exists and if so changing it. 

```
@mixin svgColorFill($newColor) {
  svg {
    path{
      &[stroke] {
        stroke: $newColor;
      }
      &[fill] {
        fill: $newColor;
      }
    }
    rect{
      &[stroke] {
        stroke: $newColor;
      }
      &[fill] {
        fill: $newColor;
      }
    }
    circle{
      &[stroke] {
        stroke: $newColor;
      }
      &[fill] {
        fill: $newColor;
      }
    }
    
  }
}
```

Duncan-Brain (approved):
Approving feel free to merge if you dont want to use the mixin!

kathyavini:
@Duncan-Brain Clever mixin!! 😍 We don't need it yet for the add animals flow, so I'll let you include the code in your icons PR and we can do the refactor when we get to an intractable button 🙏

PR #3178, opened 2024-04-02
LF-4152: Update add animals and batches ap is

SayakaOno:
**Description**

- create `validateAnimalBatchCreationBody` middleware to validate request body of add animals/batches APIs (validations were moved from animals/batches controllers)
- update batch controller to allow unspecified sex
- update animal controller to handle groups

Jira link: https://lite-farm.atlassian.net/browse/LF-4152

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/animalBatchController.js
+++ b/packages/api/src/controllers/animalBatchController.js
@@ -16,7 +16,6 @@
 import { Model, transaction } from 'objection';
 import AnimalBatchModel from '../models/animalBatchModel.js';
 import baseController from './baseController.js';
-import DefaultAnimalBreedModel from '../models/defaultAnimalBreedModel.js';
 import CustomAnimalBreedModel from '../models/customAnimalBreedModel.js';
 import CustomAnimalTypeModel from '../models/customAnimalTypeModel.js';
 import { handleObjectionError } from '../util/errorCodes.js';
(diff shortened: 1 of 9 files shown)


packages/api/src/controllers/animalBatchController.js
```diff
@@ -126,16 +72,57 @@ const animalBatchController = {
               sexCount += detail.count;
               sexIdSet.add(detail.sex_id);
             });
-            if (sexCount != animalBatch.count) {
+            if (sexCount > animalBatch.count) {
```
Duncan-Brain, line 75: Wondering if these checks could be moved out as well?

packages/api/src/models/customAnimalBreedModel.js
```diff
@@ -49,6 +50,20 @@ class CustomAnimalBreed extends baseModel {
       additionalProperties: false,
     };
   }
+
+  static async getBreedsByFarmAndTypeBreedPairs(farm_id, typeBreeds) {
```
Duncan-Brain, line 54: I think this should receive `trx`. Otherwise I think this could create a 3rd transaction in this one request.

packages/api/src/models/customAnimalTypeModel.js
```diff
@@ -65,6 +65,17 @@ class CustomAnimalType extends baseModel {
     );
     return data.rows;
   }
+
+  static async getTypesByFarmAndTypes(farm_id, types) {
```
Duncan-Brain, line 69: Same here about `trx`.

Duncan-Brain:
@SayakaOno Everything seems to be testing well for me!

I left a couple inline comments:
1. Possibly moving some more request validation out of the controller
2. Passing `trx` to the model. I am slightly concerned about having 3 db transactions be made in one request (middleware, model method, controller). There is a limit to the number of simultaneous transactions so it wouldn't scale well, and the old very tiny chance of concurrency bugs problem.

Couple other questions:
1. Is there another ticket to handle possible controller validations/creations for other animal and batch properties? The next screen has things like use, identifier, merchant, price etc.

2. Should we be using middleware for all other new and also redo existing animals endpoints to be consistent for request validation?

SayakaOno:
@Duncan-Brain Thank you for the feedback!
1. I will!
2. The debug mode wasn't working correctly, and trx might have been the cause...? Anyway, I will add it!

Other questions:
1. We don't have a ticket for it, thank you! There's a ticket to add columns to the table, so I will update it so that the APIs will also be updated.
2. What endpoints? Patch requests??

Duncan-Brain:
Other endpoints like add and patch for customBreeds customTypes etc all of them really. 

How much request validation would make it valuable to create middleware? So is addCustomAnimalBreed() too small for request validation middleware?

SayakaOno:
Ah not the same middleware! ...honestly, no idea! 😅 
Joyce, any ideas?

kathyavini (commented):
Those were some really complex cases you had to handle there 😯 And the custom and default type tables continued to _really_ not help 😅 Great job!!

I don't know at which point controller validation should be moved to middleware, but I really liked the final division here with only the really tightly CRUD-associated logic remaining in the controller. I don't know if that's a general rule that can be followed, but it read really naturally!

kathyavini (approved):

PR #3185, opened 2024-04-05
Fix failing animal test

SayakaOno:
**Description**

`faker.lorem.word()` returns non-unique strings, which causes unexpected `409 conflict` in animal types. (I believe that was the cause of the test failure: https://github.com/LiteFarmOrg/LiteFarm/actions/runs/8573143951/job/23497247488?pr=3180)
I made some changes to append a number to a generated string.

It rarely happens, so I tested using `.each()` (`test.each(Array(100).fill(null))('xxx', async () => {})`)

### DIFF
--- a/packages/api/tests/animal.test.js
+++ b/packages/api/tests/animal.test.js
@@ -531,11 +531,27 @@ describe('Animal Tests', () => {
       });
 
       test('Should be able to create animals with various types and breeds at once', async () => {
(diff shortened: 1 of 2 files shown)


Duncan-Brain:
Darn that seems annoying! 

Almost no point in using the fake word at that point!

Duncan-Brain (approved):

SayakaOno:
@Duncan-Brain 
It is annoying 😅  And no point... true...
I'll merge, thank you!!

PR #3187, opened 2024-04-08
Lf 4154 create an input for sex detail

navDhammu:
**Description**

Created component for selecting the number of animals for each sex. 

- Uses popover from material UI on desktop view and Drawer on mobile. I kept the default styles from MUI as it looks very similar but can be overridden if needed.
- The readonly input for display is not actually an input but a disguised button inside of `InputBase`.
- Modified `InputBase `to accept a `mainSection` prop which overrides the default input.
- Uses the localized number input but localization is disabled for now as was discussed in grooming.

Jira link: https://lite-farm.atlassian.net/browse/LF-4154

**Type of change**

**How Has This Been Tested?**

Manually tested in storybook. As this will be closely integrated with the parent form for creating an animal I didn't include the automated tests for this and felt its better to write the automated tests in integration with the form.

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/webapp/src/components/Animals/SexDetails/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+export default function SexDetails({ details, count, onConfirm, onCancel }: SexDetailsProps) {
+  const [anchor, setAnchor] = useState<HTMLElement | null>(null);
+
+  return (
+    <>
+      <InputBase
+        label="Sex details"
+        optional
+        mainSection={
```
navDhammu, line 73: I used a button instead of the default input here because in the designs the input display is "Male: {count} - Female: {count}" where count is in bold and there is no way to style input values like that as far as I know. Also a button is more accessible because you can open the popover/modal by tabbing to it and pressing enter or space.

packages/webapp/src/components/Animals/SexDetails/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+  // memoizing ensures this readonly input does not update while user is editing the popover
+  const anchorInput = useMemo(
+    () => (
+      <InputBase
+        label={t('ADD_ANIMAL.SEX_DETAIL')}
+        optional
+        error={total > maxCount ? t('ADD_ANIMAL.SEX_DETAIL_ERROR', { count: maxCount }) : undefined}
+        showResetIcon={false}
+        rightSection={<ChevronDown className={styles.chevronDownIcon} />}
+        mainSection={
```
navDhammu, line 73: I used a button instead of the default input here because in the designs the input display is "Male: {count} - Female: {count}" where count is in bold and there is no way to style input values like that as far as I know. Also a button is more accessible because you can open the popover/modal by tabbing to it and pressing enter or space.

packages/webapp/src/components/Form/NumberInput/useNumberInput.ts
```diff
@@ -55,6 +55,10 @@ export type NumberInputOptions = {
    * - If input value equals min then decrementing with stepper and keyboard is disabled.
    */
   min?: number;
+  /**
+   * Controls whether or not to clamp value on blur that is outside of allowed range
+   */
+  clampOnBlur?: boolean;
```
navDhammu, line 61: Added this so that the error remains after the input is blurred as it is shown in the designs

packages/webapp/src/components/Animals/SexDetails/SexDetailsCount/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+            label={label}
+            initialCount={count}
+            onCountChange={(count) => onCountChange(id, count)}
+          />
+        ))}
+      </div>
+      <div className={styles.helpText}>
+        <div className={styles.info}>
+          <Info style={{ marginTop: 0 }}>{t('ADD_ANIMAL.ANIMALS_TOTAL', { count: maxCount })}</Info>
+          <Text>{t('ADD_ANIMAL.ANIMALS_UNSPECIFIED', { count: Math.max(unspecified, 0) })}</Text>
```
navDhammu, line 56: I used the typography components here but they don't match the current designs and I was wondering if we should create the reusable components for various styles of texts?

SayakaOno: Personally I don't love using the typography components in general, and I might not have utilized them where I should have...

If you find value in using them here, I think you could pass a className to add styles. If not, I would probably mark it with `span`. I wouldn't create reusable components, do you think we should?

packages/webapp/src/components/Animals/SexDetails/index.tsx
SayakaOno: I think it's a good idea! You could probably use `<TextButton />`.

SayakaOno (commented):
I played with the story, it looks great!
Is it possible to make `SexDetailsCountInput` generic? It would be nice if we could use it not just for animal sex.

navDhammu:
> I played with the story, it looks great! Is it possible to make `SexDetailsCountInput` generic? It would be nice if we could use it not just for animal sex.

Thanks for reviewing! 

Yes it should be possible to make the input generic, something like `CountInput`? But I think we should keep it as it is for now and if the need arises create a generic one. Although I'm ok with either option.

packages/webapp/src/components/Animals/SexDetails/index.tsx
navDhammu: I didn't realize we had a text button, thanks for the suggestion!

packages/webapp/src/components/Animals/SexDetails/SexDetailsCount/index.tsx
navDhammu: Reusable text components might be useful in cases where we have the same style of text (like color, font weight), and we can just use the component directly without having to repeatedly create the same classes in css. But I think in this case its fine to just use it with span as you mentioned.

packages/webapp/src/components/Animals/SexDetails/SexDetailsCountInput/index.tsx
```diff
@@ -37,6 +37,7 @@ export default function SexDetailsCountInput({
     initialValue: initialCount,
     max,
     clampOnBlur: false,
+    useGrouping: false,
```
navDhammu, line 40: This disables the thousands grouping with comma or dot, as was discussed in grooming.

SayakaOno:
@navDhammu 
Sounds good, would it be possible to move the folder to `packages/webapp/src/components/Form/`? I suppose it would make reviews easier after converting it to general one. Since I haven't reviewed the code yet, if you think it's not necessary, feel free to leave it as is!

navDhammu:
@SayakaOno Do you mean move the entire SexDetails folder from Animals to Form? Or just the SexDetailsCountInput?

SayakaOno:
The entire SexDetails!

SayakaOno (commented):
Looks great!

We haven't told you and I'm sorry about it, but `MISSING` translation tags will be automatically added by running `pnpm i18n` in `/webapp`. It also sorts translations, so it would be appreciated if you could run it!

navDhammu:
@SayakaOno I ran the` pnpm i18n` command and it generated some keys that weren't part of this ticket, like:

  "SELECTED_COUNT": "{{count}} selected",
  "SELECTED_COUNT_one": "MISSING",
  "SELECTED_COUNT_other": "MISSING",
  
The Missing tags for _one and _other were automatically added. Do I leave these as it is? I think these are for pluralization.

packages/webapp/src/components/Form/SexDetails/styles.module.scss
```diff
@@ -0,0 +1,53 @@
...
+}
+
+.placeholder {
+  color: var(--Colors-Neutral-Neutral-200);
+  font-weight: 400;
+}
+
+.placeholderDark {
+  color: var(--Colors-Neutral-Neutral-900);
+  font-weight: 600;
```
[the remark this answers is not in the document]
navDhammu: Ya I forgot to remove that one

SayakaOno:
Thank you @navDhammu ! It would be appreciated if you could replace those `MISSING` with `{{count}} selected` for now! 🙏

SayakaOno (commented):
Thank you for running the script!
`SELECT_SEXES` and `SEX_DETAIL` are missing in `es`. Could you check if they are added by running the script?  (they were added along with three other keys for me)
I was wondering why `_many` was not added to `/en/...` and discovered the above. (`_many` was not added for me either)

It would be great if you could also resolve conflicts!

navDhammu:
Not sure why the `_many` isn't being added for en, should I add them manually? For the missing es translations I ran the script again and they were automatically added this time.

kathyavini:
@SayakaOno and @navDhammu do we need the `_many` suffix at all? I though only `_one` and `_other` are [enough for singular and plural](https://www.i18next.com/translation-function/plurals#singular-plural). I'm pretty sure I've had the opposite issue (`_many` automatically inserted by script) but I usually try to delete it.

packages/webapp/public/locales/en/translation.json
```diff
@@ -1,4 +1,21 @@
...
+    "ANIMALS_TOTAL_other": "{{count}} animals total",
+    "ANIMALS_UNSPECIFIED": "{{count}} unspecified",
+    "ANIMALS_UNSPECIFIED_one": "{{count}} unspecified",
+    "ANIMALS_UNSPECIFIED_other": "{{count}} unspecified",
+    "SELECT_SEXES": "Select sexes",
+    "SEX_DETAIL": "Sex detail",
+    "SEX_DETAIL_ERROR": "Maximum animal count: {{count}}",
+    "SEX_DETAIL_ERROR_one": "Maximum animal count: {{count}}",
+    "SEX_DETAIL_ERROR_other": "Maximum animal count: {{count}}",
+    "SEX_DETAIL_ERROR_POPOVER": "You cannot have more than {{count}} animals",
```
kathyavini, line 14: If you have included `_one` and `other` I don't think the un-suffixed version is needed.

navDhammu:
@kathyavini I don't think ` _many` is needed if `_other` handles the plural form. For zero I think there's the `_zero `suffix or if not used then the plural suffix will be used in its place.

SayakaOno:
@kathyavini I don't think `_many` is needed either, but I don't like having to delete `_many` every time we run the script...
I did a quick research but couldn't find a way to avoid it 😢 

Maybe this is the explanation of "many" in French...
https://github.com/i18next/i18next-parser/issues/460

packages/webapp/public/locales/en/translation.json
navDhammu: If you leave out the un-suffixed version then the next time someone runs the script, wouldn't all the un-suffixed versions be auto added and then they have to delete them?

kathyavini: The script is generating those un-suffixed versions for you? I can delete those and re-run the script and they aren't re-generated...

kathyavini:
@SayakaOno that link was really helpful! I didn't know that `_many` was needed in French but it looks like the behaviour of the script is exactly correct (adds it in for Fr/Sp/Pt, omits it for English) 👍

packages/webapp/public/locales/en/translation.json
navDhammu: Oh nvm it's not generating them, I don't know why I thought they would be generated

navDhammu: I removed the un-suffixed keys, although for es the `SELECTED` key was already translated so I duplicated it for _one, _many, _other.

kathyavini: > although for es the un-suffixed SELECTED key was already translated so I duplicated it for _one, _many, _other.

Okay 👍  That's a plural form though, so I it won't work for `_one`; I'll leave a suggestion. It's also totally fine to leave it all MISSING (even if it means deleting a previously translated key) to let a native speaker do the pluralization... however this is an easy one so I'll put it in.

And it looks like other the un-suffixed forms (the ones under `ADD_ANIMAL`) need to still be removed from ES/FR/PT?

navDhammu: Ok thanks for filling the es translation I had no idea about that one. And I removed the remaining un-suffixed ones from es/fr/pt - hopefully I didn't miss anything this time.

SayakaOno (approved):

kathyavini (approved):

PR #3192, opened 2024-04-15
LF-4158: Create animal creation details cards for animal

SayakaOno:
**Description**

Create a pure component for animal details view. This is such a huge PR... I would scan the code briefly and check stories if components look okay. Thank you for reviewing!

  - Update `Input` component to validate input when the field is cleared.
  - Create `ImagePicker` component and [story](http://localhost:6006/?path=/docs/components-imagepicker--docs). (it has `MediaWithAuthentication` container, so I made it a container, but does it mean a component that uses `ImagePicker` needs to be a container?)
   It is almost a copy of [FarmImagePicker.tsx](https://github.com/LiteFarmOrg/LiteFarm/blob/integration/packages/webapp/src/components/Profile/Farm/FarmImagePicker.tsx), but the major difference is a removal of `isImageRemoved`. Whether to show a preview depends on the existence of `previewUrl`, and unless `previewUrl` is updated, the preview is shown with `MediaWithAuthentication`. 
   @navDhammu It would be appreciated if you could take a look at this component! 🙏 
  - Create details cards and [stories](http://localhost:6006/?path=/docs/components-addanimalsdetails--docs). Some logic that should be in a **container** can be found in stories. Error messages should be shown for invalid inputs. (text max length)
     - General (animal, batch)
        - By selecting Sex "Male", an extra radio group shows up.
        - Inputs for sex for batch will be implemented in [LF-4159](https://lite-farm.atlassian.net/browse/LF-4159) which is a ticket for batch's details view.
     - Origin (for both animal and batch)
        - Based on the selected radio button, the rest fields show up.
     - Origin (animal, batch)
     - Unique (for animal)
        - By selecting "Other" for tag type, an extra input shows up.
        - By selecting "Other" for tag placement, an extra input shows up.
        - (Tag placement options won't change depending on the selected tag type.)
  - Create a `AddAnimalsDetails` component that holds the four details cards.
     - Only one card should be expanded at a time. (The story is just for testing this behaviour. Each card doesn't work)

NOTE:
I had to pass `t` to the four detail cards because doing `useTranslation([...])` (especially with more than two namespaces) in mui's `Collapse` caused `Maximum update depth exceeded` error.

Jira link: https://lite-farm.atlassian.net/browse/LF-4158, a little bit of https://lite-farm.atlassian.net/browse/LF-4159

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

[LF-4159]: https://lite-farm.atlassian.net/browse/LF-4159?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/webapp/src/components/Form/Input/index.jsx
```diff
@@ -71,6 +72,8 @@ const Input = ({
     input.current.value = '';
     onChange?.({ target: input.current });
     hookFormRegister?.onChange({ target: input.current });
+    // Manually trigger validation against the new value ''
+    trigger?.(name);
```
SayakaOno, line 76: I noticed that text max length error remains even if the field is cleared, so I added `trigger` to validate the field and clear the out-dated error.
(This would be a problem in details view when clearing a field in one card to clear an error, go to another card and go back to the previous card. You would see the error that was cleared once)

https://github.com/LiteFarmOrg/LiteFarm/assets/33141219/51b056a2-b2e0-4406-82c0-61373ea22b3f

packages/webapp/src/components/Animals/AddAnimalsDetails/styles.module.scss
```diff
@@ -0,0 +1,36 @@
...
+}
+
+.expandedContentWrapper {
+  margin-top: 16px;
+}
+
+.detailsWrapper, .sectionWrapper {
+  display: flex;
+  flex-direction: column;
+  gap: 16px
```
kathyavini, line 35: Minor, but there is a missing semi-colon here and a missing space on line 17 which I think prettier fixes automatically, so your auto-format might not be working for `.scss`.

packages/webapp/src/components/Animals/AddAnimalsDetails/type.ts
```diff
@@ -0,0 +1,100 @@
...
+  value: T;
+};
+
+export enum DetailsFields {
+  // GENERAL
+  NAME = 'name',
+  TYPE = 'type',
+  BREED = 'breed',
+  SEX = 'sex',
+  USED_FOR_PRODUCTION = 'used_for_production',
```
kathyavini, line 30: This field is meant to be used for REproduction, correct? It's not a separate field?

packages/webapp/src/containers/Animals/types.ts
```diff
@@ -28,3 +28,13 @@ export enum AnimalOrBatchKeys {
   ANIMAL = 'ANIMAL',
   BATCH = 'BATCH',
 }
+
+export enum AnimalSexes {
```
kathyavini, line 32: This enum is not used. But the form container might want to use it when it sets up its own `isMaleSelected`, is that correct?

packages/webapp/src/stories/Animals/Details/AnimalDetails.stories.tsx
```diff
@@ -0,0 +1,77 @@
...
+type Story = StoryObj<typeof AnimalCreationDetails>;
+
+export const Default: Story = {
+  render: () => {
+    const formMethods: FormMethods = useForm();
+
+    return (
+      <Suspense>
+        <div style={{ padding: '16px' }}>
+          <AnimalDetails
```
kathyavini, line 52: If we adopt the pattern in the [MultiStepForm component](https://github.com/LiteFarmOrg/LiteFarm/pull/3196) I think we are going to start wrapping the whole form in `<FormProvider>`.

I _think_ that means we can just use `<FormProvider>` here in the Stories, and always `useFormContext` in the form components rather than passing the `formMethods` prop, and then the components will be that much easier to fit into the way we ultimately do the form (and the fewer props passed will be the bonus!)

But am I missing something? Is there an advantage to passing `formMethods` as prop?

packages/webapp/src/stories/Animals/Details/mockData.js
```diff
@@ -0,0 +1,72 @@
...
+  { value: 'custom_1', label: 'Sheep' },
+];
+
+export const breedOptions = [
+  { value: '1', label: 'Angus' },
+  { value: '2', label: 'Cobb 5' },
+];
+
+export const sexOptions = [
+  { value: 0, label: `I don't know` },
```
kathyavini, line 29: It's not a question for this PR, but regarding my backend ticket... 'I don't know' is just a frontend implementation of `null`, right? We don't have to add a key in the `animal_sexes` table for it?

kathyavini (commented):
It looks GREAT!! (And huge!)

I left some minor comments but the only two vaguely important ones were:

1) The field named USED_FOR_PRODUCTION -- I think that one was meant to be USED_FOR_REPRODUCTION throughout
2) Whether writing for react-hook-form's `<FormProvider />` to start with might be a better idea. That would clean up the prop drilling of `formMethods` and also future-proof for using MultiStepForm once that PR is merged

packages/webapp/src/stories/Animals/Details/AnimalDetails.stories.tsx
SayakaOno: It was just my unawareness and misunderstanding, thank you so much!

navDhammu:
@SayakaOno I don't think the `ImagePicker` should be a container because it's mainly a UI component. I think we might be able to refactor the MediaWithAuthenticaion out of it and have it accept the authenticated url as a prop maybe?  I don't exactly remember how I implemented the farm image picker component but I'll have a closer look on Tuesday.

SayakaOno:
@navDhammu!
I will leave refactoring of MediaWithAuthenticaion for later but moved ImagePicker to `/component`.
Thank you for your input and review!

packages/webapp/src/stories/Animals/Details/mockData.js
kathyavini: > If we didn't have the key, "sex" would be "I don't know" in the read-only view when the user didn't select anything.

@SayakaOno yes, I think I prefer this!!

I think I also just saw something like this in another place on the AddAnimal flow -- maybe an "I don't know" breed? -- and I could have _sworn_ we scrapped the idea of an `UNKNOWN` breed key at one point in discussion 😅 At least for sure they didn't make it into the tables.

Let's revisit after soil amendment!

kathyavini (dismissed):

packages/webapp/src/components/Animals/AddAnimalsDetails/styles.module.scss
kathyavini: Huh, it still didn't add the semi-colon! So I guess it's this setting that I have checked?

Screenshot 2024-04-30 at 11 27 03 AM

Anyway it doesn't matter for that property since it's the last, but I didn't even notice that setting was configurable before 🙂

SayakaOno: Sorry... 🙏 After adding the semi-colon, "CSS Language Features" seems to have removed it. Now I should properly have set it up to use prettier!
I will add the semi-colon in my next PR!

packages/webapp/src/components/ImagePicker/index.tsx
```diff
@@ -0,0 +1,134 @@
...
+    return () => URL.revokeObjectURL(previewUrl);
+  }, [previewUrl]);
+
+  return (
+    <div>
+      {label && <InputBaseLabel label={label} optional={optional} />}
+      {previewUrl ? (
+        <div className={styles.imageContainer}>
+          {previewUrl === defaultUrl ? (
+            <MediaWithAuthentication fileUrls={[defaultUrl]} alt="image thumbnail" />
```
navDhammu, line 86: This can probably be removed along with `previewUrl === defaultUrl`. The parent component can use the `useMediaWithAuthentication` hook to fetch the private image url and pass it as `defaultUrl` to `ImagePicker`.

packages/webapp/src/stories/ImagePicker/ImagePicker.stories.tsx
```diff
@@ -0,0 +1,72 @@
...
+import ImagePicker, { type ImagePickerProps } from '../../components/ImagePicker';
+import { useController, useForm } from 'react-hook-form';
+
+// https://storybook.js.org/docs/writing-stories/typescript
+const meta: Meta<ImagePickerProps> = {
+  title: 'Components/ImagePicker',
+  component: ImagePicker,
+  decorators: [
+    ...componentDecorators,
+    (Story) => {
```
navDhammu, line 27: Do we really need this decorator? The behavior isn't dependant on react hook form and it works without it.

packages/webapp/src/stories/ImagePicker/ImagePicker.stories.tsx
```diff
@@ -0,0 +1,72 @@
...
+      return <Story {...props} />;
+    },
+  ],
+};
+export default meta;
+
+type Story = StoryObj<typeof ImagePicker>;
+
+export const Default: Story = {
+  render: (args, context) => {
```
navDhammu, line 48: Is it possible to just use args for these stories instead of the render method?

SayakaOno:
@navDhammu Thank you so much for your input, the component and stories look much better now.
It would be appreciated if you could re-review!

packages/webapp/src/stories/Animals/Details/General.stories.tsx
```diff
@@ -0,0 +1,69 @@
...
+};
+export default meta;
+
+type Story = StoryObj<typeof GeneralDetails>;
+
+const commonProps = { typeOptions, breedOptions, sexOptions, useOptions };
+
+export const Animal: Story = {
+  args: {
+    ...commonProps,
```
navDhammu, line 57: It's not necessary to spread the common props in each story - you can just put them in the global meta object and they will be applied to every story

```
import * as commonProps from './mockData

const meta: Meta<GeneralDetailsProps> = {
    args: commonProps
     // other code
}
```

packages/webapp/src/stories/Animals/Details/General.stories.tsx
```diff
@@ -0,0 +1,69 @@
...
+  },
+  render: (args, context) => <GeneralDetails {...args} {...context} />,
+};
+
+export const Batch: Story = {
+  args: {
+    ...commonProps,
+    animalOrBatch: AnimalOrBatchKeys.BATCH,
+  },
+  render: (args, context) => <GeneralDetails {...args} {...context} />,
```
navDhammu, line 68: I think these render functions on stories can be removed if you have one render in the meta object, and then also the decorator can be removed. Like: 
```

const meta: Meta<GeneralDetailsProps> = {
  title: 'Components/AddAnimalsDetails/General',
  component: GeneralDetails,
  render: (args) => {
    const formMethods = useForm({ mode: 'onBlur' });
    const sex = formMethods.watch(DetailsFields.SEX);
    const isMaleSelected = sex === 1;

    return (
      <FormProvider {...formMethods}>
        <GeneralDetails {...args} t={t} isMaleSelected={isMaleSelected} />
      </FormProvider>
    );
  },
  args: commonProps,
  decorators: componentDecorators,
  // avoid "Maximum update depth exceeded" https://github.com/storybookjs/storybook/issues/12306
  parameters: { docs: { source: { type: 'code' } } },
};
```

navDhammu (approved):
This looks good! Although I didn't review all the files but the stories are working well. I left some minor comments to possibly clean up the stories a bit. 

Note:- I created components for animal type and breed selects in (https://github.com/LiteFarmOrg/LiteFarm/pull/3201) that could be used here maybe once that is approved. This is the [file](https://github.com/LiteFarmOrg/LiteFarm/blob/d81487ea31f88a6bf3cc33e12411acf8c0d1c44e/packages/webapp/src/components/Animals/AddAnimalsFormCard/AnimalSelect.tsx) containing the inputs

navDhammu:
Something else I noticed while using the stories - the Expandable component doesn't expand if you click on it - you have to click on the chevron down icon. Is that the expected behaviour? Normally as a user I would expect that it expands anywhere I click on it, and also that I can open it with just the keyboard by tabbing and pressing enter.

SayakaOno:
@navDhammu Thank you for the comments!
I have a ticket for batch details, so I will replace type & breed selects and clean up stories when I work on it.
For the Expandable behaviour, I forgot to pass a prop! Thank you for catching. I will also create a ticket for accessibility improvement :)

PR #3199, opened 2024-04-24
LF-4169: Add animal use to DB and create API

SayakaOno:
**Description**

- Create `animal_use` and `animal_type_use_relationship` tables
- Create `GET` animal uses API
- Add tests
- Add translations for animal uses

GET: `/animal_uses`

Expected response body:
```
[
  {
    default_type_id: '1',
    uses: [
      { id: 1, key: 'MILK' },
      { id: 4, key: 'MEAT' },
      ...
    ],
  },
  ...,
  {
    default_type_id: null,
    uses: [
      { id: 1, key: 'MILK' },
      { id: 2, key: 'EGGS' },
      ...
    ],
  },
]
```

Jira link: https://lite-farm.atlassian.net/browse/LF-4169

**Type of change**

**How Has This Been Tested?**

- [x] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/api/src/models/animalUseModel.js
```diff
@@ -0,0 +1,70 @@
...
+      .join('animal_use', 'animal_type_use_relationship.animal_use_id', '=', 'animal_use.id');
+
+    const usesPerType = typeUseRelationships.reduce((map, { default_type_id, useId, useKey }) => {
+      map[default_type_id] = map[default_type_id] || [];
+      map[default_type_id].push({ id: useId, key: useKey });
+
+      return map;
+    }, {});
+
+    const response = Object.entries(usesPerType).map(([defaultTypeId, uses]) => {
```
kathyavini, line 59: I was just curious about this transform. If we are going to be using it somewhat like a lookup table on the frontend, would it not be better to keep it as a true lookup table with keys as `default_type_id` (the null key could be `''`)?

Is it to be consistent in always returning arrays from our endpoints?

kathyavini (commented):
It's really nice -- both the code to set up the join table and also the data returned! I didn't realize before I we were going to be returning a lookup table-type structure.

My only real question is about that data structure (in-line below) 🙂

packages/api/src/models/animalUseModel.js
SayakaOno: Thank you for asking!!
I went back and forth to settle on the decision, but I wasn't even sure if it was a good idea to make the response like that...

I ended up going with an array because keys like "1", "2" or "" would make it difficult to understand what they represent at a glance. I researched which approach might be better, but unfortunately, I couldn't find any resources...😞 Returning the lookup object would definitely be more efficient, so if we are okay with the structure and doing something new, I would like to give it a try. It would be great if we could discuss this further on Monday!

kathyavini: Hmmm! I like your readability argument so much I would very much like to keep it as is 🙏 The keys as `default_type_ids` would indeed become really cryptic as soon as we are a bit removed from this context.

kathyavini (approved):

PR #3139, opened 2024-02-22
LF-3893: RFC for component library

antsgar:
**Description**

Initial example setup for component library package and example migration of one component (`HoverPill` component).
To be reviewed in the context of the RFC and together with the document https://lite-farm.atlassian.net/wiki/spaces/LITEFARM/pages/1449918527/RFC%3A+Component+Library,

Jira link:
https://lite-farm.atlassian.net/browse/LF-3893
https://lite-farm.atlassian.net/browse/LF-3895

**Type of change**

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

**How Has This Been Tested?**

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [ ] Passes test case
- [ ] UI components visually reviewed on desktop view
- [ ] UI components visually reviewed on mobile view
- [ ] Other (please explain)

**Checklist:**

- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] The precommit and linting ran successfully
- [ ] I have added or updated language tags for text that's part of the UI
- [ ] I have added "MISSING" for all new language tags to languages I don't speak
- [ ] I have added [the GNU General Public License](https://lite-farm.atlassian.net/l/cp/BT0Dd7WW) to all new files

PR #3201, opened 2024-04-30
Lf 4156 create animal batch creation form card

navDhammu:
**Description**
- Created card component that holds inputs for adding animals
- Created reusable `CardV2` based on current designs, uses the `as` prop to render custom html tag other than `div` if needed. Example, `<Card as='form'>`
- Created individual `AnimalTypeSelect` and `AnimalBreedSelect` components that can be used in other places.
- Requires type options, breed options, and sex detail options to be passed in from parent
- Logic for conditionally showing breed/type options should be handled in a parent container component or inside of custom hook.
- Added search icon to react select when dropdown is open as was suggested as a usability improvement by Loic.
- Changed storybook config to allow placing story files anywhere inside src as was discussed in tech daily.

@SayakaOno There were some minor changes to NumberInput here, for example I added a `className` prop to allow changing styles for the outer `div` that wraps the input field. This might create conflicts with your work in the soil amendment components. 

Jira link: https://lite-farm.atlassian.net/browse/LF-4156

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/webapp/src/components/Form/NumberInput/index.tsx
```diff
@@ -56,13 +52,15 @@ export default function NumberInput({
...
   onChange,
   onBlur,
   ...props
-}: NumberInputProps & RhfProps) {
-  const { field, fieldState } = useController({ name, control, rules, defaultValue: initialValue });
+}: NumberInputProps<T>) {
+  const { field, fieldState } = useController({ name, control, rules, defaultValue });
   const { inputProps, reset, numericValue, increment, decrement } = useNumberInput({
-    initialValue,
+    initialValue: defaultValue,
```
navDhammu, line 63: The initial value type was giving an error so changed it use defaultValue prop that comes with react hook form

packages/webapp/src/components/Animals/AddAnimalsFormCard/AnimalSelect.tsx
```diff
@@ -0,0 +1,112 @@
...
+    <Controller
+      name={name}
+      control={control}
+      render={({ field }) => (
+        <CreatableSelect
+          options={breedOptions}
+          label={t('ADD_ANIMAL.BREED')}
+          placeholder={t('ADD_ANIMAL.BREED_PLACEHOLDER')}
+          //@ts-ignore
+          showNoTypeSelectedMessage={showNoTypeSelectedMessage}
```
navDhammu, line 79: Had to use ts-ignore here because  `showNoTypeSelectedMessage` is a custom prop not defined in ReactSelect, however ReactSelect will still accept any prop and then pass it to custom components. In this case the custom `AnimalBreedMenu` used for displaying message 'please select a type first'.

For extending the types to accept custom props, they recommend to use module augmentation, like: 
```

declare module 'react-select/dist/declarations/src/Select' {
  export interface Props<
    Option,
    IsMulti extends boolean,
    Group extends GroupBase<Option>
  > {
    showNoTypeSelectedMessage: boolean;
  }
}
```

But this makes it so that the custom props are able to be used in all ReactSelect components and according to this [discussion](https://github.com/JedWatson/react-select/issues/4804#issuecomment-945972022), its not possible to specify it per component.

packages/webapp/src/components/Animals/AddAnimalsFormCard/AddAnimalsFormCard.tsx
```diff
@@ -0,0 +1,144 @@
...
+        // @ts-ignore
+        <Input
+          label={t('ADD_ANIMAL.GROUP_NAME')}
+          optional
+          placeholder={t('ADD_ANIMAL.GROUP_NAME_PLACEHOLDER')}
+          hookFormRegister={register('group')}
+        />
+      ) : (
+        // @ts-ignore
+        <Input
```
navDhammu, line 135: This` Input` is handling text inputs, number inputs, date pickers, and passwords inputs all in one. It might be good to extract those out into different components and also use typescript to avoid ts-ignore.

packages/webapp/.storybook/main.cjs
```diff
@@ -1,24 +1,29 @@
 module.exports = {
```
SayakaOno, line 1: I can't find the difference, what's changed? 😅

packages/webapp/src/components/Animals/AddAnimalsFormCard/AddAnimalsFormCard.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+export default function AddAnimalsFormCard({
+  typeOptions,
+  breedOptions,
+  sexDetailsOptions,
+  onTypeChange,
+  onIndividualProfilesCheck,
+  showRemoveButton,
+  onRemoveButtonClick,
+  isActive,
```
SayakaOno, line 65: Do you have an idea on how to decide if this card is active or not?

packages/webapp/src/components/Animals/AddAnimalsFormCard/AddAnimalsFormCard.tsx
```diff
@@ -0,0 +1,144 @@
...
+  const watchAnimalType = watch('type');
+  const shouldCreateIndividualProfiles = watch('createIndividualProfiles');
+
+  return (
+    <Card className={styles.form} isActive={isActive}>
+      <div className={styles.formHeader}>
+        <Main>{t('ADD_ANIMAL.ADD_TO_INVENTORY')}</Main>
+        {showRemoveButton && (
+          <TextButton className={styles.removeBtn} onClick={onRemoveButtonClick}>
+            {' '}
```
SayakaOno, line 79: I think this can be removed?

packages/webapp/src/components/Animals/AddAnimalsFormCard/AddAnimalsFormCard.tsx
```diff
@@ -0,0 +1,144 @@
...
+            <XIcon /> {t('common:REMOVE')}
+          </TextButton>
+        )}
+      </div>
+      <AnimalTypeSelect
+        name="type"
```
SayakaOno, line 85: Can we use constants for names?

packages/webapp/src/components/Animals/AddAnimalsFormCard/styles.module.scss
SayakaOno, line 1: Could you please add license? 🙏

packages/webapp/src/components/Animals/AddAnimalsFormCard/styles.module.scss
```diff
@@ -0,0 +1,35 @@
...
+  flex-direction: column;
+  gap: 16px;
+}
+
+.formHeader {
+  display: flex;
+  justify-content: space-between;
+}
+
+.removeBtn {
```
SayakaOno, line 12: It looks like this button has active or hover state: https://www.figma.com/file/eMffC0h3JPlWtzNttxJpil/Litefarm-Updated?node-id=3881-23342&mode=dev

packages/webapp/src/components/Animals/AddAnimalsFormCard/styles.module.scss
```diff
@@ -0,0 +1,35 @@
...
+  display: flex;
+  align-items: center;
+  gap: 4px;
+  border-radius: 16px;
+  padding: 2px 9px;
```
SayakaOno, line 17: Let's go with `4px 8px`. If I'm not mistaken, we would like to use multiples of 4px, even if it differs slightly from the styles in the mockups.

packages/webapp/src/components/CardV2/styles.module.scss
```diff
@@ -0,0 +1,26 @@
...
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ */
+
+.card {
+  background-color: var(--White);
+  padding: 24px;
+  border: 1px solid var(--Colors-Primary-Primary-teal-50);
+  box-shadow: 0px 2px 2px 0px rgba(0, 0, 0, 0.05);
```
SayakaOno, line 20: Can border-radius be added here?

SayakaOno (commented):
I added some minor comments, but it looks great! Thank you!

navDhammu:
@SayakaOno Thanks for your review - I've made the requested changes.

packages/webapp/.storybook/main.cjs
navDhammu: Its hard to tell but I removed the` /stories` after src in `stories: ['../src/**/*.stories.mdx', '../src/**/*.stories.@(js|jsx|ts|tsx)'],`. It was matching all .stories files in the stories directory now it will look for them in src

packages/webapp/src/components/Animals/AddAnimalsFormCard/AddAnimalsFormCard.tsx
navDhammu: I guess when any field inside the form is focused, a state variable `isActive` can be set to true

packages/webapp/src/components/Animals/AddAnimalsFormCard/AddAnimalsFormCard.tsx
SayakaOno: It would be great if you could replace names and those in `watch()` too... 🙏

packages/webapp/src/components/Animals/AddAnimalsFormCard/AddAnimalsFormCard.tsx
SayakaOno: Using a listener?? I'm wondering if we can use `:focus-within`...

packages/webapp/src/components/Animals/AddAnimalsFormCard/AddAnimalsFormCard.tsx
navDhammu: Thanks for noticing that I don't know how I missed it

packages/webapp/src/components/Animals/AddAnimalsFormCard/AddAnimalsFormCard.tsx
navDhammu: :focus-within might be good as well, or even better as it's just css

SayakaOno (approved):

packages/webapp/src/components/Form/ReactSelect/CreatableSelect.tsx
```diff
@@ -16,11 +16,12 @@
 import React from 'react';
 import Select, { CreatableProps } from 'react-select/creatable';
 import { GroupBase, SelectInstance } from 'react-select';
-import { styles } from './styles';
+import { styles as baseStyles } from './styles';
 import InputBaseLabel, { InputBaseLabelProps } from '../InputBase/InputBaseLabel';
 import { useTranslation } from 'react-i18next';
 import { ClearIndicator, MultiValueRemove } from './components';
 import scss from './styles.module.scss';
+import { MenuOpenDropdownIndicator } from './components';
```
Duncan-Brain, line 24: Tiny thing: This could be combined with the other import above

Duncan-Brain (commented):
I am really impressed by this one, I learned a lot haha 

Everything looks pretty great, there are a few new things here I will still have to learn but happy to approve. I just had one question before doing so:

My only thought was that maybe since we are changing the location of our stories that we could maybe formalize it with a separate PR. I think I missed the discussion and I am curious about the pros and cons and was thinking since it could be a simple drag and drop to move the other stories we might be able to crush out the changed location for other stories so we don't support 2 systems?

navDhammu:
@Duncan-Brain Thank you for reviewing

The idea for having the stories in the same folder as the component was for better organization (atleast in my opinion and others seemed to agree). It's easier to know where a story is and usually when working on a frontend ticket you need both the component and story files and makes it easier to work on them when they're located together. It's also one less thing to have to think about when creating a new component - the story just goes in the same place as the component.

I think a separate PR is a good idea. Might be frustrating manually moving every single story file though.

Duncan-Brain:
@navDhammu  I can get on board for that change, I think it lines up well with Anto's PR #3139 

She actually put them in their own folder which maybe we should do also?

The main con that I was thinking about was for excluding stories from the server and build process. We don't currently do it but we would need to parse out the files should we decide to streamline the server at some point vs. exclude 1 folder. Very doable but I think why many choose the existing format.

The all files thing isn't necessary... but I figured if it takes like 1 hr it would be worth it so that we don't have two systems. We have a lot of transitioning architecture changes that might be hard to keep up with or confusing too haha

navDhammu:
I like the structure in Anto's PR for the component library, and I think it's fine to use the stories folder like she did. Although I feel it's better to keep the folder structure flat, especially since most components would have one story file, so having just one file inside the stories folder seems a bit overkill to me. 

Regarding the server and build process - I'm not very familiar with how vite works but isn't storybook automatically not included in the build and dev server? There's a separate dev server for storybook and for the application code so I don't think one is included in the other, but I could be wrong.

And I would prefer to just move all the story files in one PR for consistency. If thats decided then should I go ahead and change the stories for this PR back to the old structure, and create a new ticket for the change?

SayakaOno (approved):

PR #3202, opened 2024-05-01
LF-4218: Create component for composition selects

SayakaOno:
**Description**

- Create `NumberInputWithSelect` component, combining `NumberInput` and `ReactSelect`. Positioned `ReactSelect` using `position: absolute` and adjusted `padding-right` of `NumberInput` to accommodate the ReactSelect width. (I chose not to use flex layout due to difficulty in customizing border colour)
- Create `CompositionInputs` component, containing three instances of `NumberInputWithSelect`.
- Create [stories](http://localhost:6006/?path=/docs/components-compositioninput--docs).
- Create a mixin for truncating text.
- Update `InputBase` component to wrap only the "label" (Please let me know if it shouldn't be)

NOTE:
- The component is not reusable.
- After struggling with working with the type of `control`, I opted to use `any`. Any suggestions would be appreciated!

Jira link: https://lite-farm.atlassian.net/browse/LF-4218

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


SayakaOno:
I'm sorry to keep adding commits... I removed the line connecting the npk inputs, and that should be the last change!

packages/webapp/src/components/Form/CompositionInputs/NumberInputWithSelect.tsx
```diff
@@ -0,0 +1,108 @@
...
+    <div className={className}>
+      <InputBaseLabel label={label} optional />
+      <div
+        className={clsx(
+          styles.inputWithSelectWrapper,
+          hasError && styles.hasError,
+          disabled && styles.disabled,
+        )}
+      >
+        <NumberInput
```
navDhammu, line 81: I was just wondering, isn't it possible to use InputBase with right section? That way you get default styling and avoid the code duplication. 

Something like this? 

```
  ...
  const { field } = useController({ control, name });
  const { inputProps } = useNumberInput({ onChange: (value) => field.onChange(value) });

  return (
    <InputBase
      {...inputProps}
      label={label}
      optional
      rightSection={
        <Controller
          control={control}
          name={unitName}
            <ReactSelect
              options={unitOptions}
              onChange={onChange}
              value={value}
              styles={{ ...(reactSelectStyles as any) }}
              isDisabled={disabled}
            />
          )}
        />
      }
    />
  );
```
I think InputBase might need to be slightly modified to accept the error so that only the red error outline is shown and not the text, and to accept a prop to adjust the right padding.

packages/webapp/src/components/Form/CompositionInputs/NumberInputWithSelect.tsx
```diff
@@ -0,0 +1,108 @@
...
+import { useTranslation } from 'react-i18next';
+import { Controller } from 'react-hook-form';
+import NumberInput from '../NumberInput';
+import ReactSelect from '../ReactSelect';
+import useReactSelectStyles from '../Unit/useReactSelectStyles';
+import InputBaseLabel from '../InputBase/InputBaseLabel';
+import styles from './styles.module.scss';
+
+type NumberInputWithSelectProps = {
+  control: any;
```
navDhammu, line 25: Does the Control type from react-hook-form work here?

packages/webapp/src/components/Form/InputBase/index.tsx
```diff
@@ -71,18 +71,18 @@ const InputBase = forwardRef<HTMLInputElement, InputBaseProps>((props, ref) => {
...
-          leftSection={leftSection}
-          mainSection={mainSection}
-          rightSection={rightSection}
-          isError={!!error}
-          resetIcon={
-            !!error && showResetIcon ? <Cross isClickable onClick={onResetIconClick} /> : undefined
-          }
-          ref={ref}
-        />
       </label>
```
navDhammu, line 74: The `<label>` tag here is meant to wrap the inputField for accessibility so the Input can have an associated label because InputBaseLabel doesn't use the `label` tag. Does it need to be removed?

packages/webapp/src/components/Form/CompositionInputs/NumberInputWithSelect.tsx
SayakaOno: It wasn't simple at all 😢, but I used the Control type. I'm not sure if I did it correctly though 😭

packages/webapp/src/components/Form/CompositionInputs/NumberInputWithSelect.tsx
SayakaOno: I didn't feel right to use InputBase because we have the NumberInput, so that wasn't my option, but the code you suggested works well!
I added `showErrorText` prop, but I offset the right padding with a container applying `margin-right: -8px;`.

packages/webapp/src/components/Form/InputBase/index.tsx
SayakaOno: Sorry 😅, I reverted it.
`<label>` closes ReactSelect dropdown right after it opens, but I wrapped ReactSelect and added `preventDefault()` to prevent it.

SayakaOno (commented):
@navDhammu Thank you so much for reviewing!
I think I've addressed your comments. Re-review would be appreciated, thank you!
(I'll make one more change once Loïc tells me whether to display cross icons)

navDhammu (commented):
Thanks for making these changes and everything looks to be working well! I made some suggestions to simplify the code so that all those complicated prop typings can be removed, making it easier to maintain (sorry if it's a lot of work 😅 ). Another possible benefit is if we ever decide to change form libraries or not use a library then it would be easier to migrate.

packages/webapp/src/components/Form/CompositionInputs/NumberInputWithSelect.tsx
navDhammu: It looks correct to me! But also looks unnecessary and a headache to maintain so I made some comments on improvement in the review.

SayakaOno:
@navDhammu Thank you for the inputs, I think I've addressed all your comments. Could you please re-review when you have the time?
Thank you!

packages/webapp/src/components/Form/InputBase/InputBaseField/styles.module.scss
```diff
@@ -13,14 +13,14 @@
  *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
  */
 
-.input input {
+.inputWrapper .input {
   all: unset;
```
SayakaOno, line 17: Added a className to `input` so that `all: unset;` will not be applied to the input inside ReactSelect.

packages/webapp/src/components/Form/CompositionInputs/NumberInputWithSelect.tsx
```diff
@@ -0,0 +1,119 @@
...
+  label,
+  unitOptions,
+  disabled,
+  error,
+  className,
+  onChange,
+  values = {},
+}: NumberInputWithSelectProps<T, U>) => {
+  const { t } = useTranslation();
+  const unit = values[unitName] as ReactSelectOption;
```
navDhammu, line 53: Can using 'as' be avoided? I'm thinking if values is typed instead of a generic then that could work? Ex: 

```
type CompositionValues = {
    unit: Unit;
    n: number;
    p: number;
    k: number;
}

```
Then in the story it can be used to type the form like useForm<{composition: CompositionValues}>

packages/webapp/src/stories/Form/CompositionInputs/CompositionInputs.stories.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+export const Default: Story = {
+  render: (args) => {
+    const { control } = useForm({
+      mode: 'onChange',
+      defaultValues: {
+        composition: {
+          unit: {
+            label: '%',
+            value: 'percent',
```
navDhammu, line 57: Is there a reason this can't just be {unit: 'percent'} without the label? Because I see the label is already passed in with unitOptions.

navDhammu (commented):
The code is looking good I just had a couple of questions which I left in the comments. 

Other than that I noticed in the react select there is a blinking cursor that shows up when you click on it. Can that be hidden? And I'm seeing that you can't type in decimals in the inputs and thats my fault because I forgot to default the allowDecimal to true in the `useNumberInput` hook when I initially implemented it. Would you be able to fix that?

packages/webapp/src/components/Form/CompositionInputs/NumberInputWithSelect.tsx
```diff
@@ -0,0 +1,137 @@
...
+        label={label}
+        optional
+        placeholder={disabled ? '' : t('common:ENTER_VALUE')}
+        disabled={disabled}
+        error={error}
+        showErrorText={false}
+        showResetIcon={false}
+        rightSection={
+          <>
+            {error && numericValue ? <Cross isClickable onClick={() => update(NaN)} /> : null}
```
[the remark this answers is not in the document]
navDhammu: > `reset` is `reset: () => update(initialValue ?? NaN)`, so when `initialValue` is passed, it won't clear the value.
> 
> I was thinking from a UX point of view if the user expects to clear the value or reset it to initial value when they click 'X'?

packages/webapp/src/stories/Form/CompositionInputs/CompositionInputs.stories.tsx
SayakaOno: I'm not sure how to pass `value` as a `value` prop to `ReactSelect`. What is your idea??

packages/webapp/src/components/Form/CompositionInputs/NumberInputWithSelect.tsx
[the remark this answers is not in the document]
SayakaOno: Ah I see, clicking the cross icon should clear the input.
`update(NaN)` is like `setValue('xxx', NaN)`, right? I don't think we need a function to clear the value, but I can create it.

packages/webapp/src/components/Form/CompositionInputs/NumberInputWithSelect.tsx
SayakaOno: Let me try...

SayakaOno:
@navDhammu Thank you for re-reviewing!
I think I had fixed the blinking cursor by avoiding applying `all: unset;` to ReactSelect. Could you pull and try it again? 🙏

packages/webapp/src/stories/Form/CompositionInputs/CompositionInputs.stories.tsx
navDhammu: value prop is of the same type as option right? So you can get the option that has value = option.value from the unit options array. One possibility: `value={unitOptions.find(option => option.value === value}`

navDhammu:
> @navDhammu Thank you for re-reviewing! I think I had fixed the blinking cursor by avoiding applying `all: unset;` to ReactSelect. Could you pull and try it again? 🙏

Oh my bad I forgot to pull - its working great!

packages/webapp/src/components/Form/CompositionInputs/NumberInputWithSelect.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+  const { inputProps, update } = useNumberInput({
+    onChange: (value) => onChange(name, value),
+    initialValue: values[name],
+    max: 999999999,
+  });
+
+  useEffect(() => {
+    // If the value is updated from the parent, update the actual value in the input
+    if (inputProps.value !== values[name]) {
```
navDhammu, line 97: I think this is causing an issue in the `With Error` story. I'm not able to type the value its updating it as soon as I type.

Another potential issue is that inputProps.value could be a string and value[name] is a number so that might cause issues when comparing.

SayakaOno: Thank you for catching!!!

Do you think it will work if we don't compare their types?

packages/webapp/src/stories/Form/CompositionInputs/CompositionInputs.stories.tsx
navDhammu: Not to simplify the expression but to simplify the form data values when submitting to backend, but it might not matter that much so i'll leave it up to you.

packages/webapp/src/components/Form/CompositionInputs/NumberInputWithSelect.tsx
navDhammu: Sorry I should've mentioned this before but this won't work if the number is localized. For example on the 'pt' locale the number 23.5 would be 23,5 and 23,5 == 23.5 would be false. I'm just wondering is there a need for the useEffect here? If its really needed then `numericValue` can probably be used instead of inputProps.value

packages/webapp/src/stories/Form/CompositionInputs/CompositionInputs.stories.tsx
SayakaOno: Ah okay, I believe the request body will be something like:
```
{
  n: x,
  p: x,
  k: x,
  npk_unit: 'percent',
  ...
}
```
, so we will need to format the data anyways.
I'll leave it as is :)

packages/webapp/src/components/Form/CompositionInputs/NumberInputWithSelect.tsx
SayakaOno: Oh that's right...
So the problematic flow is editing an existing product.
If the existing product has N: 20 and the user changes it to 30 and then click "cancel", the value needs to be back to "20". I will use `reset` returned from `useForm`, and it makes the value "20" and it needs to be reflected in the `useNumberInput`.
`numericValue` sounds right to use, I'll try, thank you!

navDhammu: Oh ok that makes sense. 

I remember reading about how adjusting the state with an effect might be a bad idea in the react docs - here's the link https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes

They mention how its more performant to do it without the effect, just wondering if that could be applied here?

SayakaOno: Thank you so much for sharing the link, I didn't know useEffect **should be your last resort** 😰

I tried removing `useEffect` and tested using a component I'm working on in my next [ticket](https://lite-farm.atlassian.net/browse/LF-4220) and got this error:
 
Screenshot 2024-05-09 at 8 31 35 AM

It seems to be the case where `useEffect` is a solution according to this [React repo issue](https://github.com/facebook/react/issues/18178).

It would be appreciated if you could check the fixed condition!

packages/webapp/src/stories/Form/CompositionInputs/CompositionInputs.stories.tsx
SayakaOno: Actually, having an option object as unit causes a [problem](https://github.com/orgs/react-hook-form/discussions/7764), so I updated it to store string!

navDhammu (approved):
Everything looks good! Just something I had in mind for resetting the form when user clicks 'cancel' is to use the key prop to force react to reset the entire component. That might be a way to avoid useEffect but I'm not quite sure, so just something to consider when you work on that.

SayakaOno:
Thank you so much @navDhammu!
Using a key is an interesting idea, I will try!

packages/webapp/src/components/Form/CompositionInputs/styles.module.scss
```diff
@@ -0,0 +1,92 @@
...
+ *
+ *  LiteFarm is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ */
+
+@import '../../../assets/mixin';
+
+/*----------------------------------------
```
Duncan-Brain, line 18: Thats purdy 🤩 !

packages/webapp/src/assets/mixin.scss
```diff
@@ -81,3 +81,9 @@
     }
   }
 }
+
+@mixin truncateText() {
```
Duncan-Brain, line 85: ❤️

packages/webapp/src/components/Form/CompositionInputs/NumberInputWithSelect.tsx
```diff
@@ -0,0 +1,145 @@
...
+import useNumberInput from '../NumberInput/useNumberInput';
+import InputBase from '../InputBase';
+import { styles as reactSelectDefaultStyles } from '../ReactSelect';
+import styles from './styles.module.scss';
+
+export type Option = { value: string; label: ReactNode };
+
+export const UNIT = 'unit';
+
+export enum NPK {
```
Duncan-Brain, line 29: Is NPK in this component temporarily until a container is made? 

Otherwise maybe it should be generic reusable or have a Component name like NPK input?

Duncan-Brain (commented):
Looks pretty great, left one small question!

packages/webapp/src/components/Form/CompositionInputs/NumberInputWithSelect.tsx
SayakaOno: Thank you for reviewing!
I think I was trying to make it reusable, but I couldn't handle types, and this is truly non-reusable component now...
I will revisit this again later when we work on some of those additional inputs, and if I can't reuse this, I will rename to NPKInput!

Duncan-Brain (approved):
Looks great!

PR #3209, opened 2024-05-20
LF 4199: Fix undefined date range filter on Tasks page

Joaquin-M2:
**Description**

On the Tasks page, the date range filter threw `undefined` on the UI rather than applying the filter itself successfully.

Jira link: https://lite-farm.atlassian.net/browse/LF-4199

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

- [x] The precommit and linting ran successfully  ---> The linting throws 11 errors and 14 warnings, but they are not related to changes made on this PR.


SayakaOno (commented):
@Joaquin-M2 Thank you for working on this, your solution looks good!
It would be appreciated if you could look into the behaviour that an empty "To" filter is added:

https://github.com/LiteFarmOrg/LiteFarm/assets/33141219/54d6115f-0ddf-48c7-9b51-5fc23ffd74c9

packages/webapp/src/containers/Filter/Tasks/index.jsx
```diff
@@ -166,6 +166,7 @@ const TasksFilterPage = ({ onGoBack }) => {
     },
     {
       subject: t('TASK.FILTER.DATE_RANGE'),
+      filterKey: FilterType.DATE_RANGE,
```
Duncan-Brain, line 169: To keep it consistent with others should we get this from the constants file? The Filter types I could see being improved in the future.

packages/webapp/src/containers/Filter/types.ts
```diff
@@ -20,6 +20,7 @@ export interface FilterState {
   [filterOptionId: string]: /* e.g. ABANDONED, COMPLETED */ {
     active?: boolean;
     label?: string /* e.g. 'Abandoned', 'Completed */;
+    [key: string]: any;
```
Duncan-Brain, line 23: Ah it looks like this type is mostly used for multi-select right now. We might want to make the difference more explicit in the future!

Duncan-Brain (approved):
Works good for me. I left 2 small comments that I don't think need to be addressed right now. Awesome work!

SayakaOno (approved):
Thank you!

packages/webapp/src/containers/Filter/Tasks/index.jsx
Joaquin-M2: That's an option I considered too, but I wasn't sure if creating a new constant just for this was an overkill. I will create it if I work in a new ticket that requires that same value 👍

packages/webapp/src/containers/Filter/types.ts
Joaquin-M2: I can "defend" myself with TypeScript but I'm not any kind of TS wizard, so I start having troubles if typing things become a bit complex. I just wanted to get rid of the error 😆  But I agree that it should be more specific.

PR #3212, opened 2024-05-24
LF-4220: Create new add products component for soil amendment

SayakaOno:
**Description**

Screenshot 2024-06-03 at 10 39 57 AM

- Update `plus-circle` svg stroke color to `currentColor`.
- Create a mixin `flex-column-gap` that aligns its child elements in a column direction and applies a specified gap in between. 
- Create `SmallButton` component for the "x remove" button. ([stories](http://localhost:6006/?path=/docs/components-button-smallbutton--docs))
- Update `NumberInputWithSelect` and `CompositionInput` to take `onBlur`. 
- Update `InFormButtons` to make `onConfirm` function optional. (This change will allow us to submit the request using the form's onSubmit in the future if needed. I initially considered using the form's onSubmit but ended up sending the `onConfirm` prop instead because we should not nest forms.)
- `AddSoilAmendmentProducts` ([stories](http://localhost:6006/?path=/docs/page-task-addsoilamendmentproducts--docs))
- `ProductCard` component
- `ProductDetails` component
   - Takes `productId` and renders product data. In change of saving/editing products.
   - Fields not included in the initial requirements will be added in [LF-4251](https://lite-farm.atlassian.net/browse/LF-4251).
   - The NPK validation should be updated and additional unit options should be added in LF-4251.

Jira link: https://lite-farm.atlassian.net/browse/LF-4220

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

[LF-4251]: https://lite-farm.atlassian.net/browse/LF-4251?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/webapp/src/components/Form/Button/SmallButton.tsx
```diff
@@ -0,0 +1,87 @@
...
+    icon?: ReactNode;
+    format?: (text: string) => string;
+  };
+}
+
+const CONFIG: Config = {
+  remove: {
+    translationKey: 'common:REMOVE',
+    icon: <XIcon />,
+    format: (text) => text.toLowerCase(),
```
Duncan-Brain, line 45: Should we use `toLocaleLowerCase()`?

packages/webapp/src/components/Task/AddSoilAmendmentProducts/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+  const [invalidProducts, setInvalidProducts] = useState<string[]>([]);
+
+  const { t } = useTranslation();
+  const { control, setValue, watch } = useFormContext();
+  const { fields, append, remove } = useFieldArray({
+    name: FIELD_NAME,
+    control,
+  });
+
+  const productsOfType = useMemo(
```
Duncan-Brain, line 52: I think this was already here from existing add product?  I always think about: https://react.dev/reference/react/useMemo#how-to-tell-if-a-calculation-is-expensive and wonder if we can remove the memo.

Duncan-Brain (commented):
Looks great too!

Only thing I was wondering was if it was expected that the react select option details would change on save for this story? (instead of 2:0:3.3 -> 4:0:3.3)
Screenshot 2024-06-04 at 4 47 57 PM

SayakaOno (commented):
@Duncan-Brain thank you for reviewing!
For this story, the select option details are not supposed to be updated on save. It should be automatic once we integrate with API, so I didn't implement it.
Sorry for the confusion 🙏

packages/webapp/src/components/Task/AddSoilAmendmentProducts/index.tsx
Duncan-Brain: I guess all of the above haha. I am just trusting the React docs that unless looping over 1000's of items its not worth it. 🤷

SayakaOno: The doc also says "There is no significant harm to doing that" so I was wondering.
I don't like that the filtering keeps happening when products don't change, but it doesn't affect the performance so I'm okay without useMemo.
I wish we could filter the product somewhere else!

kathyavini: > I wish we could filter the product somewhere else!

How about in a new selector in `productSlice`? 😄 

I think that would end up equivalent to `useMemo()` but with Redux Toolkit handling the memoization. I might be on a selector kick, but I think that would be readability without recalculating.

SayakaOno: That's what I wanted to do! I would need to update three files, so I hesitated to go for it.
But now that you mentioned, I will create a new selector 😄

Duncan-Brain (approved):
Looks great!

packages/webapp/src/components/Task/AddSoilAmendmentProducts/ProductCard/ProductDetails.tsx
```diff
@@ -0,0 +1,266 @@
...
+      unExpand();
+    } else {
+      reset();
+      setIsEditingProduct(false);
+    }
+  };
+
+  const onSubmit = (data: FormFields) => {
+    const callback = isNewProduct(productId) ? setProductId : undefined;
+    onSave({ ...data, farm_id, product_id: productId, type: TASK_TYPES.SOIL_AMENDMENT }, callback);
```
kathyavini, line 160: Clever! So this callback will get passed to both PATCH and POST mutations? It's a bit easier to imagine with Saga but I like how it's all set up.

packages/webapp/src/stories/Pages/Task/AddSoilamendmentProducts/AddSoilAmendmentProducts.stories.tsx
```diff
@@ -0,0 +1,77 @@
+/*
```
kathyavini, line 1: There's a capitalization error in the folder name if you want to correct that.

packages/webapp/src/components/Task/AddSoilAmendmentProducts/index.tsx
```diff
@@ -0,0 +1,134 @@
...
+                setValue(`${namePrefix}.product_id`, id);
+              }}
+              setFieldValidity={getInvalidProductsUpdater(field.id)}
+            />
+          );
+        })}
+      </div>
+      {!isReadOnly && (
+        <TextButton
+          disabled={!!invalidProducts.length}
```
kathyavini, line 122: Is it correct that a farm not pursuing certification can add an infinite number of empty product cards since the `invalidProducts` array is empty?

Screenshot 2024-06-14 at 10 05 35 AM

kathyavini (approved):
🙇 🙇 🙇  Very sorry for the very late review. For yesterday's delay at least I will also blame Storybook as I had crashes on this branch all afternoon and I am still not sure what resolved them. That did however force to me carefully read the code instead of testing which I enjoyed -- these are very well-written components!!

Please merge 🙏  Since you are continuing with this component any comments that you might want to address can go on the next ticket.

packages/webapp/src/components/Task/AddSoilAmendmentProducts/index.tsx
```diff
@@ -0,0 +1,134 @@
...
+import ProductCard, { type ProductCardProps } from './ProductCard';
+import TextButton from '../../Form/Button/TextButton';
+import { type Product } from './types';
+import { defaultValues } from './ProductCard/ProductDetails';
+import { ReactComponent as PlusCircleIcon } from '../../../assets/images/plus-circle.svg';
+import styles from './styles.module.scss';
+
+export type AddSoilAmendmentProductsProps = ProductCardProps & { products: Product[] };
+
+type ProductId = 'number' | 'string';
```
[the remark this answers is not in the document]
SayakaOno: 😨!!! Thank you for catching, it's surprising for me too!
I'll import `ProductId` type!

packages/webapp/src/components/Task/AddSoilAmendmentProducts/index.tsx
SayakaOno: No, it's a bug! The button shouldn't be enabled if a product hasn't been selected!
Thank you :)

packages/webapp/src/stories/Pages/Task/AddSoilamendmentProducts/AddSoilAmendmentProducts.stories.tsx
SayakaOno: Oh, thank you for your careful eye!

SayakaOno (commented):
Thank you @kathyavini for reviewing!

There are some errors I don't like, so I'll update the PR.
I'm sorry to dismiss your approvals!

packages/webapp/src/components/Form/Button/SmallButton.tsx
```diff
@@ -0,0 +1,87 @@
...
+interface Config {
+  [key: string]: {
+    translationKey: string;
+    icon?: ReactNode;
+    format?: (text: string) => string;
+  };
+}
+
+const CONFIG: Config = {
+  remove: {
```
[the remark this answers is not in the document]
kathyavini: Oh! Thank you for keeping the release merge conflicts in mind. I would have cried a bit when I realized 🤣

SayakaOno:
Major changes I made after the approvals are to:
- fix the "Add another product" button so it's not enabled if a product hasn't been selected (for farms not pursuing certification).
- adjust the position of the remove button on mobile view (I may need to adjust breakpoints later).
- display remove buttons when the number of products is greater than one, based on the discussion in grooming.

Thank you for re-reviewing!

kathyavini (approved):
Looks great, thank you!

PR #3216, opened 2024-05-29
LF-4245 The translation for the Spanish language are "Missing"

kathyavini:
Okay should now be good!

**Description**

Almost certainly I messed something up in the merge editor and then ran `pnpm i18n` with some sort of syntax issue to create the destructive commit. I thought maybe the issue was persisting for me, but now I'm not so sure 😬 Should be good now though as `pnpm i18n` is running without issue!

(Thank you @Duncan-Brain for finding the offending commit which was https://github.com/LiteFarmOrg/LiteFarm/commit/d613163921265275a8277eb17d65abed78abc7b2)

Jira link: https://lite-farm.atlassian.net/browse/LF-4245

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

Goal was running `pnpm i18n` without creating a new diff.

**Checklist:**


kathyavini:
> How did you find "interpolated keys"?

@SayakaOno I had no method -- I just happened to notice that a few strings had been removed in the diff, so I searched them manually and found that they were constructed by interpolation. As always it would be better if there was some way to make the script aware of these!!

SayakaOno:
@kathyavini Ah okay 😂🙏 It looks like the workaround I mentioned in this [ticket](https://lite-farm.atlassian.net/browse/LF-3886) is still the only way.

kathyavini:
@SayakaOno I had forgotten all about that! By workaround you mean this part of the ticket description, right?

> It can be avoided by specifying possible static values in comments anywhere in our file as mentioned in the link above.

I don't think the ignore file would have been relevant in this case (here the keys weren't even present yet to be removed) but those comments would absolutely have done it!

```js
// t('ANIMAL.ATTRIBUTE.ANIMAL_IMAGE')
// t('ANIMAL.ATTRIBUTE.BATCH_IMAGE')
label={t(`ANIMAL.ATTRIBUTE.${animalOrBatch.toUpperCase()}_IMAGE`)}
 ```
 
Should we require implementing these comments when we write any interpolated key? It fixes both population with `pnpm i18n` and even `keepRemoved: false` for the future (ignore file still necessary for all the other old interpolated keys if we ever wanted to run that again).

SayakaOno:
@kathyavini 
> Should we require implementing these comments when we write any interpolated key?

I love the idea, it would be nice if keys were auto-generated! ❤️

Duncan-Brain (approved):
Interesting! Does that mean we should make those interpolated explicit somewhere?

PR #3217, opened 2024-05-29
LF-4240: April 2024 data policy updates (Version 6: May 31, 2024)

SayakaOno:
**Description**

Data policy updates (Version 6: May 31, 2024).
Branched out from the main branch.

Jira link: https://lite-farm.atlassian.net/browse/LF-4240

PR #3222, opened 2024-06-05
LF-4256 Old consent version is shown after setting `has_consent` to false

kathyavini:
**Description**

This PR:
- Cherry-picks @SayakaOno's commits from #3217 that created and used the new consent version constant
- Uses that constant to check against the database stored consent version and treats a mismatch as equivalent to `!has_consent`(this is actually fundamentally how the release badge works as well 🙂) in the fetchAll saga and in routes

Notes:
- It is the routes file change that had been triggering the need to re-consent upon route change. fetchAllSaga would be more relevant to initial login I think and might not be necessary, but included for completeness
- We could instead `logout()` instead of routing to `/consent`; I left it as is and I don't think the user experience is that bad, but that is an option since `.env` key logout would no longer help us here
- In the future the only change necessary will be updating the consent version in the constants file along side the updated text

Jira link: https://lite-farm.atlassian.net/browse/LF-4256

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

This demonstrates how the re-consenting is required. It should be merged together with the changes to the consent text (as will happen in the patch PR) for the timing to be correct.

**Checklist:**

### DIFF
--- a/packages/webapp/src/containers/Consent/index.jsx
+++ b/packages/webapp/src/containers/Consent/index.jsx
@@ -42,7 +42,7 @@ function ConsentForm({
 
     formState: { errors },
   } = useForm();
-  const [consentVersion] = useState('5.0');
+  const [consentVersion] = useState('6.0');
   const consent =
     role.role_id === 3 ? getLanguageConsent(language).worker : getLanguageConsent(language).owner;
   const checkboxName = 'consentCheckbox';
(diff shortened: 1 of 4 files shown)


Duncan-Brain (approved):
Looks promising nice solution you two! So happy we don't have to add migrations anymore!

SayakaOno (approved):

PR #3225, opened 2024-06-06
LF-4240 (2): April 2024 data policy updates (Version 6: May 31, 2024)

SayakaOno:
**Description**

Add back https://github.com/LiteFarmOrg/LiteFarm/pull/3217 without the migration file.


kathyavini (approved):
A bit curious to see the git log for the two cherry-picked commits which also exist on the other feature branch too now -- I guess they will be all straightened out and probably disappear once all merged?!

Duncan-Brain (approved):
WOO!

SayakaOno:
@kathyavini I forgot to reply! We have duplicate cherry-picked commits, probably because their commit hashes are different.

PR #3230, opened 2024-06-11
LF-4238: Fix cleared filters on the Finances page

Joaquin-M2:
**Description**

No results were shown on the "Finances" page when the user tries to "Clear" all the previously selected filters.

Jira link: https://lite-farm.atlassian.net/browse/LF-4238

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/containers/Finances/TransactionFilter/index.jsx
+++ b/packages/webapp/src/containers/Finances/TransactionFilter/index.jsx
@@ -44,7 +44,7 @@ const TransactionFilter = () => {
       return filter.active === true;
     });
 
-    // If filterKey has no active filters, reset its value to undefined:
+    // If filterState has no active filters, reset the value of its filterKey on 'tempTransactionsFilter' to undefined:
     if (!hasActiveFilters) {
       setTempTransactionsFilter((prevState) => {
         prevState[filterKey] = undefined;
--- a/packages/webapp/src/containers/Finances/TransactionFilter/index.jsx
+++ b/packages/webapp/src/containers/Finances/TransactionFilter/index.jsx
@@ -56,6 +56,9 @@ const TransactionFilter = () => {
   };
 
   const handleApply = () => {
+    checkActiveFilters('EXPENSE_TYPE');
+    checkActiveFilters('REVENUE_TYPE');
+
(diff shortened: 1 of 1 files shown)


packages/webapp/src/containers/Finances/TransactionFilter/index.jsx
```diff
@@ -38,7 +38,27 @@ const TransactionFilter = () => {
...
+      Object.values(tempTransactionsFilter[filterType]).some((filter) => {
+        return filter.active === true;
+      });
+
+    // If filterType has no active filters, reset its value to undefined:
+    if (!hasActiveFilters && tempTransactionsFilter[filterType]) {
+      setTempTransactionsFilter((prevState) => {
+        prevState[filterType] = undefined;
+        return prevState;
+      });
```
SayakaOno, line 54: Thank you for working on this!
Would it be a problem if we did this in `onChange` that sets `tempTransactionsFilter`? It would be nice if we could set the correct value in the first place!

Joaquin-M2: Nice suggestion. Thank you 🙂 

I just refactored the full function and moved its execution into the `onChange` event, leaving `handleApply` as it was before this PR.

packages/webapp/src/containers/Finances/TransactionFilter/index.jsx
```diff
@@ -38,6 +38,26 @@ const TransactionFilter = () => {
...
+        prevState[filterKey] = undefined;
+        return prevState;
+      });
+    } else {
+      setTempTransactionsFilter((prevState) => {
+        prevState[filterKey] = filterState;
+        return prevState;
+      });
+    }
+  };
```
SayakaOno, line 59: `prevState[filterKey] = ...` mutates the state directly, which should be avoided.
https://react.dev/reference/react/useState#my-initializer-or-updater-function-runs-twice

Also, you can combine these two by moving the conditions inside `setTempTransactionsFilter` or using a ternary operator. (I prefer the latter!)

SayakaOno (approved):
Looks good, thank you!

Duncan-Brain (approved):
👌 very nice!

PR #3234, opened 2024-06-14
LF-4216 - Add case to insert null values when quantity does not exist

Duncan-Brain:
**Description**

There is a check requiring `all null OR all not null` on quantity types volume and weight.

This sets null when quantity value does not exist. Only cases on prod where this might be true are initial task migrations mentioned in[ LF-4246 ](https://lite-farm.atlassian.net/browse/LF-4246)

EDIT Also: 
- Some record could not being inserted correctly in theory without product_quantity. I assumed defaultTo == notNullable which is not the case.
- Some records were not being rolled back correctly. Anything without product_quantity would lose other properties. On prod all without product_quantity are empty records, but I should still handle this case.

Error from docker log litefarm-api:

Screenshot 2024-06-14 at 1 10 10 PM

Jira link:

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain):

To test set a product_quantity == null on some soil_amendment_task before migrating. Error should no longer occur.

**Checklist:**

### DIFF
--- a/packages/api/db/migration/20240504185218_create_soil_amendement_task_products_table.js
+++ b/packages/api/db/migration/20240504185218_create_soil_amendement_task_products_table.js
@@ -246,6 +246,13 @@ export const up = async function (knex) {
   });
 
   // Migrate existing weight data to the new table (reversibly)
+  // This case should never happen - but if it does lets not delete data
+  await knex.raw(`
+    INSERT INTO soil_amendment_task_products (task_id, product_id, weight, weight_unit, application_rate_weight_unit)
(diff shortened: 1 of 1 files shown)


kathyavini (approved):
Looks good! This one was discussed [on Slack](https://litefarm.slack.com/archives/C05B9KJLJ82/p1718384943817789) instead of here but in summary: I'm a big fan of the non-destructive rollbacks just in case 😉

SayakaOno (approved):

PR #3236, opened 2024-06-17
LF-4248: Make composition inputs component reusable

SayakaOno:
**Description**

#### NumberInputWithSelect
- remove hard-coded field names (`npk_unit` and `NPK`)
- take `value` and `unit` separately instead of `values`
- make it compatible with one unit
- make selectWidth and justifyContent props so that it can display longer units like "mg/kg"

#### CompositionInputs
- take `mainLabel`
- fix unit's margin for disabled state
- align up to 3 items in a row
- remove the border around the inputs
- add more [stories](http://localhost:6006/?path=/story/components-compositioninput--one-unit)

#### CompositionInputs
- make adjustments

NOTE: I will change `npk_unit` to `elemental_unit` in the next PR.

Jira link: https://lite-farm.atlassian.net/browse/LF-4248

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


Duncan-Brain:
Very cool! Love this new component!

Things I noticed:
1. I think the unit options color be black instead of grey.

Screenshot 2024-06-18 at 11 10 14 AM

2. I think AddSoilAmendmentProduct Story needs to be edited, I am not getting an expected error for > 100:
Screenshot 2024-06-18 at 11 30 45 AM

Question:
1. Does the new number input, which uses new InputBase. How are we handling database stored value? In Units input component you fixed last year it had the "hidden" value I think. Are we moving this conversion to the saga or controller? Maybe I am forgetting something. (also it is maybe not applicable to this specific case where units are ratio and percent since there is no database preference)

packages/webapp/src/components/Form/CompositionInputs/NumberInputWithSelect.tsx
```diff
@@ -24,55 +24,47 @@ import styles from './styles.module.scss';
...
 
-export const UNIT = 'npk_unit';
-
-export enum NPK {
-  N = 'n',
-  P = 'p',
-  K = 'k',
-}
-
 export enum Unit {
```
Duncan-Brain, line 27: Do we still need this here? Can Units be anything now?

SayakaOno: I changed the type to `unit: Unit | string;`, so units can be anything, but I guess it's weird to use the enum as const...😅  I'll change the enum to const!

SayakaOno (commented):
Thank you @Duncan-Brain for reviewing!
1. 🙏 
2. Could you try changing the unit to %? It should still be working!
3. Good catch... I'll come up with a solution in my next PR where I implement those inputs!

The answer to your question:
As you mentioned, this isn't applicable since there's no database unit. Unless unit system conversions are needed, we can use the normal input! ❤️

Duncan-Brain (approved):
2. Oops good eye! I messed up with ratio haha

Thanks @SayakaOno for answering my question! 

For some reason I thought that any new number input that has units would need to be able to use the base units. I wonder if there should be a comment or warning not to use some types of units. 

And just checking my understanding: 
a. Composition inputs is designed to work for cases that require all units to be the same and some calculation based on inputs. 
b. Composition units must have no specific database unit (stored as entered), and the units are all stored the same. 
c. It will not work for something like the new weight/volume component where application rate, weight quantity, and area react to each other but have database units and unit values are different.

SayakaOno:
@Duncan-Brain Thank you for clarifying!

a. The first half is correct: the composition inputs are designed for a set of inputs where all units need to be the same. Calculations don't matter; they are managed by the parent component if needed.
b. Right, values and units are stored as entered. All units must be consistent.
c. Correct, this component won't work for inputs where unit options differ or selected units should not be in sync.

I'll add a comment to the component to clarify this.

packages/webapp/src/components/Form/CompositionInputs/styles.module.scss
```diff
@@ -41,44 +41,50 @@
...
       @include truncateText();
     }
+  }
 
-    .selectWrapper {
-      border-left: none;
-    }
+  &:not(.disabled) .selectWrapper {
+    margin-right: -8px;
+    border-left: 1px solid var(--grey400);
```
kathyavini, line 48: Sorry could you remind me (it feels like maybe something that we touched on in grooming a few weeks ago?) that we ended up having this line to the left of the unit? Was it to make it more consistent with the `<Unit />` component?

packages/webapp/src/components/Form/CompositionInputs/index.tsx
```diff
@@ -13,53 +13,68 @@
...
+                name={name}
+                error={error}
+                disabled={disabled}
+                onChange={onChange}
+                onBlur={onBlur}
+                unit={unit || values?.[unitFieldName]}
+                unitFieldName={unitFieldName}
+                value={values?.[name]}
+                reactSelectJustifyContent={
+                  reactSelectJustifyContent || (disabled ? 'flex-end' : 'flex-start')
```
kathyavini, line 65: And here is well; is this a `<Unit />` pattern? In Figma the content seems 'flex-end' justified in all cases.

kathyavini (approved):
The generalizations and new stories look great!

This is more of a question for the original component more than the re-usability edits, but can you refresh my memory on why the style diverges from [Figma](https://www.figma.com/design/eMffC0h3JPlWtzNttxJpil/Litefarm-Updated?node-id=13495-213049&m=dev) regarding the border to the left of the unit and the left (flex-start) justified text?

I thought maybe at some point it was discussed and considered an easier implementation, but in this case it seemed to require a bit more code, so was there another reason?

packages/webapp/src/components/Form/CompositionInputs/index.tsx
kathyavini: Oh, or it perhaps to accommodate the way the error `X` is positioned?

packages/webapp/src/components/Form/CompositionInputs/styles.module.scss
SayakaOno: I thought I would need to add a border to make it consistent... but maybe it was the white background...
Screenshot 2024-06-19 at 1 59 45 PM
https://lite-farm.atlassian.net/wiki/spaces/LITEFARM/pages/1512308854/S86+Grooming

packages/webapp/src/components/Form/CompositionInputs/index.tsx
SayakaOno: With the border, it seems strange to have the unit right-aligned...

Screenshot 2024-06-19 at 2 15 52 PM

SayakaOno (commented):
@kathyavini Thank you for reviewing!
I just assumed for some reason that the border was left out in figma 😂 
It would be great if we could allocate some time tomorrow during the planning to finalize the styling!

packages/webapp/src/components/Form/CompositionInputs/index.tsx
kathyavini: Definitely! 😂  I was only thinking in the case of no border.

kathyavini:
@SayakaOno ah yes it was the _background colour_ we discussed that one time, because it was originally grey!! Thank you for reminding me 🙏 

With the wider unit area I find the line + gap between the `%` and the caret a little jarring, but I would be very very surprised if Loïc doesn't say we can go whatever is easiest to implement.

We can still do a final style check at sprint planning tomorrow and merge before or after as you prefer.

SayakaOno:
@kathyavini Sounds good, thank you!
I'll merge this PR now and include any adjustments in the next PR!

PR #3275, opened 2024-06-26
LF-3884 check if location exists in request body

navDhammu:
**Description**

Unable to create a task when selecting a wild crop in the task creation flow.

**Issue:**
A 400 error was being returned from api due to a postgres error when quering the db to check for a deleted location before creating a task. When creating a task that targets a wild crop, the wild crop doesn't have a location associated to it so the request body has an empty locations property of `[]`, which means `req.body.locations[0]` is undefined and sending undefined to postgres results in an internal error that is caught in the catch block, sending a 400 response.

**Fix:**
To fix this I made a simple check to ensure that locations is not empty before checking if the location is deleted.

**Question:**
As a user on the app, I don't think its possible to create a task (that is not for a wild crop) without selecting a location first on the map. But isn't it possible for someone to bypass that and send a post request to create a task without a location? I tried and was able to do it, I got a success response and I see the record created in db, but it doesn't show up in the app. Just wondering if that might be a problem or if its intended to be this way?

Jira link: https://lite-farm.atlassian.net/browse/LF-3884

**Type of change**

**How Has This Been Tested?**

- [x] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/taskController.js
+++ b/packages/api/src/controllers/taskController.js
@@ -319,13 +319,14 @@ const taskController = {
     return async (req, res, next) => {
       try {
         // Do not allow to create a task if location is deleted
+        const locations = req.body.locations;
         if (
-          await baseController.isDeleted(null, Location, {
-            [Location.idColumn]: req.body.locations[0]?.location_id,
-          })
-        ) {
+          locations?.length &&
+          (await baseController.isDeleted(null, Location, {
+            [Location.idColumn]: locations[0].location_id,
+          }))
+        )
           return res.status(409).send('location deleted');
-        }
 
         // OC: the "noInsert" rule will not fail if a relationship is present in the graph.
         // it will just ignore the insert on it. This is just a 2nd layer of protection


kathyavini (approved):
Thank you so much @navDhammu!! I was hoping it was an easy fix 🙏

@SayakaOno may we scope it to release?

SayakaOno:
@navDhammu Thank you for working on this!

> But isn't it possible for someone to bypass that and send a post request to create a task without a location?

How does the user bypass it? Is it the same thing as creating a task for a wild crop?


@kathyavini Yes! I'll add to the release scope, thank you!

navDhammu:
@SayakaOno I meant like someone can just use postman for example to send the request with an empty locations array for a task that requires the user to select a location on the app. I tested it directly in the code by setting req.body.locations = [] to simulate that and it was successful in creating a task without a location.

SayakaOno:
@navDhammu Thank you for clarifying.
When creating a task for a wild crop, `locations` is `[]`, so I was wondering what the different was. (I see the created task in the app. Did you check `/tasks` to find the task you created from postman?)
Screenshot 2024-06-26 at 3 01 43 PM

navDhammu:
@SayakaOno Sorry I meant when creating a task that's **not** for a wild crop. In that case I think you have to have a location?

SayakaOno:
@navDhammu Sorry, that was my wrong assumption that a task with `locations: []` is considered as a task for wild crop. I thought that the difference between wild crops and non-wild crops was whether the location was empty or not.

I'll double check the request body for non-wild crops and create a ticket to fix the behaviour.

SayakaOno (approved):

PR #3280, opened 2024-06-27
LF-3951: Show water calculator with wild crops

Joaquin-M2:
**Description**

Users were not able to open the "Water calculator" (_Create a task_ > _Irrigation task_)  if the water measurement method was "Depth".

It is expected that the **Calculate Water Usage** link will be disabled (+ a warning) for wild crops if the measure of water use is set to **Depth**:

image

image

Jira link: https://lite-farm.atlassian.net/browse/LF-3951

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/Task/PureIrrigationTask/index.jsx
+++ b/packages/webapp/src/components/Task/PureIrrigationTask/index.jsx
@@ -197,7 +197,7 @@ export default function PureIrrigationTask({
   const waterCalculatorDisabled =
     !showWaterUseCalculatorModal ||
     !measurement_type ||
-    (measurement_type === 'DEPTH' && !(location || getValues().show_wild_crop));
+    (measurement_type === 'DEPTH' && !location);
 
   return (
     <>
--- a/packages/webapp/src/components/Task/PureIrrigationTask/index.jsx
+++ b/packages/webapp/src/components/Task/PureIrrigationTask/index.jsx
@@ -16,6 +16,7 @@ import { getIrrigationTaskTypes } from '../../../containers/Task/IrrigationTaskT
 import { useDispatch, useSelector } from 'react-redux';
 import { irrigationTaskTypesSliceSelector } from '../../../containers/irrigationTaskTypesSlice';
 import { cropLocationsSelector } from '../../../containers/locationSlice';
+import { BsFillExclamationCircleFill } from 'react-icons/bs';
 
 export default function PureIrrigationTask({
   system,
@@ -199,6 +200,7 @@ export default function PureIrrigationTask({
     !measurement_type ||
     (measurement_type === 'DEPTH' && !location);
 
+  const calculatorLinkIsDisabled = measurement_type === 'DEPTH' && !location;
   return (
     <>
       <Controller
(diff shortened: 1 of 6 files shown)


packages/webapp/src/components/Task/PureIrrigationTask/index.jsx
```diff
@@ -197,7 +197,7 @@ export default function PureIrrigationTask({
   const waterCalculatorDisabled =
     !showWaterUseCalculatorModal ||
     !measurement_type ||
-    (measurement_type === 'DEPTH' && !location);
+    (measurement_type === 'DEPTH' && !(location || getValues().show_wild_crop));
```
kathyavini, line 200: Using this form value works on task creation, but if the user wants to open the calculator when they complete the task, `show_wild_crop` won't be available -- I think it's unique to creation and maybe something else like `pinCoordinates` would have to control the logic for task completion?

kathyavini (commented):
Thank you for the fix and for your patience with our slow reviews around the soil amendment release!

I'll bring this ticket up with our designer during our weekly meeting on Wednesday (and I'll send you an invite in case you would like to join as well -- totally optional of course!) 

The ticket description as written:

> And the user is able to make the calculation both for “Volume” or “Depth” method.

doesn't make sense because there is no depth calculation that can be done without location area, and so the calculator doesn't calculate anything:

Screenshot 2024-07-29 at 3 40 24 PM

In soil amendment we went the route of prohibiting pin-based wild crops and I wonder if we might end up doing something similar here. If not, I think the way it works on your branch is good! It just needs a slight tweak as noted below, but let's discuss first with Loïc about the calculator opening + but not calculating.

Joaquin-M2:
@kathyavini I have updated the description of the PR considering what we talked about during the Dev-Design meeting from last week. I talked to Loïc on Slack and I got his confirmation about how it should look like.

On the other hand, you will see in the last commits that I've used inline styling for the additional jsx. I just tried to be consistent with the old code, but if you think I should move that into its own `module.scss` file please let me know.

Duncan-Brain (commented):
@Joaquin-M2 I think it is coming together thanks for working on this!

My main two questions are about the information label:
- Was the red color what Loic suggested? To me it looks too similar to a form validation error. If he approved please ignore me.
- on mobile view it looks a little cramped - should it be put on a newline if it reaches a touch breakpoint? 
Screenshot 2024-08-12 at 10 21 47 AM

Joaquin-M2:
@Duncan-Brain The red color was suggested by Loïc, yes. The warning icon is a personal addition, so I can remove it if you think it doesn't help.

I also moved the warning message to a new line if the viewport width is smaller than 599px:

image

Duncan-Brain (dismissed):

packages/webapp/src/components/Task/PureIrrigationTask/styles.module.scss
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+@import '../../../assets/mixin.scss';
+
+.waterCalculatorWrapper {
+  display: flex;
+  gap: 10px;
+  margin-top: 4px;
+
+  @include xs-breakpoint {
```
kathyavini, line 23:
```suggestion
  @include md-breakpoint {
```
I think I prefer the wrapping right from the tablet breakpoint (and it will probably make even more sense with the new languages coming in this month) -- does this breakpoint look okay to you too?

packages/webapp/src/components/Task/PureIrrigationTask/index.jsx
```diff
@@ -303,16 +306,42 @@ export default function PureIrrigationTask({
...
+              </Label>
+              <div className={styles.waterCalculatorWarningWrapper}>
+                <BsFillExclamationCircleFill />
+                <Label
+                  style={{
+                    marginBottom: `${disabled ? 36 : 0}px`,
+                    color: 'var(--Colors-Accent---singles-Red-full)',
+                  }}
+                >
+                  For wild crops, estimation of water usage by depth is not available.
```
kathyavini, line 339: This one too should be a translated string in i18n and not hardcoded.

kathyavini (commented):
This is fantastic! Thank you so much for implementing and doing all the coordinating with Loïc too -- it's super clear + a big improvement! The only thing I noticed missing is that the string should be translated rather than hardcoded.

Re: the width @Duncan-Brain's screenshot shows, I think I agree it would be nice to wrap right from tablet breakpoint, but Loïc's stance has always been we don't QA for tablet width so either way is good.

And I'll make sure to re-review quickly this time!! -- sorry for being pokey again 🙇

Edit: and the styling solution looks good to me!

kathyavini (approved):
Wow that was so fast! 😮 🙌

Duncan-Brain (approved):

PR #3241, opened 2024-06-19
LF-4251: Add more inputs to AddSoilAmendmentProducts component

SayakaOno:
**Description**

I rebased the branch! ⚠️ 🙏  

- Update `useExpandableItem` to handle empty props.
- Update `CompositionInputs` and `NumberInputWithSelect` to
  - handle single unit. (If there are no unit options, `unit` field will not be registered for `react-hook-form`)
  - update `justifyContent` to centre the selected unit. (It will still be right-aligned when disabled.) 
  - display error message conditionally.
- Add unit options to nutrients. (ppm, mg/kg)
- Add inputs for:
  - purpose
  - fertiliser type
  - moisture/dry matter contents
  - additional nutrients
- Improve the condition to disable "add another product" button
- Update [stories](http://localhost:6006/?path=/docs/page-task-addsoilamendmentproducts--docs)

Jira link: https://lite-farm.atlassian.net/browse/LF-4251

**Type of change**

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

**How Has This Been Tested?**

- [ ] Passes test case
- [x] UI components visually reviewed on desktop view
- [x] UI components visually reviewed on mobile view
- [ ] Other (please explain)

**Checklist:**

- [ ] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [x] The precommit and linting ran successfully
- [ ] I have added or updated language tags for text that's part of the UI
- [ ] I have added "MISSING" for all new language tags to languages I don't speak
- [x] I have added [the GNU General Public License](https://lite-farm.atlassian.net/l/cp/BT0Dd7WW) to all new files

PR #3289, opened 2024-07-05
LF-4249 Create <QuantityApplicationRate  /> component

kathyavini:
**Description**

This creates the `<QuantityApplicationRate />` component. The PR includes:
- two new [custom measures](https://github.com/LiteFarmOrg/LiteFarm/blob/ccb4a4ef7cbcbb0be3dee2bffe0874de23584dd0/packages/webapp/src/util/convert-units/applicationRates.ts) and database units for application rate, as these are not units included within the [convert-units](https://github.com/convert-units/convert-units) package
- Small revisions to `<NumberInput />` and `<Unit />` to pass 1) label styles, 2) additional onChange callbacks to Unit
- [Storybook story](http://localhost:6006/?path=/docs/components-quantityapplicationrate--docs) for isolated component
- Updates to the parent component `<SoilAmendmentProductCard />` and its [Storybook story](http://localhost:6006/?path=/docs/page-task-addsoilamendmentproducts--docs) to integrate the QAR component

Not included:
- play tests for the new kind of `<Unit />`; this was merely due to lack of time. I could revisit if the team thinks it's a good next priority
- form value handling according to the ultimate structure of the form this will be embedded in -- @SayakaOno I wasn't sure what this was going to look like exactly and the relationship of the two sets of form fields (product and task product)... will they be wrapped in one `<FormProvider />`? This calls `useForm()` in the component which is just an interim solution. Please let me know if should refactor to `useFormContext()`, and please also feel free to adjust anything to integrate with the other half of task product! 🙏 

Jira link: https://lite-farm.atlassian.net/browse/LF-4249

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/useQuantityApplicationRate.ts
```diff
@@ -0,0 +1,186 @@
...
+  const application_rate_weight_unit = watch(TASK_PRODUCT_FIELD_NAMES.APPLICATION_RATE_WEIGHT_UNIT);
+  const application_rate_volume_unit = watch(TASK_PRODUCT_FIELD_NAMES.APPLICATION_RATE_VOLUME_UNIT);
+
+  const [isWeight, setIsWeight] = useState(() => {
+    const volumeValue = getValues(TASK_PRODUCT_FIELD_NAMES.VOLUME);
+    return isNaN(Number(volumeValue));
+  });
+
+  const toggleMeasure = () => {
+    setIsWeight((prev) => !prev);
```
kathyavini, line 54: Sorry I forgot to mention that I did go with the third option for toggle behaviour regarding what we looked at in tech daily yesterday (i.e. letting both WEIGHT and VOLUME be saved within their respective components). It just saves so much code relative to converting to maintain the input.

However, if this is the UX we go with, I think this means `isWeight` will need to be added to (and later removed from) the form fields so that the selected value can be identified and the other one cleared at the point of form submission. I'm not sure whether that is preferable to nulling all fields on toggle, so right now neither is implemented, but please let me know which you think is better and of course if there is another option I'm forgetting.

packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/useQuantityApplicationRate.ts
```diff
@@ -0,0 +1,186 @@
...
+      /* @ts-ignore */
+      getUnitOptionMap()[getDefaultUnit(location_area, calculatedArea, system).displayUnit],
+      { shouldValidate: true },
+    );
+
+    /* update application rate */
+    onQuantityChange();
+  };
+
+  const onQuantityChange = () => {
```
SayakaOno, line 92: Looking at the function name, I expect the function to be called when the quantity is updated... (the same for `onPercentLocationChange`) `setQuantity` or `updateQuantity` might be more suitable.

packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/useQuantityApplicationRate.ts
```diff
@@ -0,0 +1,186 @@
...
+    const weight = getValues(TASK_PRODUCT_FIELD_NAMES.WEIGHT);
+    const weight_unit = getValues(TASK_PRODUCT_FIELD_NAMES.WEIGHT_UNIT);
+    const application_area = getValues(TASK_PRODUCT_FIELD_NAMES.APPLICATION_AREA);
+
+    const volume = getValues(TASK_PRODUCT_FIELD_NAMES.VOLUME);
+    const volume_unit = getValues(TASK_PRODUCT_FIELD_NAMES.VOLUME_UNIT);
+
+    if (weight && isWeight && weight_unit && application_area) {
```
SayakaOno, line 100: I think `0` should be handled?

packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/useQuantityApplicationRate.ts
SayakaOno: I think it's good! (I prefer whatever option results in less code for now since Loïc is away)
I would like to move up `isWeight` without adding to the form fields if possible.

SayakaOno (commented):
I'm still half way through, but I'll submit my review here!
- I'm going to wrap `<AddSoilAmendmentProducts />` with `<FormProvider />`, so `useFormContext` will be available in your component. I've added another `useForm` for product in [ProductDetails](https://github.com/LiteFarmOrg/LiteFarm/pull/3241/files#diff-938e8314a18ead14c52d6864c40837734ee221c7082c5097da1bfcc24bc75385) at line 148. I can adjust the userForm in your component, so you can leave it as is :)
- `TASK_PRODUCT_FIELD_NAMES` should be combined!

packages/webapp/src/components/Form/InputBase/InputBaseLabel/index.tsx
```diff
@@ -26,14 +26,15 @@ export type InputBaseLabelProps = {
...
+  labelStyles?: React.CSSProperties;
 };
 
 export default function InputBaseLabel(props: InputBaseLabelProps) {
   const { t } = useTranslation();
 
   return (
     <div className={styles.labelContainer}>
-      <Label>
+      <Label style={props.labelStyles}>
```
Duncan-Brain, line 37: I am wondering if it should be possible to alter baseLabelStyles? I think we want all to be consistent or to have an option for primary or secondary or something along those lines. 

So you could probably add those styles here directly if they are what we want on all from now on.

packages/webapp/src/components/Form/Switch/index.tsx
```diff
@@ -22,8 +22,8 @@ import { Main } from '../../Typography';
 interface SwitchProps {
   checked?: boolean;
   onChange?: (event: ChangeEvent<HTMLInputElement>) => void;
-  label?: string;
-  leftLabel?: string;
+  label?: ReactNode | string;
+  leftLabel?: ReactNode | string;
```
Duncan-Brain, line 26: If you end up adding the formatted labels here then this has the benefit of just being a string again.

packages/webapp/src/util/convert-units/applicationRates.ts
```diff
@@ -0,0 +1,223 @@
...
+          plural: 'Fluid Ounces per Acre',
+        },
+        to_anchor: 1 / 128,
+      },
+    },
+  },
+  anchors: {
+    metric: {
+      imperial: {
+        ratio: 0.264172 / 2.47105, // l/gal * (ac/ha)
```
Duncan-Brain, line 109: Above you used (1/ 128) I think to be clear. You can consider doing the same here.

packages/webapp/src/components/Form/Unit/index.jsx
```diff
@@ -144,7 +145,9 @@ const Unit = ({
             size={1}
             onKeyDown={onKeyDown}
             onBlur={inputOnBlur}
-            onChange={inputOnChange}
+            onChange={(event) => {
+              inputOnChange(event), additionalOnChange?.();
```
Duncan-Brain, line 149: This seems like an elegant little change with big functionality! My only thought is it should mirror `onChangeUnitOption()` and be passed to the hook instead.

packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/index.tsx
```diff
@@ -0,0 +1,249 @@
...
+      <span className={styles.switchLabel}>{label}</span>
+    );
+  };
+
+  // Unit and NumberInput accept inline style objects, rather than classnames, to style subcomponents
+  const inputLabelStyles = {
+    color: 'var(--Colors-Neutral-Neutral-500)',
+    fontSize: '16px',
+    lineHeight: 'normal',
+  };
```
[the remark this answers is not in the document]
kathyavini: Okay, removed for now as discussed this morning 🙂

packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/index.tsx
```diff
@@ -0,0 +1,249 @@
...
+                  control={control}
+                  label={t('ADD_TASK.SOIL_AMENDMENT_VIEW.PERECENT_TO_AMEND')}
+                  min={0}
+                  max={100}
+                  rules={{ required: t('common:REQUIRED') }}
+                  onChange={onPercentLocationChange}
+                  classes={{ label: inputLabelStyles }}
+                  disabled={isReadOnly}
+                  defaultValue={
+                    defaultValues?.[TASK_PRODUCT_FIELD_NAMES.PERCENT_OF_LOCATION] || 100
```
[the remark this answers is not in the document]
kathyavini: Yeah, I agree they would be better synched. Since I need to keep this `100` to populate the default/initial load, I changed the percentOfArea line -- it was only for showing 0% in the "Applied to..." string when the input was cleared, but I think I even like it better this way:

Screenshot 2024-07-09 at 9 29 23 AM

packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/useQuantityApplicationRate.ts
```diff
@@ -0,0 +1,186 @@
...
+  /* Set initial application area to area total area */
+  useEffect(() => {
+    if (!application_area) {
+      setValue(TASK_PRODUCT_FIELD_NAMES.APPLICATION_AREA, total_area);
+
+      /* display user-selected unit initially */
+      setValue(
+        TASK_PRODUCT_FIELD_NAMES.APPLICATION_AREA_UNIT,
+        getUnitOptionMap()[total_area_unit] as UnitOption,
+      );
```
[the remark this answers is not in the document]
kathyavini: That is _exactly_ where I got it 😂 I assumed if it was necessary there it was necessary here as well for reasons beyond me... 😇

packages/webapp/src/components/Task/AddSoilAmendmentProducts/types.ts
```diff
@@ -67,3 +68,37 @@ export type Product = {
...
   product_translation_key: string | null;
 };
+
+export const TASK_PRODUCT_FIELD_NAMES = {
+  WEIGHT: 'weight',
+  WEIGHT_UNIT: 'weight_unit',
+  VOLUME: 'volume',
+  VOLUME_UNIT: 'volume_unit',
+  PERCENT_OF_LOCATION: 'percent_of_location',
+  APPLICATION_AREA: 'location_area',
```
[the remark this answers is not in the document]
kathyavini: SORRY 🙇  I totally forgot to go back and check the db.

packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/useQuantityApplicationRate.ts
kathyavini: That is when it's called!

This is the function that is called when the quantity is updated, and it sets either the `application_rate_weight` or the `application_rate_volume` depending on `isWeight`. Ditto for the `onPercentLocationChange` (it's passed to the `onChange` handler the for the percent location).

If the onChange naming convention is causing confusion I can write it the other way (this would be called `updateApplicationRate` in that case). However I'm not sure I would want to rename `onPercentLocationChange`, because it would become `updateAreaAndQuantity` which seems slightly less clear to me. Do you prefer that though?

packages/webapp/src/components/Form/InputBase/InputBaseLabel/index.tsx
kathyavini: Can you remind us to discuss this with Loïc when he gets back? I love the new styles and I hope we do use them throughout app, but unfortunately we didn't think to have that discussion before Loïc left, so I don't feel quite ready to push that thorough a change now.

Separately from that, I do like having the ability to style the label differently if needed, so I would like to keep (restore?) the label classes behaviour. But you're right that it should be for true one-offs and not for a systemic change like this.

packages/webapp/src/components/Form/Switch/index.tsx
kathyavini: I actually prefer the more flexible type! It leaves open the possibility of passing one of those styled strings that requires a `<Trans />` component -- we haven't seen it yet with switches but the new form buttons had an example.

packages/webapp/src/components/Form/Unit/index.jsx
kathyavini: You mean the `useUnit` hook? Let me defer to @SayakaOno on that! I thought the hook was specific to unit-changing logic, but you're right that's how the other callback is passed. This pattern I used here is just replicating `<NumberInput />` by the way, and I meant to tag Sayaka anyway to double-check on it 🙂

packages/webapp/src/util/convert-units/applicationRates.ts
kathyavini: 1/128 is the ratio of fluid ounces to gallons, but the numbers here are the ratios of litres to gallons + acres to hectares and there are no clean proportions for those (only approximations) as they are entirely different systems.

packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/index.tsx
```diff
@@ -0,0 +1,249 @@
...
+    system,
+    watch,
+    setValue,
+    getValues,
+  });
+
+  const percent_of_location = watch(TASK_PRODUCT_FIELD_NAMES.PERCENT_OF_LOCATION);
+
+  /* Label for weight/volume switch */
+  const formatLabel = (labelKey: string, shouldBeBold: boolean) => {
```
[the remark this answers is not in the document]
Duncan-Brain: I think we can assume Loic makes things that we want to be consistent in the future. And if he changes them we can then ask about the change. But yah only if you want to ! It would clean up this file a bit so it seemed nice to move out.

packages/webapp/src/util/convert-units/applicationRates.ts
Duncan-Brain: I am saying that this would be 0.264172 *( 1 / 2.47105), since the below line is 0.264172 * 2.47105. But if thats not what was happening pls ignore!

packages/webapp/src/components/Form/Unit/index.jsx
Duncan-Brain: I though it is called useUnit() because that is what the input component is called

packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/index.tsx
```diff
@@ -0,0 +1,238 @@
...
+  },
+  APPLICATION_RATE_WEIGHT: {
+    units: applicationRateWeight,
+  },
+  APPLICATION_RATE_VOLUME: {
+    units: applicationRateVolume,
+  },
+};
+
+export interface Location {
```
Duncan-Brain, line 54: For some reason this feels like it belongs in a new types file in `/AreaDetails`

packages/webapp/src/util/convert-units/applicationRates.ts
kathyavini: Ah I see, I see 🤔 Yes I wasn't consistent with that.

I think I like how it's done in the weight anchors (system conversion) below, then. And I'll use the more intuitive litre to gallon ratio so that would be

`( 1 /  3.78541 ) * 2.47105`

on this line then

Gosh it's stressing me out how the denominator isn't flipping though, I need to check that, although calculation wise this is what worked out 😅

packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/index.tsx
```diff
@@ -0,0 +1,238 @@
...
+        <b>{label}</b>
+      </span>
+    ) : (
+      <span className={styles.switchLabel}>{label}</span>
+    );
+  };
+
+  return (
+    <div className={styles.container}>
+      <Switch
```
Duncan-Brain, line 125: I think this is missing the disabled / readonly prop. I was expecting it to be disabled when the other inputs are as well.

packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/index.tsx
```diff
@@ -0,0 +1,238 @@
...
+            hookFormGetValue={getValues}
+            hookFromWatch={watch}
+            control={control}
+            mode={'onChange'}
+            onChange={onQuantityChange}
+            disabled={isReadOnly}
+            required
+            key={isWeight ? 'weight' : 'volume'}
+          />
+          <AreaApplicationSummary
```
Duncan-Brain, line 153: I don't want to be a bummer haha, just wondering if it is going to be reused somewhere or if maybe it is small enough to just put in here.

packages/webapp/src/util/convert-units/applicationRates.ts
kathyavini: Ok, done! Made both changes -- kept the common constant between the imperial to metric and metric to imperial, and used ratio multiplication instead of division. Can you check it @Duncan-Brain?

packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/useQuantityApplicationRate.ts
```diff
@@ -0,0 +1,185 @@
...
+    const weight_unit = getValues(TASK_PRODUCT_FIELD_NAMES.WEIGHT_UNIT);
+    const application_area = getValues(TASK_PRODUCT_FIELD_NAMES.TOTAL_AREA_AMENDED);
+
+    const volume = getValues(TASK_PRODUCT_FIELD_NAMES.VOLUME);
+    const volume_unit = getValues(TASK_PRODUCT_FIELD_NAMES.VOLUME_UNIT);
+
+    if ((weight || weight === 0) && isWeight && weight_unit && application_area) {
+      setValue(
+        TASK_PRODUCT_FIELD_NAMES.APPLICATION_RATE_WEIGHT,
+        convert(weight / application_area)
```
Duncan-Brain, line 97: I think its really cool that you extended the convert units ! Its really nice to have an example how to do it as I haven't seen that done before. I am also very curious if you tried the following easy way too and if it was different:
```
 (convert(volume).from('l').to('ml') /  convert(area).from('m2').to('ha')) 
```

packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/index.tsx
kathyavini: Oh yeah, it definitely should 🤦‍♀️

packages/webapp/src/stories/Pages/Task/AddSoilAmendmentProducts/AddSoilAmendmentProducts.stories.tsx
```diff
@@ -87,6 +92,11 @@ export const ReadOnly: Story = {
   args: {
     isReadOnly: true,
     farm: { interested: true, country_id: 37, farm_id: 'xxx' },
-    defaultValues: [{ ...defaultValues, product_id: 3 }],
+    defaultValues: [
```
Duncan-Brain, line 95: Having values for the quantity might be good for the read-only specifically.

packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/index.tsx
kathyavini: Do you mean as a sub-component in the file or just insert the JSX right in there? I'm okay with either; I write it that way initially and usually extract it only for readability. If it's not more readable I'll stick it back in!

Duncan-Brain: no action needed -- but it could probably be jammed right in there if you agree/ want to

Duncan-Brain (changes requested):
Im done now!

Only thing I think was really needed was the readonly/disabled for the switch. 

Also noticed the classic uncontrolled component error in storybook when entering first value with dropdown closes. Don't know if its an artefact of being in storybook or not:
Screenshot 2024-07-09 at 2 57 55 PM

packages/webapp/src/util/convert-units/applicationRates.ts
Duncan-Brain: Whatever is clear to you, I am not exactly sure why it stood out to me. 

Thinking about it more than I did the first time, I think for precision less math operations is better. For the factor itself I think what I saw was that it was showcasing a well known conversion factor. So I might expect: (wellKnownImperialToMetricConversionFactorForX) * (1/wellKnownImperialToMetricConversionFactorForY) OR 
(wellKnownMetricToImperialConversionFactorForX) * (1/wellKnownMetricToImperialConversionFactorForY)

If that exists or make sense but its good as is!

packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/index.tsx
kathyavini: How about a new `types.ts` file that lives next to the slices (since this is a return from a selector and not specific to any one part of the app)? If so it would be a very partial type -- with a comment to that effect -- because the actual type of Location is a crazy union of location table + area/point + barn/garden/field etc and I don't even want to start to type that (and under Anto we agreed to not write types without immediate functional use).

Alternatively, I think it could also be kept here but renamed... `ApplicationRateLocation` maybe? To indicate it's only the properties necessary in `<QuantityApplicationRate />`?

What do you think?

SayakaOno (commented):
Sorry it took so long! Non-devs have no idea how much code the addition of this section requires; it was more complicated than I expected too 😂 Great job!

I left some comments on the things I noticed :)

packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/index.tsx
Duncan-Brain: No action needed, but if you think a types file makes sense somewhere else that it might be reused then I am all for that.

packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/useQuantityApplicationRate.ts
```diff
@@ -0,0 +1,185 @@
...
+
+  /* Set initial application area to area total area */
+  useEffect(() => {
+    if (!application_area) {
+      setValue(TASK_PRODUCT_FIELD_NAMES.TOTAL_AREA_AMENDED, total_area);
+    }
+
+    /* update unit of application area and string */
+    onPercentLocationChange();
+  }, []);
```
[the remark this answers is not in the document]
kathyavini: So much better! 😍 I was really led astray by copying the water use calculator.

I will set the default value on the `useForm()` level to parallel the percent location (with the fixed `<NumberInput />` that allows this); but please let me know if that doesn't work with `useFormContext()` and if the prop to `<Unit />` is necessary.

packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/useQuantityApplicationRate.ts
Duncan-Brain: Thanks for the good points about this previously discussed way not workings --- it would not be easier to handle combined units end edit Unit function. This sets us up well with a nice template for future units with even more constituent parts too.

packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/useQuantityApplicationRate.ts
```diff
@@ -0,0 +1,185 @@
...
+    onQuantityChange();
+  };
+
+  const onQuantityChange = () => {
+    const weight = getValues(TASK_PRODUCT_FIELD_NAMES.WEIGHT);
+    const weight_unit = getValues(TASK_PRODUCT_FIELD_NAMES.WEIGHT_UNIT);
+    const application_area = getValues(TASK_PRODUCT_FIELD_NAMES.TOTAL_AREA_AMENDED);
+
+    const volume = getValues(TASK_PRODUCT_FIELD_NAMES.VOLUME);
+    const volume_unit = getValues(TASK_PRODUCT_FIELD_NAMES.VOLUME_UNIT);
```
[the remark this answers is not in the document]
kathyavini: Thank you!

kathyavini:
Thank you both so much for your suggestions + discussion -- I think this is ready for re-review except for one thing noted below 🙏 

@Duncan-Brain  - sadly I couldn't get using the `onChangeUnitOption` handler to work! As far as I can understand what the issue is, it looks like that callback is run before the unit's database value is converted (I suspect there is some render cycle effect at play?), and the conversion math requires that the database value already be updated. I think there might be a workaround, but it will require a lot more code because the `onChange` handlers will need to start tracking and using the returned values and checking for them in the calculations; I think `useEffect()` is a lot cleaner in this case and will be the precedent we want.

@SayakaOno - thank you for all the suggestions, it cleaned up the hook file a _lot_!! When the `useForm()` is converted to a `useFormContext()` the Storybook story will break (there is no `<FormContextProvider />` in it), so I would like to make those changes now, but first I wanted to make sure it wouldn't mess with any code you might already have underway.

SayakaOno:
@kathyavini Thank you for asking! I haven't worked with your code yet, so please go ahead! 🙏

packages/webapp/src/components/Form/Unit/index.jsx
```diff
@@ -58,6 +58,7 @@ const Unit = ({
   onBlur,
   hasLeaf,
   autoConversion,
+  onChange: additionalOnChange,
```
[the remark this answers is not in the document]
kathyavini: What changes were you looking for in the JSDoc? There haven't been been any changes to the returned object and it doesn't appear that anything is documented about the incoming parameters, other than 

` * @param {object} props - hook properties`

so I updated the `propTypes` (a few were missing) but I'm not sure if that's what you were thinking.

packages/webapp/src/components/Form/Unit/index.jsx
```diff
@@ -58,7 +58,7 @@ const Unit = ({
   onBlur,
   hasLeaf,
   autoConversion,
-  onChange: additionalOnChange,
+  onChange: onInputChange,
```
SayakaOno, line 61: Sorry 😭 I meant renaming `onChange` itself to differentiate it from `onChangeUnitOption`.

packages/webapp/src/components/Form/Unit/index.jsx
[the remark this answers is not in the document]
SayakaOno: Sorry, my bad! I thought the props details had been defined. Thank you for updating propTypes!

packages/webapp/src/components/Form/Unit/index.jsx
kathyavini: Oh haha 😂 It won't have the same convention as the other inputs in that case though; that's okay?

SayakaOno: I guess how I think about `<Unit />` was different - it wasn't a type of input for me 😂 I'm okay as is!

SayakaOno (dismissed):
🎉 🎉 🎉

navDhammu:
@kathyavini Sorry for just jumping in here but I was randomly looking at this component and wanted to point out something I noticed. This isn't really a bug but more a UX thing so might not be a big deal, but I noticed that in the `<NumberInput />` you can type in a value above 100% and the rest of the calculations change with it even if the entered percent value is above 100. Like "Applied to 250% of your 15200 m² field" doesn't make sense. It does go back to 100 on blur but I feel this can be improved by using 100 as the max percent value in calculations. Something like `Math.min(percentValue, 100)` ?.

kathyavini:
Thanks @navDhammu, that's a good idea! I also like that UX better than applying onBlur.

packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/useQuantityApplicationRate.ts
```diff
@@ -63,7 +63,7 @@ export const useQuantityApplicationRate = ({
 
   /* Update application area + rate based on percent of location */
   const onPercentLocationChange = (percent_of_location: number) => {
-    const calculatedArea = (total_area * (percent_of_location || 100)) / 100;
+    const calculatedArea = (total_area * Math.min(percent_of_location || 100, 100)) / 100;
```
navDhammu, line 66: I'm not sure if this is the intended behavior as I don't have full context of this feature, but If `percent_of_location` is 0 then the || operator here would default it to 100. If the user enters 0 for the percent then is it just disregarded and assumed that 100% of location should be ammended?

kathyavini: Kind of -- it's not allowed to be zero, so if the user enters zero specifically, `onBlur` it will get set to a very small minimum. I don't think we want to have the application area ever calculated as zero is why I'm not too concerned if it is temporarily at 100% before it gets bumped to the minimum.

packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/useQuantityApplicationRate.ts
```diff
@@ -0,0 +1,172 @@
...
+      setValue(
+        TASK_PRODUCT_FIELD_NAMES.VOLUME,
+        convert(application_rate_volume).from('l/ha').to('l/m2') * application_area,
+        { shouldValidate: true },
+      );
+    }
+  };
+
+  /* Trigger recalculation when units are changed */
+  useEffect(() => {
```
navDhammu, line 133: Isn't it possible to do this in the change handler for Unit? 

`<Unit  onChangeUnitOption={() => updateApplicationRate()} />`

If it works then I think its cleaner and potentially more performant than useEffect. Same with quantity update below.

packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/useQuantityApplicationRate.ts
```diff
@@ -0,0 +1,172 @@
...
+        : application_area_unit;
+
+      const value =
+        total_area &&
+        roundToTwoDecimal(
+          convertFn(location_area[system], total_area, location_area.databaseUnit, unit.value),
+        );
+
+      setPreviewStringValue(value);
+      setPreviewStringUnit(unit.label);
```
navDhammu, line 159: Is this useEffect and state for previewStrings necessary? If not then I think it can be just calculated on render and if doing the calculation on every render is expensive then useMemo can be used?

Duncan-Brain (dismissed):
Looks great!

I know this will approval be cleared but there are some conflicts now -- and if you want it still might be nice to have default values in the readonly story for Page/AddSoilAmendmentProducts

packages/webapp/src/stories/Pages/Task/AddSoilAmendmentProducts/AddSoilAmendmentProducts.stories.tsx
kathyavini: @Duncan-Brain I don't believe that this story is set up to take default values for the task product (as opposed to product) fields... but I'm basing that mostly off of the purpose not being populated, for instance.

If I'm not mistaken, I don't think _task_ properties population into the `<ProductCard />` is set up yet, is that correct @SayakaOno? Maybe I'll ping on Slack because I'm a bit confused about the final form setup too for readonly.

packages/webapp/src/components/Form/Unit/index.jsx
kathyavini: I did change it; thank you!

packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/useQuantityApplicationRate.ts
kathyavini: Ummm, it wasn't about expensive calculation, but rather to freeze the value of the preview string unit to the user-selected database value, while the application area unit should adapt dynamically based on the calculation. There might be a better way but I think it's kind of low on the priority list right now as something to optimize.

kathyavini:
@Duncan-Brain and @SayakaOno I think this branch should now work and populate with correct area(s) for add task and complete task views -- and should continue to work 🤞 with the changes you showed today @SayakaOno for complete.

packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/useQuantityApplicationRate.ts
navDhammu: I see what you mean, maybe I'm missing something but I just was curious why not just compute the previewStringValue and unit without using state? For example in the component render, just do 

```
const previewStringValue = total_area &&
   roundToTwoDecimal(
     convertFn(location_area[system], total_area, location_area.databaseUnit, unit.value),
   );
```

and pass that as a prop to` AreaApplicationSummary`

My reasoning for this is that with the useEffect, its triggering an unnecessary extra render. For instance, first render prevString is null, then immediately it renders again from the useEffect with setState, and also the same thing if the useEffect dependencies change.

packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/useQuantityApplicationRate.ts
```diff
@@ -0,0 +1,151 @@
...
+  const updateApplicationRate = (updatedUnit?: UnitOption) => {
+    const weight = getValues(WEIGHT);
+    const volume = getValues(VOLUME);
+    const application_area = getValues(TOTAL_AREA_AMENDED);
+
+    if ((weight || weight === 0) && isWeight && application_area) {
+      setValue(
+        APPLICATION_RATE_WEIGHT,
+        convert(weight / application_area)
+          .from(isWeight && updatedUnit ? updatedUnit.value : 'kg/m2') // database or selected unit weight / database unit area
```
navDhammu, line 75: Sorry for nitpicking 😁 but this can be simplified to `updatedUnit?.value || 'kg/m2'.` The `isWeight `check is already done at the top in the if condition. Same for else if below.

packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/useQuantityApplicationRate.ts
kathyavini: In your code example above, where are you getting the value of `unit` from? Still from watching the value of `application_area_unit`?

packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/useQuantityApplicationRate.ts
kathyavini: Oh thank you, I will change it!

packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/useQuantityApplicationRate.ts
navDhammu: Yes unit is exactly the same as its defined:

 ```
 const unit =
    total_area_unit && location_area[system].units.includes(total_area_unit)
      ? getUnitOptionMap()[total_area_unit]
      : application_area_unit;
```

File: packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/useQuantityApplicationRate.ts (hunk 25 above)
Comment (kathyavini) [reply packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/useQuantityApplicationRate.ts:75]: Hmm, unfortunately I don't think it's going to work. As we are on quite a time crunch I'm going to just revert the handler change, but the changes needed to the calculations to work like this are actually quite a lot.

For instance, the `updateUnit` being passed by the `onChangeUnitOption` handler to `updateApplicationRate` is going to be in e.g. `lbs`, let's say for an imperial farm. But the value being pulled by `getValue()` is in 'kg/m2'.

I don't think there is clean logic for relating a change of `lb` --> `oz` on a metric-based hidden input value without re-writing Unit, which is... doesn't look worth it for avoiding two useEffects. So I'll return to the original method 😓

File: packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/useQuantityApplicationRate.ts (hunk 24 above)
Comment (kathyavini) [reply packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/useQuantityApplicationRate.ts:160]: So I'm not sure I love the idea of the calculations being re-run on every render, but as I said above the string value + unit need to be frozen. If written like this, when the `application_area_unit` changes (e.g. you can test it in the Storybook story for imperial by setting the percentage to a low value), the string will update as well. Ideally I would like to memoize the calculation and _exclude_ `application_area_unit` from the dependencies, but, as you mentioned, it's initially `undefined`.

Maybe a ref instead of component state would have been a cleaner way to make sure the values are only set one time?

**Review by SayakaOno — APPROVED** [review]
Looks great!
Thank you so much for adjusting your code to work with mine. It works perfectly!

File: packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/useQuantityApplicationRate.ts (hunk 24 above)
Comment (navDhammu) [reply packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/useQuantityApplicationRate.ts:160]: I’m not sure about ref but I tried this out on my local and was able to achieve the desired result (let me know if it doesn't work). I put the calculations inside useMemo to avoid running on each render. Now as you mentioned it needs to be frozen and that’s where the issue with using` application_area_unit` as a dependency becomes because when it changes to a different unit the calculations are re-run with the new unit value which changes the UI string. 

So what I did is turn the dependency into a boolean, it only runs when `application_area_unit` changes from falsy to truthy (or vice versa). So it doesn’t run when application unit changes, lets say from ‘ac’ to ‘ft-2’ because both are truthy. 

Here's the code I used: 

```
  const previewStrings = useMemo(() => {
    let previewStringUnit;
    let previewStringValue;

    if (application_area_unit) {
      previewStringUnit =
        total_area_unit && location_area[system].units.includes(total_area_unit)
          ? getUnitOptionMap()[total_area_unit]
          : application_area_unit;

      previewStringValue =
        total_area &&
        roundToTwoDecimal(
          convertFn(
            location_area[system],
            total_area,
            location_area.databaseUnit,
            previewStringUnit.value,
          ),
        );
    }

    return { previewStringUnit, previewStringValue };
  }, [!!application_area_unit]);
```

File: packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/useQuantityApplicationRate.ts (hunk 25 above)
Comment (navDhammu) [reply packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/useQuantityApplicationRate.ts]: I think that's just the side effect of having too much logic inside of useEffects in `Unit`. Actually re-writing Unit might not be a bad idea because if we're going to be using NumberInput going forward (with the localized numbers), then it probably makes sense to have the numbers in `Unit` behave the same way. Otherwise you have a situation with Unit and NumberInput on the same page behaving differently (commas and periods).

File: packages/webapp/src/components/Task/AddSoilAmendmentProducts/ProductCard/index.tsx [hunk 26]
```diff
@@ -182,7 +185,13 @@ const SoilAmendmentProductCard = ({
           />
         </>
       )}
-      {/* TODO: LF-4249 <QuantityApplicationRate /> */}
+      <QuantityApplicationRate
+        productId={PRODUCT_ID}
```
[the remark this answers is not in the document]
kathyavini: I absolutely did! I was only using this to pass to the `<Collapse />` id, and since it was functioning with the fixed string, I think passing product isn't necessary at all. @Duncan-Brain in the other PR I will remove the `productId` prop.

packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/useQuantityApplicationRate.ts
kathyavini: Mmmm I love the boolean value in dependency array! That is exactly what I wanted 😍

We won't be making any further changes in this PR to avoid merge conflicts with #3294 but I would like to adopt it there.

navDhammu: Also just to mention the useEffect solution works fine so its not necessary to change it. I was just pointing out what I know and from reading the react docs, the exact same thing is mentioned in this page: https://react.dev/learn/you-might-not-need-an-effect#updating-state-based-on-props-or-state

Basically whenever something can be computed during a components render, then it probably shouldn't be in state. This is more like best practices and not a hard rule though.

kathyavini:
@Duncan-Brain I think this is good to merge. After cleaning up the diff on #3294 and undrafting I will make these changes on that branch
- remove the `productId` prop
- use @navDhammu's `useMemo()` solution for the previewString

Duncan-Brain (approved):

PR #3290, opened 2024-07-09
LF-4221 (1): Update soil amendment task flows using new add products component

SayakaOno:
**Description**

I will open two PRs for the ticket so that it won't end up a huge PR.
The only things that can be tested in this branch are create/modify product. (create a new soil amendment task -> create or modify product)

- Implement API calls.
- Define `SoilAmendmentProduct` type for the APIs and replace the `Product` type.
- Update `productSlice` to return `soil_amendment_product`.
- Convert `<SoilAmendmentTask />` to TypeScript. (But the content is totally different now; purpose input was moved down to `<AddSoilAmendmentProduct />`, `<AddProduct />` was replaced with `<AddSoilAmendmentProduct />`, and new inputs were added). (It's not longer a pure component, but I'm not sure what to do. Any suggestions?)
- I had more translations 😞 

Fixes:
- Fix defaultValues for the task. (the defaultValues for the product was being misused...)
- Fix property names. (npk_unit -> elemental_unit etc.)
- Work with the right structure of product. (I wasn't considering that `soil_amendment_product` is nested within product)
- Fix the callback function in `onSubmit` to behave correctly when the API fails.
- Fix duplicated types `onSave` and `onSaveProduct` for an identical prop.
- Correct wrong console.logs

TODO:
- Adjust the structure not to nest `soil_amendment_task_products` within `soil_amendment_task`.
- Create/Complete/Read-only for task itself

Jira link: https://lite-farm.atlassian.net/browse/LF-4221

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


SayakaOno:
I added `{ shouldValidate: true }` to the `setValue` for `product_id` to enable the "Add another product" button. Thank you Joyce!

kathyavini (commented):
You're fixing the bugs faster than I can document them 😂 😂

So it sounds like in grooming it was decided (please correct if I'm missing something)
- Purpose should go at the end of the product card (above quantity application rate) instead of at the top
- Some new copy (TBD by David) needed in between method and the product cards along the lines of "Tell us what you'll be adding..."
- In-line validation needed to prevent duplicate product names
- Mixed-neutral support of product validity (and not task form validity) triggering + Add a new product
- And null NPK as `--` in the React Select? Or was that decided to be pushed? 😅 

Anyway this is wonderful and it's amazing it's already completed far enough to demo and share with the soil team tomorrow 🙏 The (hybrid) API functionality works great! The comments below are extremely minor.

packages/webapp/src/store/api/apiSlice.ts
```diff
@@ -136,6 +148,34 @@ export const api = createApi({
...
+      providesTags: ['SoilAmendmentFertiliserTypes'],
+    }),
+    addSoilAmendmentProduct: build.mutation<SoilAmendmentProduct, Partial<SoilAmendmentProduct>>({
+      query: (body) => ({
+        url: `${productUrl}`,
+        method: 'POST',
+        body,
+      }),
+    }),
+    UpdateSoilAmendmentProduct: build.mutation<SoilAmendmentProduct, Partial<SoilAmendmentProduct>>(
```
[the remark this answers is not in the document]
SayakaOno: Thank you!!!

SayakaOno:
@kathyavini Thank you for reviewing and the note!
- [x] Purpose should go at the end of the product card (above quantity application rate) instead of at the top
- [x] Some new copy (TBD by David) needed in between method and the product cards along the lines of "Tell us what you'll be adding..."
- [x] In-line validation needed to prevent duplicate product names
- [ ] Mixed-neutral support of product validity (and not task form validity) triggering "+ Add a new product"
       -> If I'm not mistaken, we will need to [partially validate the form](https://github.com/orgs/react-hook-form/discussions/3676). I think we can create an improvement ticket.
- [ ] And null NPK as -- in the React Select? Or was that decided to be pushed? 😅
        -> This will be finalized after the meeting on Jul 12.

kathyavini (approved):
You are _very_ fast! 🙏 

Looks great and all as discussed. I hope they love it at the demo tomorrow 😄🤞

Edit: Oh and thank you for the discussion link. I was assuming that `isValid` would take field names.

packages/webapp/src/components/Task/SoilAmendmentTask/index.tsx
```diff
@@ -0,0 +1,205 @@
...
+import Unit from '../../Form/Unit';
+import AddSoilAmendmentProducts from '../AddSoilAmendmentProducts';
+import { type ProductCardProps } from '../AddSoilAmendmentProducts/ProductCard';
+import { isNewProduct } from '../AddSoilAmendmentProducts/ProductCard/ProductDetails';
+import {
+  enqueueErrorSnackbar,
+  enqueueSuccessSnackbar,
+} from '../../../containers/Snackbar/snackbarSlice';
+import type { ProductFormFields, ProductId } from '../AddSoilAmendmentProducts/types';
+import { MolecularCompound, Nutrients, TASK_FIELD_NAMES } from '../AddSoilAmendmentProducts/types';
```
Duncan-Brain, line 37: Just wondering if you like this import on two lines instead of one.

packages/webapp/src/containers/Task/saga.js
```diff
@@ -101,7 +101,7 @@ export function* getProductsSaga() {
     yield put(getProductsSuccess(result.data));
   } catch (e) {
     yield put(onLoadingProductFail());
-    console.log('failed to fetch field crops by date');
+    console.log('failed to fetch products');
```
Duncan-Brain, line 104: ❤️

packages/webapp/src/containers/saga.js
```diff
@@ -577,7 +577,7 @@ export function* logUserInfoSaga() {
       yield call(axios.post, logUserInfoUrl(), data, header);
     }
   } catch (e) {
-    console.log('failed to fetch field crops by date');
+    console.log('failed to log user info');
```
Duncan-Brain, line 580: ❤️  😂

packages/webapp/src/store/api/types/index.ts
```diff
@@ -115,3 +117,64 @@ export type AnimalRemovalReason = {
...
+  id: number;
+  key: string;
+}
+
+interface Product {
+  product_id: number | string;
+  name: string;
+  product_translation_key?: string;
+  supplier?: string;
+  type?:
```
Duncan-Brain, line 141: I think this could be required after LF4260b - we checked prod database and do not have a definition for a product without type

Duncan-Brain (approved):
Everything is looking amazing so far -- this is big 👏 . I hope to have more comments in Part 2. And I guess there is some rush to have this on beta today.

Only thing I was wondering... and it is probably a question for Loic when he get back. And something I could have noticed on the PureComponent. For the mobile view only it might be too much to have name and NPK all in select option. And there should probably be some padding between the end of product name and "N".

Screenshot 2024-07-12 at 10 19 23 AM

SayakaOno:
@Duncan-Brain Thank you for reviewing and testing on the mobile view!
I think we should at least have the padding between. I've created an enhancement [ticket](https://lite-farm.atlassian.net/browse/LF-4292)!

SayakaOno (commented):
I was going to comment here... I just noticed that I hadn't submitted!

packages/webapp/src/store/api/types/index.ts
```diff
@@ -115,3 +117,64 @@ export type AnimalRemovalReason = {
...
+    | typeof TASK_TYPES.SOIL_AMENDMENT
+    | typeof TASK_TYPES.CLEANING
+    | typeof TASK_TYPES.PEST_CONTROL;
+  farm_id?: string;
```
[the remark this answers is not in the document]
Duncan-Brain: Right, there must be some type pattern to represent post vs patch. 

Don't worry about it if this works -- it is fine for me.

PR #3294, opened 2024-07-15
LF-4221 (2): Update soil amendment task flows using new add products component

SayakaOno:
**Description**

I merged Joyce's LF-4249 branch.

- Remove `checkParamAgainstBody` middleware from `/complete/soil_amendment_task/:task_id`. It didn't allow me to complete a task without modification, so I removed it for now. I will create a ticket for us to work on it later!
- Fix data structure not to nest `soil_amendment_task_products` within `soil_amendment_task`.
- Create functions to format DB task products data to form version and vice versa.
- Format defaultValues for read-only views.
- Include task products data when completing the task.
- Create SoilAmendmentTaskProducts slice.
- Create `handleGetTasksSuccess` saga to replace the duplicated code in `getTasksSuccessSaga` and `getAllTasksSuccessSaga`, and include some code to handle task products.
- Modify sagas for get/complete/post tasks.
- Update `taskEntitiesSelector` to include task products.
- Add correct defaultValues to the [story](http://localhost:6006/?path=/story/page-task-addsoilamendmentproducts--read-only). 
- Change `isWeight` to a form field. (Joyce, I understood the options you mentioned while working on this ticket. I realized that moving up the state wouldn't work, so I went with the form field. Sorry 🙏 )
- Hide remove button in read-only view. 

NOTE:
I get these warnings when loading `/tasks`, but I couldn't look into.
Screenshot 2024-07-15 at 2 29 17 AM

I think I was able to test most cases, but I don't think I exhausted all possibilities. Additionally, I couldn't do a thorough self-review, so there might be some mistakes. Sorry about that!
Since I merged Joyce's PR, I will keep this PR as a draft. Please feel free to change the base branch for a review or make any necessary commits.

Jira link: https://lite-farm.atlassian.net/browse/LF-4221

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/api/src/routes/taskRoute.js
```diff
@@ -25,7 +25,6 @@ import {
 } from '../middleware/validation/assignTask.js';
 import taskController from '../controllers/taskController.js';
 import { createOrPatchProduct } from '../middleware/validation/product.js';
-import checkParamAgainstBody from '../middleware/acl/checkParamAgainstBody.js';
```
Duncan-Brain, line 28: Will need to add this back in I think.

packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/index.tsx
```diff
@@ -145,7 +139,7 @@ const QuantityApplicationRate = ({
         leftLabel={formatLabel('ADD_TASK.SOIL_AMENDMENT_VIEW.WEIGHT', isWeight)}
         label={formatLabel('ADD_TASK.SOIL_AMENDMENT_VIEW.VOLUME', !isWeight)}
         isToggleVariant
-        onChange={toggleMeasure}
+        onChange={(e) => setValue(IS_WEIGHT, e.target.checked ? false : true)}
```
Duncan-Brain, line 142: I think the previous way of doing it was correct: `setIsWeight((prev) => !prev);`

packages/webapp/src/containers/Task/constants.js
```diff
@@ -25,3 +25,7 @@ export const TASK_TYPES = {
   HARVEST: 'harvest_tasks',
   IRRIGATION: 'irrigation_task',
 };
+
+export const TASKTYPE_PRODUCT_MAP = {
```
Duncan-Brain, line 29: Optional improvement to make this a function. All future task products should be `${taskType}_products`.

packages/webapp/src/containers/Task/saga.js
```diff
@@ -339,24 +353,44 @@ export function* getTasksSuccessSaga({ payload: tasks }) {
...
       yield put(taskTypeActionMap[task_translation_key].fail(e));
       console.log(e);
     }
   }
-  yield put(addManyTasksFromGetReq(tasks));
+
+  yield put(successAction(tasks));
+}
+
+export function* getTasksSuccessSaga({ payload: tasks }) {
```
Duncan-Brain, line 392: Do we need to create a new function if we still have` getAllTasksSuccessSaga()` with the same definition.

packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/index.tsx
kathyavini: It isn't component state anymore so it doesn't have that kind of setter (just the React Hook Form `setValue()` setter).

Duncan-Brain (commented):
I think I am seeing that:
- furrow depth and other SAT properties need to be deleted if the method changes, they seem to show up in db after not selected.
- select task location is allowing any area to be selected, is this how it was previously? or should it only be areas with soil? (ie not gates and water sources)
- ~~product form validates when the total are field is broken due to not updating the area on changed locations~~ (edit: no longer seeing validation problem for add new product link)
- ~~The edit page has all the react form issues:~~ (edit: not seeing this anymore either) 

Screenshot 2024-07-16 at 2 10 40 PM

Screenshot 2024-07-16 at 2 10 34 PM

Screenshot 2024-07-16 at 2 10 44 PM


Something to keep an eye out for: 
- I got into a weird state, when create task would not save at the end of the form I tried saving by button mashing -- the db log showed 1 user having 5 subscriptions and then save stopped working even when I reset/cancelled the form and restarted db and frontend -- hopefully that was a fluke

Duncan-Brain:
For the selectId console warning: 

What I found is that these are tasks which do not have an associated types task details. For example: 
```
task: {
  task_id: 4,
  task_type_id: 5 // pretend this is cleaning_task  
 }
``` 

There is no 
```
//Does not exist
cleaning_task: {
  task_id: 4
  }
 ``` 
 So when it is not returned line 361 of `task/saga.js` returns an empty object and that is passed to the selector for which the object does not have an id. 
 
Up to you if we should add a check here in the saga to not return an empty object.. but it may cause other issues. I will make a data debt ticket for a migration to add in missing task specific details.

**Review by kathyavini — COMMENTED** [review]
Wow, the form <> database manipulations are really complex, but the organization is very very clear so I think the few remaining bugs will be super easy to troubleshoot! 🙂

Rather than commits to this branch, I'll open a PR against it to resolve the location area persistence bug and the remaining comments from the `<QuantityApplicationRate />` which I mentioned on that PR, and also a small edit needed to `isWeight` now that it's form state.

One more major thing I also want to demo/discuss (but I'll put it on a different branch to make the diff + commits as clear as possible) is the option of not using a new soilAmendmentTaskProduct slice.

I'm of mixed minds about this because the code added to the sagas is really well-structured and the new entity is managed well, but neither the manual removal/sync nor the nesting in the products slice is necessary because the controller graph is already handling both of those things. Right now, that means we will end up doing the same data manipulation twice -- once on backend and once on frontend, and although the new code is really good and extensible, I'm not sure we want to add that level of boilerplate unless we know we really need it in the future.

Personally I don't feel soil amendment task product (and so by extension the other task products we'll likely someday add) is a true entity needing a slice, in parallel to how it doesn't have its own backend route. Unless I missing a future scenario, I don't think we will ever fetch or select on those entities in isolation from their task. In any case, I really do want to avoid the code that inserts them right back into the graph in the same place they came from because that particular immer function it's a real doozy and a known point of failure in the taskSlice!

I'll open the comparison PR to show what I mean, and maybe we can discuss in tech daily tomorrow?

**Review by kathyavini — APPROVED** [review]
It's good and stable to put on Beta for @dtrapplitefarm; corrections can be made in subsequent PRs

**Review by Duncan-Brain — APPROVED** [review]
WOO! Great work

**Review by SayakaOno — COMMENTED** [review]
Thank you for reviewing!!!

(I wonder if I forgot to submit the first three comments 2 days ago...)

File: packages/webapp/src/containers/Task/saga.js (hunk 4 above)
Comment (Duncan-Brain) [reply packages/webapp/src/containers/Task/saga.js:392]: I think I misread the diff that `getTasksSuccessSaga()` was a new function... I see the removal above now!

File: packages/webapp/src/util/task.ts [hunk 5]
```diff
@@ -0,0 +1,203 @@
...
+          formattedTaskProduct.other_purpose = other_purpose;
+        }
+        formattedTaskProduct.purposes.push(purpose_id);
+      });
+
+      const unitOptions: { [key: string]: UnitOption } = getUnitOptionMap();
+
+      return {
+        ...formattedTaskProduct,
+        weight_unit: isWeight && rest.weight_unit ? unitOptions[rest.weight_unit] : undefined,
```
kathyavini, line 100: @SayakaOno for the systems-switching bug I was thinking of passing `system` here into this function and doing the 

```
soilAmounts[system].units.includes(weight_unit)
```
Do you have a preference about whether it's done here or in the component?

SayakaOno: System switching has been handled in the useUnit, so we shouldn't need to do it anywhere...
What was the issue?

kathyavini: Ah, you know what the cause was? It's a bit of a fun/subtle one 🙂 

Unit does handle the system switching, but it needs to receive a string here, not a `UnitOption`. If the line above is changed to

`weight_unit: isWeight ? rest.weight_unit : undefined,`

then the system conversion works as expected without anything further. I'll update on my branch.

SayakaOno: Ah sorry 😓 Thank you!!!

kathyavini: I don't know why why this should be. It looks in the `useUnit` code like both `UnitOption` and string should behave the same. But the system switching wants a string, and, to make it worse, `setValue()` only works with a `UnitOption`! 😂 What an unintuitive situation!

PR #3298, opened 2024-07-16
LF-4204: Add "Previous" button to <FormNavigationButtons> component

Joaquin-M2:
**Description**

Update the <FormNavigationButtons> component to have three buttons by adding the "Previous" button:

image

Jira link: https://lite-farm.atlassian.net/browse/LF-4204

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


SayakaOno (changes requested):
Thank you for working on this, and I'm sorry for the delayed review! It would be great if you could adjust the styles as well and add `onPrevious` to the stories. (`/packages/webapp/src/stories/FormNavigationButtons/FormNavigationButtons.stories.tsx`)
Screenshot 2024-07-29 at 4 00 49 PM

(Storybook starts by running `pnpm storybook` in `./webapp`)

packages/webapp/src/components/Form/FormNavigationButtons/index.tsx
```diff
@@ -46,6 +50,18 @@ const FormNavigationButtons = ({
         <Button color="secondary-cta" onClick={onCancel} className={styles.button} sm fullLength>
           {t('common:CANCEL')}
         </Button>
+        {hasPreviousButton && (
```
SayakaOno, line 53: I think you can check if `onPrevious` exists like `contextualContent` is checked (line 48).

SayakaOno (commented):
The updated style look good! I just have one more suggestion!
Also, you can leave `MISSING` next time :)

packages/webapp/src/components/Form/FormNavigationButtons/index.tsx
Joaquin-M2: Hi @SayakaOno , thank you for your suggestions :)

Using `onPrevious` to conditionally render the **Previous** button is a bit trickier than what it seems, since `An expression of type 'void' cannot be tested for truthiness` (a warning thrown by TypeScript). So two options come to my mind:

1. We make the type of `onPrevious` be `() => true` instead of `() => void`, forcing to `return true` any time we want to use the onPrevious prop;
2. We leave the `hasPreviousButton` prop.

Both options are a bit cumbersome and I don't like them much, but I didn't find a better solution. What do you think?

SayakaOno: Did you call `onPrevious()` to use it as the condition?
Without the `()`, TypeScript shouldn't complain!

SayakaOno:
Sorry I missed one thing! Could you please remove `disabled={isDisabled}` for "Previous" so that it will always be enabled?
Screenshot 2024-07-31 at 2 00 30 PM

Mockup: 
Screenshot 2024-07-31 at 2 04 11 PM

packages/webapp/src/components/Form/FormNavigationButtons/index.tsx
Joaquin-M2: Storybook tricked me on this one. It is as you say: it works as it should without the `()`. What confused me was that every story now has the **Previous** button even though there is only one story (**WithPreviousButton**) that uses the `onPrevious` prop, so I thought that a void function was not valid for this use case. Nevertheless I tested the component on the app itself and the component works as it should.

I used Storybook in the past even though I'm not very familiarized with it, so if you have any suggestion to make the **Previous** button appear only on the **WithPreviousButton** story, please let me know.

Duncan-Brain: I am surprised we haven't run into that before! 

https://github.com/storybookjs/storybook/issues/22930

I think you could consider adding `onPrevious: undefined` to the variants . The overwriting args seems to be working correctly at least.

Joaquin-M2: It seems Storybook works fine only if it feels like it. I was working on a different ticket for which I also needed to run Storybook, then I changed to the local branch I'm using for this issue and stories were running as they should:

https://github.com/user-attachments/assets/9e56b37f-cc2f-45b3-9b6f-52a48fee2708

Then just out of curiosity I restarted Storybook and it started to show the Previous button on every story again 🤷 :

https://github.com/user-attachments/assets/07188110-2ae5-4853-89e3-132dbb292f98

I just added `onPrevious: undefined` to the variants as you proposed and now stories work fine too 👍

Duncan-Brain (approved):

SayakaOno (approved):
Thank you! 🙏

PR #3299, opened 2024-07-17
LF-4221 and LF-4249 Adjustments

kathyavini:
**Description**

These are adjustments pertaining to the QuantityApplicationRate PR #3289 and the full form integration PR #3294

Summary of changes:
1. set `is_weight` using the `defaultValues` structure appended to new field array entries
2. pass the same `defaultValues` to useForm to populate the first field array entry (thank you @SayakaOno)
3. fix total_area persistence (non-update on location selection changes) by removing the defaultValue prop on `<Unit />` and restoring the useEffect in the QuantityApplicationRate hook
4. Fix percent location non-persistence (incorrect resetting to 100 on location selection changes) by fixing `<NumberInput />` initialValue logic
5. Pass current percent location location to useEffect calculation that should run on location changes
6. Define form `<Unit />` values with strings instead of with `UnitOption` objects; this restores the correct between-system transformation when the farm measurement system does not match the database saved unit value
7. (from LF-4249 PR comments) remove non-functional `productId` prop
8. (from LF-4249 PR comments) remove `useState()` and implement `useMemo()` to calculate and freeze the summary string content

Jira link: none

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/index.tsx
+++ b/packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/index.tsx
@@ -13,7 +13,7 @@
  *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
  */
 
-import { useEffect, useState } from 'react';
+import { useState } from 'react';
 import { useFormContext } from 'react-hook-form';
 import clsx from 'clsx';
 import { useTranslation } from 'react-i18next';
@@ -86,11 +86,6 @@ const QuantityApplicationRate = ({
   const APPLICATION_RATE_WEIGHT_UNIT = `${namePrefix}.${TASK_PRODUCT_FIELD_NAMES.APPLICATION_RATE_WEIGHT_UNIT}`;
   const APPLICATION_RATE_VOLUME_UNIT = `${namePrefix}.${TASK_PRODUCT_FIELD_NAMES.APPLICATION_RATE_VOLUME_UNIT}`;
   const IS_WEIGHT = `${namePrefix}.${TASK_PRODUCT_FIELD_NAMES.IS_WEIGHT}`;
-
-  // New task products should default to weight
-  useEffect(() => {
-    setValue(IS_WEIGHT, getValues(IS_WEIGHT) ?? true);
-  }, []);
 
   const isWeight = getValues(IS_WEIGHT);
 
--- a/packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/index.tsx
+++ b/packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/index.tsx
@@ -72,7 +72,7 @@ const QuantityApplicationRate = ({
 }: QuantityApplicationRateProps) => {
   const { t } = useTranslation();
 
-  const { control, getValues, setValue, register, watch, formState } = useFormContext();
+  const { control, getValues, setValue, register, watch } = useFormContext();
 
   const WEIGHT = `${namePrefix}.${TASK_PRODUCT_FIELD_NAMES.WEIGHT}`;
   const VOLUME = `${namePrefix}.${TASK_PRODUCT_FIELD_NAMES.VOLUME}`;
@@ -190,7 +190,7 @@ const QuantityApplicationRate = ({
                   rules={{ required: t('common:REQUIRED') }}
                   onChange={updateTotalArea}
                   disabled={isReadOnly}
-                  defaultValue={formState.defaultValues?.[PERCENT_OF_LOCATION_AMENDED] || 100}
+                  defaultValue={100}
                 />
                 <SwapIcon />
                 {/* @ts-ignore */}
--- a/packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/index.tsx
+++ b/packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/index.tsx
@@ -104,7 +104,7 @@ const QuantityApplicationRate = ({
   const {
     updateApplicationRate,
     updateQuantity,
-    onPercentLocationChange,
+    updateTotalArea,
     previewStringValue,
     previewStringUnit,
   } = useQuantityApplicationRate({
@@ -188,7 +188,7 @@ const QuantityApplicationRate = ({
                   min={0.0001}
                   max={100}
                   rules={{ required: t('common:REQUIRED') }}
-                  onChange={onPercentLocationChange}
+                  onChange={updateTotalArea}
                   disabled={isReadOnly}
                   defaultValue={formState.defaultValues?.[PERCENT_OF_LOCATION_AMENDED] || 100}
                 />
@@ -206,7 +206,6 @@ const QuantityApplicationRate = ({
                   hookFromWatch={watch}
                   control={control}
                   mode={'onChange'}
-                  defaultValue={formState.defaultValues?.[TOTAL_AREA_AMENDED] || total_area}
                   disabled
                   required
                 />
--- a/packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/index.tsx
+++ b/packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/index.tsx
@@ -13,7 +13,7 @@
  *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
  */
 
-import { useState } from 'react';
+import { useEffect, useState } from 'react';
 import { useFormContext } from 'react-hook-form';
 import clsx from 'clsx';
 import { useTranslation } from 'react-i18next';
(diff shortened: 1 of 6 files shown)


packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/useQuantityApplicationRate.ts
```diff
@@ -63,8 +63,12 @@ export const useQuantityApplicationRate = ({
     APPLICATION_RATE_VOLUME_UNIT,
   ]);
 
+  useEffect(() => {
+    updateTotalArea(100);
+  }, []);
```
kathyavini, line 68: Sorry @SayakaOno, it did end up being necessary! Even (a bit to my surprise) the Unit setting logic in the called function below.

Passing to `defaultValues` caused a persistence bug when the selected locations changed. I did rename the function as it is no longer just a handler.

SayakaOno: Could you provide the steps to reproduce the bug?

kathyavini: > Could you provide the steps to reproduce the bug?

Create a soil_amendment task on a farm with multiple locations > go back and change the selected areas (add or remove one) > preview string value will update but the value in the `TOTAL_AREA_AMENDED` `<Unit />` will not update

packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/index.tsx
```diff
@@ -89,6 +87,11 @@ const QuantityApplicationRate = ({
   const APPLICATION_RATE_VOLUME_UNIT = `${namePrefix}.${TASK_PRODUCT_FIELD_NAMES.APPLICATION_RATE_VOLUME_UNIT}`;
   const IS_WEIGHT = `${namePrefix}.${TASK_PRODUCT_FIELD_NAMES.IS_WEIGHT}`;
 
+  // New task products should default to weight
+  useEffect(() => {
+    setValue(IS_WEIGHT, getValues(IS_WEIGHT) ?? true);
+  }, []);
```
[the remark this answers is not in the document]
kathyavini: Oh perfect, I was looking at the (lack of) default values passed to the overall form and didn't see the fieldArray structure. I might prefer to use for `total_area` as well 🤔

packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/useQuantityApplicationRate.ts
SayakaOno: Ohhhh 😳 I hadn't considered those scenarios at all, thank you!
And it makes sense that we need the useEffect. (although I'm not sure about the unit not being set correctly)
I think the argument should be dynamic! (It always becomes 100 regardless of the previous %)

packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/index.tsx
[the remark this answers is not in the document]
kathyavini: @SayakaOno actually it doesn't seem to be working; do you know why? Using the appended `defaultValues` I still see  `isWeight` as undefined in the component. Same for the others defined in that structure (e.g. purposes is undefined not `[]`)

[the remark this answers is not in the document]
SayakaOno: Oh! I don't think I tested the first product, let me check...

packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/useQuantityApplicationRate.ts
kathyavini: > I think the argument should be dynamic! (It always becomes 100 regardless of the previous %)

You mean we could store the previous percent value? I think here we would then have to remove the defaultValue prop on the `<NumberInput /`> 🤔 It could be done with the field array default value worked for the first product as well!

> (although I'm not sure about the unit not being set correctly)

It's weird, right? I originally wrote the useEffect as simply

```ts
  useEffect(() => {
    setValue(TOTAL_AREA_AMENDED, total_area);
  }, []);
  ```
  
but you can see (e.g. in the Imperial Storybook story, or in app depending on your area settings) that it is not picking the right unit for the value!

File: packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/index.tsx (hunk 2 above)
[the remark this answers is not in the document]
Comment (SayakaOno) [reply packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/index.tsx]: Sorry, the `defaultValues` are only used for `append()`, which applies to the second product and onwards. I opened a PR to use it for the first product too!
https://github.com/LiteFarmOrg/LiteFarm/pull/3303

File: packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/useQuantityApplicationRate.ts (hunk 1 above)
Comment (SayakaOno) [reply packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/useQuantityApplicationRate.ts:68]: I think I wasn't clear about the first point, so let me share the recording.

https://github.com/user-attachments/assets/8e57ffed-78a6-455b-be54-22c8b5cb4009

Comment (kathyavini) [reply packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/useQuantityApplicationRate.ts:68]: Ah thank you! I knew it was resetting to 100 but I didn't realize it diverged from the string.

File: packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/useQuantityApplicationRate.ts [hunk 3]
```diff
@@ -139,33 +144,37 @@ export const useQuantityApplicationRate = ({
   }, [application_rate_weight_unit, application_rate_volume_unit]);
 
   /* For the preview string, replicate the conversion the unit component would do if the value had been displayed in a <Unit /> rather than a string. However, don't update upon changes to application_area_unit beyond the initial undefined > defined change */
-  const [previewStringValue, setPreviewStringValue] = useState<number | null>(null);
-  const [previewStringUnit, setPreviewStringUnit] = useState<string | null>(null);
+  const previewStrings = useMemo(() => {
```
Duncan-Brain, line 147: Just a question about how you are using this to "freeze" the value. I don't think this needs to be optimized.

React docs:

> You should only rely on useMemo as a performance optimization. If your code doesn’t work without it, find the underlying problem and fix it first. Then you may add useMemo to improve performance.

kathyavini: Wait, I also found a new issue with this now that the percent value is retained even when switching locations so I'm pretty sure gonna change the code again! 😓 But to clarify are you asking a) how does the useMemo function to freeze the value, or b) why is it being used to freeze the value when it's intended for optimization/

SayakaOno (dismissed):

packages/webapp/src/components/Task/AddSoilAmendmentProducts/QuantityApplicationRate/useQuantityApplicationRate.ts
Duncan-Brain: Mostly b) !

kathyavini:
@SayakaOno sorry to invalidate your review (and @Duncan-Brain maybe this will also clarify your question about the useMemo?)

The preview string had been pulling the `total_application_area_unit` and using the `useMemo()` dependency array to freeze its own unit at that same one -- the correct one for 100% location area regardless of any subsequent changes to percentage that might dip the unit of total application area down to a smaller unit.

However, this relied on changing the location resetting the percentage to 100%, which has been fixed! Now the `application_area_unit` is no longer a good source of the correct unit after location area change:

Screenshot 2024-07-18 at 10 07 11 AM

To be honest, involving the application_area `<Unit />` was really an unnecessary complication when the function to get the correct unit already was available and should have just been used (particularly because system switching was handled manually anyway). I think that unnecessary bit was what made this part of the code so weird and the subject of much discussion and is now removed.

kathyavini:
Oooh thanks @Duncan-Brain for noticing the comment! Fixed. And that ts error persists even when restarting your TS server? is it new to the most recent change or from before? So weird, I don't see it, and nor did the pre-commit tsc complain...

Duncan-Brain (approved):
My ts version in intellisense is 4.9.5. It just wants a defined type. But its fine !

SayakaOno (approved):

PR #3302, opened 2024-07-17
LF-4029 - change name of Organic type selection choice to third party…

Duncan-Brain:
**Description**

Change the certification type selection from 'Organic' to 'Third-party Organic'

Jira link: [LF-4029](https://lite-farm.atlassian.net/browse/LF-4029)

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

[LF-4029]: https://lite-farm.atlassian.net/browse/LF-4029?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/api/db/migration/20240717190534_update_organic_certification_type.js
+++ b/packages/api/db/migration/20240717190534_update_organic_certification_type.js
@@ -1,0 +1,32 @@
+/*
+ *  Copyright (c) 2024 LiteFarm.org
+ *  This file is part of LiteFarm.
+ *
+ *  LiteFarm is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  LiteFarm is distributed in the hope that it will be useful,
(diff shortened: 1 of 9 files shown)


SayakaOno (commented):
Should we keep the old keys so that the old version of the app would still work when the new release is out?

(My understanding of the release strategy @kathyavini implemented might be wrong)

kathyavini:
> Should we keep the old keys so that the old version of the app would still work when the new release is out?

Yeah, it's not related to the release assets strategy, but as long as we've been PWA we've had the possible combo of non-cached translations and cached frontend.

This is a really funny one though because the frontend is just pulling the string from the database it looks like, and of course the API has no caching so it will be non-cached translations + non-cached backend, and the cached frontend will be irrelevant 😂 

I wonder if we might still see something like this  -- the way we do when we change branches locally -- but I don't think preserving the ORGANIC key would fix it, so it is as it is!

Screenshot 2024-07-17 at 4 05 00 PM

kathyavini (approved):

SayakaOno (approved):

PR #3303, opened 2024-07-17
LF-4221 Fix soil_amendment_task_products defaultValues

SayakaOno:
**Description**

Fix for https://github.com/LiteFarmOrg/LiteFarm/pull/3299#discussion_r1681600368
Use the soil_amendment_task_products `defaultValues` (currently used for `append()`) for the first product.


kathyavini (approved):
Okay thank you so much!

It doesn't work with `percent_of_location` (😭) but I'm sure the NumberInput logic is to be blamed

PR #3324, opened 2024-07-22
LF-4309 Complete form breaks when not completed fast enough allowing page to time out or something

kathyavini:
**Description**

This PR removes the passing of the purpose and method keys from the saga to the task utility functions in favour of using `{ shouldUnregister: true }` on the inputs that need to be nulled when their method_id / purpose_id is not selected

Notes:
- Removing the calls to `api.endpoints.getSoilAmendmentPurposes.select()(state),` fixes the timeout that makes the form unsubmittable after the api cache expires
- `<Unit />` component need a little bit of adjustment to accept `shouldUnregister: true`
- A cleanup before insert was on the model was still needed _on PATCH_, but I feel this is an appropriate place for it because
  - Nothing is wrong with the PATCH request, it is correctly only giving the updated data, which means it doesn't overwrite any old data in the other fields (e.g. `other_application_method`, which triggers the check constraint)
  - There is no longer any ambiguity about what data is correct
  - It is logical + readable on the model where it is defined specifically on PATCH (update) 
- `shouldUnregister: true` resolves the other persistence bugs we saw; I will re-test this momentarily then indicate their tickets below as well
  
Jira link: https://lite-farm.atlassian.net/browse/LF-4309

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

Checked on the `soil_amendment_task` product table and the `soil_amendment_task_products_purpose_relationship` table of db.

**Checklist:**

### DIFF
--- a/packages/webapp/src/util/task.ts
+++ b/packages/webapp/src/util/task.ts
@@ -137,8 +137,11 @@ type RemainingFormSATProductKeys = keyof Omit<
 >;
 
 export const formatSoilAmendmentProductToDBStructure = (
-  soilAmendmentTaskProducts: FormSoilAmendmentTaskProduct[],
-): DBSoilAmendmentTaskProduct[] => {
+  soilAmendmentTaskProducts: FormSoilAmendmentTaskProduct[] | undefined,
+): DBSoilAmendmentTaskProduct[] | undefined => {
+  if (!soilAmendmentTaskProducts) {
+    return undefined;
+  }
   return soilAmendmentTaskProducts.map((formTaskProduct) => {
     const { purposes: purposeIds, other_purpose, is_weight, ...rest } = formTaskProduct;
 
--- a/packages/webapp/src/util/task.ts
+++ b/packages/webapp/src/util/task.ts
@@ -13,8 +13,7 @@
  *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
  */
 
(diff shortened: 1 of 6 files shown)


SayakaOno (approved):
Also confirmed that https://lite-farm.atlassian.net/browse/LF-4333, https://lite-farm.atlassian.net/browse/LF-4329 and https://lite-farm.atlassian.net/browse/LF-4324 were fixed.
Thank you!

packages/webapp/src/util/task.ts
```diff
@@ -61,39 +58,29 @@ type FormSoilAmendmentTaskProduct = {
...
-  other_application_method?: string;
-  [key: string]: any;
-};
-
-type DBTask = {
   soil_amendment_task_products: DBSoilAmendmentTaskProduct[];
   [key: string]: any;
 };
 
-type FormTask = {
```
Duncan-Brain, line 82: I don't know if I would have undid the updated naming since `soil_amendment_task` has meaning and does not have property of `soil_amendment_task_products`.

Duncan-Brain (approved):
Always love a net removal of code! ❤️ 

I like what you are saying for PATCH being able to make the decision that a POST could not 👍 . I think we will need to add to this for weight and volume too. I thought I saw a similar bug there.

PR #3326, opened 2024-07-22
LF-4328: The user is not able to complete the soil amendment task after making changes

SayakaOno:
**Description**

In order to override irrelevant fields in the DB, use `null` instead of `undefined`.

Jira link: https://lite-farm.atlassian.net/browse/LF-4328

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/webapp/src/util/task.ts
```diff
@@ -119,14 +118,14 @@ export const formatSoilAmendmentTaskToFormStructure = (task: DBTask | FormTask):
 
       return {
         ...formattedTaskProduct,
-        weight_unit: isWeight ? rest.weight_unit : undefined,
-        volume_unit: !isWeight ? rest.volume_unit : undefined,
+        weight_unit: isWeight ? rest.weight_unit || undefined : undefined,
+        volume_unit: !isWeight ? rest.volume_unit || undefined : undefined,
```
kathyavini, line 122: If the database data is good (nulled properly), actually do we need these checks here? I missed it before, but now I'm wondering can it just be

```suggestion
        weight_unit: rest.weight_unit ?? undefined,
        volume_unit: rest.volume_unit ?? undefined,
```

and same for the other units below... or am I forgetting about a situation?

kathyavini (approved):
Yup, that makes total sense! Thank you for finding and fixing 🙏 

Looking at the `|| undefined : undefined` lines (a tiny bit yuck 😆 Although I understand why they are necessary!) now I'm suddenly thinking that our database data should be good enough -- units never present where they aren't meant to be -- that we can use them straight without reference to the values they are tracking?

Is that not true? It's a bit getting to the end of my thinking workday so I'm not sure 😅 But I think it should be safe!

Duncan-Brain (approved):
Nice!

packages/webapp/src/util/task.ts
SayakaOno: I'm not sure why I did these only for units... I think it should be safe to go with your suggestion. Thank you!

kathyavini (approved):

Duncan-Brain (approved):

PR #3327, opened 2024-07-22
LF-4306 No inline error for character count for describe the purpose

kathyavini:
**Description**

This PR adds/fixes the max character validation and inline validation error for the three text input fields. It does not do the React Select or the `trim()` validation, just the max character on:
1. `other_purpose` (this was the topic of both linked Jira tickets below)
2. `other_application_method` -- no Jira ticket for this one but field has same db datatype and would give same 400 error
3. `supplier` -- no Jira ticket for this one and had a maxLength, so form would be invalid and no data would get to the database, but it didn't use `hookFormMaxCharsValidation` so there was no inline error

I actually thought one of the other fields had a bug table record of its own but maybe we didn't give it blocker priority

Jira links:
https://lite-farm.atlassian.net/browse/LF-4306
https://lite-farm.atlassian.net/browse/LF-4323

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


SayakaOno (approved):
Thank you!

Duncan-Brain:
@kathyavini is this an appropriate place to also fix the model validation similar to how it is on cropModel fro the same properties?

kathyavini:
@Duncan-Brain done! Models added.

packages/api/src/models/productModel.js
```diff
@@ -36,7 +36,7 @@ class ProductModel extends baseModel {
         product_id: { type: 'integer' },
         name: { type: 'string' },
         product_translation_key: { type: 'string' },
-        supplier: { type: 'string' },
+        supplier: { type: 'string', minLength: 1, maxLength: 255 },
```
SayakaOno, line 39: I get `"error":{"message":"supplier: must NOT have fewer than 1 characters"}` for a farm not pursuing a certification!

kathyavini: Oooh thank you for spotting that @SayakaOno!!

I added `null` to the types but noticed that doesn't stop triggering the database error, so removed `minLength` too. I wonder why the other two fields work with just including `null`! 🤔🤔🤔

Duncan-Brain: @kathyavini I think we check and trim string on backend for product so an empty string is maybe supplied. (Or to handle api case). I can't see where the model is checked before `checkProductValidity()` though.

Duncan-Brain (approved):

SayakaOno (approved):

PR #3328, opened 2024-07-23
Patch/3.6.5 add uniqueness check on task product

Duncan-Brain:
**Description**

Sorry this took so long for me to pin down 😓 -- I have added a test for it now that uniqueness db check is in and removed product working for fieldArray. I was wrong that taskController did not need to be updated in previous uniqueness check PR.

Bug for "switching product ids" uniqueness check violation: 
Make soil amendment task with two task products 1 and 2.
Complete task by:
- deleting product 2,
- changing product 1 to product 2
- adding a new product with product 1

Adds:
- api test based on frontend error bug case
- modifier in different place (with old "all the time" modifier the graph does not relate deleted ids, causing "id already exists db error" ) : [objection on modifiers](https://vincit.github.io/objection.js/recipes/modifiers.html#usage-with-eager-loading)
- taskController handling avoidance of uniqueness error.

Jira link:

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


kathyavini (commented):
1. Could you elaborate on setting `deleted: false` for all entries in the update query? I don't disagree it's necessary but I can't figure out why that should be, since the frontend doesn't send it in `deleted: true` form, and there is that new middleware check as well for API requests.

2. Just to make sure I'm following the scope of what's being solved, the issue would not arise, even with the new database constraint, if the graph always deleted before updating (and thus the manual handling in the controller, that seems reasonable) OR if the index constraint were not evaluated until the `upsertGraph` was finished?

packages/api/src/controllers/taskController.js
```diff
@@ -67,6 +68,21 @@ async function updateTaskWithCompletedData(
...
+    const { soil_amendment_task_products } = data;
+
+    // Temporarily soft delete all with task_id since there is no constraint on deletions
+    await SoilAmendmentTaskProductsModel.query(trx)
+      .context({ user_id })
+      .update({ deleted: true })
+      .where('task_id', task_id);
+
+    // Set deleted false for all in update query
+    soil_amendment_task_products.forEach((taskProduct) => {
```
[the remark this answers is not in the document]
Duncan-Brain: Yup missed that, I guess there is no test!

Duncan-Brain:
@kathyavini 

1. Correct we have agreed that we do not want to use `deleted` as flag for deleting and instead prefer only the implicit "missing" element from the array. The new middleware check double checks that assertion. 

It is necessary to "undo" the deletion made in the previous line for the updated entries or else array elements that have `id` and are only updating will update the fields provided from frontend (this does not include the deleted flag as it is hidden property).

2. A partial index constraint is not "DEFERRABLE" to the end of the transaction --  I was originally hoping that would fix the issue. Essentially the problem is that ordering the changes correctly is challenging for this "unretire" problem we have run into a few times. There must be something wrong with this architecture overall.. I am becoming much more pro hard delete. Batch updating actions like this are tough.

So that challenging problem is: given an array of existing taskProducts being updated, new taskProducts, and implicit deleted task products -- order the rows to ensure the uniqueness constraint. Also given that the constraint check can't be deferred, and taskProduct ids can be updated, re-added, deleted all in one go (differing orders of the test situation above)

The final state of upsert should be ok -- but the partial index is non deferrable.

kathyavini (approved):
@Duncan-Brain Okay thank you, I think I understand now! The `deleted` property is to trigger overwrite on the PATCH, not to modify something wrong with the request 👍

Yes I was also thinking `deferred` (on a [unique constraint](https://knexjs.org/guide/schema-builder.html#unique) instead of index), but it wasn't working and I think your migration PR said the same thing, which is annoying, because there it is in the docs... 😝 

Anyway, keeping the index as is, I think this works well, and doesn't seem to be doing anything outrageous or dangerous. The static modifier is an improvement too, I think! I like it!

packages/api/tests/task.test.js
```diff
@@ -2296,6 +2296,108 @@ describe('Task tests', () => {
...
+        createdTask.soil_amendment_task_products[
+          indexOfSecondProduct
+        ] = mocks.fakeSoilAmendmentTaskProduct({
+          product_id: soilAmendmentProductOne.product_id,
+          purpose_relationships: [{ purpose_id: soilAmendmentPurpose }],
+        });
+
+        // Update first task product id to second product id
+        createdTask.soil_amendment_task_products[indexOfFirstProduct].product_id =
+          soilAmendmentProductTwo.product_id;
```
SayakaOno, line 2377: Could you help me understand this part?
1. You created a task with two products (let's say A and B).
2. "Replace second task product (B) with new taskProduct with id of first task product (A)" - Is this different from just replacing the product_id?
3. Update first task product (A) id to second product (B) id

Now the old first task product has productB and the old second task product has product A.
I expect to have two `deleted: false` task products or two `deleted: false` task products and two `deleted: true` task products.

What am I missing?

Duncan-Brain: @SayakaOno 

2. Replacing taskProduct2 with new taskProduct essentially removes returned `id`. Simulating removing a taskProduct card in ui and adding a new one with details in one line. (There is other ways to do this). But by removing id it creates a "taskProduct 3" 

This simulates what happens on the frontend scenario where id is deleted when removed. So expectation is that taskProduct1 with id 1 updates product_id, taskProduct2 is deleted, taskProduct3 is created with original taskProduct1 product_id.

packages/api/tests/task.test.js
```diff
@@ -2296,6 +2296,108 @@ describe('Task tests', () => {
...
+
+      // Add task
+      postTaskRequest({ user_id, farm_id }, 'soil_amendment_task', taskData, async (err, res) => {
+        expect(res.status).toBe(201);
+        const createdTask = res.body;
+        const { task_id } = createdTask;
+        // Delete abandonment reason to prevent validation error
+        delete createdTask.abandonment_reason;
+
+        // Find index of second task product
```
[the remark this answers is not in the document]
Duncan-Brain: Yup haha I started with just product 2 and then updated.

packages/api/tests/task.test.js
SayakaOno: I remembered about the `id` and came back to update my comment, but it was too late 😂 Thank you Duncan!

SayakaOno (approved):

PR #3330, opened 2024-07-23
LF-4341: The text overlaps the borders of the save button on mobile device

SayakaOno:
**Description**

- Truncate text in InFormButtons
   - Remove padding
   - Set display to `inline-block` 
- Update InFormButtons to take `className`
- Add max-width to the buttons in `InFormButtons`. Without max-width, the container width extends to accommodate the buttons' width.

I can't do better than this now 😣😣😣😣😣🤮 If you have any suggestions, please let me know!

Jira link: https://lite-farm.atlassian.net/browse/LF-4341

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/Form/InFormButtons/index.tsx
+++ b/packages/webapp/src/components/Form/InFormButtons/index.tsx
@@ -27,6 +27,7 @@ export interface InFormButtonsProps {
   onCancel: () => void;
   onConfirm?: () => void;
   confirmButtonType?: 'button' | 'submit';
+  className?: string;
 }
 
 const InFormButtons = ({
@@ -37,11 +38,12 @@ const InFormButtons = ({
   onCancel,
   onConfirm,
   confirmButtonType = 'button',
+  className,
 }: InFormButtonsProps) => {
   const { t } = useTranslation();
 
   return (
-    <div className={styles.container}>
+    <div className={clsx(styles.container, className)}>
       <div className={styles.buttons}>
         {statusText && <span className={styles.statusText}>{statusText}</span>}
         <Button type="button" color="secondary-cta" className={styles.button} onClick={onCancel}>
--- a/packages/webapp/src/components/Form/InFormButtons/styles.module.scss
+++ b/packages/webapp/src/components/Form/InFormButtons/styles.module.scss
@@ -25,6 +25,10 @@
 .statusText {
   font-size: 16px;
   color: var(--Colors-Neutral-Neutral-300);
+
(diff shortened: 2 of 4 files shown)


Duncan-Brain (approved):
I am fine with this fix haha 

But it is a design flaw for sure! 

I would also be happy with telling Loic we changed his design to one of these when he gets back:
- removing "Editing..." from all views
- moving "Editing..." to other text "Any changes will affect ... " for all views to become "Editing... any changes will affect ... "
- Changing "Save Product" to "Save"

kathyavini (commented):
When I looked at the ticket I actually thought the plan was just to make a super tall button 😂  Compared to that I think the truncating solution looks great!!

As @Duncan-Brain said it's a design issue now and we're going to have to make some design decisions ourselves, but I'm concerned about how neither cancel nor save is visible as a full word anymore in other languages:

Screenshot 2024-07-23 at 12 20 48 PM

Screenshot 2024-07-23 at 12 20 00 PM

Screenshot 2024-07-23 at 12 19 15 PM

I think I would vote to flex wrap or even toss "Editing..." at `xs-breakpoint` in favour of letting the functional buttons be clear:

(This is ugly! But clear)
Screenshot 2024-07-23 at 12 48 49 PM

(I think my preference as editing was always meant to be helpful text)
Screenshot 2024-07-23 at 12 49 40 PM

What do you think? Should we UX/UI channel it? It's really a design call I think...

Edit: Oh I didn't notice Duncan's edits. So my vote is closest to bullet point one but I think remove it just for mobile; no need to mess with the Desktop view which is ideal. "Product" part of save is definitely necessary as there are two forms on this page! 🥲

Duncan-Brain:
Also good with Joyces suggestion -- I just  thought maybe deleting from all views would be easier, less code, consistent but just mobile view is also cool with me!

And the other form button is "continue" vs "save" so I thought it might be distinguishable enough -- but yeah not top 
 option.

SayakaOno:
Thank you both! I'll go with the easiest option: hide "Editing..." for mobile.

Duncan-Brain (approved):

kathyavini (approved):
I think every few weeks I google how to truncate text on a flex container at the point that the parent would expand, and I still have no idea how to do it 😭 CSS can be so annoying sometimes!

At 150px "Save Product" still cuts off in Spanish and French, but this is definitely the best iteration given the design. Thank you!! 🙏

PR #3333, opened 2024-07-23
LF-4315: Magnetic supplier input field

SayakaOno:
**Description**

The cause of the strange behaviour was that `fields` returned from `useFieldArray` is updated every time any change is made on the product form inputs. This triggers re-render on `ProductCard`. The problematic was the `useEffect` in `ProductDetails`, which is supposed to be called when a new product is selected or the product is created or updated. Because of the re-render of `ProductCard`, `products` in the dependency array was updated every time any field is updated, triggering the useEffect unexpectedly.

The solution:
Replace `products` (object) in the dependency array to `productsVersion` (number) so that the useEffect is called only when productId is updated or a product is created or updated.

Changes:
- Call `setProductsVersion` when the product form is submitted.
- The actions that need to happen after `getProducts()` is complete are now passed to `getProducts()` as a callback function.
- Update `getProducts` to call the callback function.

Jira link:
- https://lite-farm.atlassian.net/browse/LF-4315
- https://lite-farm.atlassian.net/browse/LF-4343

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/Task/AddSoilAmendmentProducts/ProductCard/ProductDetails.tsx
+++ b/packages/webapp/src/components/Task/AddSoilAmendmentProducts/ProductCard/ProductDetails.tsx
@@ -89,6 +89,7 @@ export type ProductDetailsProps = {
     callback?: (id: ProductId) => void,
   ) => Promise<void>;
   fertiliserTypeOptions: { label: string; value: number }[];
+  productsVersion: number;
 };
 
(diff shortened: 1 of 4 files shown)


Duncan-Brain (approved):
Interesting solution! Solves the problem and fiddling around with products seems to still work well.

kathyavini (approved):
Amazing! Fixed so many bugs!

PR #3334, opened 2024-07-24
LF-4342 For the mobile view only it might be too much to have name and npk all in select option and there should probably be some padding between the end of product name and n

kathyavini:
**Description**

What a branch name! I put three changes into this branch in separate commits so if (2) and (3) aren't desired -- they are minor changes and whether they improve the situation is entirely subjective -- I can just revert those two changes.

The fundamental change of this PR is **number 1 below.**

Changes:

(1) Add the necessary left-hand padding between the NPK string and the rest of the React Select option:

(2) Right-align the NPK info in the dropdown options for readability:

(3) Prevent breaking between the N: / P: / K: and its percent value by wrapping each npk key-value pair in its own span:

That last one might be more code than my taste for a little change 😉 

Jira link: https://lite-farm.atlassian.net/browse/LF-4342

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/Task/AddSoilAmendmentProducts/ProductCard/index.tsx
+++ b/packages/webapp/src/components/Task/AddSoilAmendmentProducts/ProductCard/index.tsx
@@ -69,7 +69,9 @@ const formatOptionLabel = ({ label, data }: ProductOption): ReactNode => {
   return (
     <span className={styles.productOption}>
       <span key="name">{label}</span>
-      <span key="npk">{npk}</span>
+      <span className={styles.npkText} key="npk">
+        {npk}
+      </span>
     </span>
   );
 };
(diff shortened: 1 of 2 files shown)


Duncan-Brain (approved):
I like all the changes! 

Only things I noticed that is related but not obligated to think about at all we can ask Loic later:
- Name can be so long NPK doesnt appear -- so it might be good to prioritise one (imo NPK) and truncate one (imo name) for consistent options look since the max width for NPK is probably N-33.33%,P-33.33%,K-33.33%
Screenshot 2024-07-24 at 10 42 26 AM
Screenshot 2024-07-24 at 10 43 38 AM

- And maybe we should be rounding the inline text numbers further or adding ~1 instead of 1.2 
Screenshot 2024-07-24 at 11 00 24 AM

SayakaOno (approved):
Looks good!

PR #3337, opened 2024-07-24
LF-4346 The user is not able to complete a task any task for a farm worker account without an email

kathyavini:
**Description**

To summarize our discussion (please also see Jira ticket for details):
- This is not a regression; the bug is about 2 years old
- It would only be triggered in the situation Denis showed us today (completing a task assigned to a worker without an account), ONLY IF
- The managing account has multiple userFarms and the first one returned by the query in `getUserRoleId()` (see PR) happens to be non-admin. This is highly unlikely in a non-QA context, but the fix is straightforward so it has been applied
- The neighbouring function had the same bug so it was fixed at the same time

Jira link: https://lite-farm.atlassian.net/browse/LF-4346

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

To set up the situation that would cause the issue, I used a user account whose first userFarm entry (first invite) was as worker role. I verified that the original (buggy) query would return a `role_id` of 3 by running the SQL version against my database:

```sql
SELECT role.role_id
FROM "userFarm"
JOIN role ON "userFarm".role_id = role.role_id
WHERE "userFarm".user_id = <user_id here>
LIMIT 1;
```

Note: a minor remaining mystery is that running this query against Denis' user_id on beta doesn't return a `role_id` of 3, but there are something like 76 userFarm records on that ID, and I don't think there is any particular reason to trust that `.first()` will return the same one each time. It so happens that the top record (visually inspecting the records) does have `role_id` 3, for what it's worth 🙂  (Update: inspected the query and `.first()` is actually not adding a `LIMIT 1` to the SQL, interesting! So it may be grabbing the top record / first index some other way.)

I then invited this user as a manager to a different farm where I also created a user without an account and completed a task assigned to that worker, as described in the original bug documentation.

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/taskController.js
+++ b/packages/api/src/controllers/taskController.js
@@ -41,7 +41,7 @@ async function getTaskAssigneeAndFinalWage(farm_id, user_id, task_id) {
     wage_at_moment,
     override_hourly_wage,
   } = await TaskModel.getTaskAssignee(task_id, farm_id);
-  const { role_id } = await UserFarmModel.getUserRoleId(user_id);
+  const { role_id } = await UserFarmModel.getUserRoleId(user_id, farm_id);
   if (!canCompleteTask(assignee_user_id, assignee_role_id, user_id, role_id)) {
     throw new Error("Not authorized to complete other people's task");
   }
--- a/packages/api/src/controllers/taskController.js
+++ b/packages/api/src/controllers/taskController.js
@@ -40,7 +40,7 @@ async function getTaskAssigneeAndFinalWage(farm_id, user_id, task_id) {
     assignee_role_id,
     wage_at_moment,
     override_hourly_wage,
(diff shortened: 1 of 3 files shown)


SayakaOno (approved):
What a bug...😅 
Nice find!

Duncan-Brain (approved):
Looks great! Good eye on that query:

Did you so a search on similar errors? I took a quick look by searching users.user_id and found these three that looked somewhat suspicious:

- sendMissingInvitations()
	- fear: sends extra invitations
- deactivateUser()
	- fear: if you make a worker first, then added to a second farm, and the second farm deactivates you -- you lose access to the first farm.
- getTaskAssignee()
	- fear: userFarm details in response are used in some way and they are from another (first) farm

PR #3338, opened 2024-07-24
LF-4348: The 400 error will be displayed once the user will try to add an other purpose in the list of purposes during the task completion or creation

SayakaOno:
**Description**

 - Revert b697d25f5ee666afcf393f60fb69a5046da89aa1 and 4c69bfa3c6876a0cf78d8a7c5a152e1d64b67791
 - Set `keepUnusedDataFor` (https://redux-toolkit.js.org/rtk-query/api/createApi#keepunuseddatafor-1) for the three GET requests (purposes, fertiliser types, methods) to keep data cached for a week.

I'm not sure what the appropriate `keepUnusedDataFor` is... Was this an expected solution?

Wait, if the user starts the completion flow right before the cache is expired, they will have the same issue, won't they? Let me try to do better...

Jira link: https://lite-farm.atlassian.net/browse/LF-4348

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/containers/Task/saga.js
+++ b/packages/webapp/src/containers/Task/saga.js
@@ -73,6 +73,7 @@ import {
 } from '../harvestUseTypeSlice';
 import { managementPlanWithCurrentLocationEntitiesSelector } from './TaskCrops/managementPlansWithLocationSelector';
 import { formatSoilAmendmentProductToDBStructure } from '../../util/task';
+import { api } from '../../store/api/apiSlice';
 
 const taskTypeEndpoint = [
   'cleaning_task',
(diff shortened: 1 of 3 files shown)


kathyavini:
@SayakaOno it's oh so slightly gimmicky, but you could throw the value into a form field and not have to pass anything different in the data flow, nor rely on cache.

The form already has the value calculated so it would just be:

```ts
<input type="hidden" value={otherPurposeId} {...register(OTHER_PURPOSE_ID)} />
```

And `formatPurposeIdsToRelationships` would want to use loose, not strict, equality because it will come over as a string, but otherwise I don't think anything is needed besides pulling it with the other properties in `formatSoilAmendmentProductToDBStructure()`...

SayakaOno:
@kathyavini It would be great if you could open a PR with your solution!

PR #3342, opened 2024-07-28
LF-4339: Create product form validation

zagran:
**Description**

I added validation of trimmed value length for the required field Supplier. This will fix the issue on this form only, but theoretically, we can add this check on all `required` validation options.

Jira link: https://lite-farm.atlassian.net/browse/LF-4339 

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/Task/AddSoilAmendmentProducts/ProductCard/ProductDetails.tsx
+++ b/packages/webapp/src/components/Task/AddSoilAmendmentProducts/ProductCard/ProductDetails.tsx
@@ -367,7 +367,7 @@ const ProductDetails = ({
               required: interested,
               maxLength: hookFormMaxCharsValidation(255),
               validate: (value) => {
-                if (value?.trim().length === 0) {
+                if (interested && value?.trim().length === 0) {
                   return t('COMMON_ERRORS.UNIT.REQUIRED');
                 }
                 return true;
--- a/packages/webapp/src/components/Task/AddSoilAmendmentProducts/ProductCard/ProductDetails.tsx
+++ b/packages/webapp/src/components/Task/AddSoilAmendmentProducts/ProductCard/ProductDetails.tsx
@@ -366,6 +366,12 @@ const ProductDetails = ({
             hookFormRegister={register(SUPPLIER, {
               required: interested,
               maxLength: hookFormMaxCharsValidation(255),
+              validate: (value) => {
+                if (value?.trim().length === 0) {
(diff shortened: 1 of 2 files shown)


packages/webapp/src/components/Task/AddSoilAmendmentProducts/ProductCard/ProductDetails.tsx
```diff
@@ -366,6 +366,12 @@ const ProductDetails = ({
             hookFormRegister={register(SUPPLIER, {
               required: interested,
               maxLength: hookFormMaxCharsValidation(255),
+              validate: (value) => {
+                if (value?.trim().length === 0) {
+                  return t('COMMON_ERRORS.UNIT.REQUIRED');
```
Duncan-Brain, line 371: So it is actually only conditionally required if certifications `interested` is `true`.

So currently this is also validating in the `optional` case.

Duncan-Brain (changes requested):
Just looking at the video in the Jira ticket it looks like there is a special case where you can click `Create " "` in the creatable select when there are no other pre-existing products (I had to try to create a new farm to see this) . And I can still do that and receive an error on this branch. If it feels out of scope feel free to duplicate the Jira ticket and rename it focusing on supplier only for this PR.

But main issue I see currently is that the optional field shouldn't validate -- only when it is required.

packages/webapp/src/components/Task/AddSoilAmendmentProducts/ProductCard/ProductDetails.tsx
zagran: @Duncan-Brain Fixed, thanks for the feedback

zagran:
> Just looking at the video in the Jira ticket it looks like there is a special case where you can click `Create " "` in the creatable select when there are no other pre-existing products (I had to try to create a new farm to see this) . And I can still do that and receive an error on this branch. If it feels out of scope feel free to duplicate the Jira ticket and rename it focusing on supplier only for this PR.
> 
> But main issue I see currently is that the optional field shouldn't validate -- only when it is required.

Oh, you're absolutely right, thanks for pointing this out to me (especially how to test it via creating a new farm)
I think it works as expected now, I tested it on a new farm.

Duncan-Brain (approved):
Thanks for making the changes and fixing the creatable select!

kathyavini (approved):
Sorry for the slow review 🙇 

Thank you for fixing and for looking into all those edge cases too!

PR #3357, opened 2024-07-30
LF-4354: Get cypress tests passing on integration after recent patch changes

SayakaOno:
**Description**

Before the name of Organic type selection was updated in https://github.com/LiteFarmOrg/LiteFarm/pull/3302, the order of the certification types was:
Screenshot 2024-07-30 at 3 01 14 PM

and the first option of the radio group was being checked for the cypress test.
```
cy.get(Selectors.RADIO).first().check();
```

After the change, the order of the radio buttons was changed because the API now returns `PGS` first. (I have no idea how this happens)
Screenshot 2024-07-30 at 3 11 26 PM

The organic option is no longer selected, causing the test to fail.

***Solution***
Instead of selecting the first item of the radio buttons, select the radio button with `data-cy=certification_id-1`, which is the organic option.

Jira link: https://lite-farm.atlassian.net/browse/LF-4354

**Type of change**

**How Has This Been Tested?**

Ran `npm run cypress` in `packages/end-to-end`.

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/end-to-end/cypress/support/selectorConstants.ts
+++ b/packages/end-to-end/cypress/support/selectorConstants.ts
@@ -40,7 +40,7 @@ export const ROLE_SELECTION_CONTINUE = '[data-cy=roleSelection-continue]';
 export const ROLE_SELECTION_ROLE = '[data-cy=roleSelection-role]';
 
 export const CERTIFICATION_SELECTION_CONTINUE = '[data-cy=certificationSelection-continue]';
-export const CERTIFICATION_SELECTION_TYPE = '[data-cy=certificationSelection-type]';
+export const CERTIFICATION_ORGANIC_OPTION = '[data-cy=certification_id-1]';
 export const CERTIFICATION_SUMMARY_CONTINUE = '[data-cy=certificationSummary-continue]';
 export const CERTIFICATION_SELECTION_ITEM = '[data-cy=certifierSelection-item]';
 export const CERTIFIER_SELECTION_PROCEED = '[data-cy=certifierSelection-proceed]';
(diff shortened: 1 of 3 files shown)


packages/end-to-end/cypress/support/selectorConstants.ts
```diff
@@ -40,7 +40,7 @@ export const ROLE_SELECTION_CONTINUE = '[data-cy=roleSelection-continue]';
 export const ROLE_SELECTION_ROLE = '[data-cy=roleSelection-role]';
 
 export const CERTIFICATION_SELECTION_CONTINUE = '[data-cy=certificationSelection-continue]';
-export const CERTIFICATION_SELECTION_TYPE = '[data-cy=certificationSelection-type]';
+export const CERTIFICATION_ORGANIC_OPTION = '[data-cy=certification_id-1]';
```
kathyavini, line 43: Could we use THIRD_PARTY_ORGANIC since we did go through the trouble of updating across app 😂

kathyavini (commented):
Thank you so much for looking into and fixing this troublesome situation!! 🙏 

I noticed the radios weren't defined as JSX components and that was the point at which I gave up on the idea of a `data-cy`; very cool to see it working like this!

packages/end-to-end/cypress/support/selectorConstants.ts
SayakaOno: Good idea, thank you!

packages/webapp/src/components/OrganicCertifierSurvey/CertificationSelection/PureCertificationSelection.jsx
```diff
@@ -50,6 +50,7 @@ export function PureCertificationSelection({
     const certificationRadioOptions = certifications.map((certification) => ({
       label: t(`certifications:${certification.certification_translation_key}`),
       value: certification.certification_id,
+      'data-cy': `${CERTIFICATION_ID}-${certification.certification_id}`,
```
Duncan-Brain, line 53: Could also consider using the translation key since it should also be unique.

Duncan-Brain (approved):
Seems to be working great! Thanks!

kathyavini (approved):

PR #3370, opened 2024-08-06
LF-4166 (1): Implement end to end animal creation flow

SayakaOno:
**Description**

This PR contains the outline of [LF-4166](https://lite-farm.atlassian.net/browse/LF-4166).
It should make the containers tickets easier to be worked on.

- Add `<FloatingButtonMenu />` (Add animals button) to `<Inventory />`
- Update `MultiStepForm` component
  - [Stories](http://localhost:6006/?path=/docs/components-multistepform--docs)
  - Handle different types of variant (`PAGE_TITLE`, `STEPPER_PROGRESS_BAR`)
  - Add `<FormNavigationButtons />` for the new variant of the form
  - Implement logic for the new variant to:
     - submit the form on the second last page
     - hide the button on the last page
  - Implement logic to show `window.confirm` modal when leaving the form (behaviour is TBD)
- Add dummy containers
- Add `<AddAnimals />` container with the dummy containers and create a route for it

Question:
Should the URL for the page be `/animals/inventory/add_animals`?

Jira link: https://lite-farm.atlassian.net/browse/LF-4166

**Type of change**

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

**How Has This Been Tested?**

- [ ] Passes test case
- [x] UI components visually reviewed on desktop view
- [ ] UI components visually reviewed on mobile view
- [ ] Other (please explain)

**Checklist:**

- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [x] The precommit and linting ran successfully
- [ ] I have added or updated language tags for text that's part of the UI
- [x] I have added "MISSING" for all new language tags to languages I don't speak
- [x] I have added [the GNU General Public License](https://lite-farm.atlassian.net/l/cp/BT0Dd7WW) to all new files

[LF-4166]: https://lite-farm.atlassian.net/browse/LF-4166?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3371, opened 2024-08-06
LF-4163 Create animal basics container

kathyavini:
TODO:

- [x] ~Clarify on this requirement from Jira ticket: `Three cards should fit in one row at the maximum width of 1024px. Please consult with Loïc regarding the display behaviour for narrower browser sizes.` This looks quite squished as the Figma cards are designed at 400px each. Either two should fit or 1024px should be reconsidered. This can be discussed at design meeting tomorrow~ Also handled by parent 😍 Now full-width
- [x] Check on the flexbox wrapping behaviour, also for design meeting --> got the okay
- [x] ~Check fixed header variant meaning (I forget the difference between OVERFLOW and PAPER, sorry)~ No longer applicable, will not be handled by this component anymore
- [x] Double-check about supplying a count into the sex detail's options object (as required by that component's type) --> okay, got it. It's used in a reduce 👍 
- [x] Looks like there will be some redundancies with the parent form ticket; will investigate (also the floating form buttons and the organization of the form fields need to be checked against PR #3370) -- Update: integrated!
- [ ] It's not a very nice implementation of filtering...

**Description**

Here is the interim container I was using to test this component in app; need to check if it matches the true form container!

```ts
import { FormProvider, useForm } from 'react-hook-form';
import AnimalBasics, {
  defaultValues as animalBasicsDefaultValues,
} from '../AddAnimals/AnimalBasics';
import { ANIMAL_BASICS_FIELD_ARRAY_NAME, AddAnimalsFormFields } from '../AddAnimals/types';

function AnimalGroups() {
  const methods = useForm<AddAnimalsFormFields>({
    defaultValues: {
      [ANIMAL_BASICS_FIELD_ARRAY_NAME]: [animalBasicsDefaultValues],
    },
  });
  return (
    <div>
      <FormProvider {...methods}>
        <AnimalBasics />
      </FormProvider>
    </div>
  );
}

export default AnimalGroups;
```

Jira link: https://lite-farm.atlassian.net/browse/LF-4163

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


SayakaOno:
Sorry @kathyavini, it took some time to conclude...
I was uncertain about where the stepper progress bar should be, but I think it should be in `<MultiStepForm />`. 
It would be great if we could work together to make adjustments 🙏

kathyavini:
@SayakaOno okay sounds good! Let me see if I can open a PR against your branch and integrate the component into your dummy one. I'll remove the progress bar and... let's see about the form structure 😅 I think it looks compatible so far though...

SayakaOno:
Thank you Joyce!
I was also of two minds whether to nest containers under AddAnimals or not. I'm neutral about it, so I'll change the structure as necessary!

kathyavini:
@SayakaOno I think I kind of like the nesting, but it was particularly because I wanted to have a `types.ts` file shared across the form, and but I'm not sure that file will be necessary in the end! I adopted the un-nested structure for now, but I could go either way 🙂

SayakaOno:
I see, I like the nesting too if `types.ts` is imported in containers! Let's see when is the good time to make the change...

kathyavini:
Closed; please refer to #3372

PR #3374, opened 2024-08-08
LF-3416: Fix "Delete task" button clickable area

Joaquin-M2:
**Description**

The **Delete task** button clickable area was taking the full width of its container + the set padding. This PR limits the size of the button (and therefore its clickable area) to just its content width.

Jira link: https://lite-farm.atlassian.net/browse/LF-3416

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/Finances/AddExpense/styles.module.scss
+++ b/packages/webapp/src/components/Finances/AddExpense/styles.module.scss
@@ -27,7 +27,7 @@
   font-size: 14px;
 }
 
-p.iconLink {
+button.iconLink {
   display: flex;
   align-items: start;
   height: 16px;
--- a/packages/webapp/src/components/Task/TaskReadOnly/styles.module.scss
+++ b/packages/webapp/src/components/Task/TaskReadOnly/styles.module.scss
@@ -34,8 +34,7 @@
 }
 
 .deleteText {
-  width: fit-content;
-  margin-bottom: 16px;
+  padding-bottom: 16px;
 
   span {
     text-decoration: none;
@@ -76,7 +75,7 @@
 
   svg {
     transform: translate(0px, 3px);
-    margin-right: 8px;
+    margin-right: 8px,
   }
 }
 
--- a/packages/webapp/src/components/Task/TaskReadOnly/styles.module.scss
+++ b/packages/webapp/src/components/Task/TaskReadOnly/styles.module.scss
@@ -34,7 +34,8 @@
 }
 
 .deleteText {
-  padding-bottom: 16px;
+  width: fit-content;
+  margin-bottom: 16px;
 
   span {
     text-decoration: none;
@@ -75,7 +76,7 @@
 
   svg {
     transform: translate(0px, 3px);
-    margin-right: 8px,
+    margin-right: 8px;
   }
 }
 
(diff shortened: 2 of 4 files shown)


SayakaOno (commented):

The root cause is that a `p` tag is used for the button, and the task read-only page isn't the only one affected by this issue. I’m hoping that all buttons of this type (which can be found by searching `<IconLink`) can be fixed to prevent this issue from recurring. (We often copy existing code.)

Would you mind looking for a good solution?

I think I’m the one who started using IconLink for buttons - sorry about that, everyone!

Joaquin-M2:
Hi @SayakaOno . I've looked into the problem you mentioned. It seems that the `p` tag is not really the problem since it uses the `display: inline-block` css property (assigned by the `addLinkContainer` class name).

The problem seems to come from the `<Layout>` component when `<OpenLink>` is used as one of its child components. `<Layout>` uses `display: flex`, so `<OpenLink>` becomes a flex item and therefore its `display: inline-block` property is ignored. This problem doesn't happen in other places (e,g, `src\components\Crop\ManagementDetail\ManagementPlanTasks.jsx`) because `<OpenLink>` is wrapped within another element, which prevents `<OpenLink>` from being a direct child of `<Layout>` and therefore its `display: inline-block` property is not ignored.

If we don't want to manipulate too much code (for the sake of not risking to create new bugs) I see two possible solutions:

1. We wrap the jsx from `<OpenLink>` into a plain `<div>`. The drawback is that we would be dirtying the code with semantically unnecessary html;
2. We set css properties each time we use `<OpenLink>` as a flex item, as I did this time. The drawback is that it is more cumbersome from a developer perspective.

I'm open to any suggestion, so feel free to let me know if anything better comes to your mind.

SayakaOno:
@Joaquin-M2 Thank you for looking into it.
I know it's happening in `/finances/add_expense` (`packages/webapp/src/components/Finances/AddExpense/ExpenseItemInputs.jsx`) too. Do you think that's due to the same cause?

https://github.com/user-attachments/assets/3676abce-b253-4737-90fb-937baae73bef

Using a p tag as a button isn’t a good practice, so we should address that regardless.
I’m concerned that option 2 might lead to the issue recurring. The button without a background colour is tricky... Have you come across any discussions on how others avoid elements from expanding in flex by any chance? I'm also wondering how bad it would be to add a wrapper.

Joaquin-M2:
@SayakaOno In the case of `packages/webapp/src/components/Finances/AddExpense/ExpenseItemInputs.jsx`, it is because `<IconLink>` is using the class name **iconLink**, which is setting the `display` property to `flex`, which overrides `display: inline-block`.

I've pushed another commit with your nice suggestion of replacing `p` with `button`. I had to add some styling to make it look as it was before (remove border and make the background transparent). I also added the `align-self: flex-start` property for those cases where `<IconLink>` is a flex item. I also had to add the attribute `type="button"` because otherwise it was having an erratic behavior when clicked in some pages.

Just to confirm that there is not any new bug I've checked pages that are using `<IconLink>`, `<AddLink>` and `<SubtractLink>` (the last two are using `<IconLink>`, so they were also affected by the changes). They are working fine but if you want to double check feel free to do so.

And thank you for the help and suggestions 🙂

SayakaOno (approved):
I love the align-self solution, looks great! Thank you!

We've merged a lot of PRs, so please pull integration and run `pnpm install` in `webapp` at some point, as we updated Storybook version :)

Duncan-Brain (approved):
Looks great to me , thanks for fixing!

PR #3375, opened 2024-08-08
LF-4365 Upgrade Storybook to v8

kathyavini:
**Description**

This is a clean uninstall + reinstall of Storybook and its dependencies (please see Jira ticket for list). It upgrades Storybook from version 7.0.27 to 8.2.8. I have not used `--save-exact` (version fixing) as, for whatever reason, it didn't work previously and some internal dependencies continued to update despite the main Storybook version being fixed; eventually this led to Storybook being un-runnable on a fresh `pnpm install`.

Version 8 did introduce an error loading the `.stories.mdx` file format (see ticket) -- rather than resolving this, I have removed those files (and the images that were embedded in them) from the repo and @den4ik1203 has ported their contents [to Confluence](https://lite-farm.atlassian.net/wiki/x/B4DtY) (thank you Denis!)

Screenshot 2024-08-08 at 1 07 41 PM

The documentation was general to the app rather than specific to Storybook, and I think this is a better home for it as it can be viewed without having to boot up a local instance of Storybook.

Jira link: https://lite-farm.atlassian.net/browse/LF-4365

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

May need to clear your cache after the `pnpm install` because of the missing `.mdx` paths. Since doing so I can now start up Storybook first time without crash, which is a definitely improvement over the situation before!

**Checklist:**


SayakaOno (dismissed):
It's so nice to see that storybook runs on the first attempt without modifying the vite config.
Thank you!

Duncan-Brain (commented):
No issues! Looks great -- I think removing the mdx files is a great way to deal with that!

Can we delete the @mdx/rollup configuration from config in `.storybook` under the `ViteFinal()` function?

Is `@storybook/addon-mdx-gfm` still necessary?

kathyavini:
> Can we delete the @mdx/rollup configuration from config in .storybook/main.cjs under the ViteFinal() function?

@Duncan-Brain you mean this line?

```js
  async viteFinal(config) {
    return {
      ...config,
      plugins: [...config.plugins.filter(plugin => !['@mdx-js/rollup'].includes(plugin.name))],
```

I think it's not configuring it but excluding it. I would prefer to leave it in while not knowing why Jimmy added it, I don't think it necessarily has anything to do with the .mdx files and I'm not sure it's worth the investigation.

Duncan-Brain (changes requested):
Sorry I didn't see this earlier:

I am seeing that:
1. the test runner is no longer working for me (do you see this?) I couldn't find any changes made to it in the [migration guide](https://storybook.js.org/docs/migration-guide) . I am getting the following error, and the suggested command also does not work.
Screenshot 2024-08-12 at 2 25 24 PM


How it looks on integration:
Screenshot 2024-08-12 at 3 03 13 PM

2. Other small but suggested items
- Storyshots was removed in v8 according to migration guide - for some reason we have a file `storybook.test.js` that references it despite not being in package.json.
- As part of storyshots being removed storyStoreV7:true is [default and only option:](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#storyshots-has-been-removed) 
- `@storybook/testing-library` was shimmed to deprecated to `@storybook/test`
- since *.stories.mdx has been removed from V8 we can update our storybook config away from this line:
`stories: ['../src/stories/**/*.stories.mdx', '../src/stories/**/*.stories.@(js|jsx|ts|tsx)'],`
- `@storybook/addon-mdx-gfm` does indeed note it was meant for Storybook 7 and will be removed at some point. It seems like we don't need this mdx support anymore but if we did we can use [the new syntax provided in link on npm](https://www.npmjs.com/package/@storybook/addon-mdx-gfm).

kathyavini:
@Duncan-Brain can we make a separate ticket for upgrading the test package and fixing the test runner? It's not something we actively use or have finished setting up and I don't think it's worth blocking work on Add Animals to pause and get it working right now (while as an un-openable Storybook, like we used to have, is something that has to be fixed this sprint).

Duncan-Brain:
So I didn't realize it was a blocker for animals or the team (I thought it was just annoying lol .. like the vite - material ui bug)!  I did not at all see the rush for this sprint. For me storybook lets me flip flop between your branch and integration no problem despite running `pnpm i` to do it ( I would have expected to get the same error as you). But I thought you had a workaround and could just ignore (gitignore?) the files in the short term while we upgraded properly.

My personal opinion is if we upgrade packages we shouldn't lose the other dev functionality and follow the migration guide as best as possible.  The deprecations... sure we can delay that until they break/removed like `*stories.mdx` - but the test-runner shouldn't break I don't think! Approving if you and @SayakaOno think its a blocker and I am out!

Duncan-Brain (dismissed):

kathyavini:
@Duncan-Brain you don't get a Storybook crash on a fresh `git clone` of our repo? I do and I was basing my assessment on that...

Duncan-Brain:
@kathyavini I want this in too haha, I love it also! I do get the errors when I clone -- but still doesn't seem like a big deal haha since we don't re-clone very often. But lets just migrate it officially so it is all done!

Okay did you run this to upgrade storybook? `pnpn dlx storybook@latest upgrade ` ? It has auto migrations for almost all the small stuff -- super helpful and painless and complete -- [Storybook: How to upgrade](https://storybook.js.org/docs/configure/upgrading) 

It does miss like two small things because of our setup but the `pnpm run storybook` highlights the missing pieces and then its all good!

I persisted and somehow stumbled upon a slightly different suggestion to make the test-runner work:

Screenshot 2024-08-12 at 6 22 05 PM

`npx exec playwright install` fixes test runner for me. While I don't understand why it isn't a normal dependency I didn't question it too much -- adding this to our readme could help.

kathyavini:
Oh nice, thank you for migration script 🙏  Ran it (the auto bits at least) and cleaned up those last two imports. 

`npx exec playwright install` doesn't work for me -- I get the error `could not determine executable to run` so could you open another ticket about this? Also for readme updates -- `pnpm storybook-test` is not listed anywhere in the current readme anyway.

> since we don't re-clone very often

I know, it's not that `git clone` is relevant to you, it just means that new contributors also can't open Storybook if they join the project now.

Duncan-Brain:
The changes work for me! 

Ah I think I typoed ! The test runner part shouldn't need a new PR -- it does seem to eventually download the correct playwright stuff to globals between the second images `pnpm exec playwright install` and the first `npx exec playwright install`. It did fail for me too for some reason the first few times for npx. I still don't get it but there doesn't seem to be any mention of it in storybook, jest, playwright issues, and it eventually works following the instruction so 🤷

Duncan-Brain (approved):

SayakaOno (approved):
Nice! 😍 Thank you!!

I encountered the same error as Duncan.
`npx playwright install didn't install the correct version of Chromium, so I ran `npm install @playwright/test@latest` to update the version of playwright, and the expected version of Chromium was installed finally.

Most tests seem to be running, but some of the relatively new NumberInput tests are broken. I don't think this issue should be fixed at this point, but I'm bringing it to your attention 🙂
Screenshot 2024-08-13 at 10 42 39 AM

Duncan-Brain:
@SayakaOno did that get added to your `package.json` or to your local global packages?

If it is possible to add it to packages maybe it does deserve a new ticket!

SayakaOno:
@Duncan-Brain To my local global packages!!

PR #3378, opened 2024-08-12
LF-4159 Create animal creation details views for animal batch

kathyavini:
**Description**

This is the component to be imported alongside `<AnimalDetails />` as ExpandableContent for the `<AnimalsDetails />` step of the form. 
- [Storybook](http://localhost:6006/?path=/docs/components-addbatchdetails--docs)
- The same subcomponents are used by both Animal and Batch detail; Sayaka already has created the batch version of those stories -- they are nested under [Add Animal Details](http://localhost:6006/?path=/docs/components-addanimalsdetails--docs)
- Added in the disabling of the type and batch fields (and so removing the need to pass option arrays) as discussed this morning (Aug 12); updated Stories to provide a defaultValue for those two fields via form

@SayakaOno `<AnimalDetails />` was _so_ well organized! And all the subcomponents too for receiving the `BATCH` prop. We probably did not understand how well set up when we pointed this 😂 (or you had set the framework up after we had pointed)...

Jira link: https://lite-farm.atlassian.net/browse/LF-4159

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


kathyavini:
I removed the options from the type + breed ReactSelects in the `<General />` component as they weren't necessary in the context of the AddAnimalsDetails form... but realized belatedly that is it perhaps these exact 4 cards that we can re-use for the readonly + edit views. Is it so @Duncan-Brain? Then I will for sure restore. Maybe we can chat about the needs of readonly + edit in tech daily tomorrow?

Duncan-Brain:
Yeah that was my understanding is that they will be re-used but yes, whoever gets that ticket may have to figure out some form structure stuff and fill us in!

kathyavini:
@Duncan-Brain no I haven't even looked at LF-4168; I have my hands full with LF-4164. Let's discuss tomorrow because if neither LF-4168 nor this one have been merged yet and both are in review, maybe the extra fields need a separate ticket entirely since they are blocked by both.

packages/webapp/src/components/Animals/AddAnimalsDetails/type.ts
```diff
@@ -55,9 +58,9 @@ export enum DetailsFields {
 }
 
 export type Option = {
-  [DetailsFields.TYPE]: ReactSelectOption<string>; // TODO: LF-4159
-  [DetailsFields.BREED]: ReactSelectOption<string>; // TODO: LF-4159
-  [DetailsFields.USE]: ReactSelectOption<number | string>; // TODO: LF-4159
+  [DetailsFields.TYPE]: AnimalSelectOption;
+  [DetailsFields.BREED]: AnimalSelectOption;
+  [DetailsFields.USE]: ReactSelectOption<number | string>;
```
SayakaOno, line 63: I'm not sure why I didn't add this, but the value is a number, isn't it? 😓
Stories will need to be updated as well, sorry 🙏

SayakaOno (commented):
Thank you so much for taking care of this!!! 🙏 

- Should we add "Batch" to the titles?
Screenshot 2024-08-15 at 4 23 00 PM

packages/webapp/src/components/Animals/AddAnimalsDetails/type.ts
kathyavini: Fixed, thank you!

kathyavini:
> Should we add "Batch" to the titles?

Wow, I completely did not notice that extra word 👀 Thank you for your careful eyes again!

SayakaOno (approved):
Looks good, thank you!

Duncan-Brain (approved):
Looks good! 

(I personally liked the headings without "batch" -- but they are designed that way!) 

Only other thing I really see is you could consider fixing the Origin part of the batch story by passing `origin` to the Origin component similar to the Origin specific story.

kathyavini:
> Only other thing I really see is you could consider fixing the Origin part of the batch story by passing origin to the Origin component similar to the Origin specific story.

@Duncan-Brain I like that idea! Because as we saw in tech daily, I was clueless about where the interactivity was set up and not 😄 Let's just put it on all the stories.

Duncan-Brain (approved):
Still works!

PR #3385, opened 2024-08-15
Lf 4357 concurrency issue the finance page will be broken once the user will try to change the expense revenue type of the already deleted revenue or expense

gursimran-singh:
**Description**

The finance page is broken if the user changes the expense/revenue type to already deleted revenue or expense. This behaviour should be prevented and an error message should be displayed in this case.

Jira link:https://lite-farm.atlassian.net/browse/LF-4357

**Type of change**

**How Has This Been Tested?**

1. Create a new “Expense" type
2. Open the newly created “Expense" type in different tab
3. Edit any previously created Expense.
4. Delete the Expense type
5. Change the type of the Expense to the one just created
6. An error message is shown while preventing user to change to the deleted Expense type

Follow the same steps for Revenue as well to test the functionality.

- [x] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/webapp/src/containers/Finances/saga.js
```diff
@@ -111,11 +111,17 @@ export function* updateSaleSaga(action) {
...
-    } else {
-      yield put(enqueueErrorSnackbar(i18n.t('message:SALE.ERROR.UPDATE')));
+    switch (e.response.data) {
+      case 'sale deleted':
+        yield put(enqueueErrorSnackbar(i18n.t('message:SALE.ERROR.SALE_DELETED')));
+        history.push(FINANCES_HOME_URL);
+        break;
+      case 'revenue type deleted':
+        yield put(enqueueErrorSnackbar(i18n.t('message:REVENUE.ERROR.REVENUE_TYPE_DELETED')));
+        history.push(FINANCES_HOME_URL);
```
SayakaOno, line 121: Could you please remove the redirection for this case?

SayakaOno (commented):
We discussed the message and UX with Loïc. Please take a look at the comments for our requests 🙏 

Also, in case you were not aware, some tests are failing.

For the record, we also decided to persist the error message. Here is the ticket for that change:
https://lite-farm.atlassian.net/browse/LF-4390

packages/api/tests/expense.test.js
```diff
@@ -395,6 +395,7 @@ describe('Expense Tests', () => {
       const patchData = {
         value: expense.value + 5,
         note: 'patched in note',
+        expense_type_id: expense.expense_type_id,
```
SayakaOno, line 398: Instead of adjusting the tests, can we add a check to see if `expense_type_id` is included in the request body before verifying if it's deleted?

Duncan-Brain:
Does this fix the crash for you? 

I am still seeing that the iconKey is not populating for the expense that has the deleted type.  

I think you have fixed that it shouldn't be allowed to occur which you fixed -- thank you!

To fix the crash It seems like since we don't store deleted types locally and `useTransactions()` relies on the `expenseType` to choose an iconKey which is required for Icon component. 

Icon component requires an IconKey which makes sense. But since it is not in typescript and it looks like I forgot to add `propTypes.required` it does not give a good warning. We should probably either fix the logic on `useTransactions()` line 120 for example or if there is a way to avoid the crash even in this now prevented situation.

Screenshot 2024-08-21 at 9 37 04 AM

packages/api/src/controllers/saleController.js
```diff
@@ -96,6 +96,12 @@ const SaleController = {
           return res.status(409).send('sale deleted');
         }
 
+        // do not allow to change to deleted sale/revenue type
+        if (await baseController.isDeleted(trx, RevenueTypeModel, { revenue_type_id })) {
```
SayakaOno, line 100: Could you please check if `revenue_type_id` exists? 🙏

SayakaOno (approved):
Thank you!

gursimran-singh:
> Does this fix the crash for you?
> 
> I am still seeing that the iconKey is not populating for the expense that has the deleted type.
> 
> I think you have fixed that it shouldn't be allowed to occur which you fixed -- thank you!
> 
> To fix the crash It seems like since we don't store deleted types locally and `useTransactions()` relies on the `expenseType` to choose an iconKey which is required for Icon component.
> 
> Icon component requires an IconKey which makes sense. But since it is not in typescript and it looks like I forgot to add `propTypes.required` it does not give a good warning. We should probably either fix the logic on `useTransactions()` line 120 for example or if there is a way to avoid the crash even in this now prevented situation.
> 
> Screenshot 2024-08-21 at 9 37 04 AM

Yes, this fixes the crash for me. If you can elaborate little more on what exactly you want me to check/do, it will be helpful for me to look further.

gursimran-singh:
@Duncan-Brain

Duncan-Brain:
@gursimran-singh 

So I followed the instructions on how to break a farm finances page in the Jira ticket description. I expected that the farm finance page would work again and it does not work for me. I still see the whitescreen in the video.

This PR seems to fix the chance of a bad db state happening to new entries only --  I still see my finances page broken with white screen and console errors.

gursimran-singh:
@Duncan-Brain 

I understand now what your concern is that the finance page still crashes for you. This PR is intended to prevent bad DB state from happening and not to make the code to handle the bad db state. 

In my opinion we should not go in that direction rather than these kind of errors/crashes helps us know the issues with our applications. Imagine, if it was good enough to handle this situation then this error would have not been detected that the user is actually able to make a bad DB state by his actions intentionally or un-intentionally, and in long run this might have created more problems. 

I always focus on preventing the errors from happening if possible, and if not, then find a solution to handle them gracefully. Prevention is better than cure, this is what I follow. And, if you guys think that this still need more of attention then I would suggest that the API should not fetch the bad DB date and pass it to frontend. By doing this frontend won’t crash.

Duncan-Brain (approved):
@gursimran-singh 

No worries -- I can fix the frontend bug in a later PR  -- I have a few Icons improvements I should make.

Since I saw nothing about how many users it affects on Prod and Beta (but that I assume is 0) it seemed like BOTH prevention and cure would be necessary. Prevention of course being more important.

This is both an actual frontend error as well since `undefined` does not need to be passed to icon in this case. It may occur in other current or future situations.

PR #3388, opened 2024-08-19
LF-4163: Animal basics layout adjustment

SayakaOno:
**Description**

- Update AddAnimalsFormCard header translation (Add animal -> Add animals)
- Make AddAnimalsFormCard header on one line and truncate it when it doesn't fit
- Adjust AddAnimalBasics layout to align 3 cards at browser size of 1200px or more

Jira link: https://lite-farm.atlassian.net/browse/LF-4163

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


kathyavini (commented):
Thank you for completing my what I should have done in my ticket!! 🙇

I like how it looks in all the widths, but the one thing I would want to check with Loïc and show him in action next week, is how the cards shrink at each breakpoint -- it's a bit unintuitive for me that if I want more card area to work with, I should decrease my window width rather than increase it:


<br>


<br>



Normally I feel like more browser window should equal more card real estate, but it ends up being the opposite 🤔 Anyway, maybe not important, but I think it would be worth a check.

Duncan-Brain:
I think @kathyavini point makes sense for when the number of cards == 2.

But I consider the following screenshot and generally what this pr accomplishes to be correct. So if there is a way to combine the two and deal with the card count = 2 that might be more what I expect.

Screenshot 2024-08-23 at 9 37 40 AM

SayakaOno:
Thank you for reviewing!

> if I want more card area to work with

It seems like we shouldn't place three cards per row!
Maybe we can just go with two cards per row and set a max-width for the smallest breakpoint? Let's see what Loïc will say...

SayakaOno:
Based on the discussion with Loïc this morning, I made the max-width of the animal basics container 100% and set the max-width: 480px and the min-width: 360px.

Changes:
- In `FixedHeaderContainer`, apply `max-width: 1024px` if the page URL is not included in the `FULL_WIDTH_ROUTES`.
- Add max-width and min-width to the animal basics cards.
- Set animal details container's max-width 1024px. (We will need to do the same for the summary container)

Thank you!

packages/webapp/src/components/Animals/FixedHeaderContainer/styles.module.scss
```diff
@@ -24,13 +24,16 @@
...
 .paper {
   width: 100%;
-  max-width: 1024px;
   overflow-y: auto;
   border: none;
   border-radius: 0;
   box-shadow: none;
   flex: 1;
 
+  &.hasMaxWidth {
```
kathyavini, line 33: ❤️

packages/webapp/src/containers/Animals/AddAnimals/AddAnimalDetails/styles.module.scss
```diff
@@ -17,6 +17,8 @@
 
 .container {
   padding: 16px;
+  max-width: 1024px;
```
kathyavini, line 20: 🙏

packages/webapp/src/components/Animals/FixedHeaderContainer/index.tsx
```diff
@@ -22,12 +22,17 @@ import React, {
...
+import { matchPath } from 'react-router-dom';
 import clsx from 'clsx';
 import { Paper } from '@mui/material';
+import history from '../../../history';
+import { ADD_ANIMALS_URL } from '../../../util/siteMapConstants';
 import styles from './styles.module.scss';
 
 const PAPER_BORDER = 2;
 
+const FULL_WIDTH_ROUTES = [ADD_ANIMALS_URL];
```
kathyavini, line 34: It would be a few more strings for the `.some()` to check (and that might be enough reason), but is there another downside to exporting the App `FULL_WIDTH_ROUTES` into this component so we are defining those routes in just one place instead of two?

kathyavini (dismissed):
Thank you ❤️  And sorry for all the back and forth I think these cards have gotten about this max-width thing... I think since April, even 🫨 

The ability to set max width on `<FixedHeaderContainer />` is awesome though, thank you! I think we will really like having that option available going forward 🙏

packages/webapp/src/components/Animals/FixedHeaderContainer/index.tsx
SayakaOno: I initially thought about this too but realized that these `FULL_WIDTH_ROUTES` serve two different purposes.
For example, the inventory page needs to be full width in the App because of the KPI, but the content itself still needs to be max-width of 1024px. (it will be cumbersome when needing to add a route to both files though...)

Maybe renaming it to `CONTAINER_FULL_WIDTH_ROUTES` could be less confusing? Or passing a CSS class to `FixedHeaderContainer` might be better??

kathyavini: Ohhhh, I see! In that case, yes maybe a different constant name because the identical naming made me automatically think of the other one and assume the function would be the same too.

Can you tell me more about passing a CSS class? I might have thought just a prop `isFullWidth` or `fullWidthContent` would work too, but in both cases this would have to be passed all the way through e.g. MultiStepPageForm, is that right? Since some forms will I suppose be full-width and others not?

SayakaOno: Passing a CSS class would be similar to passing `isFullWidth` or `fullWidthContent` - instead of a boolean, we would pass something like `{ classes.paper: <className that has max-width: 100%> }`. But you are right, it wouldn't be an enjoyable approach! I'll rename the constant instead. Thank you! ❤️

kathyavini (approved):

Duncan-Brain (approved):

PR #3390, opened 2024-08-21
Lf 3898 an error message is shown once the user tries to upload the custom photo of the farm which is more than 5 mb

navDhammu:
**Description**

- Displays an error message modal when user selects an image file that exceeds 5mb. 
- Some minor refactoring in farm profile form.

Jira link: https://lite-farm.atlassian.net/browse/LF-3898

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


Duncan-Brain (changes requested):
Looking great! I love that the extra `FarmImagePicker` is now consolidated!

Only change I see needed is that the drag and drop action also needs to be handled the same way.

Extras:
1. Instead of shouldRemoveImage -- should we just be setting `imageFile = null`?
2. Not related to your work but: there is a tiny hover tooltip that says "no file chosen" - we might be able to remove it.

Screenshot 2024-08-23 at 10 31 01 AM

navDhammu:
@Duncan-Brain I made the change for handling drag and drop. Do you have any ideas on how to remove that hover tooltip?

Something else I had in mind, would it be better to allow setting the max size dynamically on ImagePicker incase there's a need to change it in some places, rather than a fixed 5mb?

navDhammu:
Regarding the shouldRemoveImage:

Since imageFile = null is the initial state of the form, then in the case where a user already has a custom farm image, and only the other form fields are edited without changing the image file, then the existing image would end up being deleted when the form is submitted.

Also, I actually don’t feel the UX is intuitive here. For example, clicking remove image immediately makes it disappear, so I’m not sure users will understand that it’s not actually deleted until they press save. And when adding the image, once it’s selected or dropped in it may give the impression that image is uploaded and saved, but they may not know ‘Save’ needs to be pressed.

packages/webapp/src/components/ImagePicker/index.tsx
```diff
@@ -75,7 +75,7 @@ export default function ImagePicker({
       dropContainerRef.current?.classList.toggle(styles.dropContainerActive);
     } else if (e.type === 'drop') {
       const file = e.dataTransfer?.files[0];
-      if (file) showImage(file);
```
Duncan-Brain, line 71: Any chance of `file` not existing?

Duncan-Brain (approved):
Looks good!

For variable size - I think hard coded is good for now until we set a max on backend (which if we are not already we should be setting) then the natural limit would be what the backend accepts. Could allow a prop to specify less than that but unless theres a storage concern why limit the user.

For `imageFile= null` -- sorry I meant `imageFile = prevImageFile` then -- main point being is the `shouldRemoveImage` a useful piece of code.

navDhammu:
Now sure what you mean by` imageFile = prevImageFile` because my understanding is that only the url of the remote file (stored in aws-bucket) is stored in the frontend and not the actual file. The imageFile field contains the file data (from the users device) which the server then uses to upload to aws. In the case where a user wants to delete the file, we mark it with a boolean flag with shouldRemoveImage so the backend knows it should be deleted, and if imageFile is null and shouldRemoveImage is false, we know that nothing should be done to the image.

packages/webapp/src/components/ImagePicker/index.tsx
navDhammu: It's weird because when I first wrote that, TypeScript was complaining that file is possibly undefined but now its not. So I don't think it's possible for file to be undefined but I'm not 100% sure.

Duncan-Brain:
Sorry completely missed the backend piece! Thought it might just be able to compare files and determine it without a special boolean.

SayakaOno (approved):
Looks good, thank you!

PR #3393, opened 2024-08-23
Image picker update

SayakaOno:
**Description**

The current `ImagePicker` is designed to save the file itself rather than the image URL. This PR introduces a hook and updates the `ImagePicker` to support saving the image URL instead.

- Create `useImagePickerUpload` hook, with logic extracted from `ImagePickerWrapper`. (Instead of compressing the file when the file size exceeds, `FileSizeExceedModal` is displayed as Navdeep implemented it)
- Update `ImagePicker` to use the function returned from the `useImagePickerUpload` hook.
- Add animals URLs to imageRouteURL for `uploadImageSaga`.
- Integrate `useImagePickerUpload` hook in `AnimalDetails`.
- Update stories
  - [animals](http://localhost:6006/?path=/docs/components-addanimalsdetails--docs)
  - [batches](http://localhost:6006/?path=/docs/components-addbatchdetails--docs)

Jira link: N/A

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/ImagePicker/index.tsx
+++ b/packages/webapp/src/components/ImagePicker/index.tsx
@@ -37,15 +37,26 @@ export type OnFileUpload = (
   event: FileEvent,
 ) => Promise<void>;
 
-export type ImagePickerProps = {
-  onSelectImage?: (file: File) => void; // Required if `shouldGetImageUrl` is false or falsy.
+type CommonProps = {
   onRemoveImage: () => void;
   label?: string;
   optional?: boolean;
   defaultUrl?: string;
-  shouldGetImageUrl?: boolean;
-  onFileUpload?: OnFileUpload; // Required if `shouldGetImageUrl` is true.
 };
+
+type ImageUrlUpload = CommonProps & {
+  shouldGetImageUrl: true;
+  onSelectImage?: never;
+  onFileUpload: OnFileUpload;
+};
+
+type DirectImageUpload = CommonProps & {
+  shouldGetImageUrl?: false;
+  onSelectImage: (file: File) => void;
+  onFileUpload?: never;
+};
+
+export type ImagePickerProps = ImageUrlUpload | DirectImageUpload;
 
 export default function ImagePicker({
   onRemoveImage,
--- a/packages/webapp/src/components/ImagePicker/index.tsx
+++ b/packages/webapp/src/components/ImagePicker/index.tsx
@@ -13,7 +13,7 @@
  *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
  */
 
-import { ChangeEvent, useEffect, useRef, useState } from 'react';
+import { ChangeEvent, DragEvent, useEffect, useRef, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { AddLink } from '../Typography';
 import PureFilePickerWrapper from '../Form/FilePickerWrapper';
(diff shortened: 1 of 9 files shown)


packages/webapp/src/components/ImagePicker/index.tsx
```diff
@@ -25,12 +25,26 @@ import { ReactComponent as EditIcon } from '../../assets/images/farm-profile/edi
...
+export type OnFileUpload = (
+  e: ChangeEvent<HTMLInputElement> | DragEvent,
+  setPreviewUrl: (url: string) => void,
+  setFileSizeExceeded: (exceeded: boolean) => void,
+  event: FileEvent,
+) => Promise<void>;
+
 export type ImagePickerProps = {
-  onSelectImage: (file: File) => void;
+  onSelectImage?: (file: File) => void; // Required if `shouldGetImageUrl` is false or falsy.
```
kathyavini, line 41: What do you think about enforcing what is written in the comments via types?

```ts
type CommonProps = {
  onRemoveImage: () => void;
  label?: string;
  optional?: boolean;
  defaultUrl?: string;
};

type ImageUrlUpload = CommonProps & {
  shouldGetImageUrl: true;
  onSelectImage?: never;
  onFileUpload: OnFileUpload;
};

type DirectImageUpload = CommonProps & {
  shouldGetImageUrl?: false;
  onSelectImage: (file: File) => void;
  onFileUpload?: never;
};

export type ImagePickerProps = ImageUrlUpload | DirectImageUpload;
```
I really don't know personally -- union types are so long and I dislike the `never` requirement -- but maybe being even more explicit about the two kinds of upload would be handy? (And you might already have better names in mind than these to describe them).

kathyavini (dismissed):
This is very cool! `onFileUpload` and how it's integrated with the component is very neat and clean. But the way that `getOnFileUpload` is set up so it can be mocked in the Storybook story?! 🤯 That I look forward to hearing about tomorrow because it's not a pattern I would have ever thought of.

packages/webapp/src/components/ImagePicker/index.tsx
SayakaOno: This is what I wanted to do! Thank you Joyce!!

kathyavini (dismissed):

Duncan-Brain (commented):
Looking good and at least the upload part seems to work for me! I think it is an interesting idea to make a helper function instead of a container to keep the logic in one place and I think this works well for that case. And focusing on the URL makes a lot of sense.

But I would like to test it on Crops or somewhere the whole flow is working including saving the url to the db.

For me I just wonder if we should also be:
- removing `ImagePickerWrapper` container + logic and use this one component always moving forward. Since there is only a couple components like crops and farm that might use this. 
- should `handleRemoveImage`,`handleChangeImage` also be included?

SayakaOno (commented):
Thank you Duncan for reviewing!

> But I would like to test it on Crops or somewhere the whole flow is working including saving the url to the db.

The crop image picker is built with `ImagePickerWrapper`, so to test it, you would need to either replace `ImagePickerWrapper` with `ImagePicker` + the hook or replace the `onFileUpload` in `ImagePickerWrapper` to use the `onFileUpload` returned by the hook...

> - removing ImagePickerWrapper container + logic and use this one component always moving forward. Since there is only a couple components like crops and farm that might use this.

The `onFileUpload` logic in `ImagePickerWrapper` and the hook is not exactly the same; `ImagePickerWrapper` compresses images over 5MB rather than displaying a warning modal, so we would need to modify the hook first. Since the UI/UX differs between `ImagePickerWrapper` (farm image picker) and `ImagePicker` (crop image picker), I think it's beneficial to have variants of image pickers, but centralizing the logic is a good idea!!


> - should handleRemoveImage,handleChangeImage also be included?

In the hook? If we need more complex behaviour, we could return `handleRemoveImage` from the hook, but defining it in a parent component and passing it to `ImagePicker` should suffice for now. `handleChangeImage` seems to be the same as `onFileUpload`...  I might have misunderstood your question!

SayakaOno:
I've updated the JSDoc and parameter names related to `event` and `eventType` and addressed Duncan's comment.
Thank you in advance for re-reviewing!

Duncan-Brain (approved):
Cool! 

@SayakaOno I just meant the UX variants are just "tech debt" that we haven't updated styles right? So it would be nice to just migrate everything and delete `ImagePickerWrapper` if we are making improvements anyways -- but not a blocker for me.

I will just wait until animals is finished to check!

kathyavini (approved):

SayakaOno:
@Duncan-Brain So you think:

Screenshot 2024-08-28 at 12 40 08 PM

should be changed to:

Screenshot 2024-08-28 at 12 40 28 PM

!??

I like the current crop image picker, but we can definitely update it if needed :)

Duncan-Brain:
@SayakaOno I think that will eventually happen ya. I don't think it 'should' but that seems to be the style Loic is going for. Maybe I am wrong but I thought that was the new version of the component! Otherwise we should have reused `ImagePickerWrapper` for animals if we like that style better?

SayakaOno:
Understood! Just to clarify, I thought the crop image picker is a better fit as the first input in the form, but I’m not opposed to using the newer version :)

PR #3395, opened 2024-08-26
LF-4272 trim sign-up name whitespace

kevin-wu01:
**Description**

Previously, when the user tries to create an account with leading spaces in the name field it would cause an error on creation. The issue has been fixed by trimming any leading spaces in the field on blur. 

[Jira link](https://lite-farm.atlassian.net/browse/LF-4272)

**Type of change**

**How Has This Been Tested?**

- [x] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/CreateUserAccount/index.jsx
+++ b/packages/webapp/src/components/CreateUserAccount/index.jsx
@@ -105,6 +105,9 @@ export default function PureCreateUserAccount({ onSignUp, email, onGoBack, isNot
         label={t('CREATE_USER.FULL_NAME')}
         placeholder={'e.g. Juan Perez'}
         hookFormRegister={register(NAME, { required: true })}
+        onBlur={(e) => {
+          e.target.value = e.target.value.trimStart();
+        }}
       />
       <Controller
         control={control}
--- a/packages/webapp/src/components/CreateUserAccount/index.jsx
+++ b/packages/webapp/src/components/CreateUserAccount/index.jsx
@@ -106,7 +106,7 @@ export default function PureCreateUserAccount({ onSignUp, email, onGoBack, isNot
         placeholder={'e.g. Juan Perez'}
         hookFormRegister={register(NAME, { required: true })}
         onBlur={(e) => {
-          e.target.value = e.target.value.trimStart();
+          e.target.value = e.target.value.trim();
         }}
       />
       <Controller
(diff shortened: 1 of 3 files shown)


packages/webapp/src/components/CreateUserAccount/index.jsx
```diff
@@ -105,6 +105,9 @@ export default function PureCreateUserAccount({ onSignUp, email, onGoBack, isNot
         label={t('CREATE_USER.FULL_NAME')}
         placeholder={'e.g. Juan Perez'}
         hookFormRegister={register(NAME, { required: true })}
+        onBlur={(e) => {
+          e.target.value = e.target.value.trimStart();
```
Duncan-Brain, line 109: seems to be working well for me -- any reason we don't want to just use `.trim()` to prevent trailing spaces as well?

SayakaOno (commented):
I think we typically trim strings on the backend using the `checkAndTrimString` utility function, so we should do the same here. (@Duncan-Brain please correct me if I'm wrong!)

Duncan-Brain:
Good call @SayakaOno forgot all about that -- if that fixes it too that would be more robust. Could also do both FE and BE for visual input correctness in which case the trimStart() on the FE makes sense too while keeping it fully trimmed on BE!

kevin-wu01:
Thanks for the suggestions, I added the backend trimming for consistency. However, after looking at it some more, adding backend trimming by itself wouldn't fix the issue. There is a regex in the frontend [here](https://github.com/LiteFarmOrg/LiteFarm/blob/3857643b1d44a957952df346c7b485c5480d3126/packages/webapp/src/containers/CustomSignUp/saga.js#L112) that splits the `Full Name` field on whitespace, which means any leading whitespace causes an empty string `first_name` to be sent to the API.

I found the same bug on the invite user to farm page and checked Google SSO out of curiosity too, but Google thankfully trims whitespace in names. The combined full name field seems to make handling the data a bit awkward.

Duncan-Brain (approved):
Awesome thank you for finding that second instance! 

@SayakaOno has gone on holiday so I will request the second reviewer from @kathyavini

kathyavini (approved):
> The combined full name field seems to make handling the data a bit awkward.

Yeah, I think it has definitely come up in discussion before whether this is the right way to be doing it!

Thank you @kevin-wu01, it looks great!

PR #3398, opened 2024-08-28
LF-4397 Adjustments on add animals

kathyavini:
**Description**

As looked at together today. Copied from Jira ticket:

- Add `brought_in_date`, `birth_date`, and `origin_id` to animal_batch table.
- Move the birth date input to the origin details.
- Remove the validation for `Batch count must be greater than 1` on the backend.
- Organic status on the frontend:
  - Default: Non-Organic
  - Remove placeholder
  
 The `animalBatch` model was also updated and `sire` and `dam` (present on table already) were also added to it.
 
 @SayakaOno I snagged one of your commits from WIP PR (the types correction -- thank you!! 🙏) via cherry-pick so I could write the correct `.find()` on ORGANIC_STATUS

Jira link: https://lite-farm.atlassian.net/browse/LF-4397

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

Inspected DDL of animal and animal_batch tables to verify the new properties were the same datatype on both tables.

**Checklist:**


packages/webapp/src/components/Animals/DetailCards/Other.tsx
```diff
@@ -73,11 +73,9 @@ const OtherDetails = ({
         render={({ field: { onChange, value } }) => (
           <ReactSelect
             label={t('ANIMAL.ATTRIBUTE.ORGANIC_STATUS')}
-            optional
-            value={value}
+            value={value ?? organicStatusOptions.find(({ value }) => value === 'Non-Organic')}
```
SayakaOno, line 76: That was super quick 😳, thank you!!!
I think this default value isn't integrated with react-hook-form and needs to be set through `usePopulateDetails`...

kathyavini: Oh, thank you! I assumed setting the value in the ReatSelect would update the form value as well, but it was lazy of me not to check 🙇 

It can definitely be done in `usePopulateDetails` but do you prefer that pattern rather than some sort of setting in the component? It's a little off from the other logic in that hook (which is all basics -> details) and I think it would have be replicated in the next container too 🤔

SayakaOno: Sorry, that was just the first approach that came to mind, and I'm not particular about it!!

kathyavini: The other option would be like this, so... it's not that beautiful either really 😂 

```ts
// Other.tsx

  useEffect(() => {
    if (!getValues(`${namePrefix}${DetailsFields.ORGANIC_STATUS}`)) {
      setValue(
        `${namePrefix}${DetailsFields.ORGANIC_STATUS}`,
        organicStatusOptions.find(({ value }) => value === 'Non-Organic'),
      );
    }
  }, []);
  ```

Comment (SayakaOno) [reply packages/webapp/src/components/Animals/DetailCards/Other.tsx]: But all the changes would be in the `Other.tsx`, right? 😂 
I'm good with either way!

**Review by SayakaOno — DISMISSED** [review]

File: packages/webapp/src/components/Animals/DetailCards/Other.tsx (hunk 1 above)
Comment (kathyavini) [reply packages/webapp/src/components/Animals/DetailCards/Other.tsx]: Correct, only that change added to `Other.tsx`

@Duncan-Brain can pick when he reviews 😆

**Comment by Duncan-Brain** [comment]
@kathyavini 

I am seeing when I add animals and go forward default is set correctly. When I go back and add new animals or batches -then go forward again -- default value for organic status is not set. 

Do you see that?

File: packages/webapp/src/containers/Animals/AddAnimals/AddAnimalDetails/usePopulateDetails.ts [hunk 2]
```diff
@@ -86,7 +92,11 @@ export const usePopulateDetails = (
 
       return updatedData.map((entity, index) => {
         const origData = trimmedOrigGroup[index] || {};
-        return { ...origData, ...entity };
+        return {
+          ...origData,
+          ...entity,
+          [DetailsFields.ORGANIC_STATUS]: origData[DetailsFields.ORGANIC_STATUS],
```
Duncan-Brain, line 98: Its too bad this goes here -- it was so clean before haah

packages/webapp/src/components/Animals/DetailCards/Other.tsx
Duncan-Brain: It seems like hooks are preferred place for storing this initial value logic -- `Other.tsx` looks pretty clean right now!

packages/webapp/src/containers/Animals/AddAnimals/AddAnimalDetails/usePopulateDetails.ts
kathyavini: It doesn't have to go here! We can use the `useEffect()` method (https://github.com/LiteFarmOrg/LiteFarm/pull/3398#discussion_r1735330899) instead!!

Let's see if that fixes the bug too because that would be easiest (yes, I do see it)!

packages/webapp/src/components/Animals/DetailCards/Other.tsx
kathyavini: I think the mess added to `Other.tsx` by putting it there is less than the mess added to `usePopulateDetails` 😂  Also I do think it fixes the bug if you could please re-check 🙏 

Thank you for spotting that @Duncan-Brain!!!!

SayakaOno (approved):
Good catch Duncan!
Looks good 💖

Duncan-Brain (approved):
Looks good!

PR #3399, opened 2024-08-28
Fix icons

Duncan-Brain:
**Description**
******My diff got messed up by pre-commit hooks on backport. -- please follow commits to actually understand!

A few icon errors were popping up -- just trying to improve the component for the future. 

- Fixes error described in https://github.com/LiteFarmOrg/LiteFarm/pull/3385 
  - finances page breaks due to bad logic if `expenseType` fails conditional due to non existence: `expenseType?.farm_id ? 'OTHER' : expenseType?.expense_translation_key` 
  - also fixes revenue types version of same logic error
- Fixes expenseType custom icon to `'OTHER'` (tiny minus icon) from `'CUSTOM'` (tiny plus icon)
- Fixes spelling error on `PIG` to match database `PIGS`
- Adds comment on `svgColorFill()` to make more clear the use case
- converts `/Icon` to typescript and fixes all associated type errors (I think I messed up -- but I first thought since the pre-commit fails when `tsc` fails that I had to push all type error fixes at the same time in one big lump commit)

Jira link: LF-4357, [LF-4406](https://lite-farm.atlassian.net/browse/LF-4406)

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

[LF-4406]: https://lite-farm.atlassian.net/browse/LF-4406?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kathyavini (commented):
Alright, they are great changes 🙇 

Very good to get rid of so much type casting and the ts version of these files is so much better! Love those red squiggles 😂  And the type guard function!

I suggested some changes in how things were written that shouldn't have any effect on logic. Seems to work great in context -- could you merge in the most recent version of integration (you have a conflict) and then maybe again after LF-4166 merge to double-check on the full animal flow though?

SayakaOno (commented):
I didn't test it, but looks good to me overall.
I think we could have a function to return an animal type icon key to make it simpler.

After resolving the conflict, you will have errors in `formatDBAnimalsToSummary` function. I did `typeString.toUpperCase()` to create iconKeys, so it requires some adjustments.
Thank you! 🙏

packages/webapp/src/components/Icons/icons.tsx
```diff
@@ -63,8 +63,64 @@ import { ReactComponent as SoilAmendmentTask } from '../../assets/images/task/So
...
+  'SHEEP',
+] as const;
+export type AnimalTypeIconKey = (typeof AnimalTypeIcons)[number];
+// TypeGuard for animal type icons
+export const isAnimalTypeIconKey = (iconKey: string): iconKey is AnimalTypeIconKey => {
+  return AnimalTypeIcons.includes(iconKey as AnimalTypeIconKey);
+};
+
+// Follows structure of https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeys-type
+export type IconName =
```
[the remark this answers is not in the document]
Duncan-Brain: I think I am just still learning typescript haha -- feel free to check out what I think solves that issue.

packages/webapp/src/containers/Animals/Inventory/KPI/index.tsx
```diff
@@ -35,13 +35,14 @@ const formatAnimalTypes = (
   types.forEach((type) => {
     if (type.count) {
       const label = 'type' in type ? type.type : t(`animal:TYPE.${type.key}`);
-      const iconKey = ('key' in type ? type.key : '') as keyof typeof iconNames;
+      const typeKey = 'key' in type ? type.key : 'CUSTOM_ANIMAL';
+      const iconKey = isAnimalTypeIconKey(typeKey) ? typeKey : 'CUSTOM_ANIMAL';
```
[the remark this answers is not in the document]
Duncan-Brain: Haha nice -- I need to remember && lol  I was looking at a inline double conditional 🤦

packages/webapp/src/components/Icons/icons.tsx
```diff
@@ -63,8 +63,64 @@ import { ReactComponent as SoilAmendmentTask } from '../../assets/images/task/So
...
+import { FunctionComponent } from 'react';
 
-const iconMap = {
+const AnimalTypeIcons = [
+  'CATTLE',
+  'CHICKEN',
+  'PIGS',
+  'BATCH',
+  'BATCH_GREEN', // svgColorFill does not work with this icon
+  'CUSTOM_ANIMAL',
```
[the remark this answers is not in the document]
Duncan-Brain: I think it makes sense to be in the same grouping -- there may be a case where you set the CUSTOM_ANIMAL explicitly instead of as a default.

packages/webapp/src/containers/Finances/useTransactions.js
```diff
@@ -116,7 +116,7 @@ const buildExpenseTransactions = ({ expenses, expenseTypes, dateFilter, expenseT
         (expenseType) => expenseType?.expense_type_id === expense.expense_type_id,
       );
       return {
-        icon: expenseType?.farm_id ? 'CUSTOM' : expenseType?.expense_translation_key,
+        icon: expenseType?.farm_id ? 'OTHER' : (expenseType?.expense_translation_key ?? 'OTHER'),
```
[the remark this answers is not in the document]
Duncan-Brain: Nooo, for some reason we make translation keys for custom types, so farm id is the only way to differentiate. 

An alternative might be to typeguard the possible type translation keys.

Duncan-Brain:
@SayakaOno there is actually no errors because you asserted the AnimalSummary type using `as` ..this might be dangerous. In my last commit I reverted my use of 'as' because I just learned 'as' skips type checking so I was still seeing errors causing crashes on FE: 

 [A quick resource I found on 'as'](https://dev.to/alexanderop/the-problem-with-as-in-typescript-why-its-a-shortcut-we-should-avoid-2km4)

Duncan-Brain:
Thanks @kathyavini for the typing suggestions!

packages/webapp/src/containers/Finances/useTransactions.js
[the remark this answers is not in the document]
kathyavini: Oh, so we switched how we used translation keys at some point? 😂 That's going to be infinitely confusing.

packages/webapp/src/components/Icons/icons.tsx
```diff
@@ -63,8 +63,29 @@ import { ReactComponent as SoilAmendmentTask } from '../../assets/images/task/So
...
+  BATCH_GREEN: BatchIconGreen, // svgColorFill does not work with this icon
+  CUSTOM_ANIMAL: CustomAnimalIcon,
+  ALPACA: AlpacaIcon,
+  GOAT: GoatIcon,
+  RABBIT: RabbitIcon,
+  SHEEP: SheepIcon,
+};
+
+const AnimalTypeIcons = Object.keys(animalIconMap);
+export type AnimalTypeIconKey = (typeof AnimalTypeIcons)[number];
```
kathyavini, line 82: I think I personally preferred the original way (keeping a separate string array for the animal icon names in particular, since they are the ones with a type guard).  And as I said, you can enforce that they all exist in `iconMap` by typing them as `IconName[]`.

But if you prefer this way to avoid the redundancy, it should work, but I don't think you can type off of the `Object.keys()` -- it will just be a string array. Typing off of the map should be okay, though:

```ts
export type AnimalTypeIconKey = keyof typeof animalIconMap;
```

kathyavini (commented):
I'm not sure if you drafted the PR in order to solve the type error crash on a custom animal in the Summary Screen (I _think_ that's all Sayaka was referring to, and that should be solvable with the same pattern you've established elsewhere):

```ts
iconKey: isAnimalTypeIconKey(typeString) ? typeString : 'CUSTOM_ANIMAL',
```

-OR- the deeper question of why `typeString.toUpperCase()` doesn't trigger a type error even after you remove _all_ the type casting in those utility functions 😂 

If the second, I think it's okay to leave -- if you fix the way `AnimalTypeIconKey` is calculated, the utility functions still trigger on completely wrong strings (e.g. if you supply 'BOO' to `iconKey`) and making it more robust than that might be more trouble than it's worth.

But I tested just now and it will pass `tsc` with all type casting removed from `AddAnimals/utils.ts`, so I think you can safely do that without having to go down a typescript rabbit hole.

packages/webapp/src/components/Icons/icons.tsx
Duncan-Brain: I tried enforcing it with `IconName[]` but it is a circular reference -- its slightly tricky but it is technically already enforced by `AnimalTypeIconKey` being used in the definition for `IconName`.

Decided to revert the change and keep the strings. It seems easier to convert an array to a type than the opposite.

packages/webapp/src/components/Icons/icons.tsx
```diff
@@ -63,8 +63,29 @@ import { ReactComponent as SoilAmendmentTask } from '../../assets/images/task/So
...
+// TypeGuard for animal type icons
+export const isAnimalTypeIconKey = (iconKey: string): iconKey is AnimalTypeIconKey => {
+  return AnimalTypeIcons.includes(iconKey as AnimalTypeIconKey);
```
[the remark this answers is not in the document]
Duncan-Brain: So it can't just be removed for me -- causes a type error on iconKey in includes wanting to be of that type. I am mixing two examples:

Has casting:
[Typescript TypeGuard:](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates)

Uses type `any`:
[StackOverflow example:](https://stackoverflow.com/questions/51528780/typescript-check-typeof-against-custom-type)

So I could remove the cast but then the function prop would want something like `(iconKey: string)` => `(iconKey: any)` and I think typing the string could be valuable for the includes function. So I am not quite sure.

packages/webapp/src/components/Icons/icons.tsx
kathyavini: Sorry to not be more clear. This is what I was originally suggesting:

```ts
const AnimalTypeIcons: IconName[] = [ // <-- type as IconName[] so you can't add just any string
  'CATTLE',
  'CHICKEN',
  'PIGS',
  'BATCH',
  'BATCH_GREEN',
  'CUSTOM_ANIMAL',
  'ALPACA',
  'GOAT',
  'RABBIT',
  'SHEEP',
] as const;

export type AnimalTypeIconKey = (typeof AnimalTypeIcons)[number];

export const isAnimalTypeIconKey = (iconKey: string): iconKey is AnimalTypeIconKey => {
  return AnimalTypeIcons.includes(iconKey as AnimalTypeIconKey);
};

export type IconName = keyof typeof iconMap; // <-- this is my main suggestion

export const iconMap = {  // <-- remove the type here to avoid circular reference
  // ...
```

The point is to let the `IconName` type be automatically generated from the keys of iconMap, and you would not type iconMap in that case to avoid the circular reference.

However, your second method (creating an `animalIconMap` and spreading it into `iconMap`) was also totally fine -- you didn't have to revert it! It's nice because it has no redundancies and let's you type iconMap -- maybe it's the best of the three. But in any case, as I said in my comment, you just needed to change a single line (the type export) and it would behave as before too.

```ts
const animalIconMap = {
  CATTLE: CattleIcon,
  CHICKEN: ChickenIcon,
  PIGS: PigIcon,
  BATCH: BatchIcon,
  BATCH_GREEN: BatchIconGreen,
  CUSTOM_ANIMAL: CustomAnimalIcon,
  ALPACA: AlpacaIcon,
  GOAT: GoatIcon,
  RABBIT: RabbitIcon,
  SHEEP: SheepIcon,
};

const AnimalTypeIcons = Object.keys(animalIconMap);
export type AnimalTypeIconKey = keyof typeof animalIconMap; <-- only change needed

// TypeGuard for animal type icons
export const isAnimalTypeIconKey = (iconKey: string): iconKey is AnimalTypeIconKey => {
  return AnimalTypeIcons.includes(iconKey); <-- bonus: type casting not necessary this way
};

export const iconMap: Record<string, FunctionComponent> = { <-- another bonus: can still keep this type
  // ...
```

Either of those solutions will work without manually listing out the strings in `IconName`.

Duncan-Brain:
@kathyavini  - So I don't see a type error crash on animals summary -- the deeper question remains!  But I drafted it because it was wrong and I had to go for an hour -- it was as you thought in a comment the typing changes to `string` and `string[]` with the updated types to remove redundancy -- I think my intellisense was slow to update or something -- so I think it is maybe harder to figure how to remove that redundancy and an improvement for a later date.

Removed those extra type castings in utils as requested 🤷  -- deeper question remains.

packages/webapp/src/components/Icons/icons.tsx
Duncan-Brain: So I tried that but the following happens: 

Screenshot 2024-09-03 at 6 39 07 PM

IconName needs to be specific.

Duncan-Brain: I think then removing the `Record` typing fixes that issue -- but then iconMap isn't typed haha

kathyavini: Ahh your right, sorry -- in both cases you can't retain `Record<string, FunctionComponent>` without that breaking `keyof typeof iconMap` 🤔 That seems strange -- there has to be a way! 

I think I still prefer `iconMap` untyped rather than the behemoth of manual types (that will keep having to grow), but maybe Anto can rescue us with a solution that's best of both worlds 😂

kathyavini (commented):
> So I don't see a type error crash on animals summary

Actually, it is the icon line that crashes but I don't think it's affected by this PR -- do you see this when you add a new animal with the creatable select?

Screenshot 2024-09-03 at 5 07 04 PM

It's fixed by your typeguard code but it exists on integration too so I don't know if this is what Sayaka's comment was referring to -- doesn't sound like 🤔

packages/webapp/src/components/Icons/icons.tsx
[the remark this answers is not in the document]
kathyavini: Sorry, that was only referring to the version of the code that you reverted!

kathyavini:
> will fix that also in this pr one way or another

I gotcha covered! That crash had its own ticket:
- https://github.com/LiteFarmOrg/LiteFarm/pull/3478

packages/webapp/src/components/Icons/icons.tsx
Duncan-Brain: @kathyavini I like this solution actually! 

I was looking for more info about it and the terminology I found is "Constrained Identity Function": https://kentcdodds.com/blog/how-to-write-a-constrained-identity-function-in-typescript

There might also be an option to use `satisfies` if our ts is up to date!

Duncan-Brain:
Merge "conflicts":
- locked icon was merged before this and would have been lost
- supportedLngs in i18n was duped -- backport merge didnt catch pre-commit hooks

packages/webapp/src/components/Icons/icons.tsx
```diff
@@ -114,6 +131,9 @@ const iconMap = {
   TRASH: TrashIcon,
   // Input
   LOCKED: LockedIcon,
-};
+} satisfies Record<string, FunctionComponent>;
```
kathyavini, line 134: 😍

kathyavini (approved):
Thank you for that article link and what a great find in `satisfies` ❤️ 

I don't know what to make of the diff anymore -- we've really made a messy situation with our partial backport I guess -- but `icons.tsx` looks fantastic!!

PR #3400, opened 2024-08-30
LF-4405: floating menu button looks different from what is designed

SayakaOno:
**Description**

Screenshot 2024-08-29 at 3 22 24 PM

- Fix the position of the floating menu by adjusting the `anchorRef`.
- Implement the active state of the button as designed.
   - replace hex with `var`.
   - switch CSS for hover style and active style so that hover style won't override the active state CSS. 
   - sorry for the formatting!

Screenshot 2024-08-30 at 13 28 00

Jira link: https://lite-farm.atlassian.net/browse/LF-4405

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/assets/colors.scss
+++ b/packages/webapp/src/assets/colors.scss
@@ -14,6 +14,7 @@
   --secondaryGreen800: #495c51;
   --secondaryGreen200: #c1e6d2;
   --secondaryGreen50: #f2faf5;
+  --yellow900: #6b4d00;
   --yellow700: #ffb800;
   --yellow400: #fed450;
   --yellow300: #fce38d;
--- a/packages/webapp/src/components/Button/FloatingActionButton/styles.module.scss
+++ b/packages/webapp/src/components/Button/FloatingActionButton/styles.module.scss
@@ -17,11 +17,11 @@
   position: relative;
 
   &[aria-expanded='true'] {
-    background-color: #e3b000;
+    background-color: var(--Btn-primary-hover);
   }
   @media (hover: hover) {
     &:hover {
-      background-color: #e3b000;
+      background-color: var(--Btn-primary-hover);
     }
   }
 
--- a/packages/webapp/src/components/Button/FloatingActionButton/styles.module.scss
+++ b/packages/webapp/src/components/Button/FloatingActionButton/styles.module.scss
@@ -16,13 +16,13 @@
 .floatingActionButton {
   position: relative;
 
-  &[aria-expanded='true'] {
-    background-color: var(--Btn-primary-hover);
-  }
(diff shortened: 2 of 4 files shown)


Duncan-Brain:
Careful eye! I never would have noticed this. 

Looking at the figma I am seeing a margin or spacing between the button and the menu on desktop only. I bet that is a mistake in figma and they should be the same. Do you know which is correct?

Figma - desktop
Screenshot 2024-08-30 at 1 09 41 PM

Local - desktop

Screenshot 2024-08-30 at 1 09 11 PM

kathyavini:
Oh dear, there really is some inconsistency in that [Desktop design](https://www.figma.com/design/eMffC0h3JPlWtzNttxJpil/Litefarm-Updated?node-id=3735-40055&m=dev) you found @Duncan-Brain.

I see also that the mobile and component design show the active state without any shadow, but the desktop design has retained it; I suspect that's an error as well (because it's not on the component Figma even for desktop).

I do think the requirement for the inner text area to have some amount of border radius is consistent through all the animal views, though.

If the curved corners are added to the `.menuItem` here, it will also back-propagate to finances, but I think that's OK as this is the more recent version of `<FloatingMenu />`:

Screenshot 2024-08-30 at 11 00 18 AM
Screenshot 2024-08-30 at 11 00 28 AM

SayakaOno:
I should've been more careful when I created it, sorry...

https://www.figma.com/design/eMffC0h3JPlWtzNttxJpil/Litefarm-Updated?node-id=91-14447&m=dev
Screenshot 2024-08-30 at 2 24 27 PM

https://www.figma.com/design/eMffC0h3JPlWtzNttxJpil/Litefarm-Updated?node-id=3093-39506&node-type=INSTANCE&m=dev
Screenshot 2024-08-30 at 2 24 41 PM

I think no spacing between the menu and the button is correct.

Also, font-weight was 700 on the finance screen 😅 
Do you mind confirming what's right with Loïc? 🙏 

I will remove the shadow for the active state!

kathyavini:
@SayakaOno 

> Do you mind confirming what's right with Loïc? 🙏

Alright, will do. We can ask Loïc about the font-weight and the desktop spacing. Should we confirm about the border-radius too? I'd be surprised if he didn't want it on animals (it's in all the views) but we can check about what he wants on Finances.

SayakaOno:
Yes, it would be great if you could check those three points. Thank you! 🙏

kathyavini (approved):

Duncan-Brain (approved):
Looks good to me! Thanks for aligning the style!

PR #3402, opened 2024-08-30
Add animal fixes

SayakaOno:
**Description**

Address Duncan's [comment](https://github.com/LiteFarmOrg/LiteFarm/pull/3397#pullrequestreview-2270361189) in #3397.  There might be better solutions and it might require adjustments. Please feel free to update or close as needed. Thank you! 🙏 

- Filter out animal uses that are not for the currently selected animal type when coming back to the details view.
- Adjust summary image width (The image might look too small at some breatkpoints.)

Jira link: N/A

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/Animals/AddAnimalsSummaryCard/styles.module.scss
+++ b/packages/webapp/src/components/Animals/AddAnimalsSummaryCard/styles.module.scss
@@ -75,8 +75,9 @@
 
 .imageContainer {
   padding-top: 32px;
-  width: 488px;
+  max-width: 488px;
   overflow: hidden;
+  flex: 1;
 
   @include md-breakpoint {
     order: 1;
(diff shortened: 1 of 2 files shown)


kathyavini:
Thank you @SayakaOno for working on this before you leave (and thanks for making a separate PR -- that definitely helps with the review!)

Sorry, I should have read Duncan's comments closer as I didn't realize the uses bug was from my hook and not related to LF-4166 at all. Thank you for suggesting a fix 🙏 

Have a great vacation!! ❤️ 🛫  See you in October

SayakaOno:
No worries!!
Thank you Joyce! ❤️ Itte kimasu 🛫 See you in October!

kathyavini (approved):
I had never given any thought to the idea of filtering the user's already input uses based on the current type, but I think this is the most logical approach. As it had not been discussed before I think we could give it a quick check with Loïc tomorrow, but I'm more than happy to merge as is first.

If Loïc likes the behaviour, we might also need it in the component anyway for readonly/edit, but I'm happy to have it reside here until that is built out!

To sum up a Slack conversation, I believe @Duncan-Brain actually wasn't requesting this adjust of previous selected items, but instead was just seeing the core filtering logic not working at some point on the other branch. But this is a real use case worth addressing, I think.

Duncan-Brain:
It seems weird to keep any of the use data if switching types, maybe in the future we can just clear it. 

I definitely just meant it was not filtering! haha But this makes sense too.

PR #3411, opened 2024-09-06
Lf 4399 translate rrule files into new languages rebase

Duncan-Brain:
**Description**

Moves the files to the locales folder for compile time use. 

Notes:
- Eager imports were chosen over dynamic imports - choice was made to avoid running async await up to the component. Maybe theres another way to do this but its working well now. They seem like small files at 1KB but maybe if theres too many it will be a problem?
- Crowdin: Will be using this to test how it handles array based translations -- will update the `rruleTranslation/ index.js` if this fails.

Jira link: [LF-4399](https://lite-farm.atlassian.net/browse/LF-4399)

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/webapp/src/locales/en/rrule.json
```diff
@@ -0,0 +1,49 @@
+{
+  "__comment": "THIS IS A FILE COMMENT FOR rrule.json. This locale file is not for use in 'i18n' and only for use with 'rrule' software package. The english file is only used as a base for translation to other languages -- DO NOT MODIFY.",
```
kathyavini, line 2: Sorry why do we need this? CrowdIn translators can't modify the source file anyway and there is a high likelihood of someone translating these instructions if we put them up. As for devs, do you think there is more danger of the files getting modified in their new location than there was in their previous location?

kathyavini (approved):
It looks good to me! I don't know much about importing or the difference between eager and dynamic import, so I can't really evaluate that 😂 I wouldn't mind getting a primer on this in tech daily, though if we have some extra time this week!

Looks good to me though and I think we can coordinate the up + down, then? My only question was about the necessity of adding that comment string.

PR #3415, opened 2024-09-09
Lf 4402 translate sensor csv js into new languages

Duncan-Brain:
**Description**

This adds the sensors translations to a locales directory to be translated in crowdin and resolves the historical BUG of a mismatch between the template csv and the parsed csv.

Notes:
- the main challenge was around the dynamic import, import syntax is different on the frontend from back due to vite + typescript compiler with no such thing on the backend. Therefore this one function was split up between the frontend and backend or else would have also been placed in the shared utility script.

Jira link: [LF-4402](https://lite-farm.atlassian.net/browse/LF-4402)

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

[LF-4402]: https://lite-farm.atlassian.net/browse/LF-4402?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/api/src/controllers/sensorController.js
+++ b/packages/api/src/controllers/sensorController.js
@@ -40,10 +40,16 @@ import { sensorErrors, parseSensorCsv } from '../../../shared/validation/sensorC
 import syncAsyncResponse from '../util/syncAsyncResponse.js';
 import knex from '../util/knex.js';
 
-const getSensorTranslations = async (language) => {
-  const translations = await import(`../../../shared/locales/${language}/sensorCSV.json`, {
+const getSensorTranslations = async (language, res) => {
+  // Remove country identifier from language preference
+  const parsedLanguage = language.includes('-') ? language.split('-')[0] : language;
+  const translations = await import(`../../../shared/locales/${parsedLanguage}/sensorCSV.json`, {
     assert: { type: 'json' },
   });
+  // Handle case where user language not supported
+  if (!translations) {
+    res.status(422).send('User language not yet supported');
+  }
   return translations.default;
 };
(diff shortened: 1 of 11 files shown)


packages/api/src/controllers/sensorController.js
```diff
@@ -40,6 +40,19 @@ import { sensorErrors, parseSensorCsv } from '../../../shared/validation/sensorC
...
 
+const getSensorTranslations = async (language, res) => {
+  // Remove country identifier from language preference
+  const parsedLanguage = language.includes('-') ? language.split('-')[0] : language;
+  const translations = await import(`../../../shared/locales/${parsedLanguage}/sensorCSV.json`, {
+    assert: { type: 'json' },
+  });
+  // Handle case where user language not supported
+  if (!translations) {
+    res.status(422).send('User language not yet supported');
```
antsgar, line 51: On the front end side you defaulted to English if the language isn't supported, should we do the same here for consistency?

Duncan-Brain: Yeah maybe I overthought this one! 

I put a "TODO" below about the issue. We base the csv parser off of the users display language in the db. I was thinking it should really be passed as a param as part of the request body to state the language of the csv explicitly without relying on the users db value.

In any case, I was thinking if the the users db language is not supported -- can we assume the csv file they have attached is in english? 

I will put up a PR

PR #3416, opened 2024-09-09
LF-4056 Language selector needs to provide list of language in their native form

kathyavini:
**Description**

This removes the language selector strings from i18n so they can be hardcoded rather than translated according to the currently selected language. The order of languages is up to us and we can confirm with product, but right now it is alphabetical by language code for roman alphabet languages, and then alphabetical by language code for non-roman alphabet languages (i.e. the three Indic languages)

@Duncan-Brain I know there is the possibility we discussed of still using the translated string in some capacity but the image Loïc provided looked like this. This is also the easier one to implement as it will involve merely removing, rather than adding, strings to CrowdIn.

Jira link: https://lite-farm.atlassian.net/browse/LF-4056

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


Duncan-Brain (approved):
@kathyavini Looks good, would say we could not delete the translated languages just in case but -- they are here if we flip flop back!

PR #3427, opened 2024-09-11
LF-4221 investigate moment date locale and modify if necessary

kathyavini:
_**Please see Jira ticket [LF-4428 Never translated string on crop flow creation](https://lite-farm.atlassian.net/browse/LF-4428) for a visual table showing where in the app we will now start seeing localized date strings.**_

(LF-4428 is the bug ticket related to this issue, but the underlying cause is as I described in the comment of [LF-4421](https://lite-farm.atlassian.net/browse/LF-4421))

**Description**

**BACKGROUND/EXPLANATION**

Our codebase used `moment(date).locale()` in several places. I'm not sure how we missed that it never worked. In order to use `moment(date).locale()` you need to import the locales, for instance like this

```js
import moment from 'moment/min/moment-with-locales';
```

(The [moment.js docs](https://momentjs.com/docs/#/i18n/loading-into-nodejs/) also describe importing needed locales individually, but that I did not find to work.)

Elsewhere in the app we have been localizing with the native `Intl.DateTimeFormat()` so for consistency I have applied at here as well, rather than importing and using moment locales.

**SUMMARY OF CHANGES**

- anywhere in app that `moment(date).locale()` was called, I made sure that the `Intl.DateTimeFormat()` was called instead
  - one exception is a function used in taskCard, because a downstream function had been built expecting a non-localized string
  - I also tried to to verify (hopefully thoroughly) that the moment-formatted date had always been used directly in the UI and not sent to the backend for any kind of storage (which would start messing with the data) -- all strings did look safe that I saw
- we had two functions named `getLocalizedDateString` so I renamed one of them
- cleaned up a comment that happened to call `moment().locale()` rather than cleaning up the code in the comment

Jira links:
https://lite-farm.atlassian.net/browse/LF-4421 (code ticket / subtask)
https://lite-farm.atlassian.net/browse/LF-4428 (bug ticket for verification)

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

[LF-4421]: https://lite-farm.atlassian.net/browse/LF-4421?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/webapp/src/components/CropCatalogue/CropStatusInfoBox/index.jsx
+++ b/packages/webapp/src/components/CropCatalogue/CropStatusInfoBox/index.jsx
@@ -37,27 +37,9 @@ export default function CropStatusInfoBox({
 }) {
   const classes = useStyles();
   const { t } = useTranslation();
-  const onDateChange = (e) => setDate?.(e.target.value);
 
   return (
     <Card color={'info'} className={clsx(classes.container)} {...props}>
-      {/* <NativeDatePickerWrapper
-        style={{
-          position: 'absolute',
-          right: '0',
-          transform: 'translateX(-12px)',
-        }}
-        value={date}
-        onChange={onDateChange}
-      >
-        <Underlined>{t('common:EDIT_DATE')}</Underlined>
-      </NativeDatePickerWrapper> */}
-      {/* <Text>
-        {t('CROP_CATALOGUE.CROP_STATUS')}{' '}
-        <span className={classes.semibold}>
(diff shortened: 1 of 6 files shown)


Duncan-Brain:
Looks good to me, moment is deprecated so any opportunity to get rid of them is great!

PR #3428, opened 2024-09-11
Lf 4395 setup typescript on backend

navDhammu:
This sets up typescript on the backend. New and existing files in `src` and `tests` should work now when converted to .ts. Anything in` /db` including the migrations still need to use .js but that can be setup with typescript as well if needed. I tested it to make sure its good for development and adjusted the npm scripts for production and integration, but will need some help to ensure nothing is broken in production as I'm not fully aware of how everything is setup with docker, for example there's prod.DockerFile which uses some npm commands that may not work.

### Overview
In order to get typescript working on the backend, it has to first be transpiled to js so it can be executed by node as node doesn't support running ts files natively. Although that may change in the near future with this new experimental feature in latest [release](https://nodejs.org/en/blog/release/v22.6.0).

Tsc is the official compiler shipped with typescript that does the transpilation from ts to js, and it's what I started working with initially but then switched to using [swc](https://swc.rs/) after running into problems with jest. Swc is an alternate to tsc that's written in rust so its much faster and provides a better dev experience, there are others like esbuild which is used by vite on the frontend. I chose swc because they provide a package [swc/jest](https://swc.rs/docs/usage/jest) that takes care of the problems I was having initially getting typescript and es modules working with our current test setup with jest. A nice side effect of using this package is that tests are running faster now, I'm seeing api tests in github actions completing in about 3 minutes instead of 6.

For development, theres still this issue of having to first compile all source code to .js before running the node server. So the workflow needs to be: make a change -- run the compiler -- execute the compiled server.js file. To make all this more straightforward there's this popular package [ts-node](https://github.com/TypeStrong/ts-node) that allows executing .ts files directly, so it makes it possible to directly execute `ts-node server.ts`. It does this by compiling typescript to javascript on the fly and passing the resulting javascript code to node. I found a similar well maintained package [swc-node](https://github.com/swc-project/swc-node) that does the same thing but uses the swc compiler, so I ended up going with that.  

To make things fast during development, this setup doesn't run type-checking on every code change, but instead its done in the pre-commit just like on frontend.

### Changes
- Requires node version >=20.6. This is enforced using the `engines` option in package.json and `engine-strict=true` in `.npmrc`. So when using an npm command like npm install, it will give an error if using the node version doesn't match this.
- For development - `npm run dev` replaces `npm run nodemon`. Nodemon is not needed as node.js comes with watch mode after node v18
-  For production - the source code is compiled to `/dist` via tsc, and the server can be started by executing `node dist/api/src/server.js`
- Babel is replaced with swc
- There are 2 separate tsconfigs, one for root dir `/api` and one for `/api/src`. The root tsconfig is only responsible for type-checking which is run in precommit, and this includes the tests as well. The api/src tsconfig is for building the source code for production.
- When importing typescript files, you need to import them with the `.js` extension like `import something from file.js` - not `file.ts`. This is because when building the code for production tsc will change all typescript code to javascript including the file extension to .js, but won't change the import statements. So anywhere an import is requiring a file with a .ts extension, it won't exist in the production build which will make the app crash.

Jira link:

**Type of change**

**How Has This Been Tested?**

- [x] Passes test case
- [ ] Other (please explain)

**Checklist:**


antsgar:
Wow, thank you so much for putting this together! The PR description was super helpful to understand the changes and the decision making process, amazing work 👏 I tested it out and it all seems to be working properly -- for the Dockerfile, I'm assuming we'll need to update the Node version and the last piece to remove installing nodemon and using it to start the server. I'm thinking we could check out this branch on the beta server and test the changes.

Right now we're using the same Node version for all the packages, so I think it might be good to update all of them to 20 which is LTS anyway.

We should also update the Readme instructions on the parts it refers to Nodemon.

navDhammu:
@antsgar Thank you for testing this out! 

I'm a bit confused with the dockerFile and what its meant to do. It's named as prod.DockerFile which suggests its for production, but then its using nodemon which is only meant for development. So if we remove nodemon from the dockerFile, then does it needs to be replaced with `npm run dev` to keep the same functionality? But if its meant for production then this wouldn't be ideal and we should be using a production command instead.

Duncan-Brain:
Hey @navDhammu sorry it has taken me some time to get to this, I am still excited this could be available soon!

The npm settings for `engine` and `engine-strict` are cool! We have definitely had people on the wrong version before causing bugs so that is cool! Also awesome that swc reduces test time, that is super handy because its like 13mins on CI! Only other replacement for nodemon I have tried is `tsx` which is pretty easy to work with.

I tested and everything worked well on:
  - dev script
  - migrations on the test database 

I tested and it fails:
- local docker compose file - specifically the export service on the node engine flag

I expect to fail
- deployment - due to prod.Dockerfile

So my other current thoughts are: 

1. Node version - totally agree with @antsgar that it would be good to update this across packages if we are updating (I might gently suggest a prequel PR for that and also Typescript version -- see below) 
   - one reason for this is I think most devs use the local commands instead of docker and manually changing node version when working between packages could be troublesome
   - Files I saw it needs updating version were, `beta_export_deploy.sh` and `prod.Dockerfile`
 
2. Nodemon - yes it is used, yes it shouldn't be - but we do need the command to work in the `prod.Dockerfile` line. As long as it does then we can punt the bad nodemon use issue and discuss on a different ticket.
 
3. Typescript version - Same as node version it might be good to update across packages if we are updating
     - I saw a lot of TS errors on your branch because my vscode settings uses Litefarms `./node_modules/typescript/lib` as the Intellisense parser - and for dev experience it would be great to have them all match so that we don't have to manually switch across packages to see others errors
     - This could maybe even be set in our codebase even under the `/.vscode/settings.json`
    
4. Tsconfig files
    - There is two - If there is any way we can have just one by using command inline flags like --noEmit or other I think that could be good
    - It's interesting that there is a package for a base tsconfig - curious about the value/risks vs copy pasting and attribution?
  
 5. Import syntax and `.js` vs `.ts`. 
     - I thought there was  a tsconfig flag for this like [--allowImportingTsExtensions](https://www.typescriptlang.org/tsconfig/#allowImportingTsExtensions) that requires the bundler to handle the conversion.

Duncan-Brain:
Sorry if I was unclear on a couple things!

**Node version**

I think we want to set the explicit version of node 20 something -- not --lts.  I think our last tech lead Ivan stressed the importance of not setting a moving target for core engines to avoid surprises down the road and have a known dependency list. It is handy to not update... but could potentially break things without a traceable action on our side.

**nodemon**

`npx` and `npm i -g` implementations of nodemon should still work since they are not part of package.json-- so we can put the scope of that onto a different ticket I think. I think only thing in prod.Dockerfile I was sure wouldn't work was the old node version. nodemon start line itself.. might have worked.

For your 

**Export server**

Ahh true I forget that is there sometimes -- will ask Joyce if there is a best way to update those and make a ticket.

**Typescript**

Agreed on long term solution

**Tsconfig files**

We can go with two for now -- I think we typically only need to `build` for beta/production. So that would be done in a dockerfile and we would select the folders we copy to the image -- which would exclude tests and other things not used on prod. 

**Import .js vs .ts**

Yah I might be wrong on that and not a big deal really to keep -- but I think babel (or in this case swc) should have an extension/plugin like[ --rewriteImportExtensions](https://www.npmjs.com/package/babel-plugin-add-import-extension) for either using `.ts` . Or even using extensionless (which might be handier actually.. at least for me my vscode autocompletes import, it assumes I am using react and does not add the extension to the import which results in errors :( haha)

navDhammu:
@Duncan-Brain  I agree on setting the node version to an exact number like 20.18, I realized this afterwards when the CI was failing due to pnpm-lock error which happened because the node version changed.

Regarding nodemon:
- I don't think some of those existing scripts with nodemon will work even though they're used with npx.  For example if you look at `npx nodemon --inspect=0.0.0.0:9229 src/jobs/index.js` in the export script, if any typescript files are being imported by `src/jobs/index.js` then it'll probably crash since the typescript wouldn't have been compiled.

Tsconfig:
- It might be possible to keep one tsconfig that way with docker, but will have to test it out.
- To answer your other question about using an external package for the base tsconfig - I thought its good to have a set of base recommended options based on the node version we're using. Ofcourse we can just copy paste it but then whenever we change versions, we have to look it up and copy paste again - its easier to just update the package with npm.

For typescript version:
- I was wondering if I should just make the temporary change in this PR and update typescript across all packages? 
- Or if we want to go for a long term solution, I think something like pnpm workspaces would be ideal because they have great built-in support for monorepos and we're already using pnpm in webapp, so we just need to update the api and other packges to use pnpm, and this isn't just useful for typescript but other common dependencies like prettier, eslint, etc. I did a quick search in jira and found this old [ticket](https://lite-farm.atlassian.net/browse/LF-2234) for migrating to pnpm workspaces. Any thoughts on this?

Duncan-Brain:
Node version 👍 

Nodemon -- right it would need to have the build command first then it would change to something using the `/dist` right?  like `npx nodemon --inspect=0.0.0.0:9229 dist/src/jobs/index.js`

Tsconfig - a) no worries about figuring it out now! b) I asked because like the node version it could change without us knowing -- just nice to be in charge of base config and reference where needed maybe -- open to either way though

Typescript version - a) I don't know how much work that would be so maybe -- the suggestion to make a prequel PR with just node and typescript might help separate things nice but no obligation just things I am noticing b) pnpm workspaces -- Sounds interesting -- I think we were looking at migrating back to `npm` lol... based on a suggestion by LiteFarm Alumni Jimmy he shared this - https://x.com/ruyadorno/status/1621186886470361093?t=qQbp9OVEKTBpdmWkmpGBOQ&s=19  either way probabaly out of scope for the typescript for backend improvement

navDhammu:
Nodemon - Yeah it has to be built first then run through the dist folder, or if its for dev only then can be run directly with swc-node.

The reason I suggested workspaces is because its very easy to setup and would fix the typescript version issue for good, among others. The amount of work to make a prequel PR with node and typescript version for every package would be about the same as setting up something like pnpm workspaces. All that's needed is to create a pnpm-workspace.yaml file, and adjust scripts from npm to pnpm, then install typescript in the root and it'll be available for each package in the same version. 

From my experience npm's workspace feature isn't as good compared to pnpm workspace having tried both - I found pnpm to be better in terms of speed and ease of use. npm also has some other issues like being able to import packages you didn't insall, for example, it's possible in npm to do something like `import something from 'package-x'` even if you don't have package-x listed as dependency, simply because package-x was installed by a dependency of some other dependency.

kathyavini (commented):
Thank you so much for your patience with this one @navDhammu!! 🙏 

We wanted release 3.7.0 completed before making any large-scale codebase changes, but with that now finished, we're really excited to get this merged in!

I've updated the node version in the local docker image and tagged it with `:latest`. I think we can keep using a latest tag for simplicity, so could you please update the `docker-compose.yml` file as follows?

- `image: litefarm/node-awscli:latest`

Other than that, I think the only two things we need before merge would be:

1. Removing references to `lts` in favour of `20.17.0` so there is no unanticipated version bumping (I think you and @Duncan-Brain have already discussed)
2. There are some merge conflicts, but I think resolving them should be a simple as regenerating the lockfiles

And please let me know if I've missed something else that is a must-have before merge.

Again thank you so much for your patience and also for putting this together!! 🙏 🙏 🙏

navDhammu:
Hey @kathyavini, thanks for making that change.  Its been a while since I worked on this so I'm trying to refresh my memory, but from what I remember in the discussion with Anto and Duncan is that node and typescript version should be updated and consistent across all packages. I felt like since we’re using a monorepo, we shouldn’t have to be updating each package manually every time there needs to be a version update (or any update that needs to reflect in each package). So I created a [proof of concept](https://github.com/LiteFarmOrg/LiteFarm/pull/3523) for adding pnpm with its workspaces feature that handles this in a more maintainable way. I've updated node to 20.18 in that PR, so please let me know if the team decides to go with that, and if yes, then I can revert changes to node version in this PR and make any other necessary updates.

navDhammu:
Yeah updating TS and node for every package isn't necessary for this one. But it'll be annoying because you would have to remember to change node version to 20 every time when working on the backend, then change it back to 18 for frontend, and also remember to point the vs-code's version of typescript to the package you're working on.  

Regarding the discussion of package manager POC PR - I just think if we go that route, then it makes sense to have that one merged first because it makes this one easier to manage. Otherwise it's just unnecessary work to first manually do the changes here and then have to revert them again once/if the team decides to go with the pnpm workspaces.

kathyavini:
> But it'll be annoying because you would have to remember to change node version to 20 every time when working on the backend, then change it back to 18 for frontend, and also remember to point the vs-code's version of typescript to the package you're working on.

Ah, sorry if I was unclear. We DO need the node and typescript versions matched across the backend and frontend, but for now, this can be done via manually specifying in multiple .nvmrc files (for node, as you already did), and specifying the same version of typescript in pnpm and npm (i.e. bumping the pnpm install up to match backend). I happen to use VSCode set to a higher typescript install than either of those without errors anyway, but @Duncan-Brain had a point about using `.vscode/` to make sure that workspace settings are consistently applied without dev input -- that sounds like a good idea, but I don't think this is a blocking issue, since it's a really simple setting to change for the dev.

The problem with waiting on the other PR is that the team has already gotten on board with the backend conversation to TypeScript, but the question of migrating the package manager is conceptually separate and we don't have team consensus on that yet. This PR has been waiting for a while, and we really do want to get it merged as soon as possible so we can start testing on beta and ironing out any issues. (As an aside, the core team holiday starts tomorrow, so ASAP will mean first thing in January when we return!)

navDhammu:
Ok that sounds reasonable - I can try to do these changes over the winter break so that when the core team is back it can be reviewed right away.

navDhammu:
Hi everyone - sorry I wasn't able to work on this earlier as planned, but I've now made some updates to address the node version and setting the docker image in `docker-compose.yml` to `litefarm/node-awscli:latest` as suggested by @kathyavini  . I resolved some of the merge conflicts but there's still conflicts for the lock files in `/api` and `/webapp` even after I regenerated them. I think it might be because I'm using node v20.17 which is resulting in a slightly different lock-file. Any ideas on how to resolve this?

kathyavini:
@navDhammu 

Ack sorry Nav, I read your question but was caught up in something else and it completely slipped my mind! Sorry 🙇 

> I resolved some of the merge conflicts but there's still conflicts for the lock files in /api and /webapp even after I regenerated them. I think it might be because I'm using node v20.17 which is resulting in a slightly different lock-file. Any ideas on how to resolve this?

I think with a version bump like this, that would be expected. As long as your pnpm version is on latest I think that's totally fine; you can just resolve the conflicts entirely in favour of your branch. It looks like it's just pnpm too for some reason (when I check out your branch locally and test merge integration, I see only one conflict in api lockfile and 1109 in webapp 😂).

navDhammu:
@kathyavini 

I've merged the changes in favour of this branch. Also I was thinking it might be good to add a specific pnpm version to engines in package-json to restrict anyone for running pnpm install on a lower version.

kathyavini (approved):
> Also I was thinking it might be good to add a specific pnpm version to engines in package-json to restrict anyone for running pnpm install on a lower version.

We agree 😅 There was even a [ticket](https://lite-farm.atlassian.net/browse/LF-4652) to do so recently created, but it hasn't made it to the priority queue.

But yeah... I think we're good to merge! I'll just give @antsgar a heads up so she can take a look. Thank you so much, this is so exciting!! 🙏 🎉

navDhammu:
> We agree 😅 There was even a [ticket](https://lite-farm.atlassian.net/browse/LF-4652) to do so recently created, but it hasn't made it to the priority queue.

Thought I'll take this chance to promote my other PR 😄 and mention that https://github.com/LiteFarmOrg/LiteFarm/pull/3523  addresses the pnpm version issue as well.

antsgar:
@navDhammu thank you so much for working on this beast of a PR!! 🙌

PR #3437, opened 2024-09-12
Lf 4426 days of the week component repeat crop plan is not understandable in punjabi or hindi

Duncan-Brain:
**Description**

This PR adapts the component to welcome longer abbreviations by no longer requiring a square button checkbox and allowing varying size.

Screenshot 2024-09-12 at 3 03 40 PM
Screenshot 2024-09-12 at 3 03 18 PM

Jira link:[LF-4426](https://lite-farm.atlassian.net/browse/LF-4426)

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

[LF-4426]: https://lite-farm.atlassian.net/browse/LF-4426?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kathyavini:
Thank you for getting rid of the horrible absolute positioning code!! 🙏 It's much improved and definitely adaptable. I have a minor request and some questions:

1) Request: the border disappears when the checkbox is selected. With the previous layout (distributed across a fixed width) that didn't really matter, but with the current solution there is a slight shifting around of the options depending on what's checked -- can you either move to outlines over border OR preserve a transparent border in the checked state so that there is no layout shift when selecting?

https://github.com/user-attachments/assets/93e3eb3e-7405-447c-8d38-30f04c319157

2) Question: I see the space between items has been decreased, is that intentional?

Original:
Screenshot 2024-09-12 at 12 28 16 PM

New:
Screenshot 2024-09-12 at 12 28 09 PM

I don't mind too much either way, but in terms of Hindi and Punjabi, they probably will break awkwardly onto two lines with the smaller gap so we don't have to tighten it up that much if we don't want to.

Screenshot_20240912-121926

3) You showed us this morning all items taking the width of the longest item. Even if we don't square it up (i.e. do aspect ratio 1/1), do we still want the width calculated like that (all rectangles equal width)? It's much improved in e.g. French, but that's a good example of where I think equal width would look better than variable width:

Original (squished, yuck!)
Screenshot 2024-09-12 at 12 51 41 PM

New -- much better, but I think it would look nicer if all rectangles were equal width? @antsgar curious what you think?
Screenshot 2024-09-12 at 12 51 27 PM

Hindi/Punjabi the variable width looks fine to me, as the difference between the shortest and longest day isn't as dramatic and they aren't bunched in the middle either:

Screenshot 2024-09-12 at 1 07 05 PM

Duncan-Brain:
1. yeah, I wondered if you would notice that! Outline it is!
2. I can't find a good gap spacing using Loics factors of 8px rule and I think it would be trying to "magic number" it -- which I don't like doing. So overall I think the smaller gap looks more professional ... 16px maybe max --- feel free to open something if you want a different spacing.  
3. If you like the equal width it will also require setting that px value as a magic max-width number explicitly based on what works for existing languages max chars. Grid handles that nicer I think but doesn't do overflow as well. And I personally don't love the idea of language specific styling if we can help it. Especially if Loic is just going to redesign this at some point for repeat task flow.

kathyavini:
@Duncan-Brain by fixed width, you mean that the overall width gets fixed and then flexbox handles the even distribution? So it would require passing a different appropriate max width to the the component based on current language, like a map `{ 'fr': '100px', 'en': '80px' }` etc? 🤔

Yeah I guess I also do not want to start creating language-based style maps. I'll just wait to see if @antsgar has a thought about it, but I'm okay with straight variable width then.

Duncan-Brain:
@kathyavini What I meant by width is that I personally had trouble finding a solution that used the intrinsic content width as the equal width for all flex items. I tried working on the flex basis to achieve that following this thread (https://stackoverflow.com/questions/29503227/how-to-make-flexbox-items-the-same-size) but flex basis 0 with flex grow requires some width property of some sort to grow up to "fixed max width", and the flex basis 100% I also couldn't get to work I think commenters on that had the same issue.

There might be more complex react refs to watch the width possible to craft a responsive solution use that but didn't seem worth it for this quick fix...

I would have preferred grid but it doesn't work with fit content according to this thread: https://stackoverflow.com/questions/52764726/css-grid-auto-fit-with-max-content

kathyavini (approved):

PR #3442, opened 2024-09-13
LF-4429 Gender dropdown which by default has prefer not to say selected does not update when changing language

kathyavini:
**Description**

This ticket references the gender options select in the Profile/Account page (Top menu > My info), but the same bug existed on new account creation (non-SSO), which I _believe_ is the only other page in app where we want the translations to shift dynamically based on what is entered in the language dropdown -- in Profile/Account that happens on save, on CreateUserAccount that happens onChange for the language select. Invites were not modified because when sending invites there is no translation behaviour based on languageOptions, and when creating an account from invitation, there is no option to change the language.

In this PR:
- On both pages needing this behaviour I added a useEffect that tracks the `genderOptions`
- I also memoized the gender options array because it had been previously re-generating on every render (incidentally that included on every keystroke of the email input on the login page!) which would have prevented using it in the dependency array
- In CreateUserAccount, moved the default gender value to the form (matching Profile/Account); the way it had been set before it was `undefined` from the form's perspective

Jira link: https://lite-farm.atlassian.net/browse/LF-4429

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/webapp/src/hooks/useGenderOptions.jsx
```diff
@@ -12,17 +12,23 @@
...
 const useGenderOptions = () => {
   const { t } = useTranslation();
 
-  return [
-    { value: 'MALE', label: t('gender:MALE') },
-    { value: 'FEMALE', label: t('gender:FEMALE') },
-    { value: 'OTHER', label: t('gender:OTHER') },
-    { value: 'PREFER_NOT_TO_SAY', label: t('gender:PREFER_NOT_TO_SAY') },
-  ];
+  const genderOptions = useMemo(
```
Duncan-Brain, line 21: Use memo should not be needed as an optimization here -- your solution seems to work without it.

Duncan-Brain (changes requested):
I am suspicious that the genderOption updating is the issue -- when I console log directly in the Controller before the React Select render the options seem to be correct without any of these changes. Even the other useEffect with the language thing that you removed half of in the other PR -- I can delete and see no issues I am not sure why it is necessary if the useFarm entity is changed as a prop.

If you replace the `value={value} ` with `value={genderOptions.find(({ value }) => value === userFarm[userFarmEnum.gender])}` that that'll do it.

packages/webapp/src/hooks/useGenderOptions.jsx
kathyavini: What do you mean by loading a const?

So I didn't initially add it as an optimization, but because of this behaviour in the Profile/Account page (you can test by removing the useMemo from the hook and not altering the other code)

1. Change the language and hit save
2. Immediately (before snackbar completes) change your gender option to something else
3. When your snackbar completes your option is "bumped back" to what it had used to be (this is the useEffect hook running because the genderOptions were re-generated when the snackbar completes)

I guess it made the most sense to me to memoize on the options because... well, the options only need to be regenerated when the language changes. But your comment made me review the React Docs on this one and I have to say I'm actually more confused about this particular case, as it says both:

Screenshot 2024-09-13 at 4 31 28 PM

and my code actually didn't work without it 🤔

In any case it also says this, where the third bullet point is this particular case.

Screenshot 2024-09-13 at 4 31 44 PM

So yeah, in this particular case the useEffect dependency array could be used instead (not empty, but `t` same as the useMemo dependency array) but I thought that third bullet point was somewhat leaning towards keeping it in the useMemo?

Anyway, I'm interested in what you said about useEffects below (if the useEffect code were excluded -- and I think I would prefer to that too! -- I don't think that particular bug would be a problem), but as for the optimization, I really have no idea. As I said in the PR description, the genderOptions array is getting regenerated on _every keystroke_ of the login email field (I suppose we watch that field because we provide error validation onChange? And nevermind how crazy it is that login + account creation are one component!) I agree it can't be an expensive calculation, but... it calls `t`... it can't be totally free, right? 🤔 🤔

kathyavini:
@Duncan-Brain so I totally copied the other `useEffect` when I decided on this pattern and you're saying they're both unnecessary???!

I love the prop route but I think I'm missing something... if I just make the change above, sometimes changing gender options and then changing languages bumps my gender selection to something different. (`getValues()` logged in the component also isn't returning an up-to-date value for gender but I'm not sure if that is an artifact of something else). I can keep digging, but do you see that as well?

This is all in the Profile/Account by the way, I haven't tried the prop in CreateUserAccount.

packages/webapp/src/hooks/useGenderOptions.jsx
Duncan-Brain: Not totally free no but not worth optimizing according to react ahah. We should optimize with intention to improve bad performance not as a preventative... according to react. 

Keystrokes - It sounds to me like the whole component is rendering so the one function is probably least of problems. Could either fix the over rendering or also pass anything as a prop from a less rendered container as an alternative to memorizing.

Duncan-Brain:
K will check on Monday, re language I just thought latest changes hadnt been pulled haha probably needs same solution as gender

Duncan-Brain:
Sorry about the half baked answer -- I think theres something to it based on the fact the correct options are in there and its how react should work. I think the following works -- but if not I approve this pr since yours definitely does work -- react expectations aside! And the language stuff can be done in some other PR.. I still saw the timeout so thought the branch wasnt up to date. Also all on edit profile...
```
default values:[userFarmEnum.gender]: genderOptions.find(
        (opt) => opt.value === userFarm[userFarmEnum.gender],
      )?.value,
...
onChange={(val) => onChange(val.value)}
value={genderOptions.find((opt) => opt.value === value)}
```

Demo since I am not in today much:


https://github.com/user-attachments/assets/8a2a86ad-e55b-436c-a959-5072eb327ad5



Rabbit hole I went down last week trying to figure out why react-hook-form controller wasn't playing nicely even though it is unmounted correctly. At first I checked reset and shouldUnregister (on controller level):
https://stackoverflow.com/questions/67327916/react-hook-form-v7-value-is-not-cleared-after-unmounting-its-registered-compone
Then found this other person using controller with react select to double check our implementation. https://codesandbox.io/p/sandbox/react-select-with-react-hook-form-sc2xz?file=%2Fsrc%2FApp.js%3A56%2C1-57%2C1

kathyavini:
@Duncan-Brain no rush (please do it when you feel better!) but could you open a POC/discussion branch with your solution and we'll take a look with Anto when you're back?

antsgar:
I looked into this! I think there should be an easier way to fix this which is essentially we don't need the value of the dropdown to change when the language changes -- we just need the value to be something representative of what's selected, doesn't need to be translated.

So if we change the `useGenderOptions` hook to return just the keys

```
const useGenderOptions = () => {

  return [
    { value: 'MALE', label: 'gender:MALE' },
    { value: 'FEMALE', label: 'gender:FEMALE' },
    { value: 'OTHER', label: 'gender:OTHER' },
    { value: 'PREFER_NOT_TO_SAY', label: 'gender:PREFER_NOT_TO_SAY' },
  ];
};
```

we can then use React Select's `getOptionLabel` prop to do the actual translation

```
const getGenderOptionLabel = (option) => t(option.label);

<Controller
  control={control}
  name={userFarmEnum.gender}
    <ReactSelect
      label={t('CREATE_USER.GENDER')}
      options={genderOptions}
      onChange={onChange}
      value={value}
      toolTipContent={t('CREATE_USER.GENDER_TOOLTIP')}
      getOptionLabel={getOptionLabel}
    />
  )}
/>
```

`getOptionLabel` will take care of ensuring we're showing the label in the right language, but the actual value and options of the dropdown will never change regardless of the selected language -- and so even if the language changes we won't need to change any of the contents of the select component.

kathyavini:
@antsgar could you double-check that I implemented as you were thinking?

I'm oh-so-slightly nervous to scope for release because I didn't have a chance to go through the existing components which all use `useGenderOptions` (there are five in total it looks like), and I have to head out now. But if we can get Denis to take a look at each, then I think it should be okay.

This will be a new pattern -- all the other ReactSelects option objects are formatted as `genderOptions` used to be, with a value and a translated key. I think using `getOptionLabel` is the better pattern by far, but I don't think we want to have two patterns running in the codebase. Maybe after release (because we do need some fix in today) we can discuss what to do moving forward?

packages/webapp/src/components/Profile/EditUser/index.jsx
```diff
@@ -44,6 +44,8 @@ export default function PureEditUser({
   const adminRoles = [1, 2, 5];
 
   const genderOptions = useGenderOptions();
+  const getGenderOptionLabel = (option) => t(option.label);
+
```
antsgar, line 48: Not in scope for this PR, but in this file there's a `getDefaultGender` function which could be simplified now to simply

```js
const getDefaultGender = () => genderOptions.find(({ value }) => value === userFarm.gender)
```

similarly to what you did in the first component for default values

antsgar:
@kathyavini yeah, this is what I had in mind! I didn't find any other component not listed in this PR using this hook, so should hopefully be fine. You don't need to change this to merge, but one additional improvement would be to have the `useGenderOptions` hook return the `getGenderOptionLabel` function to avoid having to repeat it in multiple files

antsgar (approved):

kathyavini:
> but one additional improvement would be to have the useGenderOptions hook return the getGenderOptionLabel function to avoid having to repeat it in multiple files.

Oh duh 😅 I will open another PR after release and refine with both of your suggestions, thank you 🙏

Duncan-Brain:
Love the idea for removing the hook part with translations from the options (now truly a const) very nice!

PR #3462, opened 2024-09-19
LF-4436 polyfill datetimeformat

antsgar:
**Description**

Uses `formatjs` polyfill functions for `Intl.DateTimeFormat`. `formatjs` correctly supports Punjabi fully.

**Summary of findings** (also added this to the Jira ticket)

There seem to be two different bugs in Chrome for Punjabi in the DateTimeFormat functions:

- Months and days of week strings are returned incorrectly. Months are returned as M1, M2, etc. and days of week are returned in English.
- 
- Despite the above which would mean Punjabi isn’t fully supported, executing DateTimeFormat.supportedLocalesOf(['pa']) returns ['pa'], meaning Chrome is telling us Punjabi is fully supported.

Similar bugs I found in other languages (didn’t find any report for Punjabi in particular):

- https://fotis.xyz/posts/a-tolocalestring-mystery/#a-past-chrome-bug-with-supportedlocalesof
- https://issues.chromium.org/issues/41471281
- https://issues.chromium.org/issues/339666802
- https://issues.chromium.org/issues/40872393
- https://issues.chromium.org/issues/40172622

Solutions explored:

- Replacing all uses of native Intl.DateTimeFormat functions with an external library such as dayjs. 

  - Pros:
    - Fixes the problem for Punjabi and potentially other future languages as well.
    - Allows us to consolidate and replace uses of both moment and Intl functions with a library that’s supported.
  - Cons:
    - Using an external library instead of native functions when our use case is pretty simple and probably doesn’t warrant it.
    - Effort to roll this change back to native functions in the future when Chrome solves its bug.
    - Effort to replace uses of Intl and moment before translations release. 
    - dayjs doesn’t support a format like the one we’re using for the welcome string in its localized formats [Format · Day.js](https://day.js.org/docs/en/display/format#localized-formats) . We could construct it by concatenating week day, day, month, but we would lose the capability to adjust the order of these items depending on the locale (e.g. some languages use month first and others display day first).

- Writing a custom function that replaces the wrong strings (M1, M2, etc) with the right Punjabi words.

  - Pros:
    - Fixes the problem without adding any extra dependency.
  - Cons:
    - Not as quick of a solution as I thought. As I started coding this, I realized there are multiple ways to pass a locale to toLocaleDateString. You can pass a string or an array of locales in the locales argument, or you can pass it as a property in the options argument. Having to cover for all these different use cases ended up feeling like I was re-writing the toLocaleDateString function itself. We also need to manually define a set of rules to replace the wrong portions of the string with what we wanted which seemed pretty brittle. In the future, if there were other languages with minor bugs we’d keep adding to this function and it’d likely grow enough to be a pain to maintain and debug.
    - Because of the above, too much effort for something so hacky.

- Using formatjs which is a library that polyfills Intl functions for unsupported browsers (this PR).

  - Pros:
    - Really easy to implement, just import the library and the locale data from the library and done.
    - Really easy to roll back if Chrome fixes its issues, just remove the imports and done.
    - The locale data might help fix other bugs as well (like the rc-calendar one).
    - Transparent to both users and devs. The functions should behave in the same exact way as the native ones. Devs don’t even need to know this polyfill exists, they can keep on calling toLocaleDateString and other functions as usual.

  - Cons:
    - Adding an extra dependency.
    - formatjs has a function called s`houldPolyfill` which detects if a locale is supported in a browser and polyfills only if it isn’t. Because Chrome is incorrectly saying it supports Punjabi, this function doesn’t work properly for our case, so we need to force polyfill regardless of what it returns. This means that we’ll be using formatjs’s polyfill functions at all times and for all locales, even those that aren’t broken and in browsers that do have the right support.

Jira link:

https://lite-farm.atlassian.net/browse/LF-4436

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


Duncan-Brain:
@antsgar Did you happen to see this error while testing? 
Screenshot 2024-09-19 at 6 33 41 PM

The polyfill worked well for a while but navigating around both @kathyavini  and I started seeing this one and it completely crashes -- wondering if you happend to see it as well?

antsgar:
@Duncan-Brain oof really good catch, thanks!! Nope I did not encounter that, does it happen randomly? Did you notice any patterns like it happening for a particular language? Will test this a bit more today to see if I manage to run into it!

Duncan-Brain:
The only common thing both @kathyavini and I tried was commenting out some language imports. But it's mostly unrecoverable if that was the cause which we aren't certain.

antsgar:
Not sure how this happens spontaneously, but pretty sure it's due to the date being passed to `toLocaleDateString` in the Weatherboard being `undefined`. If I force an `undefined` on the `formatDate` function, I can repro the error message. This likely was also happening without the polyfill as well, except the native function throws a concrete error in this scenario instead of crashing, and so in that case an "Invalid date" message is shown in the app instead of completely breaking it (see this screenshot where I forced the same situation on integration).

Screenshot 2024-09-20 at 2 32 56 PM

Apparently there's a bug in formatjs where this error message should be returned but isn't https://github.com/formatjs/formatjs/issues/3508. Someone even fixed it but I don't think it ever got released.

We should probably be able to avoid the situation by catching the undefined date ourselves before calling Intl, will look into it a bit more.

antsgar:
Update -- the fix actually was released, but it was only added for `polyfill` and not for `polyfill-force`. I might be able to add a PR to their repo to implement the same thing for `polyfill-force`.

antsgar:
Opened a PR on their repo here https://github.com/formatjs/formatjs/pull/4533/files. If it doesn't get merged I'll see if we can add the error handling on our side. As an extra precaution I also added the current date as a fallback on the Weatherboard, but let me know if you still see the same crash happening as I'm testing by forcing it but wasn't able to repro it myself.

kathyavini:
@antsgar I only see the crash after I log out and log back in, and I do still see it with the most recent change. Did you test it with re-log?

It's the selector that's returning `undefined` initially, right? I think it produces the correct date eventually, so crash resolves for me if format date just returns an empty string until that time (why the check in the fall to `formatDate` doesn't work equivalently I'm not really sure)

```js
  formatDate(lang, dte) {
    return dte
      ? new Date(dte).toLocaleDateString(lang, {
          weekday: 'short',
          day: 'numeric',
          month: 'long',
        })
      : '';
  },
  ```

**Comment by antsgar** [comment]
@kathyavini thank you so much for the repro steps, that helped a lot! So the fix didn't work because at the last minute I decided to make it so that the format date function received millis instead of seconds, and with the `* 1000` multiplication the fallback in the component wasn't making any sense and returning `NaN` if the date was undefined. Should be fixed now!

File: packages/webapp/src/containers/WeatherBoard/index.jsx [hunk 1]
```diff
@@ -29,7 +20,7 @@ export default function WeatherBoard() {
   const formattedForecast = {
     humidity: `${t('WEATHER.HUMIDITY')}: ${humidity}`,
     iconName,
-    date: utils.formatDate(language_preference, date),
+    date: utils.formatDate(language_preference, date ? date * 1000 : new Date()),
```
Duncan-Brain, line 23: I think one of these two are throwing a new warning from moment as I don't see it on the base branch.

Screenshot 2024-09-24 at 11 16 19 AM

packages/webapp/src/polyfillDateTimeFormat.ts
```diff
@@ -0,0 +1,12 @@
+// We are polyfilling because Chrome desktop has incomplete support for some of the languages (even if in theory they are supported)
+// Once Chrome starts returning the right thing these imports (and the package) should be removed
+import '@formatjs/intl-datetimeformat/polyfill-force';
+import '@formatjs/intl-datetimeformat/add-all-tz';
+import '@formatjs/intl-datetimeformat/locale-data/en';
+import '@formatjs/intl-datetimeformat/locale-data/es';
```
Duncan-Brain, line 6: If there is any way to future proof this using a vite glob import or something that could be useful to not have to manually update in the future. Not a blocker of course!

packages/webapp/src/polyfillDateTimeFormat.ts
```diff
@@ -0,0 +1,12 @@
+// We are polyfilling because Chrome desktop has incomplete support for some of the languages (even if in theory they are supported)
```
Duncan-Brain, line 1: Missing the copyright for new files.

Duncan-Brain (commented):
It seems to be working well for me! Added a couple notes above but can approve.

A few things I tested:
- Having an unsupported language (setting browser and db language preference to `kn-IN` which can happen) Basically I was worried that by not adding all polyfilled languages we might have an issue for people coming to us with an unsupported system/browser language.  - default to english worked
- commenting out some polyfilled languages - default to english worked

kathyavini:
I think the mix of translated (on Widget) and non-translated (in Document expiry) dates were indeed either an artifact from jumping around branches, or MAYBE even my morning brain looking at this situation??!

Screenshot 2024-09-24 at 12 34 23 PM

(Here it's the location and not the time that is translated!) That would be embarrassing but at 8:30am anything is possible!

So I agree with both points points from discussion, that:
- if this library change could be behaviourally silent, that would be ideally
- BUT ALSO dates falling back to the same language as app (i.e. `en`) makes sense and looks consistent (we could even get a quick eval by Loïc tomorrow morning)

As for the main con of the approach and thank you @antsgar for spelling all of the options out like that!)

> This means that we’ll be using formatjs’s polyfill functions at all times and for all locales, even those that aren’t broken and in browsers that do have the right support.

It IS pretty unfortunate. There is a [code example in the docs](https://formatjs.io/docs/polyfills/intl-datetimeformat#dynamic-import--capability-detection) which makes this behaviour look so tantalizingly modifiable (as if we should be able to write `shouldPolyfill(locale) || (locale === 'pa')`,  but is this a different use of the library than we have set up? -- like having to call `polyfill()` manually each time we use it? If so, I think the current over polyfilling is worth it and at least we know under what conditions we can someday remove it.

antsgar:
@kathyavini I played around with that example a bit and it does allow to import the locales only if necessary -- however, the library code itself also checks for the `shouldPolyfill` function to see if it should replace the native functions (check here https://github.com/formatjs/formatjs/blob/main/packages/intl-datetimeformat/polyfill.ts#L10). So unfortunately if `shouldPolyfill` returns `false` for a language, it'll never overwrite the native functions no matter what we do on our side. The only workaround is forcing the polyfill.

packages/webapp/src/polyfillDateTimeFormat.ts
antsgar: I'm not sure, since in this case we don't want to import all the locale data but only the languages we support, so glob probably wouldn't work? (I haven't used it before so I might be wrong!). We could maybe do a dynamic import with `supportedLanguages`

packages/webapp/src/polyfillDateTimeFormat.ts
antsgar: Ty! Will add 🙂

packages/webapp/src/containers/WeatherBoard/index.jsx
antsgar: This `formatDate` function doesn't use moment so it shouldn't be the source of this warning, right? I'm not using moment at any point in this PR -- we should look into that warning though!

kathyavini (approved):

packages/webapp/src/containers/WeatherBoard/index.jsx
Duncan-Brain: Right wrong spot! 

I think our `util/moment.js` file is where the new Intl.DateTime format interact with moment then maybe:

Screenshot 2024-09-25 at 10 44 27 AM

I didn't see this warning before on base branch to this but .. it is just a warning.

Duncan-Brain (approved):

PR #3466, opened 2024-09-19
Update API test Github Actions Workflow on patch/transations

kathyavini:
**Description**

This is both a test and something we want to do anyway 🙂 After merge, expectation is that all new PRs made against patch/translation will trigger the API test suite (and hopefully pass!)

There is the slightest chance that the API tests will even run on this PR, but I would be a bit surprised 🤔 

Jira link: none

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/.github/workflows/automated_tests.yml
+++ b/.github/workflows/automated_tests.yml
@@ -4,6 +4,7 @@ on:
   pull_request:
     branches:
       - integration
+      - "patch/**"
 
 jobs:
   automated_tests:
--- a/.github/workflows/automated_tests.yml
+++ b/.github/workflows/automated_tests.yml
@@ -8,7 +8,7 @@ on:
 jobs:
   automated_tests:
     runs-on: ubuntu-20.04
-    container: node:16.15.0
+    container: node:18.16.1
     services:
       postgres:
         image: postgres:13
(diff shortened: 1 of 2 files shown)


Duncan-Brain:
@kathyavini  -- I think the default is to check the branch it is on. Is that what you meant?

The error I see in the logs there is that for some reason the package and package.lock are out of sync.

kathyavini:
@Duncan-Brain ah okay, I just did a quick glance and assumed the missing packages were a discrepancy with integration. I don't know why the package and package.lock would be out of synch. `patch/translations` is still building + deploying fine on beta 🤔

Duncan-Brain:
@kathyavini the log mentioned it was specifically `npm ci` that didn't like it. I am guessing we use `npm i` in those other situations.

Maybe beta and deploy should also actually be using `npm ci` based on some listed benefits: https://stackoverflow.com/questions/52499617/what-is-the-difference-between-npm-install-and-npm-ci

kathyavini:
Maybe it would, but in this particular case doesn't it make more sense to have our API tests match deployment and not the other way around?

Duncan-Brain:
I think what I gather is that `npm ci` should provide consistent results whereas `npm i` is inconsistent so matching is not really useful in this case if inconsistent.  

I think we can just try fixing package-lock for this PRs goals but might be good to make an improvement ticket for this topic in the future.

Duncan-Brain (approved):
aws-sdk always be changin' -- test are passing nice!

PR #3473, opened 2024-09-23
Lf 4414 the user is able to create an animal with the incomplete date of birth or date brought in

dariaedicas:
**Description**
The user is able to create an animal with the incomplete date of birth or date brought in. An error message should be displayed. It should describe that this is not right format. You should select the date.
For the input type=date there is no change event when the user enters date partially, the value I receive will always be an empty string, no matter if the field is empty or the date is incomplete. Only when all the day-month-year parts were touched I receive the input and can do the validation.
The workaround would be to check e.target.validity.badInput onKeyUp. 

Jira link:
https://lite-farm.atlassian.net/browse/LF-4414
**Type of change**
 Bug fix 
**How Has This Been Tested?**
UI components visually reviewed on desktop view


packages/webapp/src/components/Animals/DetailCards/Origin.tsx
```diff
@@ -116,14 +136,28 @@ const Origin = ({ t, currency, originOptions, namePrefix = '' }: OriginProps) =>
...
+            return moment(value, 'YYYY-MM-DD', true).isValid()
+              ? isNotInFuture(value)
+              : t('common:INVALID_DATE');
+          },
+        })}
+        onCleared={() => {
+          setIsBirthDateValid(true);
+          clearErrors(`${namePrefix}${DetailsFields.DATE_OF_BIRTH}`);
+        }}
+        onKeyUp={(e: any) => {
```
antsgar, line 157: Not a blocker for merge but if we could annotate these functions properly like so

```suggestion
        onKeyUp={(e: React.ChangeEvent<HTMLInputElement>): void => {
```

it'd be great!

antsgar (approved):
Solution looks good and seems to be working properly, thank you for contributing @dariaedicas! 🚀 I didn't know `onChange` didn't fire for keyboard input on a date field, it's cool to learn that (although inconvenient in this case).

My only suggestion for next time would be to add the context that you commented in the Jira ticket in the PR description as well, it helps reviewers get a notion on the reasoning behind the change and makes reviews easier.

Duncan-Brain (changes requested):
Cool change ! I never heard of e.target.validity.badInput before!

Only issue for me is technically `moment` is deprecated and we prefer not to use it in new code if possible. If too challenging to replace that specific function with a regular Date function I can approve.

packages/webapp/src/components/Animals/DetailCards/Origin.tsx
```diff
@@ -13,19 +13,18 @@
  *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
  */
 
-import { useMemo, useState } from 'react';
+import React, { useMemo, useState } from 'react';
 import { useFormContext } from 'react-hook-form';
 import Input, { getInputErrors } from '../../Form/Input';
 import RadioGroup from '../../Form/RadioGroup';
-import { isNotInFuture } from '../../Form/Input/utils';
+import { isNotInFuture, parseISOStringToLocalDate } from '../../Form/Input/utils';
```
Duncan-Brain, line 20: Looks like parseISO.. is unused.

Duncan-Brain (commented):
That was fast, I am happy that there was something around plain Date that worked out!

Just made one small inline comment.

antsgar (approved):

Duncan-Brain (approved):

PR #3485, opened 2024-10-02
LF-4388 animal details sub components handle edit and readonly modes

kathyavini:
**Description**

This adds two more modes (`readonly` and `edit`) to the subcomponent cards that display animal details (the original has been designated `add`). This PR also:
- Adds disabled states (for `readonly` mode) to custom form elements that were missing them:
  - [SexDetails](http://localhost:6006/?path=/story/form-sexdetails--disabled)
  - [ImagePicker](http://localhost:6006/?path=/story/components-imagepicker--with-default-url-disabled)
  - Note: still waiting to hear back from Loïc whether there is a visually distinct form of the upload button itself (when the user has not yet added an animal image), as there is not one on Figma. 
- Migrates the type and breed selects of the General Card to the dedicated components that were used in AddAnimalsFormCard (AnimalTypeSelect/AnimalBreedSelect)
- Adds breed-clearing on type change functionality to the General Card, and migrates that behaviour from a useEffect to an onChange in both the General Card and the Basics Card (= AddAnimalsFormCard), as the onChange is a cleaner solution
- Updates the Storybook stories + adds mock data to represent a selected animal:
  - [General](http://localhost:6006/?path=/docs/components-addanimalsdetails-general--docs)
  - [Origin](http://localhost:6006/?path=/docs/components-addanimalsdetails-origin--docs)
  - [Other](http://localhost:6006/?path=/docs/components-addanimalsdetails-other--docs)
  - [Unique](http://localhost:6006/?path=/docs/components-addanimalsdetails-unique--docs)

Jira link: https://lite-farm.atlassian.net/browse/LF-4388

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


SayakaOno (dismissed):
Looks great!!

Just a small thing I noticed: the sex details input gets highlighted when clicked, even though it's disabled.
Screenshot 2024-10-03 at 11 26 02 AM

packages/webapp/src/components/Animals/AddAnimalsFormCard/AddAnimalsFormCard.tsx
```diff
@@ -111,6 +101,7 @@ export default function AddAnimalsFormCard({
         onTypeChange={(option) => {
           trigger(`${namePrefix}${BasicsFields.TYPE}`);
           onTypeChange?.(option);
+          breedSelectRef?.current?.clearValue();
```
antsgar, line 104: We possibly don't need a ref here at all, 

```js
resetField(`${namePrefix}${BasicsFields.BREED}`, { defaultValue: null });
```

seems to do the trick.
Also, not related to this PR, but do we need a ref for the UUID value? Couldn't we set it in the `useEffect` hook without storing it in a ref?

packages/webapp/src/components/ImagePicker/index.tsx
```diff
@@ -42,6 +43,7 @@ type CommonProps = {
   label?: string;
   optional?: boolean;
   defaultUrl?: string;
+  disabled?: boolean;
```
antsgar, line 46: Should we call this `isDisabled` to make it consistent with the other changes?

packages/webapp/src/components/Animals/DetailCards/General.tsx
```diff
@@ -125,7 +148,16 @@ const GeneralDetails = ({
   }, [animalOrBatch, t, sexOptions, control, watchBatchCount]);
 
   return (
-    <div className={styles.sectionWrapper}>
+    <div className={clsx(styles.sectionWrapper, mode === 'edit' && styles.edit)}>
+      {(mode === 'readonly' || mode === 'edit') && (
+        <>
+          {/* @ts-ignore */}
```
antsgar, line 154: I'm not getting any errors when removing this, do we need it?

antsgar (approved):
Couple of minor comments but code's looking great! 🔥 🔥

packages/webapp/src/components/Animals/AddAnimalsFormCard/AddAnimalsFormCard.tsx
kathyavini: Ah thank you!! I remember we concluded at one point that ReactSelects were impervious to hook form functions, but it was based on `setValue` not working, and I don't think I ever tried `resetField` before.

The UUID ref was meant for persisting the uuid between the component unmounting and remounting... but of course what also persists between remounts is... the form 😁🤦‍♀️ I have removed both refs.

packages/webapp/src/components/ImagePicker/index.tsx
kathyavini: Do you know if there is a rule we can follow here?

I assumed `<Input />`, `<NumberInput />` , and `<RadioGroup />` all use `disabled` because they pass that prop eventually to a real html `<input />` (as does this component), and the ReactSelects use `isDisabled` because that's what the underlying ReactSelect expects.

I was trying to keep that, but I see now that we have about an equal amount of `disabled={isDisabled}` and `isDisabled={disabled}` in the codebase; so we appear all mixed up 😜

kathyavini:
Sorry to be slow to get back to this branch! I think I have addressed all comments above.

@SayakaOno that focus border was the behaviour of `<InputBaseField />` so I have updated that... but actually I'm totally stumped why `<LockedInput />` doesn't show the same behaviour... they both pass `disabled` to `<InputBase />` and don't apply any other styles to the border 😳

packages/webapp/src/components/ImagePicker/index.tsx
antsgar: That seems like a good way to reason it! There's no rules per se, and tbh I've seen similar mixes in every codebase I've seen 😅 I personally always prefer to do `is...` for boolean variables because I think it makes it clearer that they are booleans, but that's my preference (and I've probably not even followed my own rule more than once)

antsgar (approved):

packages/webapp/src/components/Form/SexDetails/index.tsx
```diff
@@ -71,8 +73,12 @@ export default function SexDetails({
         error={total > maxCount ? t('ADD_ANIMAL.SEX_DETAIL_ERROR', { count: maxCount }) : undefined}
         showResetIcon={false}
         rightSection={<ChevronDown className={styles.chevronDownIcon} />}
+        disabled={isDisabled}
         mainSection={
-          <TextButton onClick={(e) => setAnchor(e.currentTarget)} className={styles.button}>
+          <TextButton
+            onClick={isDisabled ? () => {} : (e) => setAnchor(e.currentTarget)}
+            className={styles.button}
```
SayakaOno, line 80: Thank you for looking into the focus style!
I was curious about what was going on and looked into the code.
This TextButton is still being focused while `InputBase` itself is disabled, so adding `disabled={isDisabled}` here should fix the behaviour without the CSS!

I'll approve but won't merge in case you want to make the change :)

SayakaOno (approved):

packages/webapp/src/components/Form/SexDetails/index.tsx
kathyavini: Thank you so much for investigating that!! Yes I definitely prefer that and will push it now 🙏

SayakaOno (approved):

PR #3486, opened 2024-10-02
LF-4374 Badge component

gursimran-singh:
**Description**

A reusable badge component that shows the content when onHover - desktop, onClick - mobile, desktop. Usage: 
`<Badge title='Beta' content='This is the content which will be displayed onHover - desktop, onClick - mobile, desktop' />`

Jira link: https://lite-farm.atlassian.net/browse/LF-4374

**Type of change**

**How Has This Been Tested?**

- [x] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/webapp/src/components/Badge/index.tsx
```diff
@@ -0,0 +1,100 @@
...
+  const popperProps = {
+    modifiers: [
+      {
+        name: 'offset',
+        options: {
+          offset: [0, -10],
+        },
+      },
+    ],
+  };
```
antsgar, line 63: This offset is causing the popover to overlap with the tooltip contents, which causes the hover styling not to work properly for a small portion of the tooltip when it's open. To get rid of the spacing between the popover and the tooltip, I'd suggest replacing this with the margin change alternative documented here in MUI docs https://v5.mui.com/material-ui/react-tooltip/#distance-from-anchor (where it says "Alternatively, you can use the slotProps prop to customize the margin of the popper."). According to the designs, there should still be a subtle margin between the badge and the popover, of 2px.

packages/webapp/src/components/Badge/styles.module.scss
```diff
@@ -0,0 +1,47 @@
...
+
+}
+.badge {
+  //color: #E8A700;
+  color: var(--Colors-Accent-Accent-yellow-600);
+  background: #FFFBF2;
+  border-radius: 16px;
+  height: 24px;
+  font-size: 14px;
+  padding: 4px;
```
antsgar, line 32: NIT, but I think the left padding should be `6px`

antsgar (commented):
@gursimran-singh thank you so much for working on this, we really appreciate it. It looks great! 🙌

I left a couple of comments throughout, and additionally:
- In the designs, when the popover is open, the badge's bottom border has a smaller border radius (half of the normal radius) to sort of align with the popover.
- The ticket description includes adding the Beta badge to different Animals components, were you planning on tackling that in a separate PR? If you can't get to it it's no problem, we can split it up to a separate ticket.

gursimran-singh:
@antsgar 

I will do the mentioned changes. And regarding adding Beta badge to different Animals components, to what i remember when having discussion with team is that these can be added later(i might have got it wrong). I will add the Badge component to required places and will get in touch if i have any questions/issues doing that. Thanks.

packages/webapp/src/components/Badge/index.tsx
```diff
@@ -42,33 +47,32 @@ const Badge: React.FC<BadgeProps> = ({ content, title = 'Beta', showIcon = true
...
+        e.stopPropagation();
+        handleTooltipContentHover(true);
+      }}
       onMouseEnter={() => handleTooltipContentHover(true)}
       onMouseLeave={() => handleTooltipContentHover(false)}
       className={styles.tooltip}
-    >
-      {content}
-    </div>
+      dangerouslySetInnerHTML={{ __html: content }}
```
antsgar, line 57: Could you explain the reason behind this change? Instead of setting the link as HTML within the content string, I think we should keep this as is and set the content up as a React Node including the <a> tag

packages/webapp/public/locales/en/translation.json
```diff
@@ -2059,5 +2059,11 @@
   },
   "YEAR_SELECTOR": {
     "TITLE": "Select Year"
+  },
+  "BADGE": {
+    "BETA": {
+      "TITLE": "Beta",
+      "CONTENT": "Beta version: A pre-release version for testing and feedback. We're excited to bring you a sneak peek of our new animals management feature, now in beta! As we embark on this beta journey together, your insights are more valuable than ever. While we've worked hard to bring you this feature, the road to perfection is paved with feedback - and sometimes, a few bumps! If you encounter any quirks or bugs, please let us know. Your reports are crucial in helping us refine and improve. Thank you for your patience and for being an integral part of our community's growth.You can learn more about beta <a href=\"#\">here</a>"
+    }
```
antsgar, line 2067: There's a missing space right before "You can..."

antsgar (commented):
Looks great! Not a blocker for merging (we can fix later), but on second thought I think on desktop the tooltip content should appear/disappear on click instead of on hover. It seems strange to me that the badge changes appearance on click and when clicking outside of it, but the tooltip interaction happens on hover. I'll double check with design tomorrow though.

On the other hand, I think there's a different design for the tooltip on mobile, which is full screen. I'll double check this tomorrow as well and let you know!

packages/webapp/src/components/Badge/index.tsx
gursimran-singh: My intentions when developing the component was to make it reusable with different title and content in any other section. So, do not want to hard code anything in the component itself. I will change it to use <Trans> component from 'react-i18next' package(which i found later), is safer way to handle these situations and set the link when passing the content to the <Badge> component. 

Also, it would be nice if you can tell the link address, then i will update that as well.

gursimran-singh:
I had discussion with team and showed them the functionality and they were ok with it. Only change was that the badge in the menu items should no show the content popup in any case. 

` I think there's a different design for the tooltip on mobile, which is full screen`
 We had a discussion on this as well, showing on full screen is a part of spotlight feature and in mobile content should only be shown on click as we do not have hover there.

Do let me know if there are changes and i will do it accordingly.

packages/webapp/src/components/Badge/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ */
+
+import React, { useState } from 'react';
+import { Tooltip, ClickAwayListener, IconButton, tooltipClasses } from '@mui/material';
+import InfoOutlinedIcon from '@mui/icons-material/InfoOutlined';
+import styles from './styles.module.scss';
+import { Trans } from 'react-i18next';
```
Duncan-Brain, line 20: This seems to be unused in this file ... I thought pre-commit hooks would catch this though!

packages/webapp/src/components/Badge/styles.module.scss
```diff
@@ -0,0 +1,73 @@
...
+  @include xs-breakpoint {
+    height: 16px;
+    position: relative;
+    font-size: 14px;
+    top: initial;
+    left: initial;
+    padding: 6px;
+    margin-right: 6px;
+  }
+}
```
SayakaOno, line 73: Would it be a bad idea to pass a className with these styles to the Badge component instead?
I've never seen a raw CSS className used here before, so it would be nice if we could keep things consistent!

packages/webapp/src/components/Badge/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+// Define the Props
+interface BadgeProps {
+  content?: string | React.ReactElement;
+  title: string;
+  showIcon?: boolean;
+  style?: React.CSSProperties;
+}
+
+const Badge: React.FC<BadgeProps> = ({ content = '', title, showIcon = true, style }) => {
```
SayakaOno, line 30: We're aiming to avoid inline CSS, so could you please change the `style` prop to `className` instead? 🙏

packages/webapp/src/components/Badge/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+  };
+
+  const tooltipContent = (
+    <div
+      onClick={(e) => {
+        e.stopPropagation();
+        handleTooltipContentHover(true);
+      }}
+      onMouseEnter={() => handleTooltipContentHover(true)}
+      onMouseLeave={() => handleTooltipContentHover(false)}
```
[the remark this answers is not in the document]
gursimran-singh: The reason for this is the tooltip should still be open if the user click inside the open box. Without this, it closes immediately if clicked inside the box because `ClickAwayListener` is listening to all the click event happened outside of the button and will trigger a function which closes the tooltip.

packages/webapp/src/components/Badge/index.tsx
gursimran-singh: Its good we are avoiding the inline css, i have seen a lot here. And sure, i will update accordingly.

packages/webapp/src/components/Badge/styles.module.scss
gursimran-singh: It is one of the way by which we can access the parent class in the child component. But, i will see these styles can be attained by passing class to the component and also avoiding the inline css.

packages/webapp/src/components/Badge/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+const Badge: React.FC<BadgeProps> = ({ content = '', title, showIcon = true, style }) => {
+  const [open, setOpen] = useState<boolean>(false);
+  const [tooltipHover, setTooltipHover] = useState<boolean>(false);
+
+  const handleTooltipOpen = () => {
+    if (!tooltipHover) {
+      setOpen(true);
+    }
+  };
+  const handleTooltipClose = () => {
```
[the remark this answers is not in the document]
Duncan-Brain: Honestly its mostly fine the way you have it as an only onHover on desktop and onClick on mobile. 

But specifically the style of the button on desktop should not change onClick on desktop though as shown in the first video... unless the full click behaviour opens the tooltip. So either we need to disable click styles on desktop, or add the full onClick open behaviour.  I would personally approve either of those two options.

I will tag Loic on the Jira ticket to hopefully get you clarification on this though so you feel confident. Sorry about that lack of clarity -- please always feel free to tag Loic with your UI/UX questions on the Jira ticket too!

packages/webapp/src/components/Navigation/useSectionHeaders.tsx
```diff
@@ -48,7 +48,7 @@ export function useSectionHeader(path: Pathname): string | React.ReactElement |
     <>
       {t('SECTION_HEADER.ANIMALS_INVENTORY')}
       <Badge
-        style={{ marginLeft: '10px' }}
+        className="positionRight"
```
antsgar, line 51: Since the styles are contained in the child component instead of being part of the parent, here I'd instead pass a "position" property set to either left or right, and then in the `Badge` component I'd set the appropriate class names doing something like `className = styles[position]` with .left and .right classes added to styles, or `className = position === 'left' ? styles.positionLeft : stylesPositionRight`.

Usually when a component accepts a `className` prop as shown here, it means that the parent can pass a className pointing to styles specified in the parent itself, so basically the parent can "customize" the way the child component looks. If we attempted to do this with the way this is set up it wouldn't work because Badge is looking for this class name in its own styles file. Hopefully that makes sense! Not a blocker for merging though.

gursimran-singh: Initially, i coded it that way, not sure why i went with more complex solution. I was trying to handle two use case(menu-item and positioning) with one parameter. Thanks for correcting me though.

antsgar (dismissed):
Looks great, thank you so much for contributing! 🙌 

@Duncan-Brain @SayakaOno if you could prioritize reviewing this one so we can start completing the beta status story, that'd be great 🙏

packages/webapp/src/components/Badge/index.tsx
```diff
@@ -83,7 +90,12 @@ const Badge: React.FC<BadgeProps> = ({ title, content = '', showIcon = true, cla
         slotProps={slotProps}
       >
         <IconButton
-          className={clsx(styles.badge, focus && styles.focus, className ? styles[className] : '')}
+          className={clsx(
+            styles.badge,
+            focus && styles.focus,
+            !isMenuItem && styles[position + 'Position'],
```
Duncan-Brain, line 96: cool!

packages/webapp/public/locales/en/translation.json
```diff
@@ -2061,5 +2061,11 @@
   },
   "YEAR_SELECTOR": {
     "TITLE": "Select Year"
+  },
+  "BADGE": {
+    "BETA": {
+      "TITLE": "Beta",
+      "CONTENT": "Get a sneak peek at our new animal management feature—now in beta! Your feedback is crucial as we refine this feature. Please share any bugs or suggestions you encounter. We appreciate your help in improving liteFarm! You can learn more about beta here <a>here</a>"
```
Duncan-Brain, line 2068: Small thing -- there are two "here" at the end of the sentence.

Duncan-Brain (commented):
Looks great! Thanks for your patience with us figuring out the behaviour.

I added some suggestions for typos, but otherwise I approve!

antsgar (approved):

PR #3489, opened 2024-10-04
LF-4378 Readonly/Edit animal details framework - add form framework for animal details

kathyavini:
**Description**

This PR adds the form framework for `<SingleAnimalView />` with placeholders for the remaining tickets (header component and form container, which will populate the form with data). This ticket lays out the page, hooks it up to the inventory, and controls the readonly/edit toggle.

Notes:
- Following @antsgar's suggestion, the readonly + edit modes have been added right into `<MultiStepForm />` (now called `<ContextForm />` and `<WithStepperProgressBar />`. So the components used are actually identical to the AddAnimals case 👍
- There is no URL change between readonly + edit (as discussed with @Duncan-Brain and @SayakaOno on Oct. 4) and so the cancel modal is also now triggered manually via component in addition to the `transition.unblock()` behaviour from before
  - I converted `<ContextForm />` to (pretty loosely typed) TypeScript. After the fact I did wonder if we will need it in JS if we ever start transitioning over the old forms -- but for now as we have been using it in new forms, I think this will be better for our purposes.
- animals and batches do not have different kinds of URLs, but rather are all `/animals/{internal_identifier}` under this pattern. I have updated the `path` property in Inventory to reflect this

DID NOT:
- Add the wording for the cancel modal. There is a [suggestion from Loïc](https://www.figma.com/design/eMffC0h3JPlWtzNttxJpil/Litefarm-Updated?node-id=20375-159081&node-type=frame&m=dev) that would involve reworking the modal string logic. We have a ticket to do so anyway so (from the Language Epic) so I think it should be done then

Jira link: https://lite-farm.atlassian.net/browse/LF-4378

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


antsgar:
This looks like it's on a great track! 🔥

I could be mistaken, but I think it could be further simplified by baking in the "editing" logic into the MultiStep variant and accounting for the single step scenario in that component as well. In that way, we'd be making that variant more flexible and we'd be able to consume it without having to think about which modes we want to allow or how many steps we have.

I also agree that the state handling for the confirmation modal is currently a bit convoluted, where in the current single step variant where the parent component triggers a `checkIsFormDirty` state update, passes that prop down to the form component and that triggers an effect in the form component to in turn update a state variable to show the confirmation modal. Since the current form component already triggers the transition block and shows the modal when the form is dirty, shouldn't that be enough?

I pushed a branch `LF-4378-readonly-edit-animal-details-framework-add-form-framework-for-animal-details-2` where I tested out what I describe above -- I tested it briefly but it might not be fully functional, just a POC. I might also be missing something so it's definitely up for discussion, no need to use that.

kathyavini:
@antsgar I love re-using `<WithStepperProgressBar />`! (And I think it can be safely renamed back to original this way with that `isSingleStep` property).

> Since the current form component already triggers the transition block and shows the modal when the form is dirty, shouldn't that be enough?

So the issue is that the current form triggers the modal the when the browser URL changes. The reason your branch triggers the modal on canceling the Edit is because this code was removed in the container:

```ts
  // Override the onCancel in ContextForm because we don't want to navigate away
  const onCancel = () => {
    setCheckIsFormDirty(true);
  };
 ```

and the default `onCancel()` in `<ContextForm  />` triggers `browser.back()`. That means cancelling the editing state returns the user to the Inventory (which does on the plus side, trigger the modal), while we need it to just return the user to `readonly` mode -- but without a URL change! (No URL change because we don't want the user to be able to then go back to the edit state using the back button). This was the reason for the `checkIsFormDirty` state & prop. We could remove it if we were okay with cancelling edit sending the user back to the inventory, but I don't think that is correct UX 🫤

**Comment by antsgar** [comment]
@kathyavini ah, that makes total sense! You're 100% right, a URL change wouldn't be correct UX in this case. In that case, I'd do something similar to what you had before, but what if instead of triggering the modal manually from outside the form we assumed this to be the desired behavior for any editable forms? -- that is, if you are exiting editing mode, then you'd want to display the confirmation modal. So we'd let the form components itself handle what to do with the cancel flow in the case where the form is editable. I pushed another POC of this to a branch `LF-4378-readonly-edit-animal-details-framework-add-form-framework-for-animal-details-3` (bear in mind I commited with `--no-verify` because I had a TS warning)

**Comment by antsgar** [comment]
@kathyavini I just noticed you pushed some more changes an hour ago and I did my branch before then so it's probably outdated now! But maybe it helps anyway, just FYI it's branched off from commit [3e66175](https://github.com/LiteFarmOrg/LiteFarm/pull/3489/commits/3e66175d8e6003daceca8922fed4f44030ed70b9)

**Comment by kathyavini** [comment]
@antsgar okay we did kind of similar things but I liked the use of the `onCancel()` more than a `useEffect()` in the variant, so I went with that! I think it might fix the remounting problem too! 🤞🙏 

The only thing I changed is making the `setIsEditing` prop the conditional, because I think for the old form (AddAnimals) to work and display the form nav buttons, the form has to be `isEditing=true` by default.

> (bear in mind I commited with --no-verify because I had a TS warning

Was it on passing `setIsEditing`? 😂 That is what annoyed me into converting the component into TS...!

**Comment by antsgar** [comment]
@kathyavini oooh why does the other form break if we don't default `isEditing` to `true`? Isn't that a bit confusing since that form isn't actually editable?

**Comment by kathyavini** [comment]
@antsgar it just won't show the form navigation buttons! So then you can't go forward or submit the form.

Yeah it was a bit unintuitive for me too, but if you think about it another way, a regular form IS editable by default 😂  What it doesn't have is a readonly state.

**Comment by antsgar** [comment]
@kathyavini ah you're right 😂 that makes sense! So thinking about that I think it might be a little bit more readable like so

```js
  const onCancel = () => {
    if (form.formState.isDirty) {
      setShowCancelFlow(true);
    } else {
      if (isEditing && setIsEditing) {
        setIsEditing(false);
      } else {
        history.back();
      }
    }
  };
```js

The first if will trigger the cancel flow on all forms except if on readonly mode (there's no way the form would be dirty if it's in readonly mode). And then the second portion after the else will check if it needs to get out of edit mode, and if not it will just go back to the previous page. Gets rid of the `return` before `history.back` that I had added -- it's essentially the same code though so you don't have to change it if you prefer it the current way!

Oh and yes, the TS warning was due to `setIsEditing` 😅

**Review by antsgar — DISMISSED** [review]

File: packages/webapp/src/components/Form/ContextForm/WithStepperProgressBar.tsx [hunk 1]
```diff
@@ -108,17 +118,26 @@ export const WithStepperProgressBar = ({
     } catch (e) {
       console.error(`Error during canceling ${cancelModalTitle}: ${e}`);
     }
+    setIsEditing?.(false);
+  };
+
+  const handleDismissModal = () => {
+    setTransition({ unblock: undefined, retry: undefined });
+    setShowCancelFlow?.(false);
+    setIsEditing?.(true);
```
SayakaOno, line 127: Are there any situations where `handleDismissModal` is called while `isEditing` is `false`?

SayakaOno (commented):
Wow, this is impressive! Looks great!!! ❤️

packages/webapp/src/components/Form/ContextForm/WithStepperProgressBar.tsx
kathyavini: Ah no, thank you. That is a holdover from a previously implementation of `onCancel()` that flipped the boolean first but I don't Anto's implementation of `onCancel` could ever get to that state; I'll remove it! 🙏

SayakaOno (approved):
❤️

PR #3204, opened 2024-05-02
LF-4219: Create component for weight application rate tabs

SayakaOno:
**Description**

- Update `StateTab` component to
  - take `variant`.
  - replace `Semibold` with `button` so that tabs become focusable. (I plan to use `:focus-within` to style cards when focused, which will require this change. The outline is "none" for now.)
- Update `RoterTab` to be compatible with the new CSS.
- [Stories](http://localhost:6006/?path=/docs/components-routertab-statetabs--docs)

Jira link: https://lite-farm.atlassian.net/browse/LF-4219

**Type of change**

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

**How Has This Been Tested?**

- [ ] Passes test case
- [x] UI components visually reviewed on desktop view
- [x] UI components visually reviewed on mobile view
- [ ] Other (please explain)

**Checklist:**

- [ ] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [x] The precommit and linting ran successfully
- [ ] I have added or updated language tags for text that's part of the UI
- [ ] I have added "MISSING" for all new language tags to languages I don't speak
- [x] I have added [the GNU General Public License](https://lite-farm.atlassian.net/l/cp/BT0Dd7WW) to all new files

PR #3490, opened 2024-10-04
LF-4382: Tabs component - Create component for handling tabbed views

SayakaOno:
**Description**

- Create `Tab` presentation component by extracting the shared UI logic from `RouterTab` and `StateTab`.
   - Each `tab` item can now have an optional `format` function, allowing the rendered tabs to include icons or styled text.
   - Add styles for the `underline` variant. (reused some code from the closed tab PR https://github.com/LiteFarmOrg/LiteFarm/pull/3204)
   - (Previously, there was no UI difference between the two components, making it difficult to maintain or update their look.)
- Refactor `RouterTab` and `StateTab` components and migrate them to TypeScript.
   Both now utilize the `Tab` component, passing in `onClick` and `isSelected` functions.
- Rename `RouterTab` stories to `Tab` stories and add more [stories](http://localhost:6006/?path=/story/components-tab--underline).

Please let me know if I'm overcomplicating the `Tab` type 🙏 I'm not sure if it needs to be generic.

Jira link: https://lite-farm.atlassian.net/browse/LF-4382

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/webapp/src/components/RouterTab/Tab.tsx
```diff
@@ -0,0 +1,64 @@
...
+ *  LiteFarm is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ */
+import { ReactNode, CSSProperties } from 'react';
+import clsx from 'clsx';
+import styles from './styles.module.scss';
+
+export enum VARIANTS {
```
antsgar, line 19: NIT, but I think usually the convention for TS enums is to name them in singular and with PascalCase, so in this case `enum Variant`

packages/webapp/src/components/RouterTab/Tab.tsx
```diff
@@ -0,0 +1,64 @@
...
+  PILL = 'pill',
+  UNDERLINE = 'underline',
+}
+
+export type BaseTab = {
+  label: string;
```
antsgar, line 25: I noticed in the designs sometimes we have something else in the tab other than the label (e.g. in the Tasks tab there's a counter next to it), not a blocker at all but should we allow passing a React element here on top of a string to account for that scenario?

antsgar (approved):
Tabs are looking awesome! 🚀

packages/webapp/src/components/RouterTab/Tab.tsx
SayakaOno: Oh, thank you so much!! I'll update it!

packages/webapp/src/components/RouterTab/Tab.tsx
SayakaOno: I originally considered making `label` a React element as well, but I noticed that in both `RouterTab` and `StateTab`, the label is currently used as part of the `id` (`id={tab.label + index}`). To avoid potentially breaking this, I kept the `label` as a string and added a `format` function (on the next line) to include additional elements like counters or icons. (You can see an example of this in the `FormattedLabel` story.)

In such situation, I'm not really sure if I should:
1. take the time to make sure if it's safe to change `id` and make the `label` React element.
2. extend the feature without making the change to what we have.

Any guidance on which direction would be best?

antsgar: For some reason I totally missed the formatted label example -- that's perfect as is! It's even better than making label more flexible, it gives the option to customize the display of the tab without altering the label itself. Love that solution!

antsgar (approved):

Duncan-Brain (approved):
Love it, think it works great, think it looks great!

Tab type - yeah it is complex as you mention, I don't know enough to help -- only thing I can think of is it looks defined from the bottom up (Tab > RouterTab) rather than top down (RouterTab > Tab)  -- but I think you did it right! 🤷 

Only other thing not related to your PR --  it is weird that `/StateTab` folder is nested under `/RouterTab` folder. ....It would ruin your nice clean diff but I would support renaming it `/Tab` and reorganizing contents.

SayakaOno:
Thank you @Duncan-Brain!
I really don't like the folder structure either, it doesn't make sense... I'll create a tech debt ticket!

PR #3496, opened 2024-10-11
Fix layout

SayakaOno:
**Description**

This PR makes the `mainColumn` fully responsive to the browser size, preventing it from pushing the side menu.

Here’s the problem:

https://github.com/user-attachments/assets/99af173a-b018-4cec-82a2-fb88689de4a1

I learned about the solution while struggling with truncating text properly for the animal header ([LF-4381](https://lite-farm.atlassian.net/browse/LF-4381)). (It took me a few days to realize that `mainColumn` was the element that needed this fix 😭)

This article was very helpful for understanding the purpose of `min-width: 0`!
https://css-tricks.com/flexbox-truncated-text/#aa-the-solution-is-min-width-0-on-the-flex-child

Jira link: N/A

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

[LF-4381]: https://lite-farm.atlassian.net/browse/LF-4381?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/webapp/src/components/Navigation/styles.module.scss
+++ b/packages/webapp/src/components/Navigation/styles.module.scss
@@ -21,4 +21,5 @@
   min-height: 100vh;
   min-height: -webkit-fill-available;
   overflow: clip;
+  min-width: 0;
 }


antsgar (approved):
Thanks for linking the article, super useful!

Duncan-Brain (approved):
Approving !

I don't know the answer but I have this question:

In the article it mentions setting min-width to some value (or 0). Are there any other values for min-width that make sense here? Sometimes it is better to push stuff aside or make a horizontal scroll rather than unusable/unreadable flex items

SayakaOno:
@Duncan-Brain I didn't know there was a related ticket, thank you!! I'll let Denis know!

Regarding the min-width, we currently don't have a specific minimum screen size we support, so... should we discuss with Loïc? I think it's around 350px, though.

Duncan-Brain:
I don't know if we need to ...unless you are interested or think it will help in the future to know when it is acceptable for UI to break. Up to you!

PR #3501, opened 2024-10-16
LF-4470: Write database migration for animal movement task

SayakaOno:
**Description**

Create migration for animal movement task.
- Create `animal_movement_task` table
- Create `animal_movement_task_purpose_relationship` table
- Create `animal_movement_purpose` table
- Create `animal_movement_task_animal_relationship` table
- Create `animal_movement_task_animal_batch_relationship` table
- Add `ANIMAL_MOVEMENT` type to `task_type` table
- Add `location_id` column to `animal` and `animal_batch` tables

Jira link: https://lite-farm.atlassian.net/browse/LF-4470

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/api/db/migration/20241016204145_create_animal_movement_task_tables.js
+++ b/packages/api/db/migration/20241016204145_create_animal_movement_task_tables.js
@@ -20,7 +20,7 @@
 export const up = async (knex) => {
   // Add new animal_movement entry to task_type table
   await knex('task_type').insert({
-    task_name: 'Movement Task',
+    task_name: 'Movement',
     task_translation_key: 'MOVEMENT_TASK',
   });
 
--- a/packages/api/db/migration/20241016204145_create_animal_movement_task_tables.js
+++ b/packages/api/db/migration/20241016204145_create_animal_movement_task_tables.js
@@ -80,22 +80,12 @@ export const up = async (knex) => {
   await knex.schema.createTable('animal_movement_task_animal_relationship', (table) => {
     table.integer('task_id').references('task_id').inTable('task').notNullable();
     table.integer('animal_id').references('id').inTable('animal').notNullable();
-    table.string('created_by_user_id').references('user_id').inTable('users');
-    table.string('updated_by_user_id').references('user_id').inTable('users');
-    table.dateTime('created_at').notNullable();
-    table.dateTime('updated_at').notNullable();
-    table.boolean('deleted').defaultTo(false);
   });
 
   // Create animal_movement_task_animal_batch_relationship table (similar to above but for batches)
   await knex.schema.createTable('animal_movement_task_animal_batch_relationship', (table) => {
     table.integer('task_id').references('task_id').inTable('task').notNullable();
     table.integer('animal_batch_id').references('id').inTable('animal_batch').notNullable();
-    table.string('created_by_user_id').references('user_id').inTable('users');
-    table.string('updated_by_user_id').references('user_id').inTable('users');
-    table.dateTime('created_at').notNullable();
-    table.dateTime('updated_at').notNullable();
-    table.boolean('deleted').defaultTo(false);
   });
 };
 
--- a/packages/api/db/migration/20241016204145_create_animal_movement_task_tables.js
+++ b/packages/api/db/migration/20241016204145_create_animal_movement_task_tables.js
@@ -76,14 +76,14 @@ export const up = async (knex) => {
     );
   });
 
-  // Create animal_movement_task_animal_relationship table (links the task to the animals that are to be moved)
-  await knex.schema.createTable('animal_movement_task_animal_relationship', (table) => {
+  // Create task_animal_relationship table (links the task to the animals that are to be moved)
+  await knex.schema.createTable('task_animal_relationship', (table) => {
     table.integer('task_id').references('task_id').inTable('task').notNullable();
     table.integer('animal_id').references('id').inTable('animal').notNullable();
   });
 
-  // Create animal_movement_task_animal_batch_relationship table (similar to above but for batches)
-  await knex.schema.createTable('animal_movement_task_animal_batch_relationship', (table) => {
+  // Create task_animal_batch_relationship table (similar to above but for batches)
+  await knex.schema.createTable('task_animal_batch_relationship', (table) => {
     table.integer('task_id').references('task_id').inTable('task').notNullable();
     table.integer('animal_batch_id').references('id').inTable('animal_batch').notNullable();
   });
@@ -102,8 +102,8 @@ export const down = async (knex) => {
     'animal_movement_task',
     'animal_movement_task_purpose_relationship',
     'animal_movement_purpose',
-    'animal_movement_task_animal_relationship',
-    'animal_movement_task_animal_batch_relationship',
+    'task_animal_relationship',
+    'task_animal_batch_relationship',
   ];
 
   for (const table of tablesToDelete) {
--- a/packages/api/db/migration/20241016204145_create_animal_movement_task_tables.js
+++ b/packages/api/db/migration/20241016204145_create_animal_movement_task_tables.js
@@ -20,8 +20,8 @@
 export const up = async (knex) => {
   // Add new animal_movement entry to task_type table
   await knex('task_type').insert({
-    task_name: 'Animal Movement',
-    task_translation_key: 'ANIMAL_MOVEMENT',
+    task_name: 'Movement Task',
+    task_translation_key: 'MOVEMENT_TASK',
   });
 
   // Add location_id column to animal and animal_batch tables (foreign key to location table)
--- a/packages/api/db/migration/20241016204145_create_animal_movement_task_tables.js
+++ b/packages/api/db/migration/20241016204145_create_animal_movement_task_tables.js
@@ -94,7 +94,10 @@ export const up = async (knex) => {
  * @returns { Promise<void> }
  */
 export const down = async (knex) => {
-  await knex('task_type').where({ task_translation_key: 'ANIMAL_MOVEMENT' }).del();
+  await knex('task_type')
+    .where({ task_translation_key: 'MOVEMENT_TASK' })
+    .andWhere({ farm_id: null })
+    .del();
   await knex.schema.alterTable('animal', (table) => table.dropColumn('location_id'));
   await knex.schema.alterTable('animal_batch', (table) => table.dropColumn('location_id'));
 
--- a/packages/api/db/migration/20241016204145_create_animal_movement_task_tables.js
+++ b/packages/api/db/migration/20241016204145_create_animal_movement_task_tables.js
@@ -80,12 +80,22 @@ export const up = async (knex) => {
   await knex.schema.createTable('animal_movement_task_animal_relationship', (table) => {
     table.integer('task_id').references('task_id').inTable('task').notNullable();
     table.integer('animal_id').references('id').inTable('animal').notNullable();
+    table.string('created_by_user_id').references('user_id').inTable('users');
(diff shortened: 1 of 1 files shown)


antsgar:
@SayakaOno can we setup this one as ready for review? I want to approve it, it's looking great! And since this one will be the foundation for the upcoming backend tickets it'd be good to merge it soon

SayakaOno:
Thank you @antsgar, sure! I'll undraft this now!

antsgar (dismissed):

packages/api/db/migration/20241016204145_create_animal_movement_task_tables.js
```diff
@@ -0,0 +1,112 @@
...
+      'BEHAVIORAL_AND_SOCIAL_MANAGEMENT',
+      'EXERCISE',
+      'QUARANTINE',
+      'OTHER',
+    ].map((key) => ({ key })),
+  );
+
+  // Create animal_movement_task table
+  await knex.schema.createTable('animal_movement_task', (table) => {
+    table.integer('task_id').references('task_id').inTable('task').primary();
```
Duncan-Brain, line 61: should this table have a `toLocation` ?

Duncan-Brain (commented):
Looks good only main question is about the translation key.

@antsgar I cant help but wonder if we should create a separate join table for `animal_location` `animal_batch_location` . Only because all animal location changes will then need to go through the get and patch of animal/animal_batch.

While we have tried to add early returns and everything in animal patch -- there might be value or efficiency in separating location. 

For example:
-  to retrieve animals on a specific location for map view all animal details may not be needed (if in animals or map), 
- patching animal location ( to avoid creating new middleware location logic or running any extra animal patch logic),

packages/api/db/migration/20241016204145_create_animal_movement_task_tables.js
```diff
@@ -0,0 +1,112 @@
...
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+export const up = async (knex) => {
+  // Add new animal_movement entry to task_type table
+  await knex('task_type').insert({
+    task_name: 'Animal Movement',
+    task_translation_key: 'ANIMAL_MOVEMENT',
```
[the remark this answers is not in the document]
SayakaOno: Is there still a possibility that we'll go with `ANIMAL_MOVEMENT_TASK`? If not, I'll update it to `MOVEMENT_TASK`!

packages/api/db/migration/20241016204145_create_animal_movement_task_tables.js
SayakaOno: Previously I thought it should (and I added), but I realized that we're using `location_tasks` as "toLocation". Do you think it'll work?

SayakaOno (commented):
Thank you @Duncan-Brain for reviewing!
Just to make sure, are you thinking about calling the animal/batch patch API separately when completing the movement task? I was going to update animals/batches' `location_id` as part of the task completion API, which wouldn't go through the validations.
(Also, I thought the location of an animal or batch should never be editable through the animal/batch patch API...)

packages/api/db/migration/20241016204145_create_animal_movement_task_tables.js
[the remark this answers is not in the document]
Duncan-Brain: I don't know really -- might be worth asking Loic

packages/api/db/migration/20241016204145_create_animal_movement_task_tables.js
Duncan-Brain: yes good ok that makes sense! Thanks for clarifying!

Duncan-Brain:
@SayakaOno Good point about not using the patch endpoint! haha I probably thought that yes.... will strike that one out. The other scenario that we wont be implementing until v2 (or maybe never) was locate functionality : 
Screenshot 2024-10-29 at 2 20 08 PM

antsgar:
@Duncan-Brain I think I see what you mean, it's a good question! But even if there are scenarios where we don't need all the attributes, that's something we can define at the API level, that is, e choose what to query and what to return on the API, regardless of how the database is designed. Since there's a 1:1 relationship between animal/batch and location, I don't see more reason to separate location into its own separate table than any of the other animal/batch attributes.

Duncan-Brain (dismissed):

SayakaOno:
I discussed with Anto in tech daily on how we can handle removed animals/batches in movement tasks. What we're aiming is ([Loïc's message on slack](https://litefarm.slack.com/archives/CAH7PMYE4/p1729854705335289?thread_ts=1729806103.149959&cid=CAH7PMYE4)):

> **For an incomplete task.**
Animal deleted: Animal does not appear in the task, nor the read only view.
Animal removed: Animal does not appear in the task, nor the read only view.
**Completed/Abandoned task**
Animal deleted before completion/abandonment: Animal is simply removed from the task and wiped from the db.
Animal removed before completion/abandonment: Animal is simply removed from the task.
Animal deleted after completion/abandonment: Since we defined the case of deleting an animal as “Created by error”(ie. Not appearing in any of the farm historical records). The animal is removed from the task and the db.
Animal removed after completion/abandonment: Animal remains in the task as part of historical data.

I added base properties to `animal_movement_task_animal_relationship` and `animal_movement_task_animal_batch_relationship` to support soft-delete.

Duncan-Brain:
Are base properties only useful for this case? : 

> Animal deleted after completion/abandonment: Since we defined the case of deleting an animal as “Created by error”(ie. Not appearing in any of the farm historical records). The animal is removed from the task and the db.

For this case, at least in crops, we only allow abandoning of the crop (not deletion) if tasks have been completed. 

Sorry I missed yesterdays discussion -- I do not think we should allow created in error deletions after tasks have been completed with that animal. So by extension I am not sure I see the need yet for base properties here.

SayakaOno:
Base properties are necessary for:

>    **Completed/Abandoned task**
>    Animal removed before completion/abandonment: Animal is simply removed from the task.
>    Animal removed after completion/abandonment: Animal remains in the task as part of historical data.

For example, if we create a task to move animal A and B, remove animal A, complete the task, and then remove animal B, we should show only animal B in the read-only view.

Without a deleted flag, we would have to compare `removal_date` (timestamp) for animal A and B against `complete_date` (YYYY-MM-DD) of the task. This approach only works if these actions occur on different dates, but if they happen on the same day, it won’t be accurate.

(Another scenario: if we set the complete_date to a day before animal A was removed, animal A would appear in the read-only view after completion, even though it wasn't moved.)

So, I believe we should choose either hard-delete or soft-delete to ensure proper implementation (and we agreed on soft-delete yesterday).

> I do not think we should allow created in error deletions after tasks have been completed with that animal. So by extension I am not sure I see the need yet for base properties here.

Loïc says ([thread](https://litefarm.slack.com/archives/CAH7PMYE4/p1729854705335289?thread_ts=1729806103.149959&cid=CAH7PMYE4)):

> Removing or deleting an animal is designed as an intentional, multi-step workflow to prevent accidental actions. So in general we should always allow user to do so, even if tasks are associated with the animal(s) on the go.

If you spot any issues with this approach, please bring them up!

packages/api/db/migration/20241016204145_create_animal_movement_task_tables.js
```diff
@@ -0,0 +1,122 @@
...
+    table.string('other_purpose');
+    table.check(
+      `(other_purpose IS NOT NULL AND purpose_id = ${otherPurposeId}) OR (other_purpose IS NULL)`,
+      [],
+      'other_purpose_id_check',
+    );
+  });
+
+  // Create animal_movement_task_animal_relationship table (links the task to the animals that are to be moved)
+  await knex.schema.createTable('animal_movement_task_animal_relationship', (table) => {
```
Duncan-Brain, line 80: Now that I am seeing this again -- should we keep this table generic for other animal tasks by calling it simply `task_animal` or `task_animal_relationship`;

Duncan-Brain:
Okay I might like to learn more...I agree we need to not compare dates and instead choose a deletion method. 

But I thought base object properties (and by extension soft delete) were for base objects like task or animals. I might like to ask @antsgar about what is our standard for choosing soft-delete vs hard delete. Task sub-details do not seem useful to keep in an "un-delete" scenario. I would propose to treat it like other changeable properties on a task like date, task_purpose -- `location_tasks` which seems analogous. 

I think removing is similar to abandoning so I agree with Loic there. But might have to ask Loic about deleting even if tasks are completed with animal. This might have implications regarding animal task based wages in the future which is why I think we disallowed it in the past. There might be a cost per animal (medicine) that we cannot allow deletion from task.

packages/api/db/migration/20241016204145_create_animal_movement_task_tables.js
```diff
@@ -0,0 +1,122 @@
...
+      'TRANSPORTATION',
+      'BEHAVIORAL_AND_SOCIAL_MANAGEMENT',
+      'EXERCISE',
+      'QUARANTINE',
+      'OTHER',
+    ].map((key) => ({ key })),
+  );
+
+  // Create animal_movement_task table
+  await knex.schema.createTable('animal_movement_task', (table) => {
```
Duncan-Brain, line 60: Since we aren't storing toLocation ... should we still keep this table for future? its sort of empty right now.

SayakaOno: We need it to have `{ animal_movement_task: { purposes... } }` in tasks!

packages/api/db/migration/20241016204145_create_animal_movement_task_tables.js
SayakaOno: I don't see any problems with that. @antsgar What do you think?

packages/api/db/migration/20241016204145_create_animal_movement_task_tables.js
Duncan-Brain: Haha just for the graph data structure that is interesting!

Duncan-Brain:
Ah true, Anto mentioned we could maybe migrate the animal batch sex detail to be hard delete in this comment: https://github.com/LiteFarmOrg/LiteFarm/pull/3484#issuecomment-2423380321 

If I found some spare time I was thinking about doing that

SayakaOno:
I pushed two commits to:
1. remove the base properties from the relationship tables. (reverted)
2. change the names of the task_animal relationship tables to be more generic.

Thank you! 🙏

packages/api/db/migration/20241016204145_create_animal_movement_task_tables.js
```diff
@@ -0,0 +1,112 @@
...
+    table.integer('animal_batch_id').references('id').inTable('animal_batch').notNullable();
+  });
+};
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+export const down = async (knex) => {
+  await knex('task_type').where({ task_translation_key: 'ANIMAL_MOVEMENT' }).del();
```
Duncan-Brain, line 97: This should be `MOVEMENT_TASK` and sorry I did not notice before -- because task_type is user-editable we should check `farm_id = null` before deleting.
Screenshot 2024-11-04 at 3 29 24 PM

SayakaOno: Oh good catch, thank you!

Duncan-Brain (approved):
Awesome!

antsgar (approved):

PR #3510, opened 2024-10-22
[Not-urgent] LF-4380 - address comments and improvement suggestions

Duncan-Brain:
**Description**

- simplify setFalsyValuesToNull

Jira link: [LF-4380](https://lite-farm.atlassian.net/browse/LF-4380)

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

[LF-4380]: https://lite-farm.atlassian.net/browse/LF-4380?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/api/src/middleware/validation/checkAnimalOrBatch.js
```diff
@@ -171,78 +171,71 @@ const checkAnimalBreed = async (
...
+      setFalsyValuesToNull(breedKeyOptions, animalOrBatch);
+    }
+
+    const isNotNullingAllBreedOptions = !(
+      someExists(breedKeyOptions, animalOrBatch) &&
+      !someTruthy([default_breed_id, custom_breed_id, breed_name])
+    );
+    const isCreatingWithBreed =
+      creating && someTruthy([default_breed_id, custom_breed_id, breed_name]);
+    // Do checks on breed unless removing breed from existing record or
```
SayakaOno, line 194: "Do checks on breed unless removing breed from existing record" - it seems to me that the next line can simply be `if(someTruthy([default_breed_id, custom_breed_id, breed_name])`.
I got confused here and I may have misunderstood something.

packages/api/src/middleware/validation/checkAnimalOrBatch.js
```diff
@@ -171,78 +171,71 @@ const checkAnimalBreed = async (
...
           preexistingAnimalOrBatch,
-          customBreed,
+          default_breed_id,
           default_type_id,
-          custom_type_id,
         );
       }
+      // Check if custom breed or new type is present
+      // Skip checks in some cases where new custom type or breed is added
+      if ((custom_breed_id && !type_name) || ((default_type_id || custom_type_id) && !breed_name)) {
```
SayakaOno, line 206: The combination of breed ID and type_name is wrong... isn't it already handled earlier? I'm wondering if it can be `if (custom_breed_id || ((default_type_id || custom_type_id) && !breed_name))`

SayakaOno (commented):
The updated code is much more readable and easier to follow! 👏 
I left a few comments on places where I got confused.

packages/api/src/middleware/validation/checkAnimalOrBatch.js
Duncan-Brain: So if you comment out the logic and replace it with the suggested you will see that 3 tests fail. The tests that fail are all tests that edit the `type`. They fail because they successfully allow the user to change the type to one that does not match the breed (eg honey bee breed and cow type would be an allowed combination). 

So I will fix/update the comment since I forgot to finish the sentence. 
> Do checks on breed unless removing breed from existing record or creating animal without breed

This describes when it should not run the checks but can also be read as: 

> Check breed when creating new animal with breed or when editing an animal that has type or breed being edited.. but not when the breed is being nulled because you don't need to check it if you are nulling it.

It was a logic structure decision to make breed tests be more complex. Since breeds must have types but types do not have to have breeds. It is equally valid ways to structure the logic so that breed tests were mixed into the type middleware, but I found that hard to understand.

Please let me know if you have suggestions for better comments!

packages/api/src/middleware/validation/checkAnimalOrBatch.js
Duncan-Brain: So this is one of my failures of having enough time to structure it how I wanted. It is actually currently handled later.... and you will see that if you change the logic to your suggestion, you are right it will be caught here!.. However it is not the desired error message. 

What I would like to do is chunk up and move both `checkAndAddCustomTypesOrBreeds()` & `checkCustomTypeAndBreedConflicts()` inside the `checkType()` and `checkBreed()` functions.

I did not feel I could do this a the time because:
- This had already taken me an uncomfortably long amount of time to understand and adapt the logic.
- These functions were mostly pre-existing and I did not want to change the order of other work unless necessary, one of my personal goals is to understand when I should make improvements vs leaving the properly functioning logic as is. I think I have received some peer feedback related to this. 

So I chose to name and reuse existing logic but I really did not add or re-order much. I only added a little bit of new logic for comparing existing record and updates and tried to keep the order of logic the same ..as best I could to understand it there are exceptions.

packages/api/src/middleware/validation/checkAnimalOrBatch.js
SayakaOno: Ah, I forgot about the existing breed! The previous comment should've been clear enough. Apologies for the misunderstanding!

packages/api/src/middleware/validation/checkAnimalOrBatch.js
SayakaOno: Understood, thank you for the detail!

SayakaOno (approved):

PR #3511, opened 2024-10-23
LF-4489: nav menu changes for malayalam

antsgar:
**Description**

- Setup specific font for Malayalam so that it looks the same across devices.
- Make nav menu wider for Malayalam (this is a hacky and ugly solution but it is what it is).
- Add ellipsis for menu items that overflow.
- Show entire text on hover for text that overflows.

We can decide to apply all of these changes or some of them, it's up for discussion!

Jira link:
https://lite-farm.atlassian.net/browse/LF-4489

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


SayakaOno:
Thank you Anto! I'm curious how it looks on Denis' screen now, but all the solutions look good to me!

I opened a PR to correct the incorrect mixin import paths I came across: 
https://github.com/LiteFarmOrg/LiteFarm/pull/3512

packages/webapp/src/components/CardWithStatus/ManagementPlanCard/styles.module.scss
```diff
@@ -8,7 +9,7 @@
   font-size: 16px;
   line-height: 24px;
   font-style: normal;
-  font-family: 'Open Sans', 'SansSerif', serif;
+  @include fontFamily();
```
Duncan-Brain, line 12: Maybe a silly question but is it possible or a bad idea to set this globally and remove font family from all components?

antsgar: 100%! But it seems for some components we need to specify it at the element level because we need to overwrite existing MUI styles -- it's probably only a handful of cases and not all of the different instances that we have, but I figured I'd figure that out later 😅 but you're totally right, we shouldn't be specifying this a hundred times

Duncan-Brain: Oh yeah I forgot about MUI!

antsgar:
@SayakaOno thank you so much for the import fixes!! I thought VSCode would give me some sort of warning if the path was wrong but I guess not? 😞 I ended up adding an alias so we don't have to worry about writing the right path, because it seems pretty error prone. Let me know what you think!

SayakaOno:
I love the alias!! I also didn't like that we always had to import the right path, so it's nice!

SayakaOno (approved):

kathyavini (approved):
Looks good on beta. Love the new font!

PR #3518, opened 2024-10-25
LF-4471: Add or update models for animal movement task

SayakaOno:
**Description**

- Add models for:
   - animal_movement_purpose
   - task_animal_relationship
   - task_animal_batch_relationship
   - animal_movement_task
   - animal_movement_task_purpose_relationship
- Update `taskModel` to add relations to `animal_movement_task`, `animals` and `animal_batches`
- Update `getTasksByFarmId`
   - format `withGraphFetched` for better readability
   - add `animal_movement_task` for test (this will be implemented properly in LF-4472)
      <details>
        <summary>Sample response</summary>
        
        ```
        [
          {
            "task_id": 1132,
            "animal_batches": [
                {
                  "id": 24,
                  "farm_id": "474069c6-22a9-11ee-a59f-e66db4bef552",
                  "default_type_id": 1,
                  ...
                },
                ...,
              ],
              "animals": [
                {
                  "id": 34,
                  "farm_id": "474069c6-22a9-11ee-a59f-e66db4bef552",
                  "default_type_id": 1,
                  ...,
                },
                ...,
              ],
            ...,
            "animal_movement_task": {
              "task_id": 1132,
              "purpose_relationships": [
                {
                  "task_id": 1132,
                  "purpose_id": 1,
                  "other_purpose": null,
                },
                {
                    "task_id": 1132,
                    "purpose_id": 14,
                    "other_purpose": "ooooo"
                },
              ],
            }
          },
          ...
        ]
        ````
      </details>
- Update `animalModel` and `animalBatchModel`
   - add `location_id`
   - add a relation to `tasks`
      <details>
        <summary>Sample response when including "animal_movement_tasks" in GET request</summary>
        
        ```
        {
          "id": 34,
          "tasks": [
              {
                  "task_id": 1132,
                  "due_date": "xxx"
              },
              ...
          ],
          "internal_identifier": 47,
          ...
        }
        ````
      </details>

NOTE:
~~I got confused by `relationMapping`, so I read the Objection's document and ended up with something a little different from what we've done so far. Please let me know if we should stick with the previous approach. The main difference, for instance, is whether to join the "purpose relationship" table or the "purpose" table directly.~~

The documents and code I referred:
- many-to-many relationship with extra properties
   https://vincit.github.io/objection.js/recipes/extra-properties.html
- relationMappings for a join table (code)
   https://github.com/strues/boldr/blob/f8daa99ea626947c6809e1c3683afc637f7a38d0/packages/boldr-api/src/models/join/menuMenuDetail.js

Jira link: https://lite-farm.atlassian.net/browse/LF-4471

**Type of change**

**How Has This Been Tested?**

Added test data to "task", "location_tasks", "animal_movement_task", "animal_movement_task_animal_relationship", "animal_movement_task_animal_batch_relationship" and "animal_movement_task_purpose_relationship" tables and called the get tasks API. 

- [ ] Passes test case
- [x] Other (please explain)

**Checklist:**


packages/api/src/models/animalMovementTaskModel.js
```diff
@@ -0,0 +1,103 @@
...
+          from: 'animal_movement_task.task_id',
+          through: {
+            modelClass: AnimalMovementTaskAnimalBatchRelationshipModel,
+            from: 'animal_movement_task_animal_batch_relationship.task_id',
+            to: 'animal_movement_task_animal_batch_relationship.animal_batch_id',
+          },
+          to: 'animal_batch.id',
+        },
+      },
+      purposes: {
```
Duncan-Brain, line 85: @SayakaOno Regarding your note on relation mappings in case you wanted feedback/discussion. 

If it helps.. I think my understanding of `relationMappings` is that you only need to specify it when you expect to use relation methods like query.relate(), graph insert, upsert etc. 

In the past, I have only thought of it as the upsert path. No upsert needed(or other relation method), no path (relationMapping) needed.

For your use case here, previously we do not intend on using a `POST: /api/animal_movement_task` to update `animal_movement_purpose` so we wouldn't add a relation mapping, and nesting shared enum table inside a task object would also be not helpful. 

But if we follow through on Antos idea to replacing the `purpose_id ` with `purpose_key` on GET endpoint then what you have here should work correctly! The consequence of this I think will be is that we may need to add something like `{unrelate: ['purposes']}` to task upsert options.

SayakaOno: Thank you Duncan for your comment, it's much clearer now.

I used ChatGPT's help on this (as always 😅), and the code is based on two key points:
- Joining the relations tables is not correct.
- Adding `relationMappings` is typically beneficial. It can make relationships much clearer for other developers. (However, I also learned from the Objection documentation that relation tables don’t even require a model!)

I was wondering when to define `relationMappings`, so your input is definitely helpful! If I understand correctly, adding `relationMappings` may sometimes require extra work, such as including `{unrelate: ['purposes']}`...

Replacing the `purpose_id`  with `purpose_key` was also something I wanted to discuss in tech daily.

This is a great opportunity for me to deepen my understanding of relationships, and I appreciate your insights! Were there any resources that helped you become familiar with Objection? I'd like to come back with more polished questions.

Duncan-Brain: I mostly have just been getting familiar indirectly due to the different places we use graph + relations and reading objection documentation. (I only use chatGPT for simple things like simplify this function or make jsdoc-- but it still fails me on that lol -- it will be a while before it takes my job I think 😉 )

- task + management plan use withGraphFetched to explicitly specify individual relations.
- upsertGraph() when specified with a model, I think will try to update all related (so that means I am curious whether it would be possible to update the enum table because of the `through` mechanism. If so then maybe an option like unrelate is necessary).
- graph option like `noInsert`, `noUpdate`

> Joining the relations tables is not correct.

I would be curious to learn more about this point. Our LiteFarm definition of a "relations" table may not be the same as Objections definition.  In the example with actors, movies, and movies_actors -- both actors and movies seem like what I loosely name "modules". They are "important". The enum table will only be used in one context so I do not know if that is the same kind of relations table.

SayakaOno:
Thank you @Duncan-Brain for your input!

After experimenting with joining the `purpose` table to the `animal_movement_task` table, I decided to use the `animal_movement_task_purpose_relationship` table, as we have consistently done. (Although I prefer to return purposes rather than relationships, we can cook the data as needed, so I'm convinced this is a better approach :))

If you're interested, here’s my thought:
For `upsertGraph` to work with the PTACH request, we would need to `additionalProperties: true` in the Purpose model or add `other_purpose` to its jsonSchema. This is because `other_purpose` is treated as an "extra" field in the `animal_movement_task_purpose_relationship` table. However, since the `purpose` table is a lookup table, it didn't feel appropriate to add `other_purpose` there.

reference:
- https://github.com/Vincit/objection.js/issues/1288
- https://stackoverflow.com/questions/57510549/objection-js-cant-update-many-to-many-relation-with-join-table-extra-property

Duncan-Brain (changes requested):
@SayakaOno it looks good! 

1. I think the main thing I noticed is that not all possible `relationMappings` should need to be specified... only the Models which we intend on acting upon in the controllers. So for example we probably will not make a `GET` endpoint on `AnimalMovementTaskPurposeRelationshipModel` and so no relations should need to be specified from this one. I might be wrong about this: but a possible downside is it might be possible to create circular references with relationMappings leading to expensive queries.

2. Regarding :

> Although I prefer to return purposes rather than relationships,

I think I have a new way to explain that may (or may not) help ease your mind!

It mostly has to do with the idea that enum tables are not really "relations" like described in objection as incorrect code style -- this previous comment:
 
> In the (objection) example with actors, movies, and movies_actors -- both actors and movies seem like what I loosely name "modules". They are "important". The enum table will only be used in one context so I do not know if that is the same kind of relations table.

An alternative way to represent `purpose_relationships` is with a simple purposes enum array ['GRAZING, FEEDING'] column directly on the `animal_movement_task`. If it were a simple enum array like this it is easier to understand as _**not really**_ a "relationship".

So, at LiteFarm, we have chosen to have a mostly [normalized database](https://en.wikipedia.org/wiki/Database_normalization) and by expanding that simple enum array out -- this satisfies the 1st normal form. A side benefit of this normalization is the enum can be added to without migration using a admin user POST (if we decided that was useful). 

What we could do to make that normalization easier on ourselves (and sorry I keep bringing this up it is just very related lol) -- on the enum table we could use a [natural key](https://en.wikipedia.org/wiki/Natural_key) instead of what we use now a [surrogate key](https://en.wikipedia.org/wiki/Surrogate_key). Then instead of being what looks like a "relationship" table `{animal_id: 1, purpose_id: 1}` we would have a regular "details" table `{animal_id: 1, purpose: GRAZING}` -- but I think technically it is more of a details table now but looks like a relationship.

Sorry for the long discussion here!

packages/api/src/controllers/taskController.js
```diff
@@ -713,8 +713,22 @@ const taskController = {
       const graphTasks = await TaskModel.query()
         .whereNotDeleted()
         .withGraphFetched(
-          `[locations.[location_defaults], managementPlans, soil_amendment_task, soil_amendment_task_products(filterDeleted).[purpose_relationships], field_work_task.[field_work_task_type], cleaning_task, pest_control_task, harvest_task.[harvest_use], plant_task, transplant_task, irrigation_task.[irrigation_type]]
```
Duncan-Brain, line 716: ❤️

packages/api/src/models/taskModel.js
```diff
@@ -266,6 +305,7 @@ class TaskModel extends BaseModel {
       taskType: 'omit',
       plant_task: 'edit',
       transplant_task: 'edit',
+      animal_movement_task: 'omit',
```
Duncan-Brain, line 308: I think `animals`, and `animal_batches` also needs `omit` for now.

SayakaOno (commented):
Thank you Duncan!

1. I intentionally added some relationMappings that aren't currently in use, mainly for practice. I didn't think it would interfere, but I can remove them.

> a possible downside is it might be possible to create circular references

Are you referring to [this](https://vincit.github.io/objection.js/guide/relations.html#require-loops-non-ecmascript-modules-only)? If so, it says "non ECMAScript modules only", so we would be fine :)

2. Thank you for the detailed explanation! The current design looks solid, so I have no concerns there. My preference is mostly for a simpler response structure like this:
```
[{
  id: 1,
  purposes: [
    {purpose: GRAZING},
    {purpose: OTHER}
  ]
})
```
over:
```
[{
  id: 1,
  purposes: [
    {animal_id: 1, purpose: GRAZING},
    {animal_id: 1, purpose: OTHER}
  ]
})
```
One solution could be to select only the columns we need?

packages/api/src/models/animalBatchModel.js
```diff
@@ -141,6 +144,19 @@ class AnimalBatchModel extends baseModel {
           to: 'animal_batch_use_relationship.animal_batch_id',
         },
       },
+      tasks: {
```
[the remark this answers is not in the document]
Duncan-Brain: Is it more efficient if we have already retrieved all tasks on frontend? I am worried about nesting base objects into other base objects like existing saga complexity. 

When we invalidate tags if we add a new animal task, we would need to invalidate both [Animals + Tasks] ?

packages/api/src/models/animalMovementTaskModel.js
```diff
@@ -0,0 +1,65 @@
...
+      properties: {
+        task_id: { type: 'integer' },
+      },
+      additionalProperties: false,
+    };
+  }
+
+  static get relationMappings() {
+    return {
+      task: {
```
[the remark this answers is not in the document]
Duncan-Brain: Good point ! I am not sure I understand why it was put in on the others -- I am maybe missing something so maybe you should keep?

Duncan-Brain:
@SayakaOno 
1.  I think my thought was just wrong but that is also good to know thanks!

2.  Yeah I would bet a modifier directly on the relationMapping could help us out there! That would be a handy template to have. It could maybe even return one step further:

`[{
  id: 1,
  purposes: [
    GRAZING,
    OTHER
  ]
})`

Duncan-Brain (approved):

packages/api/src/models/animalBatchModel.js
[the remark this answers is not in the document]
SayakaOno: So when a user removes an animal, would tasks associated with that animal be included in the API request as well? It feels more natural and easier to handle it on the backend...

The cashing is a good point - if we were using RTK query for tasks, I would think that invalidating tasks might reflect changes on animals. But since animals wouldn't automatically receive new or updated tasks with the current set up, joining them in API responses doesn't seem like a good idea.

SayakaOno:
`[{ id: 1, purposes: [ GRAZING, OTHER ] })` This is much better!!

Duncan-Brain:
chatGPT seems to think all we need is something like this on the relationMapping haha:

```
modify: (query) => {
          query.select('purpose'); // Select only the `purpose` field
        }
 ```

packages/api/src/models/animalBatchModel.js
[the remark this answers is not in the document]
Duncan-Brain: Good point about removal cascade and agreed it should be handled on backend .

And I think for frontend I must have thought that we would be modifying tasks slice to receive the `animal_task_relationship` table -- to your earlier point about not returning relationships -- then its not a good idea... Okay I think I understand now 🙏

antsgar (approved):
@SayakaOno looks great!! Neat job with the models 🥳

antsgar:
Oh and re the relationMappings -- I'm far from an expert on Objection and not very familiar with it, but my take is that it doesn't hurt to specify the model as thoroughly as possible and set it up so that any potential relationships can be correctly fetched/inserted in the future. The model definition itself doesn't do anything, the actual `withGraphFetched` or `upsertGraph` calls will define what happens.

SayakaOno:
Thank you Anto!

> The model definition itself doesn't do anything, the actual withGraphFetched or upsertGraph calls will define what happens, right?

That's my understanding!

Duncan-Brain:
@antsgar My understanding was it depends how you call it. Here is an example of where we let the model definition insertGraph whatever it can which looks very clean:

Screenshot 2024-11-07 at 7 06 30 PM


Here is an example where we explicitly define what we want which can get messy:
Screenshot 2024-11-07 at 7 09 02 PM

It is possible to work around anything unwanted yes

antsgar:
@Duncan-Brain I see what you mean, thank you for the examples! But I don't think it's messy necessarily, more verbose for sure but also more explicit? We can discuss it in more depth for future tickets though, I don't have a super strong opinion about it.

PR #3524, opened 2024-11-11
LF-4472: Update controllers and middleware to support animal movement task

SayakaOno:
**Description**

- Update TaskController
   - Update `createTask`
      - Format animal and batch IDs
      - Update `withGraphFetched` to handle animal movement tasks
      - Sample API request
   - Update `completeTask`
      - Format animal and batch IDs
      - Update `if` to `switch`
      - Validate and format animal movement task data and `upsertGraph`
- Update `checkTask`
    - Refactor `checkProductsMiddlewareMap`
    - `checkAnimalTask` in `checkCreateTask` and `checkCompleteTask`
- Modify Animal and AnimalBatch models
   - Add more relationships so no API calls will be required in the task read-only page. (This can be finalized later) 
   - Add modifiers for `withGraphFetched`.
   
Sample request and response body:
- POST & PATCH: https://lite-farm.atlassian.net/browse/LF-4479?focusedCommentId=17820
- GET: https://lite-farm.atlassian.net/browse/LF-4483?focusedCommentId=17819

If calling the APIs makes the review easier, you can find the branch in https://github.com/LiteFarmOrg/LiteFarm/pull/3525. 

Jira link: https://lite-farm.atlassian.net/browse/LF-4472

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/api/src/controllers/taskController.js
```diff
@@ -67,48 +117,120 @@ async function updateTaskWithCompletedData(
   nonModifiable,
   typeOfTask,
 ) {
-  if (typeOfTask === 'soil_amendment_task') {
-    const { soil_amendment_task_products } = data;
+  switch (typeOfTask) {
```
SayakaOno, line 120: Converted without any changes.

packages/api/src/controllers/taskController.js
```diff
@@ -58,6 +62,52 @@ async function getTaskAssigneeAndFinalWage(farm_id, user_id, task_id) {
...
+      const formattedPurpose = {
+        purpose_id: purposesMap[key],
+        ...(other_purpose ? { other_purpose } : {}),
+      };
+
+      formattedPurposes.push(formattedPurpose);
+    }
+  }
+
+  data.animal_movement_task.purpose_relationships = formattedPurposes;
```
antsgar, line 105: Is there a reason we can't have the FE send out the purpose IDs instead? I was thinking of something similar to `related_animal_ids`, where we would have `purpose_ids = [1, 2, 3]` and then the BE wouldn't need to do a search by key, since the FE already has both IDs and keys

packages/api/src/controllers/taskController.js
```diff
@@ -58,6 +62,52 @@ async function getTaskAssigneeAndFinalWage(farm_id, user_id, task_id) {
...
+      `Invalid field: "purpose_relationships" should not be included. Use "purposes" instead`,
+    );
+  }
+
+  if (!data.animal_movement_task.purposes) {
+    delete data.animal_movement_task.purposes;
+    return data;
+  }
+
+  checkIsArray(data.animal_movement_task.purposes, 'purposes');
```
antsgar, line 85: I think these kind of checks would be better placed in the middleware, in the `checkAnimalTask` function along with the checks you added there for the other fields. There aren't any custom entities being inserted/handled here like it happens for other tasks, so I think it wouldn't be correct to add this to `checkCustomDependencies`

packages/api/src/controllers/taskController.js
antsgar: Thank you for doing this! ❤️

packages/api/src/controllers/taskController.js
```diff
@@ -1098,3 +1272,7 @@ async function filterOutDeletedManagementPlans(data, req) {
     validPlantingMangementPlans.includes(planting_management_plan_id),
   );
 }
+
+const flattenInternalIdentifier = (animalOrBatch) => {
+  animalOrBatch.internal_identifier = animalOrBatch.internal_identifier.internal_identifier;
```
antsgar, line 1277: Why do we have nested internal identifiers? 🤔

packages/api/src/util/animal.js
```diff
@@ -142,3 +147,82 @@ export const upsertGroup = async (req, animalOrBatch, farm_id, trx) => {
...
+  if (invalidAnimalIds.length || invalidBatchIds.length) {
+    throw customError(
+      'Some animal IDs or animal batch IDs do not exist, are removed, or are not associated with the given farm.',
+      400,
+      { invalidAnimalIds, invalidBatchIds },
+    );
+  }
+};
+
+export async function isOnOrAfterBirthAndBroughtInDates(date, animalIds, batchIds) {
```
antsgar, line 196: This function could be simplified by querying for the earliest date both for animals and batches, and then doing a comparison with those two dates only

packages/api/src/util/animal.js
```diff
@@ -142,3 +147,82 @@ export const upsertGroup = async (req, animalOrBatch, farm_id, trx) => {
...
+    brought_in_date && birthAndBroughtInDatesSet.add(brought_in_date);
+  });
+
+  if (birthAndBroughtInDatesSet.size) {
+    const [y, m, d] = date.split('-');
+    const dateTimestamp = new Date(y, m - 1, d);
+
+    for (const dateToCompre of birthAndBroughtInDatesSet) {
+      const broughtInOrBirthdate = new Date(
+        dateToCompre.getFullYear(),
```
antsgar, line 216: NIT -- spelling in `dateToCompare`

antsgar (commented):
Left a few comments/questions, but it's looking good! 🥳

SayakaOno (commented):
Thank you @antsgar for reviewing!
I believe I've addressed all your comments. I’d appreciate it if you could take another look! 🙏

packages/api/src/controllers/taskController.js
antsgar: I don't entirely remember the conversation exactly so sorry if I said something different, but here's what I'm thinking:
- We fetch the list of purposes on the FE from the API, containing both IDs and keys, so we get something like `purposes = [{ id: 1, key: 'GRAZING' }, { id: 5, key: 'OTHER' }]`
- On the FE, we check for the "OTHER" key in case where we have some logic to display certain fields if that option was selected.
- Once we're ready to send the info back to the API and save the task, we send only the purposes IDs, and the "other purpose" note, if it has been shown to the user per the logic above and filled out.
- The BE takes the list of purpose IDs and checks if the key for any of them is "OTHER". If it is, it allows the "other purpose" field and stores it in the DB. If none of the purposes matches that key, it either returns a bad request to inform the client the request doesn't make sense, or simply ignores the other purpose string and doesn't store it. The DB constraint will disallow the BE from storing something that doesn't make sense.
This way we're preventing inconsistencies on all 3 layers:
- On the FE, by only showing the field when it's relevant.
- On the BE, by not passing through data to the DB that doesn't make sense.
- On the database, by disallowing storing data that doesn't make sense.
The FE however doesn't need to worry about checking what the ID is for "Other", it just deals with keys on the visual side and sends IDs once it sends the info for the BE to take over.
Let me know if this makes sense or if I'm missing something!

SayakaOno: You may have meant not to send relationships from the beginning. Sending IDs and other_purpose separately works well, so I'll make the necessary adjustments!

packages/api/src/controllers/taskController.js
antsgar: Ah, I see!! I know this is pre-existing but I think we should change the name either on the relation or the column to make it less confusing

SayakaOno:
Changes made:
- PATCH and POST request bodies:
   `purpose_relationships` has been changed to `purpose_ids` and `other_purpose`.

POST sample request body:
```
{
    "due_date": "2024-11-30", 
    "task_type_id": 48,
    "locations": [{"location_id": "7e7fce6c-7ffc-11ee-8813-e66db4bef551"}],
    "related_animal_ids": [2,3,4],
    "related_batch_ids": [2,3],
    "animal_movement_task": {
        "purpose_ids": [1, 14],
        "other_purpose": "xxx"
    },
    "assignee_user_id": "3c8ddad6-22a9-11ee-b54e-e66db4bef552",
    "override_hourly_wage": false,
    "wage_at_moment": null
}
```

- Relation name change:
   The relation name with `animal_union_batch_id_view` is now updated to `animal_union_batch`.

Thank you!

kathyavini (commented):
It looks fantastic and it reads so elegantly 😍 I approve, but I will await the small API change if you end up tracking it down!

packages/api/src/controllers/taskController.js
```diff
@@ -67,48 +104,120 @@ async function updateTaskWithCompletedData(
...
+
+      if (!data.animal_movement_task) {
+        data.animal_movement_task = {};
+      }
+      // Prevent deletion of animal_movement_task and allow proper updates on purpose relationships
+      data.animal_movement_task.task_id = task_id;
+
+      // If the request body do not have purposes, add 'animal_movement_task.purpose_relationships'
+      // to prevent existing purpose relationships from being deleted
+      const noDelete = data.animal_movement_task.purpose_relationships
```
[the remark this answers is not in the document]
SayakaOno: Thank you for reviewing!

I didn't refer old code, so I didn't remember that "we need to always send something for these relationship table properties" to retain the data...😅 
But I tried to make it work for typical PATCH requests, where you send only the fields that need updating.

Another option is to manually add the existing data in the backend, but I think adding the property to `noDelete` is the most efficient option!

SayakaOno:
I've updated the logic to remove data when sending `null` in PATCH requests. Specifically, I updated the condition to determine if the data was provided or not.

For animals and batches, there's a variable called `animalsOrBatchesProvided`. Previously, I considered `null` as "provided", but now having the property (even if it's `null`) is considered "provided" instead. (https://github.com/LiteFarmOrg/LiteFarm/pull/3524/commits/6c111714528ee6f18e4d186f0204d3ffa5e83031)

The same change applies to `purpose_ids`; I previously treated `purpose_ids: null` as "not provided", but I've updated the condition accordingly. (https://github.com/LiteFarmOrg/LiteFarm/pull/3524/commits/25a551faefd3d942a56db532ed27824b0607d470)

The behaviour now feels right! Thank you Joyce for bring this up!

kathyavini (approved):

antsgar (approved):

PR #3527, opened 2024-11-12
LF-4481:  Add Location to animal inventory UI

SayakaOno:
**Description**

Inventory table
- Move the Location type to a common file
- Add `location_id` to Animal and AnimalBatch types and adjust mockData
- Update `useAnimalInventory` to include location
- Add Location column to animal inventory table
- Update `useFilteredInventory` to handle locations filter
- Add `location` string to animal inventory searchable string
- Update `Plain` cell component to take `content`(ReactNode) for greater flexibility
- Update "none" text style in the group column to normal font  (it was previously in italic)
- Fix broken [story](http://localhost:6006/?path=/story/components-tables-cells--fake-table-with-cells)

Header
- Modify `AnimalSingleViewHeader`'s props (separated `locationText` from `animalOrBatch`) and adjust [stories](http://localhost:6006/?path=/docs/components-animals-animalsingleviewheader--docs)
- Send locationText to `AnimalSingleViewHeader`

Jira link: https://lite-farm.atlassian.net/browse/LF-4481

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/Table/Cell/CellTypes/Plain.tsx
+++ b/packages/webapp/src/components/Table/Cell/CellTypes/Plain.tsx
@@ -17,11 +17,11 @@ import clsx from 'clsx';
 import styles from '../styles.module.scss';
 
 export type PlainCellProps = {
-  text: ReactNode;
+  content: ReactNode;
 };
 
-const Plain = ({ text }: PlainCellProps) => {
-  return <div className={clsx(styles.text, styles.overflowText)}>{text}</div>;
+const Plain = ({ content }: PlainCellProps) => {
+  return <div className={clsx(styles.text, styles.overflowText)}>{content}</div>;
 };
 
 export default Plain;
--- a/packages/webapp/src/components/Table/Cell/CellTypes/Plain.tsx
+++ b/packages/webapp/src/components/Table/Cell/CellTypes/Plain.tsx
@@ -12,11 +12,12 @@
  *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
  *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
  */
+import { ReactNode } from 'react';
 import clsx from 'clsx';
 import styles from '../styles.module.scss';
 
 export type PlainCellProps = {
-  text: string | number | null;
+  text: ReactNode;
 };
 
 const Plain = ({ text }: PlainCellProps) => {
--- a/packages/webapp/src/components/Table/Cell/CellTypes/Plain.tsx
+++ b/packages/webapp/src/components/Table/Cell/CellTypes/Plain.tsx
@@ -17,11 +17,11 @@ import clsx from 'clsx';
 import styles from '../styles.module.scss';
 
 export type PlainCellProps = {
-  content: ReactNode;
+  text: ReactNode;
 };
 
-const Plain = ({ content }: PlainCellProps) => {
-  return <div className={clsx(styles.text, styles.overflowText)}>{content}</div>;
+const Plain = ({ text }: PlainCellProps) => {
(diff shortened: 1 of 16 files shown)


antsgar:
Looks/works great, it's all coming together!! 🙌

antsgar (approved):

packages/webapp/src/components/Table/Cell/CellTypes/Plain.tsx
```diff
@@ -12,15 +12,16 @@
...
 
 export type PlainCellProps = {
-  text: string | number | null;
+  content: ReactNode;
 };
 
-const Plain = ({ text }: PlainCellProps) => {
-  return <div className={clsx(styles.text, styles.overflowText)}>{text}</div>;
+const Plain = ({ content }: PlainCellProps) => {
+  return <div className={clsx(styles.text, styles.overflowText)}>{content}</div>;
```
Duncan-Brain, line 24: There are also these styles which were meant to only specified for text.

packages/webapp/src/containers/Animals/Inventory/index.tsx
```diff
@@ -145,6 +147,27 @@ function AnimalInventory({
         ),
         sortable: false,
       },
+      {
+        id: isDesktop ? 'location' : null,
+        label: t('ANIMAL.ANIMAL_LOCATIONS').toLocaleUpperCase(),
+        format: (d: AnimalInventory) => (
+          <Cell
+            kind={CellKind.PLAIN}
```
Duncan-Brain, line 155: There is a cell kind of "ICON_TEXT" did it not work?

Duncan-Brain (approved):
Everything works great! 

I might have preferred to keep the `PLAIN` cell kind very text specific, and adapt `ICON_TEXT` instead, or not use standardized Cell at all. `<Cell />` is not necessary if making something single use and custom. But I will leave it up to you!

packages/webapp/src/containers/Animals/Inventory/index.tsx
SayakaOno: I just tried it out. We would need to work around the text (the component requires `ICON_NAME`, and the location needs to be the text `Unknown` for undefined locations) and pass custom styles. I think it's better not to use the `Cell`.

Duncan-Brain: Yup up to you!

I think 

```
d.location ? <Cell kind={ICONTEXT} text iconName  ../> : <Cell kind={PLAIN} text /> 
```

Would also work to keep text styles same across cells but I am good with anything.

SayakaOno:
Thank you for your input Duncan!

I've reverted the changes to the Plain cell component and updated the location column to avoid using `Cell`.
Thank you in advance for re-reviewing!

packages/webapp/src/containers/Animals/Inventory/index.tsx
SayakaOno: The location text style is unique 😅

Duncan-Brain: Hmm 🤔 .  Yeah, it is unique.... Is it designed to be different than all other cells? It seems strange to me to have one cell have different size, color, and weight of text. 

The whole point of `<Cell />` was to enforce the same text and other styles -- its almost useless otherwise haha.

SayakaOno: I believe the difference is intentional. I don’t have a strong sense of design, but it doesn’t seem strange to me :)

antsgar (approved):

Duncan-Brain (approved):

SayakaOno:
I've just resolved conflicts. Thank you! 🙏

antsgar (approved):

PR #3530, opened 2024-11-16
Lf 4483 create selected animals summary component

Duncan-Brain:
**Description**

This creates the animal summary component --  ~~found out yesterday Loic considers this a nice to have. But it is complete now so~~ 🤷  It seems back in scope for read-only + completion views.

Notes:
- There is a google maps crash on timeout + expanding item -- based on a previous encounter I think it will resolve in production as relating to react strict mode and the fact our google maps package is maybe out of maintenance - https://github.com/LiteFarmOrg/LiteFarm/pull/3059#issuecomment-1877138127
- Things I am not super comfortable with about the solution:
  - a) re-using the AnimalInventory container and making variants (views) of it, I think containers are meant to be single use for the most part so I might have preferred to split this up more.
  - b) double use of `useAnimalInventory()` -- one for ExpandableAnimalInventory to calculate animalCount and one for regular animal inventory use. I tried to conditionally render the expandable in a single component but the container height got twitchy when selecting an animal on previous animal selection screen.

Jira link: [LF-4483](https://lite-farm.atlassian.net/browse/LF-4483)

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

[LF-4483]: https://lite-farm.atlassian.net/browse/LF-4483?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


antsgar:
@Duncan-Brain when you branch off of another feature branch, my suggestion is that you set that branch as target for your PR. That way it'll show the right diff and it can be reviewed even if the other PR hasn't been merged yet. Once the other PR is merged, GH will automatically change the target branch to integration for you 🙌

packages/webapp/src/containers/Animals/Inventory/useFilteredInventory.ts
```diff
@@ -22,7 +22,19 @@ import { animalMatchesFilter } from '../../Filter/Animals/utils';
...
 import { isInactive } from '../../Filter/utils';
 
-export const useFilteredInventory = (inventory: AnimalInventory[]) => {
+export const useFilteredInventory = (
+  inventory: AnimalInventory[],
+  showOnlySelected: boolean,
+  selectedInventoryIds: string[],
+) => {
+  if (showOnlySelected) {
+    return useMemo(() => {
```
antsgar, line 31: Doesn't this break the rules of hooks since you're calling a hook within a conditional? https://react.dev/reference/rules/rules-of-hooks#only-call-hooks-at-the-top-level

I think we could just add a condition to `animalOrBatchMatches` like so

```js
const animalOrBatchMatches =
        showOnlySelected && selectedInventoryIds.includes(entity.id) ||
        isInactive(animalsOrBatchesFilter) ||
        (entity.batch
          ? animalsOrBatchesFilter[AnimalOrBatchKeys.BATCH]?.active
          : animalsOrBatchesFilter[AnimalOrBatchKeys.ANIMAL]?.active);
```

packages/webapp/src/containers/Animals/Inventory/index.tsx
```diff
@@ -288,4 +340,43 @@ function AnimalInventory({
   );
 }
 
+export function ExpandableAnimalInventory(props: ExpandableAnimalInventoryProps) {
+  const { t } = useTranslation();
+  const { inventory } = useAnimalInventory();
```
antsgar, line 345: Following up on the idea I mentioned above of having different parent components for the different "versions" of the inventory, we could set `inventory` as a prop for the `AnimalInventory` container, and then call `useInventory` from the parent components. So we'd end up with something like:

```js
const AnimalsInventorySummary = () => {
  const { inventory, isLoading } = useAnimalInventory();
  const animalCount = ...;
  <ExpandableAnimalInventory
    ...
    expandedContent={
      <AnimalsInventory
        inventory={inventory}
        isLoading={isLoading}
        tableMaxHeight={...}
        tableSpacerRowHeight={...}
        ...
       />
    }
  />
}

const AnimalsInventoryTaskSelection = () => {
  const { inventory, isLoading } = useAnimalInventory();
  <AnimalsInventory
    inventory={inventory}
    isLoading={isLoading}
    tableMaxHeight={...}
    tableSpacerRowHeight={...}
    ...
   />
}
```
so the double call of `useAnimalInventory` would also be solved

antsgar (commented):
Left a few suggestions re code structure but this is looking amazing, I love it!

packages/webapp/src/containers/Animals/Inventory/index.tsx
```diff
@@ -288,4 +340,43 @@ function AnimalInventory({
...
+  const animalCount = props.preSelectedIds.reduce((acc, cur) => {
+    return acc + (inventory.find((animalOrBatch) => animalOrBatch.id === cur)?.count || 0);
+  }, 0);
```
antsgar, line 348: This is totally okay but I think I personally would find this a bit more readable

```js
const animalCount = inventory.find((animalOrBatch) => props.preSelectedIds.includes(animalOrBatch.id)).length;
```

packages/webapp/src/components/Animals/Inventory/index.tsx
```diff
@@ -83,71 +83,118 @@ const PureAnimalInventory = ({
...
+          tableMaxHeight: !isDesktop || !containerHeight ? undefined : containerHeight - usedHeight,
+          tableSpacerRowHeight: isDesktop ? 96 : 120,
+          showInventorySelection: isAdmin,
+          showSearchBarAndFilter: true,
+          alternatingRowColor: true,
+          showTableHeader: isDesktop,
+          showActionFloaterButton: isAdmin,
+        };
+    }
+  };
```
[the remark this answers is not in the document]
Duncan-Brain: Yeah I hated this part of it the solution. Good call out. I will do that first suggestion.

My instinct was to make more containers and bundle the reused logic into a hook -- but it seemed liek it would take too long. But ideally we would reuse components and not containers. That would make components like KPI and FloatingActionBar not even be possible for task animals Inventory.

packages/webapp/src/containers/Animals/Inventory/useFilteredInventory.ts
Duncan-Brain: Wow great point about breaking the hook rules that was silly of me. I dont know why I didn't get a warning either. Will do that!

packages/webapp/src/containers/Animals/Inventory/index.tsx
Duncan-Brain: Its got to be `count` not `length` because of batches so I don't think I can do this.

antsgar: Ah!! I totally missed the count part, makes sense

packages/webapp/src/components/Expandable/styles.module.scss
```diff
@@ -36,3 +37,15 @@
...
 }
+
+.pill {
+  height: 24px;
+  padding: 8px 16px;
+  background: var(--Colors-Neutral-Neutral-50);
+  color: var(--Colors-Neutral-Neutral-600);
+  width: 100%;
+  justify-content: center;
+  @include lg-breakpoint {
```
kathyavini, line 48: The mobile layout should extend to 1200px? Or should `@include` here and in .alwaysVisibleContent be sm-breakpoint?

kathyavini (commented):
I think this is in progress based on the comments thread, and I'm not done reading the files either, but just wanted to leave a few small notes! It looks good!

packages/webapp/public/locales/en/translation.json
```diff
@@ -1941,6 +1941,7 @@
     "ADD_TASK_FLOW": "task creation",
     "AMOUNT_TO_ALLOCATE": "Amount to allocate",
     "ANIMAL_MOVING_TO_LOCATION": "Moving to:",
+    "ANIMAL_MOVEMENT_EXPANDING_SUMMARY_TITLE": "See detail list of animals to move",
```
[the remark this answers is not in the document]
Duncan-Brain: I think a list of details could be a "detail list" where "detail list" is the object of the sentence. "See a detailed list.. " makes just "list" the object of the sentence with detailed describing it. Its not common but not wrong either I don't think.

packages/webapp/src/containers/Animals/Inventory/useFilteredInventory.ts
Duncan-Brain: Tried it the suggested way, but the other matches returned extra results which was not desired. 

Because selectedIds is likely to change more frequently I just moved it out of the conditional but kept its own separate useMemo.

packages/webapp/src/components/Expandable/styles.module.scss
Duncan-Brain: Yes because thats what animal inventory is 🤷 .. it needs to match.

packages/webapp/src/components/Expandable/ExpandableItem.jsx
```diff
@@ -52,31 +54,35 @@ export default function ExpandableItem({
...
+            {mainContent}
+          </div>
+          <div
+            onClick={onIconClick}
+            className={clsx(styles.iconWrapper, iconClickOnly && styles.clickable, classes.icon)}
+          >
+            {icons[isExpanded ? 'up' : 'down']}
+          </div>
         </div>
+        {pillBody && leftCollapseIcon && <Pill body={pillBody} className={styles.pill} />}
```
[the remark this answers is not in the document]
Duncan-Brain: It was a real pain to get styles right... I really tried to avoid editing expandable item.... :( 

I have added stories and moved styles to expandable item. But now I am thinking it would be better served by offering variants...

packages/webapp/src/components/Animals/Inventory/index.tsx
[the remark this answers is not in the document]
Duncan-Brain: So I am still not happy with the solution, but I think it is at least better if you could check it out.

Duncan-Brain:
So I am still not super happy with the size of the `AnimalInventory` container -- it almost feels like I would rather duplicate code than have it be used for all these cases... But it works and I think it somewhat resembles `TaskLocations` style component switching.

packages/webapp/src/containers/Animals/Inventory/index.tsx
Duncan-Brain: Let me know what you think of the newer refactor... its still not optimal. Theres just so much shared logic I found it hard to split away without making it look more complex.

packages/webapp/src/components/Expandable/styles.module.scss
kathyavini: Okay I'll make sure to leave a note in the bug bash table about the mobile view inventory at 800-1200px; looks totally bonkers to me (bonkers on the headers + padding I mean, not the table rows)

packages/webapp/src/components/Expandable/ExpandableItem.jsx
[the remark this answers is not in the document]
kathyavini: Thank you for the Storybook! ❤️ 

Yeah, I think ultimately the variant we want would maybe just handle the two line wrapping + centring without being particular that the wrapping bit has to be a pill, but I think it's fine to leave like this until we know in what context (if any) we'll see it again.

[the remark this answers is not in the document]
Duncan-Brain: Thanks for the reminder!

packages/webapp/src/containers/Animals/Inventory/useFilteredInventory.ts
```diff
@@ -79,4 +94,10 @@ export const useFilteredInventory = (inventory: AnimalInventory[]) => {
     groupsFilter,
     locationsFilter,
   ]);
+
+  if (showOnlySelected) {
```
kathyavini, line 98: So if I'm understanding correctly, the filterMatches are never used in this case? Then I think this conditional could go at the top of the hook (above the useMemo) so that the useMemo logic doesn't have to run at all. It's also not necessary to include showOnlySelected in the useMemo because when it's true, the useMemo return is irrelevant, and when it's false, I think the new conditional line is skipped anyway, right? So it behaves as before?

kathyavini (commented):
The component looks totally 🔥 (well except for the mobile header -- which will get fixed 😂 ) and it's hard to imagine how many CSS problems you had to solve to get this working! Personally I like how the Inventory variants are set up now -- agree its so much code in `<AnimalInventory />` but I think each variant reads clearly. And they are definitely easy to consume, having tested them in the task readonly + complete 👍

antsgar:
The refactor looks awesome, much better than before IMO! Excellent work 🔥 I'd take it one step further by letting the "variant" components own passing to BaseAnimalInventory the props that are specific to each variant and not common to all of them. I'm not sure if I'm explaining it clearly, so I pushed out an alternative branch showcasing this `LF-4483-create-selected-animals-summary-component-2` (I haven't tested it thoroughly though so please do if you decide to move forward with that change)

kathyavini:
Sorry I think I missed something -- it might be what you referred to here

> here is a google maps crash on timeout + expanding item

but I don't think so, because it is happening on the screens that don't include the expandable, and timeout doesn't seem to be required. Actually I'm not even sure what the error is because my console is only showing warnings but there is a full crash on the location selecting step of tasks OTHER than movement, could you check if you see that?

Screenshot 2024-11-29 at 11 29 52 AM

packages/webapp/src/containers/Animals/Inventory/useFilteredInventory.ts
Duncan-Brain: Yeah thanks I moved it up and removed the excess return statement!

packages/webapp/src/containers/Animals/Inventory/styles.module.scss
```diff
@@ -29,8 +29,66 @@
...
+
+.section {
+  margin-top: 24px;
+  padding: 8px 0;
+  background: var(--White);
+  border-radius: 8px;
+  border: 1px solid var(--Colors-Neutral-Neutral-50);
+  gap: 8px;
+
+  @include lg-breakpoint {
```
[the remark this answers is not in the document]
Duncan-Brain: Makes sense! Having to use full width routes and apply weird padding styles for some components in some widths got weird.  I will let you do it since you know what you need!

packages/webapp/src/containers/Animals/Inventory/useFilteredInventory.ts
```diff
@@ -32,9 +46,10 @@ export const useFilteredInventory = (inventory: AnimalInventory[]) => {
     [AnimalsFilterKeys.LOCATION]: locationsFilter,
   } = useSelector(animalsFilterSelector);
 
-  return useMemo(() => {
+  const filterMatches = useMemo(() => {
     return inventory.filter((entity) => {
       const animalOrBatchMatches =
+        (showOnlySelected && selectedInventoryIds.includes(entity.id)) ||
```
kathyavini, line 52: This line is unnecessary too because it will always be false at this step (due to early return)

Duncan-Brain: yup forgot that in too when I was testing the other solution! 🙏

kathyavini (approved):
> I was using logic for styles like it was className but doesnt work for style or classes obviously 

I actually got confused in the exact same way looking at the console from the crash! I saw the console warning related to the styling, but thought "that's not the reason for the crash; classnames can't crash" 😂  But I missed that it was a style object!

Looks great to me!

antsgar (approved):

PR #3532, opened 2024-11-18
LF-4371 Feedback - Add feedback drawer and icon

kathyavini:
**Description**

Thank you @kevin-wu01 for doing the bulk of this ticket, including the integration into App! 🙏

This ticket covers the side drawer component and the menu icon that calls it up. I did throw the existing feedback form (with no alternations) into the Drawer, but there is a **separate ticket** for the feedback from refactor, including removing the old routes and adjusting the form content, that I will open separately.

Notes:
- The Jira ticket called for `SideDrawer` to be its own component that is used in Desktop view, and the existing `Drawer` to be used in mobile view; however in the end I made SideDrawer a variant of Drawer to
  - keep the form content from being cleared upon resize across the breakpoint 
  - reduce code duplication between form and side drawer
  - not repeat the mobile calculations in both the container and the Drawer itself
- I couldn't locate a working Storybook story for the existing Drawer (there is one for NavBarDrawer but it doesn't open for me and it's not precisely for this component itself).
  - Here is the new [Storybook story for Drawer](http://localhost:6006/?path=/story/components-drawer--default), which also hopefully makes the variant organization clear
  

Jira link:  https://lite-farm.atlassian.net/browse/LF-4371

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/webapp/src/components/Navigation/SideMenu/index.jsx
```diff
@@ -233,7 +233,7 @@ const PureSideMenu = ({
         isOpen={isDrawerOpen}
         onClose={onDrawerClose}
         fullHeight
-        responsiveModal={false}
+        desktopVariant={DesktopDrawerVariants.DRAWER}
```
kathyavini, line 236: @antsgar I don't know if you remember back well enough to when the `responsiveModal` property was defined, but it doesn't seem necessary in this context because the Drawer is never a modal in mobile view anyway! Maybe Drawer or SideMenu also got a refactor since originally written?

packages/webapp/src/components/Drawer/index.tsx
```diff
@@ -46,23 +51,19 @@ const Drawer = ({
...
 }: DrawerProps) => {
   const theme = useTheme();
   const isDesktop = useMediaQuery(theme.breakpoints.up('sm'));
 
-  if (!isOpen) {
-    return null;
-  }
-
-  return isDesktop && responsiveModal ? (
+  return isDesktop && desktopVariant === DesktopDrawerVariants.MODAL && isOpen ? (
```
kathyavini, line 66: I actually destroyed the mobile drawer slide-up animation with this change, which we did discuss a little in this [PR comment thread](https://github.com/LiteFarmOrg/LiteFarm/pull/3150#discussion_r1521902923) but kind of landed on thinking it was innocuous! At the time I totally missed that applying conditional rendering to the Drawer = no animation! Luckily it never went to App, but you can see the slide up animation broken on beta/integration now!

SayakaOno (approved):
Looks great to me! Nice stories ❤️

packages/webapp/src/components/Drawer/index.tsx
antsgar: Oooh great catch!!

packages/webapp/src/components/Navigation/SideMenu/index.jsx
antsgar: Yeah since this only renders in mobile view I don't think the prop makes sense, I'm not sure why I passed it along initially or if something got refactored along the way

kathyavini:
@antsgar the styling of the form had it's own ticket in Jira and the PR for that is open here:
- https://github.com/LiteFarmOrg/LiteFarm/pull/3535

(It's only in draft because it's branched off of / depending on this one, but actually I'll undraft it because the code is done). Could you check the points you mentioned against that PR branch?

antsgar:
@kathyavini ahhh that makes sense, thanks! I'll approve this one and review the other one

antsgar (approved):

Duncan-Brain:
@kathyavini Any reason you can think of why the Happy path tests might be failing haha?

kathyavini:
@Duncan-Brain I see that same failure locally 🤔 

Screenshot 2024-11-21 at 4 19 19 PM

Could it be that cypress is still reading the elements of the off-viewport form (closed state is off viewport to the right)? The selector that is failing is suuuuuuper general:

```js
    cy.get(Selectors.REACT_SELECT)
      .find('input')
```

Duncan-Brain:
@kathyavini That makes sense to me if it is being added to both forms yah!

PR #3533, opened 2024-11-18
LF-4531: Set up new task abandonment reason for tasks with removed animals

SayakaOno:
**Description**

This PR adds support for the new abandonment reason when all animals are removed from a task. The key will be added as part of delete/remove animals/batches APIs.

- Add a migration to add constraint to validate the `abandonment_reason` in the `task` table. (This wasn't necessary, but I added it to avoid adding a wrong key. Wrong keys don't exist in the prod or beta DB, so it should be safe to add this constraint.)
- Add a translation key for the new reason
- Add the new reason to `TaskModel`
- Add the `checkAbandonTask` middleware not to allow the new key to be added from clients
- Add a test 
 
This task was simpler than expected since the keys are hard-coded in the webapp, and the read-only page and task creation page don't share the same code.

- Creation
   https://github.com/LiteFarmOrg/LiteFarm/blob/aa4be4fbba5419348083870c6046839dbcf26b5f/packages/webapp/src/components/Task/AbandonTask/index.jsx#L164-L178
   
- Read-only

https://github.com/LiteFarmOrg/LiteFarm/blob/aa4be4fbba5419348083870c6046839dbcf26b5f/packages/webapp/src/components/Task/TaskReadOnly/index.jsx#L350-L371

Jira link: https://lite-farm.atlassian.net/browse/LF-4531

**Type of change**

**How Has This Been Tested?**

Replaced abandoned task's `abandonment_reason` with `NO_ANIMALS` in the DB and checked the read-only page in the webapp.

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/api/db/migration/20241118191954_add_constraint_to_task_abandonment_reason.js
+++ b/packages/api/db/migration/20241118191954_add_constraint_to_task_abandonment_reason.js
@@ -1,0 +1,43 @@
+/*
+ *  Copyright 2024 LiteFarm.org
+ *  This file is part of LiteFarm.
+ *
+ *  LiteFarm is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  LiteFarm is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ */
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+export const up = async (knex) => {
+  const reasons = [
+    'OTHER',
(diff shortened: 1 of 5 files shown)


antsgar:
Similarly to how we do this now for other entities, I think the best way to solve the problem of disallowing wrong keys at the database level would be to have an enum table for the abandonment keys, and have the reason column have a foreign key to that table. Would be similar to what's happening now with animal movement purposes, or animal uses. I'm not sure if that migration would be trickier though due to us already having existing data, so we don't need to do this now! But just proposing that as a longer term solution

antsgar (approved):

SayakaOno:
Thank you for reviewing!

@kathyavini I thought we should have the constraint, so I added it! I think the model validation is enough.
Is there any risk in deleting and re-creating the check??

@antsgar Sounds good, I will add a note to the bug ticket about the other auto-abandon reason to consider the option!

kathyavini (approved):
@SayakaOno

> Is there any risk in deleting and re-creating the check??

I don't think there is any danger, but it would render this migration redundant (there would be no difference between having run it and not having run it). If we are trying to minimize migrations run against prod, that didn't seem ideal. I also thought though that the other key fix was coming out this release, though, which I don't think is the case...

PR #3534, opened 2024-11-19
LF-4384: End to end animal details

SayakaOno:
**Description**

- Add missing `desiredKeys` to `animalBatchController`'s `editAnimalBatches` so that all fields sent to the backend are updated
- Update `getRecordIfExists` middleware helper to exclude removed animals
  - Prevents removed animals/batches from being updated
- Update `WithStepperProgressBar` not to use `FixedHeaderContainer` for single step forms which don't have headers
  - Prevents two scrollbars from appearing in `SingleAnimalView`
- Update `MeatballsMenu` and `DropdownButton` to accept disabled prop
- Create `CustomRouteComponentProps` type to override the default history from `react-router` with the history object from the `History` library
- Update `parseUniqueDefaultId` to handle exception (it was returning `NaN` when typing a new animal type, causing `filteredUses` - the animal use options - to be undefined)
- Adjust the batch count width in `AnimalSingleViewHeader`. (Counts with more than five digits are not properly handled, but if necessary, we can create a new ticket to address this)
- Update `useAnimalOrBatchRemoval` to return the API result
  - Allows `SingleAnimalView` to handle actions based on the result
- Update `SingleAnimalView` to:
  - make the page full-width
  - use `FixedHeaderContainer`
  - implement functionality for animal removal
  - replace `history.push` with `history.back` for navigation
  - redirect to `UnknownRecord` when no animal or batch is found for the URL
  - disable the menu button when editing
  - hide meatballs menu for removed animals to prevent invalid actions (edit, remove)

NOTE: I didn't implement the floating button, as it was mentioned as a nice-to-have in the Jira ticket.

Jira link: https://lite-farm.atlassian.net/browse/LF-4384

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


SayakaOno:
Denis has found bugs in Beta.
1. Farm worker is able to enter the Animal Edit mode through UI
2. Crash

https://github.com/user-attachments/assets/fca9e0d1-ad67-4779-b108-1eb0fa5efe75

I'll try to address these in this PR 🙏

SayakaOno:
Two additional changes were made:
- Hide the meatballs menu for non-admin users
- Fix the crash seen in the recording (035f8e788056aaa9200efa98c5bf210ce4fea7b5):
  - The cause was `steps[1]` in `ContextForm`. Although it's a single form, `activeStepIndex` was incorrectly becoming `1`.
  - To resolve this, I removed the `onGoForward` call that was incrementing `activeStepIndex` in `onSave`.
  - Since `onGoForward` also handled form resetting, I moved the resetting logic to `onContinue` in `WithStepperProgressBar`, ensuring it is called when the form is saved.

Duncan-Brain:
Sorry I haven't gotten very far yet but the `uses` in read-only seems to not be present on first visit. On refresh it shows up.

packages/webapp/src/containers/Animals/SingleAnimalView/index.tsx
```diff
@@ -149,45 +164,79 @@ function SingleAnimalView({ isCompactSideMenu, history, match, location }: AddAn
       console.error(e);
       dispatch(enqueueErrorSnackbar(t('message:ANIMALS.FAILED_UPDATE_BATCH')));
     }
+  };
+
+  const getInventoryId = () => {
+    if (!selectedAnimal && !selectedBatch) {
```
antsgar, line 170: Since there's a redirect to another page on the effect above if the batch or animal doesn't exist, is this check necessary?

packages/webapp/src/containers/Animals/SingleAnimalView/index.tsx
```diff
@@ -149,45 +164,79 @@ function SingleAnimalView({ isCompactSideMenu, history, match, location }: AddAn
...
+      return '';
+    }
 
-    onGoForward();
+    const animalOrBatch = AnimalOrBatchKeys[selectedAnimal ? 'ANIMAL' : 'BATCH'];
+    return generateInventoryId(animalOrBatch, (selectedAnimal || selectedBatch)!);
+  };
+
+  const { onConfirmRemoveAnimals, removalModalOpen, setRemovalModalOpen } = useAnimalOrBatchRemoval(
+    [getInventoryId()],
```
antsgar, line 179: Perhaps we could modify `useAnimalOrBatchRemoval` to receive the actual entity IDs and whether they're a batch or animal instead of the inventory IDs? Seems more straight forward since here we're generating the inventory ID just to pass it to the hook, and then the hook has to re-parse the actual IDs from it

antsgar (commented):
Left a few minor comments but this is great, thanks for catching all the little details that were missing to make the feature work!

When an animal is removed from the single animal view, the success toast reads "Successfully removed selected animals". Should we change this to "animal" singular when there's only one animal?

packages/webapp/src/containers/Animals/SingleAnimalView/index.tsx
SayakaOno: I think I was hoping to avoid doing `(selectedAnimal || selectedBatch)!` on line 175. It didn't work as expected, so I removed it!

packages/webapp/src/containers/Animals/SingleAnimalView/index.tsx
SayakaOno: I hesitated to refactor since it would involve changes to an unrelated file, but it’s true that I wasn’t happy with the inefficiency. Let me open another PR for this!

packages/webapp/src/containers/Animals/SingleAnimalView/index.tsx
```diff
@@ -149,45 +164,79 @@ function SingleAnimalView({ isCompactSideMenu, history, match, location }: AddAn
...
+    () => ({}),
```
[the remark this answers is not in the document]
SayakaOno: I didn't feel like adding `?` all over the hook 😅 but I agree that's the right thing to do!
I’ve fixed this for now, and the refactor should tidy things up a bit!

SayakaOno (commented):
Thank you for reviewing!!

The bug Duncan found was caused by `defaultValue` being set after the ReactSelect was rendered. I’ve updated it to render the form only after the default values are ready.([commit](https://github.com/LiteFarmOrg/LiteFarm/pull/3534/commits/e599d65c9a722bfe4ee94b7c3bbc0592a1f4e0f5))

Regarding making the success message singular, I noticed Denis created a bug ticket for the inventory view, so I'll leave the fix for that [ticket](https://lite-farm.atlassian.net/browse/LF-4190).

I plan to refactor the `useAnimalOrBatchRemoval` hook in a separate PR, but this one should now be ready for re-reviews. Thank you! 🙏

antsgar (approved):

packages/webapp/src/types/index.ts
```diff
@@ -13,8 +13,19 @@
...
+import { History } from 'history';
+
 export enum OrganicStatus {
   'NON_ORGANIC' = 'Non-Organic',
   'TRANSITIONAL' = 'Transitional',
   'ORGANIC' = 'Organic',
 }
+
+// Custom RouteComponentProps with a custom history type from the 'history' library
+export type CustomRouteComponentProps<T extends { [K in keyof T]?: string | undefined }> = Omit<
```
kathyavini, line 26: Thank you!! ❤️

Duncan-Brain (dismissed):
@SayakaOno 

Everything is looks great to me thanks for wrapping up! Left a couple small inline comments which aren't blockers.

~~Only other thing I saw not working was the upload image functionality -- not sure if that is addressable here.~~  nvm gotta start bucket 😂 

Feel free to merge!

packages/webapp/src/components/Form/ContextForm/WithStepperProgressBar.tsx
```diff
@@ -109,6 +117,7 @@ export const WithStepperProgressBar = ({
     if (isFinalStep) {
       setIsSaving(true);
       await handleSubmit((data: FieldValues) => onSave(data, onGoForward, setFormResultData))();
+      reset(getValues());
```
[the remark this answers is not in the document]
SayakaOno: `reset()` will reset the form to the default values. However, `reset(getValues())` updates the default values to the current form values before resetting, so this behaviour is expected. You can find a brief explanation in the RULES section on this [page](https://react-hook-form.com/docs/useform/reset).

Thank you Duncan for thoroughly testing this, I don't think `reset`, `setIsSaving` and `setIsEditing` shouldn't be called unless the API call is successful. Let me see if I can fix this!

packages/webapp/src/containers/Animals/Inventory/useAnimalOrBatchRemoval.ts
```diff
@@ -120,40 +125,42 @@ const useAnimalOrBatchRemoval = (
...
 
     setRemovalModalOpen(false);
+    return result;
   };
 
-  const onConfirmRemoveAnimals = (formData: FormFields) => {
+  const onConfirmRemoveAnimals = async (formData: FormFields) => {
     if (Number(formData.reason) === CREATED_IN_ERROR_ID) {
-      handleAnimalOrBatchDeletion();
+      return handleAnimalOrBatchDeletion();
```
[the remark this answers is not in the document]
Duncan-Brain: Is there an article or something you read about it? It makes sense but I tried looking it up and cant find anything

packages/webapp/src/components/Form/ContextForm/WithStepperProgressBar.tsx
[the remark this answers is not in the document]
Duncan-Brain: Interesting use of reset! and sounds good!

packages/webapp/src/containers/Animals/Inventory/useAnimalOrBatchRemoval.ts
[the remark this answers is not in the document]
SayakaOno: This page was convincing enough for me: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function
_"Async functions always return a promise"_ and code examples don't have `await`.

SayakaOno:
I made some additional changes, thank you!
- Conflicts were resolved.
- Revert the `onSave` change in `ContextForm`. (removing `onGoForward` should prevent the crash)
- Update `AnimalBreedSelect` to replace `undefined` with `null` as value

packages/webapp/src/components/Animals/AddAnimalsFormCard/AnimalSelect.tsx
```diff
@@ -96,7 +96,7 @@ export function AnimalBreedSelect<T extends FieldValues>({
           placeholder={isTypeSelected ? undefined : t('ADD_ANIMAL.BREED_PLACEHOLDER_DISABLED')}
           isDisabled={!isTypeSelected || isDisabled}
           onChange={(option) => onChange(option)}
-          value={value}
+          value={value || null}
```
SayakaOno, line 99: `undefined` doesn't clear the select, but `null` does.

https://github.com/user-attachments/assets/6d3d6086-49bb-4eb3-a5d5-1fcae0f43a11

packages/webapp/src/components/Form/ContextForm/index.tsx
```diff
@@ -108,10 +107,6 @@ export const ContextForm = ({
       setIsEditing={setIsEditing}
       showCancelFlow={showCancelFlow}
       setShowCancelFlow={setShowCancelFlow}
-      onSave={(...args: any) => {
-        setActiveStepIndex(0);
```
SayakaOno, line 112: I found this has changed behaviour of the other form (it displays the first page of the form for a second). I believe it's working without it, but if any problems are found, we will need another solution.

https://github.com/user-attachments/assets/f10a38c1-b5e1-418b-958c-27a457b4e79d

Duncan-Brain (approved):
Thanks for fixing error behaviour! 

Only thing I think I see now is that for some reason the sex_detail input is not allowing enabled in edit mode unless count is changed. We could probably make a bug ticket and just merge this though!

SayakaOno:
Thank you @Duncan-Brain! I think that's one of the things Denis was mentioning yesterday, let me check with him!

antsgar:
@SayakaOno I'm merging this one and we can create a separate one for the bug if that's okay!

PR #3509, opened 2024-10-22
Lf 4476 create new animal inventory step component and route

Duncan-Brain:
**Description**

Disables 
- KPI, 
- actionMenu, 
- floater button, 
- links in existing inventory columns, 
- space at bottom of table for actionMenu (would be nice to remove that in the future for all views)

Adds 
- new route for animal inventory selection in tasks
- rowClick select instead of following link
- to persisted paths
- view specific styling

Notes:
- some menus are rendered in container some in component
- Thanks @antsgar for the in depth Jira ticket instructions!

Jira link: [LF-4476](https://lite-farm.atlassian.net/browse/LF-4476)

**Type of change**

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

**How Has This Been Tested?**

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [ ] Passes test case
- [ ] UI components visually reviewed on desktop view
- [ ] UI components visually reviewed on mobile view
- [x] Other (please explain): 
- overwrite transplant task flow
- console log persisted form data on taskLocations page

**Checklist:**

- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] The precommit and linting ran successfully
- [ ] I have added or updated language tags for text that's part of the UI
- [ ] I have added "MISSING" for all new language tags to languages I don't speak
- [ ] I have added [the GNU General Public License](https://lite-farm.atlassian.net/l/cp/BT0Dd7WW) to all new files

[LF-4476]: https://lite-farm.atlassian.net/browse/LF-4476?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3535, opened 2024-11-19
LF-4372 Feedback - move feedback form and delete old routes

kathyavini:
**Description**

This is the second part of the feedback form and is branched off the first part (#3532). It removes the old route and menu that directed to the feedback form, and adjusts the styling of the inner form.

Styling adjusts were a compromise between what is in Figma and not over-scoping the refactors for this particular ticket.

Implemented
- `'md'` size `<Button />` I think maybe this size wasn't yet present in Figma when the new colours were updated?
  - Also [Updated Button Storybook](http://localhost:6006/?path=/story/components-button--small-with-icon) to `sm`, `md`, and default clearer
- File upload now uses the new component
- Inline styles were removed from individual components. `<Form />`/`<Layout />` still expect a style object instead of CSS classes, however -- I did not touch that behaviour

Not implemented
- Anything that touches on the base form components used throughout App
  - Label colours in Figma have been off for a really long time (= are a lighter shade of grey) relative to what we're keeping in App
  - Didn't implement bold styling on the selected radio or the email input, although I like how those look!
- Ideally I would have liked to see Button size as an enum and medium set as default, but there are so many Buttons in app using the current prop setup with `'sm'` and I didn't want to touch refactoring all of them

Jira link: https://lite-farm.atlassian.net/browse/LF-4372

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/webapp/src/components/Form/form.module.scss
```diff
@@ -2,4 +2,5 @@
   display: flex;
   flex-direction: column;
   flex: 1;
+  height: 100%;
```
kathyavini, line 5: @Duncan-Brain I saw in the git history that you were just recently working in the styling of Form & Layout in #3509? I used your new prop `fullWidthContent` to drop the padding on the form (👍) but to put the footer buttons at the bottom of the screen -- which it seems they are intended for, based on the JSX below -- I also needed this property.

```js
<Footer
  style={{
    bottom: 0,
    position: hasWhiteBackground ? 'sticky' : 'relative',
    ...classes.footer,
  }}
  hasWhiteBackground={hasWhiteBackground}
>
```

Just wanted to check that this doesn't break whatever adjusts you were just making. Seems okay on the movement task flow, but I'm not sure what I should be checking for!

packages/webapp/src/components/Drawer/style.module.scss
```diff
@@ -67,6 +67,7 @@
 
 .sideDrawer {
   top: var(--global-navbar-height);
+  max-height: calc(100vh - var(--global-navbar-height));
```
kathyavini, line 70: I missed this on the previous PR, but this is used in conjunction with 100% content (on Drawer + on the Form it contains) to make sure that the Layout footer (= the form buttons) end up on the bottom of the screen, rather than slightly below it!

packages/webapp/src/containers/Help/index.jsx
```diff
@@ -1,30 +1,34 @@
...
 
   const handleSubmit = (file, data) => {
     dispatch(startSendHelp());
     dispatch(supportFileUpload({ file, form: data }));
   };
-  const handleBack = () => {
-    history.push('/');
+  const onCancel = () => {
+    closeDrawer?.();
+    setKey((prev) => prev + 1); // Remount form when closing drawer
```
kathyavini, line 17: This was not my first choice, but `reset()` on the form populated the error object (and so all the validation error messages), which persisted upon opening the form again. I tried all the [properties on reset](https://react-hook-form.com/docs/useform/reset) before deciding to remount.

packages/webapp/src/components/Form/form.module.scss
Duncan-Brain: I added it just for mobile view to make the inventory rows full width. It looks good on your branch I think! 

The only thing height could maybe mess up is the max-height of the inventory flex box. It is possible I used a calc() to keep the page from overflowing.

packages/webapp/src/containers/Help/index.jsx
antsgar: `reset` actually works properly here because it does reset the form state in its entirety. The problem is on the rendering logic. If we don't do this key update, when the form is cleared and the values change, the component will rerender, but it won't remount. Every time the form rerenders, React Hook Form will re calculate the errors (so the problem isn't solved by clearing out the errors when canceling, either). The key update fixes the problem, but in a drastic way -- it just unmounts the whole thing, remounts and starts over.

I think the correct way to solve this is to fix the error rendering logic in the component by only showing the error messages if the fields are actually dirty, e.g.

```js
        {errors[SUPPORT_TYPE] && SUPPORT_TYPE in dirtyFields ? (
          <Error>{t('HELP.REQUIRED_LABEL')}</Error>
        ) : (
          ''
        )}
```

where `dirtyFields` is a property in `formState`.

You'll also need to set a defaultValue of `null` for the support type field so that it resets and clears out properly, since `undefined` won't work.

antsgar (commented):
Looks awesome, love it! 😍 

On submit, the drawer should also close and the form should clear out -- when I tested it this only happened for me on cancel.

packages/webapp/src/components/ImagePicker/index.tsx
```diff
@@ -44,16 +44,19 @@ type CommonProps = {
...
 type CustomFileUpload = CommonProps & {
   onSelectImage?: never;
   onFileUpload: OnFileUpload;
+  shouldReset?: never;
 };
 
 type DirectImageUpload = CommonProps & {
   onSelectImage: (file: File) => void;
   onFileUpload?: never;
+  shouldReset?: boolean;
```
kathyavini, line 59: @SayakaOno from what you remember of working in ImagePicker, does this look right? In my understanding of DirectImageUpload, the previewUrl is never controlled, just the file blob itself? So if the parent form needs to reset the preview from the outside (on form clear) it needs something like this?

Duncan-Brain (commented):
Looks mostly good to me! I left a couple comments. 

Main thing I saw that I was not sure about was the onSubmit behaviour. Essentially, on submission the drawer immediately closes -- I can open it back up and see the "is Submitting" button text. Then the pop-up appears. 

I think I would expect the drawer to only close and clear after the successful submission. Then the success popup appears.

When I test a failed submission (api broken/off), I would expect the values to remain in the form and the form stay open. I see this instead with a not informative snackbar:

Screenshot 2024-11-26 at 3 10 09 PM

packages/webapp/src/components/ImagePicker/index.tsx
```diff
@@ -66,12 +69,19 @@ export default function ImagePicker({
...
   onFileUpload,
   isDisabled = false,
+  shouldReset,
 }: ImagePickerProps) {
   const [previewUrl, setPreviewUrl] = useState(defaultUrl);
   const [showFileSizeExceedsModal, setShowFileSizeExceedsModal] = useState(false);
   const dropContainerRef = useRef<HTMLDivElement>(null);
   const { t } = useTranslation();
 
+  useEffect(() => {
```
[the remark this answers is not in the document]
kathyavini: Hmm, yes that was my original thought with the key, this is basically the exact use case for 'fresh form'! I'm curious what Sayaka thinks, but I thought the hook was specific to CustomImageUpload and it wasn't necessary or desired that a hook always be used with the component.

packages/webapp/src/components/Drawer/style.module.scss
kathyavini: Hmmm, so you mean you would expect the buttons fixed to the bottom of the drawer the way the header of the drawer is fixed to the top? I think it could definitely be done, but the refactor would be pretty significant because all the form is set up in that inner component. How important do you think that distinction is? Should we check with Loïc tomorrow morning?

kathyavini:
> I think I would expect the drawer to only close and clear after the successful submission. Then the success popup appears.

@Duncan-Brain that is a really good point, thank you! I was thinking the behaviour should be the same as onCancel, but really for submission it needs be tied to the saga, same as the snackbar. I'll see if I can rework that!!

packages/webapp/src/components/ImagePicker/index.tsx
SayakaOno: Sorry I needed some time to think this through...

If we had more time, I think it would be better to refactor the component to either:
1. Accept a `ref` and clear the file from the parent, similar to how it’s done for ReactSelect here:
     Screenshot 2024-11-26 at 1 19 15 PM
2. Take file and/or previewUrl as props

I don’t think the hook can handle image removal without these refactors.
That said, refreshing the form seems like the best approach in this case!

SayakaOno:
I didn’t review the code or read the comments, so I'm sorry if this has already been reported.
When clicking the "Cancel" button, `onError` is being called; it seems like the form is being submitted upon canceling.

packages/webapp/src/components/Drawer/style.module.scss
Duncan-Brain: No, I think I expect the scroll bar to extend the full height of the drawer and the bottom white bar not to exist. 

This is not a blocker for me at all -- if you think its not a problem please ignore me haha

kathyavini:
> I didn’t review the code or read the comments, so I'm sorry if this has already been reported.
When clicking the "Cancel" button, onError is being called; it seems like the form is being submitted upon canceling.

@SayakaOno ah no I don't think it was brought up yet, thank you 🙂 It's our old friend type='button' missing from the cancel!

packages/webapp/src/containers/Help/index.jsx
kathyavini: > Since reset is using the same defaultValues, it seems there should be a way to get the full form state (errors inclusive) to the original pristine state, as if remounting.

Oh my goodness @SayakaOno solved the mystery -- the cancel button was submitting 🙇 🙇 🙇 🙏🙏

antsgar:
Ohhhhh good catch with the button change causing the errors to be triggered, fascinating!! Shouldn't buttons always have a type=button? Anyways, this looks good to me ☺️

antsgar (dismissed):

packages/webapp/src/components/Form/Button/index.tsx
```diff
@@ -21,6 +21,7 @@ type ButtonProps = {
   color?: 'primary' | 'secondary' | 'secondary-2' | 'secondary-cta' | 'warning' | 'error' | 'none';
   children?: ReactNode;
   sm?: boolean;
+  md?: boolean;
```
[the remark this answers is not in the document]
kathyavini: Oh sorry I thought I responded to this! The md style will override (except for font-size). It's not ideal, but it's not very dangerous either.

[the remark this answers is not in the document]
Duncan-Brain: I've seen some glitchy issues with overrides, normally with regards to an `auto` setting like height in combination with other properties. Not very dangerous but ideally not possible either.

Duncan-Brain (approved):

antsgar (approved):

PR #3538, opened 2024-11-20
LF-4529 update default types count when adding animals

antsgar:
**Description**

Update KPIs when animals are created. This is slightly different from the approach in #3500 because it only updates the count endpoint for default types, thus preventing having to re-fetch all of the types all the time. This way we keep one of the advantages of separating default and custom types which is considering default types somewhat "static" and not having to re-fetch them after user updates.

Jira link:
https://lite-farm.atlassian.net/browse/LF-4529

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/containers/Animals/Inventory/KPI/index.tsx
+++ b/packages/webapp/src/containers/Animals/Inventory/KPI/index.tsx
@@ -18,6 +18,7 @@ import { useTranslation, TFunction } from 'react-i18next';
 import {
   useGetDefaultAnimalTypesQuery,
   useGetCustomAnimalTypesQuery,
+  useGetDefaultAnimalTypesCountQuery,
 } from '../../../../store/api/apiSlice';
 import { PureTileDashboard, TypeCountTile } from '../../../../components/TileDashboard';
 import useQueries from '../../../../hooks/api/useQueries';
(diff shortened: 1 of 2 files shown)


Duncan-Brain:
This seems to work fine... but I am not quite sure I understand the benefit here vs how it was done previously. I don't think that just the count is being updated. What I am seeing is that now there are two state entries in the store with same data:
 
Screenshot 2024-11-20 at 10 50 59 AM

I definitely see value in the "keeping default types somewhat static" -- because it is an enum table and so farm specific logic feels odd to me on this specific endpoint.  

But the only way I understand that happening is to somehow move count to a new route or option on /animals + /batches that returns derived statistics or something based on animal inventory.

antsgar:
@Duncan-Brain thank you for the feedback! I think you're right and I didn't think this through. The reason this isn't really beneficial is IMO not related to the routes, but to a combination of:
- The endpoint returning all of the type entities (id + key) even if the count parameter is sent. Ideally when the count parameter is sent we should return only id + count, since that's all we need for counts.
- The type entity is really small and only has one field additional to the id, so even if we changed the above, there wouldn't be much of a benefit in not re-fetching just the key. If types were a giant entity, then it would be beneficial to only re-fetch the ID and the count and not the rest of the fields.
Based on that we can keep this as is and close this PR!

PR #3540, opened 2024-11-20
LF-4474 [Nice to have] convert create task button to CTA floating button

kathyavini:
**Description**

This PR removes the "+ Create a task" text link on TaskPage and moves that functionality to a FloatingActionButton. There was also a check for locations and a creation location modal tucked into the `<TaskCount />` component that were lifted up to the parent `<TaskPage />`. This was by necessity, but I think it's also more logical; I was personally surprised to find that logic co-located with the count display!

Jira link: https://lite-farm.atlassian.net/browse/LF-4474

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/webapp/src/containers/Task/styles.module.scss
```diff
@@ -0,0 +1,21 @@
...
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ */
+
+.buttonWrapper {
+  position: fixed;
+  bottom: 16px;
+  right: 16px;
+  z-index: 100;
```
antsgar, line 20: I wonder if we should set a z-index on the floating button component itself since it's supposed to always float over everything else?

packages/webapp/src/containers/Task/index.jsx
```diff
@@ -170,47 +187,60 @@ export default function TaskPage({ history }) {
...
+            isAdmin ? t('LOCATION_CREATION.TASK_BODY') : t('LOCATION_CREATION.TASK_BODY_WORKER')
+          }
+          dismissModal={dismissLocationCreationModal}
+          isAdmin={isAdmin}
+        />
       )}
-    </Layout>
+
+      <div className={styles.buttonWrapper}>
+        <FloatingActionButton type={'add'} onClick={handleAddTask} />
```
antsgar, line 242: I think you brought this up before, but do we want the button to trigger the action directly, or to open the floating menu like we're doing with Animals? It does make sense to just trigger the action if there's only one option, but we could double check with Loïc (if we haven't) and if this is the way make it consistent in Animals as well.

antsgar (approved):
A small change but makes such a difference! Love it ☺️

Duncan-Brain:
Looks like the position is maybe different on animals or something? Only other thing is it would be good to keep the "Add a task" translated text for screen readers and hover or something as discussed in design.

Screenshot 2024-11-20 at 11 12 30 AM
Screenshot 2024-11-20 at 11 12 42 AM

packages/webapp/src/containers/Task/styles.module.scss
kathyavini: That makes sense to me. Do you think the positioning CSS should even be controlled by a prop on the component if we're using it so often in this corner? 🤔

packages/webapp/src/containers/Task/index.jsx
kathyavini: Thanks for bring it up! As discussed with Loïc today, I'll make them both one-click (no menu).

kathyavini:
@Duncan-Brain that's a great idea about the screen readers! I kept the string deletion in one commit just in case it had to be reverted 😉 

I'm not sure what you're seeing with the positioning though hmm 🤔 For me if I switch between task and inventory the button doesn't move at all (which should be as expected; I took the positioning CSS from FloatingMenuButton). Can you check you're definitely on the base branch, updated version?

kathyavini:
@Duncan-Brain interesting, I can't even reproduce at tablet widths. I just updated the component in inventory as we are no longer using the menu. Does it make any difference?

Duncan-Brain (dismissed):
No.. same issue -- I think it need to be moved up one spot in the html element hierarchy.

Aria label looks good thanks for adding it to inventory too!

packages/webapp/src/containers/Task/styles.module.scss
antsgar: Oh yes, that's a really good idea!

antsgar (approved):

PR #3541, opened 2024-11-20
LF-4415 Animals IDs not sorted in order

Tbrid:
**Description**

Fix the issue of animals IDs not sorted in ascending order in the inventory page.

Jira link: https://lite-farm.atlassian.net/browse/LF-4415

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


Duncan-Brain (commented):
Hey @Tbrid thanks for doing this one! 

I am not quite sure I know what is the expected order... so I will ask some questions on the Jira ticket. 

I think the length check works well specifically for the ID##'s but because the names are sometimes chosen as an identifier it is no longer quite alphabetical but based on length for those ones. The Del 6 in my picture must have been randomly sorted based on chance.

Screenshot 2024-11-21 at 3 26 20 PM


I will tag design who might have an opinion too just to get clarification.

packages/webapp/src/containers/Animals/Inventory/useAnimalInventory.tsx
```diff
@@ -214,6 +217,22 @@ interface BuildInventoryArgs {
...
+    if (a.name && !b.name) {
+      return -1;
+    } else if (!a.name && b.name) {
+      return 1;
+    } else {
+      if (a.identification.length > b.identification.length) {
+        return 1;
+      } else {
+        return a.identification.localeCompare(b.identification);
+      }
```
antsgar, line 231: @Tbrid thank you so much for contributing!!

Unfortunately I don't think this portion of the sorting function is correct, and in testing it I'm not getting the right sorting order. On one hand, we shouldn't be using the `identification` property to sort since that's a string that can be either the name, the identifier or the internal identifier, but in this case we want to sort granularly, one property at the time. On the other hand, I'm not sure I understand the reasoning behind comparing the length of the strings?

I think this is more or less what the function could look like

```js
return inventory.sort((a, b) => {
    if (a.name && !b.name) {
      return -1;
    }
    if (b.name && !a.name) {
      return 1;
    }
    if (a.identifier && !b.identifier) {
      return -1;
    }
    if (b.identifier && !a.identifier) {
      return 1;
    }
    return (
      (a.name && b.name && a.name.localeCompare(b.name)) ||
      (a.identifier && b.identifier && a.identifier.localeCompare(b.identifier)) ||
      a.internal_identifier - b.internal_identifier
    );
  });
```

but I haven't tested it thoroughly!
If you have the time, a good idea could be to add some unit tests to verify that it's working properly. Otherwise we can just test it manually and make sure it works with a variety of records.

Duncan-Brain: I get that the length check was for the internal_identifier string that includes 'ID###' . 

If the `identification` property is no longer used here (it is just for label display now) then it should be a plain number as Anto described 👍

SayakaOno:
I've updated `TableV2` to accept a custom comparator directly, instead of embedding it within `columns`, for better simplicity.

antsgar:
Works like a charm, thank you both @Tbrid and @SayakaOno for working on this! ❤️

antsgar (approved):

PR #3542, opened 2024-11-20
LF-4410: Add limit on creating animals and batches and tests

Tbrid:
**Description**

Put 1000 limit on creating animals and animal batches. Added tests on api package. Added max limit on UI count input.

Jira link: https://lite-farm.atlassian.net/browse/LF-4410

**Type of change**

**How Has This Been Tested?**

- [x] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/webapp/src/components/Animals/AddAnimalsFormCard/AddAnimalsFormCard.tsx
```diff
@@ -112,6 +114,7 @@ export default function AddAnimalsFormCard({
         <NumberInput
           name={`${namePrefix}${BasicsFields.COUNT}`}
           control={control}
+          max={ANIMAL_COUNT_LIMIT}
```
SayakaOno, line 117: Thank you for working on this!

Screenshot 2024-11-20 at 3 55 23 PM

Sorry, we should've explained this first... When the checkbox is selected and the count is 1000, 1000 animals will be created. If the checkbox is not selected, a batch with 1000 animals (one record in the DB) will be created instead. The latter is a valid case, so it would be great if you could add a condition to set a maximum limit when the checkbox is selected 🙏

Tbrid: Hi, Sayaka. So I put a hook to deal with this, but then I found it wouldn't change the displayed value in the input when I only use setValue, (i.e.  if number is 2000 and user clicks the checkbox, the value will be set to the limit, but still displayed as 2000) so I put a "value" prop for the NumberedInput, let me know what you think about it.

SayakaOno: Sorry for the late reply! Sending `value` to `NumberInput` seems good, but it looks like `value` will be automatically passed through without modifying `NumberInput`, since `...props` is added to `InputBase`. Could you double check?

Would it work if we move the logic from the `useEffect` to the checkbox's `onChange` instead?

SayakaOno: Oh right, `inputProps`'s value overrides it.

@navDhammu The number in the `NumberInput` doesn't update when the value is changed via `setValue`. Do you think we should update the logic in `useNumberInput` to handle this?

navDhammu: @SayakaOno The reason it doesn't update with `setValue` is because `NumberInput` manages its own state, so when changing value with setValue it doesn't update NumberInput's internal state, which is why you still see the previous number. 

I don't think its a good idea to update the logic inside the `useNumberInput` to handle this because then you would probably have to add a useEffect to update the state based on changes to some prop, and this isn't best practices for react. 

Better ways of handling this in my opinion:
1. Lift the state up to where its needed. This would involve refactoring out the useNumberInput hook from NumberInput and using it in each component that uses NumberInput.
2. Use a key to rerender NumberInput.

Also, sending `value` to `NumberInput` is not a good idea because it overrides the value given by `useNumberInput` which is a formatted string based on the locale. So overriding that breaks the localization feature.

SayakaOno: Thank you for your input, I'll discuss the solution with the team!

Tbrid: @navDhammu Thanks for explaining.  @SayakaOno I think the 2nd way of using a key is the easiest.

SayakaOno: I agree, but I'm hesitant to go with that approach because we may encounter the same issue in the future. Needing to add a key each time doesn't seem like a sustainable solution...

I'll bring this up in tech daily early next week (feel free to join if you'd like) and get back to you!

antsgar: A bit late to the party but chiming in here! I actually think changing the value of the input ourselves wouldn't be a good UX practice, because this happens inadvertently for the user and they get no context on why the number is changing and no visibility that there's a limit on the field. Instead, we should validate if the user's input is higher than the limit and if it is, show the input as errored and show an error message. Per [this comment](https://lite-farm.atlassian.net/browse/LF-4410?focusedCommentId=17718) the message should be “You can only add up to 1000 animals at a time”.

packages/webapp/src/components/Animals/AddAnimalsFormCard/AddAnimalsFormCard.tsx
```diff
@@ -82,6 +86,22 @@ export default function AddAnimalsFormCard({
...
+
+  useEffect(() => {
+    if (shouldCreateIndividualProfiles && watchAnimalCount > ANIMAL_COUNT_LIMIT) {
+      setError(`${namePrefix}${BasicsFields.COUNT}`, {
+        type: 'manual',
+        message: t('animal:COUNT.MAX', { max: ANIMAL_COUNT_LIMIT }),
+      });
+    } else {
+      clearErrors(`${namePrefix}${BasicsFields.COUNT}`);
+    }
```
SayakaOno, line 102: Could we use `trigger`?
```
trigger(`${namePrefix}${BasicsFields.COUNT}`)
```

packages/webapp/src/components/Animals/AddAnimalsFormCard/AddAnimalsFormCard.tsx
```diff
@@ -121,6 +141,7 @@ export default function AddAnimalsFormCard({
               value: true,
               message: t('common:REQUIRED'),
             },
+            max: shouldCreateIndividualProfiles ? countMaxValidation() : undefined,
```
SayakaOno, line 144: Could we use `hookFormMaxValidation`?

SayakaOno (commented):
We've merged some PRs, so it would be great if you could merge integration!

We can create another ticket, but I'll mention it here so we won't forget... the "X" icon doesn't clear the input.

packages/webapp/src/components/Animals/AddAnimalsFormCard/AddAnimalsFormCard.tsx
```diff
@@ -83,6 +87,14 @@ export default function AddAnimalsFormCard({
     }
   }, []);
 
+  useEffect(() => {
+    if (shouldCreateIndividualProfiles && watchAnimalCount > ANIMAL_COUNT_LIMIT) {
+      trigger(`${namePrefix}${BasicsFields.COUNT}`);
+    } else {
+      clearErrors(`${namePrefix}${BasicsFields.COUNT}`);
+    }
```
SayakaOno, line 95: The error will be cleared automatically by `trigger`, so `trigger(`${namePrefix}${BasicsFields.COUNT}`);` without any conditions should work.

It would be even better if this could be handled in the `onChange` for the checkbox. Have you tried it?  I found it was possible by adding a setTimeout like this:
```
onIndividualProfilesCheck?.((e.target as HTMLInputElement).checked);
setTimeout(() => {
  trigger(`${namePrefix}${BasicsFields.COUNT}`);
}, 0);
```

SayakaOno (commented):
Could you please also check the translation files? The only change we expect is `COUNT.MAX`. Thank you 🙏

SayakaOno:
We discussed this ticket with the team this morning, and here's my follow up.

@antsgar 
The default error message for the max value is `Please enter a value less than {{value}}`. Considering the chances that users see this error message is low, I think we can go with the default message for now to avoid the need for translation.
Also, limiting the number for batch is easy. For example:
```
hookFormMaxValidation(shouldCreateIndividualProfiles ? 1000 : 9999999)
```

antsgar:
@Tbrid we are planning a release for tomorrow and would love to include this fix in it, so due to that time constraint the core team will be taking care of applying the suggested changes to this ticket if you don't mind. Thank you so much for working on it and excellent work!! 🙌🙌🙌

Tbrid:
@antsgar Ok, no problem 👍

SayakaOno:
I've also added `max` to the general detail card to address: [LF-4596](https://lite-farm.atlassian.net/browse/LF-4596)

antsgar (approved):
Looks good! ❤️

PR #3544, opened 2024-11-22
Lf 4375 spotlight animals add animals beta spotlight

Duncan-Brain:
**Description**

Adds animals beta spotlight 

Notes:
- I filled in missing for only some strings as I learned other aren't adding `MISSING` anymore!
- I don't love prop drilling the open Feedback Survey state -- if we get more of these it might be worth making a provider.
-  600px issue (see video) really stumps me
    - there is some interplay with the animation of the side menu (commenting out the keyframes prevents this) -- in most circumstances the side menu animation has been run before getting to the animals screen (eg on home page). One can track the `animationend` event but its messy for not much value.
  

https://github.com/user-attachments/assets/e81a0c50-81b9-4195-b5f1-aafa29cff1e2

Jira link:[LF-4375](https://lite-farm.atlassian.net/browse/LF-4375)

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

[LF-4375]: https://lite-farm.atlassian.net/browse/LF-4375?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/webapp/src/containers/Animals/Inventory/AnimalsBetaSpotlight.jsx
```diff
@@ -0,0 +1,74 @@
...
+import { useDispatch, useSelector } from 'react-redux';
+import { TourProviderWrapper } from '../../../components/TourProviderWrapper/TourProviderWrapper';
+import { setSpotlightToShown } from '../../Map/saga';
+import { showedSpotlightSelector } from '../../showedSpotlightSlice';
+import { Trans } from 'react-i18next';
+import Badge from '../../../components/Badge';
+import { ReactComponent as SendIcon } from '../../../assets/images/send-icon.svg';
+import styles from './styles.module.scss';
+
+export default function AnimalsBetaSpotlight({ children, setFeedbackSurveyOpen }) {
```
antsgar, line 26: Since we'll likely be needing similar "beta" spotlights for upcoming features, I could see value in having a `BetaSpotlight` reusable component with only the more specific parameters as props, and then having that component be a child of this one. Not a blocker for merge though, can be done when we have an upcoming beta feature!

antsgar (approved):
Looks great! 🎉 left one minor comment but not a blocker.

My only thought is that the location for this spotlight means that the user won't see it unless they navigate to the Animals section, and by that point they have already noticed it so there's less impact to having a spotlight. It seems to me it'd be more useful to have it on the "Animals" text in the menu?

kathyavini:
@antsgar did you have any thoughts on the failure at 599-600px Duncan mentioned because it has me totally stumped! Or is it safe to say we don't care if one particular width is non-functional?

Screenshot 2024-11-25 at 1 02 37 PM

Screenshot 2024-11-25 at 12 50 01 PM

kathyavini (approved):
I also had the same initial thought as Anto regarding the spotlight requiring Animals to be clicked (it's almost too late at that point) but I guess if it targets the menu, there wouldn't be anything to highlight in mobile view? Well, I guess until the menu is opened for the first time? 🤔 

@Duncan-Brain I like your idea about a provider for those props! For the next iteration / beta feature!

packages/webapp/public/locales/en/translation.json
```diff
@@ -2066,7 +2068,7 @@
   "BADGE": {
     "BETA": {
       "TITLE": "Beta",
-      "CONTENT": "Get a sneak peek at our new animal management feature—now in beta! Your feedback is crucial as we refine this feature. Please share any bugs or suggestions you encounter. We appreciate your help in improving liteFarm! You can learn more about beta <a>here</a>"
+      "ANIMALS_CONTENT": "Get a sneak peek at our new animal management feature—now in beta! Your feedback is crucial as we refine this feature. Please share any bugs or suggestions you encounter. We appreciate your help in improving liteFarm! You can learn more about beta <a>here</a>"
```
kathyavini, line 2071: Okay, it's a little too late but we'll want to avoid this in the future -- renaming strings that right now are only in CrowdIn -- because CrowdIn UX is not good for it.

I *think* the original translation was preserved (only and cryptically) in the AI suggestion, but its lost the usual authorship metadata and there appears to be no way to search for it 🫤

Duncan-Brain: Yeahhh, I knew it as I was doing it but figured we will need to be able to change key names. This one is particularly long I should have checked first. But was maybe hoping they hadn't got to it in the couple days badge was in for.

Duncan-Brain:
@antsgar @kathyavini 

In the designs it seems to be shown on inventory, I think one of the nice thing about the placement is that it isn't obtrusive unless you navigate to animals. Until then it is a curiosity you can find out about when you click the menu.

For the 600px thing -- yeah I tried to describe its a very weird  thing that only happens essentially when you navigate directly to the inventory screen bypassing the menu animation on all other screens. If you see the menu animation anywhere else, it shouldn't happen.

PR #3545, opened 2024-11-22
LF-4371a Fix Cypress tests

kathyavini:
**Description**

This was extremely annoying!

The only issue with the Cypress tests was that the non-specific selectors, e.g. `cy.get(Selectors.RADIO).first()` or `cy.get(Selectors.REACT_SELECT)'` were also returning the off-screen elements included in the closed `sideDrawer` form.

Filtering on `':visible'` should have worked in theory (and did for the React Select), but two of the Radios stopped returning the visible radios once the off-screen radios were found (??!) I.e. `cy.get()` used to return the on-screen radios, but now returned only the off-screen ones before the on-screen ones were even rendered.

In fixing the radio selectors, I accidentally changed one selected value, causing an inscrutable missing element in a different test file, which wanted a termination task instead of a harvest task, although that was not clear from the error message.

Jira link: none

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

Ran Cypress locally + in CI on GitHub actions. The googlemapsAPI call wait seems to only work in CI now.

**Checklist:**

### DIFF
--- a/packages/end-to-end/cypress/e2e/crops.js
+++ b/packages/end-to-end/cypress/e2e/crops.js
@@ -49,7 +49,7 @@ describe('Crops', () => {
       .find('input')
       .type(crops['CEREALS'] + '{enter}');
 
-    cy.get(Selectors.RADIO).first().check({ force: true });
+    cy.get(Selectors.RADIO).filter('[name="can_be_cover_crop"]').first().check({ force: true });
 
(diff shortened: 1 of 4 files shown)


antsgar (approved):
I can feel your pain reading the PR description 😂 glad you were able to solve it but this seems really annoying!

PR #3547, opened 2024-11-23
LF-4475 [Nice to have] Update layout for manage custom tasks

kathyavini:
**Description**

This adds the "CantFindCustomTask" block we use in Expense + Revenue Type Selection into Tasks. It was a lower priority nice-to-have, but with the discussion around adding animals to custom tasks, it might be especially nice. There is no code complexity to the change so I have opened it.

**Done in this PR:**
- Simple conversion to TS of `<CantFindCustomTask />` because even I was confused by the meaning of the props I had set up on this component ("miscellaneousConfig"??)
- Simple [Storybook story](http://localhost:6006/?path=/docs/components-cantfindcustomtype--docs) since previously it had only been integrated within the PureFinanceTypeSelection story. This is also to make the props a bit clearer
- breakpoint adjust to fix this small layout bug which I think had just gone under our radar since the menu release:

Screenshot 2024-11-22 at 3 49 01 PM

**Not done:**
- Did not migrate `<CantFindCustomType />` out of Finances, despite its new use in tasks
- Did not make the miscellaneous functionality more general (i.e. the string is hard-coded for Expense types right now, which is still fine because it's the only place that uses that functionality)
- This was to avoid making any changes that would involve touching finances code & thereby to keep scope slim

Jira link: https://lite-farm.atlassian.net/browse/LF-4475

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

I SKIPPED ADDING MISSING! It seems the consensus is that we can do this now 😁 

### DIFF
--- a/packages/webapp/public/locales/en/translation.json
+++ b/packages/webapp/public/locales/en/translation.json
@@ -117,6 +117,7 @@
     "ASSIGN_TASK": "Assign task",
     "ASSIGNEE": "Assignee",
     "CANCEL": "task creation",
+    "CANT_FIND_INFO_TASK": "You can also create your own custom task types!",
     "CLEANING_VIEW": {
       "ESTIMATED_WATER": "Estimated water usage",
       "IS_PERMITTED": "Is the cleaning agent in the permitted substances list?",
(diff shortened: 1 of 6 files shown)


SayakaOno (approved):
Looks good to me!

antsgar (approved):
Lovely!! 🙌

PR #3552, opened 2024-11-27
LF-4545 Adjust read only task view for movement tasks

kathyavini:
**Description**

This PR completes the readonly task view for animal movement tasks. Modifications:
- The readonly animal inventory should be shown in TaskReadOnly
- The task details component (PureMovementTask) was updated for compatibility with the readonly view
- The incoming (GET) task data needed adjust to be shown in TaskReadOnly (the animal inventory) and PureMovementTask (the movement purpose)
  - As a reminder, the animals + animal batches are top level task properties while the purpose is nested under `animal_movement_task` (incoming from the backend) / `movement_task` (outgoing towards backend, and adjusted in saga)
  
Although some of the work of complete (e.g. putting the task details into StepOne and common data formatting) has been added to this branch, please consider this branch as **_only for populating readonly._** -- edit/complete will be done on the other branch.

Jira link: https://lite-farm.atlassian.net/browse/LF-4545

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/webapp/public/locales/en/translation.json
```diff
@@ -1952,7 +1952,7 @@
     "ADD_TASK_FLOW": "task creation",
     "AMOUNT_TO_ALLOCATE": "Amount to allocate",
     "ANIMAL_MOVING_TO_LOCATION": "Moving to:",
-    "ANIMAL_MOVEMENT_EXPANDING_SUMMARY_TITLE": "See detail list of animals to move",
+    "ANIMAL_MOVEMENT_EXPANDING_SUMMARY_TITLE": "Animals in this task",
```
kathyavini, line 1955: The original string had just gone up to CrowdIn less than an hour ago, so I went ahead and _manually adjusted_ this on CrowdIn (so translators will not have to translate the now irrelevant string).

Since this is already on CrowdIn, I will remove the "new translations" tag from this branch.

packages/webapp/src/components/Task/TaskReadOnly/index.jsx
```diff
@@ -347,6 +355,20 @@ export default function PureTaskReadOnly({
         </div>
       )}
 
+      {isAnimalTask && (
+        <div style={{ marginBottom: '24px' }}>
```
SayakaOno, line 356: Could we use className?

packages/webapp/src/components/Task/TaskReadOnly/index.jsx
```diff
@@ -347,6 +355,20 @@ export default function PureTaskReadOnly({
...
+          <Semibold>{'Animals'}</Semibold>
```
SayakaOno, line 357: This text needs to be a translation key :)

SayakaOno (commented):
Looking great!!

I had a similar bug in the animal read-only view where "uses" are not shown right after logging in. (it's not very useful, but here is the [commit](https://github.com/LiteFarmOrg/LiteFarm/pull/3534/commits/e599d65c9a722bfe4ee94b7c3bbc0592a1f4e0f5) for reference) I think we need to render the purposes after the options are fetched here too!

kathyavini:
@SayakaOno thank you so much for taking on more than your share of reviews 🙇❤️ And thank you for the heads up about the purposes bug!! 🙏 

Since I set the purposes in the component, my impulse would be to just add the options to the setting dependency array (I'll push commit of which one I mean), so it will re-run once the options go from empty array to populated. Do you think there is a disadvantage to doing it that way vs delaying render?

SayakaOno:
@kathyavini Ah yes, you're calling `setValue`, so that works perfectly! I was thinking about cases where `defaultValue` is used :)

SayakaOno (approved):

antsgar (commented):
Looks great!! 🙌 

Only one minor comment that is a more generic change (not only applicable to this screen). I think we shouldn't show the Remove button on the pills and the dropdown indicator when the field is disabled (for the multiselect component on Purposes here). We could hide them easily by editing the styles in the component

```js
   multiValueRemove: (provided, state) => ({
    ...
    display: state.isDisabled ? 'none' : 'block',
  }),
  multiValue: (provided, state) => ({
    ...
    padding: state.isDisabled ? '0 18px 0 12px' : '0 4px 0 12px'
  }),
  dropdownIndicator: (provided, state) => ({
    ...
    display: state.isDisabled ? 'none' : 'block',
  }),
```

kathyavini:
@antsgar oh that's a very good point -- much better!!

Unrelated to anything, I kind of love writing CSS in TS in like this 😂

kathyavini:
Ah please hold off on re-review, I think there is a bug!! 🐞🙏

antsgar (approved):

kathyavini:
Had to resolve merge conflict 🙏

antsgar (approved):

SayakaOno (approved):

PR #3560, opened 2024-12-03
LF-4554 Adjust frontend task view to handle tasks with no locations

kathyavini:
**Description**

Minor frontend view changes needed to work with tasks with no locations:
1. Task card selector logic should specify "No location" instead of the fallback "Multiple locations"
2. Readonly view should only show the map if pins or locations are associated with the task
3. ~[TBD, code not yet pushed] Filter by location logic should exclude tasks that have no location instead of matching them to every location~ - Will create a fresh ticket to address the nice-to-have filter updates; not included here

Jira link: https://lite-farm.atlassian.net/browse/LF-4554

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

You can view tasks without locations on the branch without GET, or by adding this line to the taskEntitiesSelector to remove locations from custom tasks once they have already been fetched:

```js
if (farm_id) {
  taskEntities[task_id].locations = [];
}
```

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/Task/TaskReadOnly/index.jsx
+++ b/packages/webapp/src/components/Task/TaskReadOnly/index.jsx
@@ -233,7 +233,7 @@ export default function PureTaskReadOnly({
         </div>
       )}
 
-      {task?.locations?.length || task?.pinCoordinates?.length ? (
+      {task.locations?.length || task.pinCoordinates?.length ? (
         <>
           <Semibold style={{ marginBottom: '12px' }}>{t('TASK.LOCATIONS')}</Semibold>
           {isTaskType(taskType, 'TRANSPLANT_TASK') && (
--- a/packages/webapp/src/components/Task/TaskReadOnly/index.jsx
+++ b/packages/webapp/src/components/Task/TaskReadOnly/index.jsx
@@ -233,26 +233,30 @@ export default function PureTaskReadOnly({
         </div>
       )}
 
-      <Semibold style={{ marginBottom: '12px' }}>{t('TASK.LOCATIONS')}</Semibold>
-      {isTaskType(taskType, 'TRANSPLANT_TASK') && (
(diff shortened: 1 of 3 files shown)


packages/webapp/src/components/Task/TaskReadOnly/index.jsx
```diff
@@ -233,26 +233,30 @@ export default function PureTaskReadOnly({
...
-        }}
-        readOnlyPinCoordinates={task.pinCoordinates}
-        style={{ minHeight: '160px', marginBottom: '40px' }}
-        locations={task.locations}
-        selectedLocationIds={task.selectedLocationIds || []}
-        farmCenterCoordinate={user.grid_points}
-        maxZoomRef={maxZoomRef}
-        getMaxZoom={getMaxZoom}
-      />
+      {task?.locations?.length || task?.pinCoordinates?.length ? (
```
SayakaOno, line 236: The null checks for task don't seem necessary, as they aren't needed elsewhere in the code, but it works great!

SayakaOno (dismissed):

packages/webapp/src/components/Task/TaskReadOnly/index.jsx
kathyavini: I was hit with this irrational fear of being the cause of a new type error during bug bash 😅 But you're absolutely right -- there are about 20 lines of code that would crash before even getting here. I will remove them, thank you! 🙏

SayakaOno (approved):

packages/webapp/src/components/Task/TaskReadOnly/index.jsx
kathyavini: I looked properly and the whole Pure component is conditionally rendered based on task existing 😜

PR #3582, opened 2024-12-06
LF-4564 unnest tasks from animals

antsgar:
**Description**

This was initially suggested here https://github.com/LiteFarmOrg/LiteFarm/pull/3554#issuecomment-2515869484
Updates logic in the UI that verifies if animals/batches have finalized tasks to check the tasks themselves instead of relying on tasks being nested within animals.
Eliminates nested tasks from animals in API.
Removes need to invalidate animal/batch tags on task creation/deletion.

Jira link:
https://lite-farm.atlassian.net/browse/LF-4564

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/containers/Animals/Inventory/useAnimalOrBatchRemoval.ts
+++ b/packages/webapp/src/containers/Animals/Inventory/useAnimalOrBatchRemoval.ts
@@ -29,7 +29,7 @@ import { AnimalOrBatchKeys } from '../types';
 import { useDispatch, useSelector } from 'react-redux';
 import { useTranslation } from 'react-i18next';
 import { completedTasksSelector, abandonedTasksSelector } from '../../taskSlice';
-import { Animal } from '../../../store/api/types';
+import { Animal, AnimalBatch } from '../../../store/api/types';
 import { getLocalDateInYYYYDDMM } from '../../../util/date';
 import { getTasks } from '../../Task/saga';
 
@@ -174,8 +174,8 @@ const useAnimalOrBatchRemoval = (
     }
   };
 
-  const getFinalizedTaskIdsSet = useCallback(() => {
-    return new Set([...completedTasks, ...abandonedTasks].map(({ task_id }) => task_id));
+  const getFinalizedTasks = useCallback(() => {
+    return Array.from(new Set([...completedTasks, ...abandonedTasks]));
   }, [completedTasks, abandonedTasks]);
 
   const hasFinalizedTasks = useMemo(() => {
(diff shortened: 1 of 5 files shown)


Duncan-Brain:
I am currently seeing that the created in error option is not shown for all animals.... looking into it.

I like this solution! I think Sayakas concern was with efficiency of querying tasks. My main concern was nesting, and I didn't originally know there was a backend check going in. So that was a bug-concern and probably influenced my thoughts on recopying the existing pattern, but that is in now with #3546! So last thing was concurrency and IMO who cares about that as long as we check on the backend before writing to db! haha We could consider refetching tasks + invalidating animal tags to cover that on deletion/removal failure for the future... but I don't need it now.

But yah looking into the bug I am seeing.

antsgar:
@Duncan-Brain thank you so much for testing thoroughly!! I didn't have much time honestly and I checked for an animal that did have finalized tasks and figured it was working correctly 🤦🏻‍♀️

packages/webapp/src/containers/Animals/Inventory/useAnimalOrBatchRemoval.ts
```diff
@@ -174,32 +174,26 @@ const useAnimalOrBatchRemoval = (
...
-        }
-      }
-      // Stop iterating once all selected inventory IDs have been processed.
-      if (inventoryFoundCount === selectedInventoryIdsSet.size) {
-        return false;
-      }
-    }
+    const finalizedTasks = getFinalizedTasks();
+
+    return selectedInventoryIds.some((id) =>
```
Duncan-Brain, line 188: @antsgar I think this would do it -- but its not pretty:

return selectedInventoryIds.some((animalOrBatchId) => {
        const {id, kind} = parseInventoryId(animalOrBatchId);
        return finalizedTasks.filter(
          ({ animals, animal_batches }: { animals: Animal[]; animal_batches: AnimalBatch[] }) => {
            const animalIds = animals.map(({ id }) => `${id}`);
            const batchIds = animal_batches.map(({ id }) => `${id}`);
            return kind === AnimalOrBatchKeys.ANIMAL ? animalIds.includes(`${id}`) : batchIds.includes(`${id}`);
          },
        ).length > 0;
      }
    );

Duncan-Brain: Made a PR for it if helpful -- .. removed redundant `> 0`

Duncan-Brain (approved):
Re-running tests but pre-approve thinking it was a flakey one

SayakaOno (approved):

kathyavini (approved):
Fixes both the type error crash and the deletion-ablity bug that would have needed tag invalidation so I approve 👍 

Just wanted to make sure @SayakaOno had a chance to take a look, too.

I do still agree in general that we want to move heavy filtering to the backend, and I would be in favour of revisiting the other solution if this looks like it's making the app slow when tasks pile up!

Duncan-Brain:
> I would be in favour of revisiting the other solution if this looks like it's making the app slow when tasks pile up!

This lines up with my understanding of optimisation 😝  : https://wiki.c2.com/?RulesOfOptimization

PR #3585, opened 2024-12-09
LF-4569 It is not possible to complete any kind of task if you choose that you have done some changes to the task

kathyavini:
**Description**

Error was occurring ~because the all task GET was returning  `{ animals: [], animal_batches: [] }` for all tasks now (they are part of task graph), but the frontend complete flow was assuming they would only be sent when relevant and doesn't purge them.~

Update: Sorry I take it back, the extra properties are innocuous until the function run in the useEffect in StepOne, which will return animalIds regardless of whether the properties are empty arrays are not.

Jira link: https://lite-farm.atlassian.net/browse/LF-4569

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/taskController.js
+++ b/packages/api/src/controllers/taskController.js
@@ -900,6 +900,12 @@ const taskController = {
         if (task.task_type_id !== soilAmendmentTypeId) {
           delete task.soil_amendment_task_products;
         }
+        if (task.animals?.length === 0) {
+          delete task.animals;
+        }
+        if (task.animal_batches?.length === 0) {
+          delete task.animal_batches;
+        }
         task.animals?.forEach(flattenInternalIdentifier);
         task.animal_batches?.forEach(flattenInternalIdentifier);
       });
--- a/packages/api/src/controllers/taskController.js
+++ b/packages/api/src/controllers/taskController.js
@@ -900,12 +900,6 @@ const taskController = {
         if (task.task_type_id !== soilAmendmentTypeId) {
           delete task.soil_amendment_task_products;
         }
-        if (task.animals?.length === 0) {
-          delete task.animals;
-        }
-        if (task.animal_batches?.length === 0) {
-          delete task.animal_batches;
-        }
         task.animals?.forEach(flattenInternalIdentifier);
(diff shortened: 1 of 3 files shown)


Duncan-Brain (changes requested):
@kathyavini  Maybe I am missing something but I am still getting an error on trying to complete cleaning task. I think it need to be purged on post or patch or FE format:
Screenshot 2024-12-10 at 11 21 05 AM

kathyavini:
@Duncan-Brain thank you for showing me the bug! I think I wasn't testing properly with the "made changes" radio which was the key part!

The bug was in the useEffect generating animalIds (as an empty array) whether or not the task had associated animals.

Given that, I'm not sure if I want to remove the empty arrays from the all farm task GET anymore. Now that animals are part of the task graph, there are other API responses (e.g. the POST response on creation) that returns them on all types of tasks anyway... for consistency I think we'll just consider them as a property of all API responses now? What do you think?

Duncan-Brain:
@kathyavini I don't know for sure the best option. It seems reasonable to both ignore on FE and make nonModifiable for !ANIMAL_TASKS. 

But I think the error would be fixed by not making running this function on ln 445 and ln 749 of task controller _only_ for animal tasks

```
if (
          [...ANIMAL_TASKS, CUSTOM_TASK].includes(typeOfTask) &&
          ('related_animal_ids' in data || 'related_batch_ids' in data)
        ) {
          data = this.formatAnimalAndBatchIds(data);
        }
```

The risk is then that technically on the API you can indeed add animals to a cleaning task..... So I think maybe it should actually be kept.. but then maybe adding an `else { delete related_animal_ids.... }`


I also like returning the normalized relationship table animalTasks as its own separate GET but still allowing related_ids for PATCH and POST. But that won't fix the problem... just an alternate to not returning conditionally on GET.

antsgar (approved):

antsgar:
I had approved but saw the ongoing discussion so will wait to see which direction we want to go for!

Duncan-Brain (approved):
It works so I approve too! I read your question before seeing the new changes haha

antsgar (approved):

PR #3168, opened 2024-03-22
LF-4174 - Add checkbox and transition to complete flow

Duncan-Brain:
**Description**

Thanks Sayaka for the speedy help! Adding the persisted paths to the step one like pages vs the page before solved the problem.
~~I might need some help figure out the persistForm flow. I did not expect it to have these problems.~~

~~Custom tasks flow should actually be working. But not distinct complete flow groups for Harvest, and All others.~~

Essentially copying the `onComplete()` function from` /containers/Task/TaskReadOnly`. 

~~Problem is once I click continue on the first complete flow page - the persist data present is purged and errors on saving/ continuing the complete flow.~~

~~The `useHookFormPersist()`  seems to run the unlisten() function on the push from the `onReqSuccessSaga` because the persisted paths weren't in the original call.~~

~~I have tried `resetAndUnlockForm() ` and adding new paths but I can't get past the problem.~~

~~Current workaround ideas are:
- redirect after going to task readonly (ugly flash of the page visited)
- make some new complete task component to sit in between~~

Jira link: [LF-4174](https://lite-farm.atlassian.net/browse/LF-4174)

**Type of change**

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

**How Has This Been Tested?**

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [ ] Passes test case
- [ ] UI components visually reviewed on desktop view
- [ ] UI components visually reviewed on mobile view
- [ ] Other (please explain)

**Checklist:**

- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] The precommit and linting ran successfully
- [ ] I have added or updated language tags for text that's part of the UI
- [ ] I have added "MISSING" for all new language tags to languages I don't speak
- [ ] I have added [the GNU General Public License](https://lite-farm.atlassian.net/l/cp/BT0Dd7WW) to all new files

[...]

PR #3587, opened 2024-12-10
LF-4551 App crashes if trying to complete a custom task via the "task has already been completed" flow

kathyavini:
**Description**

This was a `hookFormPersistReducer` bug that was specific to directing to the `/complete` part of the task complete flow (non-custom tasks direct to `/before_complete` when the "This task has already been completed checkbox" is selected).

The persistedFormData was being cleared -- probably, based merely on what fixed it -- due to a URL change off of a persisted path. There is a useEffect that updates persisted paths  in `/before_complete` that I tried replicating in `/complete` that did not fix the issue; I think to use with the setting of formData in the saga, I guess the paths should be enqueued before that?

PR additionally
- routes the "already completed" custom task to `/before_complete` if animals are present, so they can be adjusted (mirroring the normal complete flow and non-custom tasks)
- adds optional chaining to the component so it doesn't crash when the `persistedFormData` is `{}` until the saga `setFormData` runs

Jira link: https://lite-farm.atlassian.net/browse/LF-4551

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/containers/Task/saga.js
+++ b/packages/webapp/src/containers/Task/saga.js
@@ -82,7 +82,7 @@ import {
   createCompleteHarvestQuantityTaskUrl,
   createCompleteTaskUrl,
 } from '../../util/siteMapConstants';
-import { setFormData } from '../hooks/useHookFormPersist/hookFormPersistSlice';
+import { setFormData, setPersistedPaths } from '../hooks/useHookFormPersist/hookFormPersistSlice';
 import { formatSoilAmendmentProductToDBStructure, getSubtaskName } from '../../util/task';
 import {
   formatAnimalIdsForReqBody,
@@ -613,12 +613,22 @@ export function* createTaskSaga({ payload }) {
         task_translation_key === 'HARVEST_TASK' ? result.data[0] : result.data;
       yield call(getTasksSuccessSaga, { payload: isHarvest ? result.data : [result.data] });
       if (alreadyCompleted) {
-        const isCustomTaskWithAnimals =
-          isCustomTask && (result.data.animals?.length || result.data.animal_batches?.length);
-        yield call(onReqSuccessSaga, {
-          message: i18n.t('message:TASK.CREATE.SUCCESS'),
-          pathname: getTaskCompletePathname(task_id, task_translation_key, isCustomTaskWithAnimals),
-        });
+        if (isCustomTask) {
+          const isCustomTaskWithAnimals =
+            isCustomTask && (result.data.animals?.length || result.data.animal_batches?.length);
+          yield call(onReqSuccessSaga, {
+            message: i18n.t('message:TASK.CREATE.SUCCESS'),
+            pathname: getTaskCompletePathname(
+              task_id,
+              task_translation_key,
+              isCustomTaskWithAnimals,
+            ),
+          });
+          yield put(
+            setPersistedPaths([`/tasks/${task_id}/complete`, `/tasks/${task_id}/before_complete`]),
+          );
+          yield put(setFormData({ task_id, taskType }));
+        }
       } else {
         yield call(onReqSuccessSaga, {
           message: i18n.t('message:TASK.CREATE.SUCCESS'),
--- a/packages/webapp/src/containers/Task/saga.js
+++ b/packages/webapp/src/containers/Task/saga.js
@@ -613,17 +613,13 @@ export function* createTaskSaga({ payload }) {
         task_translation_key === 'HARVEST_TASK' ? result.data[0] : result.data;
       yield call(getTasksSuccessSaga, { payload: isHarvest ? result.data : [result.data] });
       if (alreadyCompleted) {
+        const isCustomTaskWithAnimals =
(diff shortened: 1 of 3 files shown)


Duncan-Brain:
@kathyavini Just taking a look this morning, @SayakaOno is definitely the persistedForms expert ahah. I think she helped me with that PR https://github.com/LiteFarmOrg/LiteFarm/pull/3168

What I am seeing is that it was working at some point as I have a commit that says "custom tasks work" haha. And even now, when I see the type error crash... a simple refresh fixes the page, so there seems to be either an extra or a missing render cycle. 

Ideas: 

a) Find the missing/extra render

b) Could be a good opportunity to use `git bisect` again? 

c) I personally like the idea of routing everything to `/before_complete` for consistency. If you wanted to add something there you could add the `notes` section. If we are completing right away I do see value in skipping that screen but.. we could decide that for all tasks in the future.

kathyavini:
> I personally like the idea of routing everything to /before_complete for consistency

@Duncan-Brain this was my favourite idea too but in the end it felt wrong to adjust unrelated UX to fix a saga/state bug 🤔

I could never re-create the "fixed on reload" you showed me, so I guess it has something to do with the order in which async code was being called. I threw in a persisted paths into the saga, which is different than how `/before_complete` handles it (with the useEffect) but I also don't think `/before_complete` needs setFormData at all, so... maybe that's why.

Anyway, I guess it's working?

Duncan-Brain:
@kathyavini 

> Anyway, I guess it's working
 
 😂  such confidence! lol It is working! 
 
 ... for custom tasks. The reqSuccessSaga also needs to be put in the else condition for the other tasks! 
 
 I guess the mainly surprising thing is that you couldn't reproduce my refresh success .. so it seems like we found an ordering problem or race condition somewhere. Looking at it, I wonder if we should be changing path after setting persisted data (onReqSuccess and its route changing after the other data setting) but I am not sure if that works... just that it makes sense in my head.

kathyavini:
> The reqSuccessSaga also needs to be put in the else condition for the other tasks

@Duncan-Brain wait, what do you mean? It was never broken for the other tasks. They all route to `before_complete` which doesn't have this issue.

packages/webapp/src/containers/Task/saga.js
```diff
@@ -609,11 +613,16 @@ export function* createTaskSaga({ payload }) {
...
         yield call(onReqSuccessSaga, {
           message: i18n.t('message:TASK.CREATE.SUCCESS'),
-          pathname: getTaskCompletePathname(task_id, task_translation_key),
+          pathname: getTaskCompletePathname(task_id, task_translation_key, isCustomTaskWithAnimals),
         });
         if (isCustomTask) {
+          yield put(
+            setPersistedPaths([`/tasks/${task_id}/complete`, `/tasks/${task_id}/before_complete`]),
+          );
           yield put(setFormData({ task_id, taskType }));
```
SayakaOno, line 626: Are you able to create and complete a custom task without animals? The solution looks solid, but `persistedFormData` is still not updating for me, and I can't complete the task.
I tried adding `yield delay(500);` before `setFormData`, and that seemed to work.
My laptop feels very slow today, so the issue might not be reproducible on your end.

kathyavini: @SayakaOno yeah, I'm able to complete it without a delay (the form data does populate). I hate that it's system dependent, that sucks 😦

> My laptop feels very slow today

Unrelated but did you update to Sequoia recently? 😅

SayakaOno: `500` was just a random number, but I guess it's reasonable?

I think it's been a few weeks since I updated to Sequoia. I didn't do anything, but my laptop is running much faster now, and it works fine without a delay 😅

SayakaOno (approved):

antsgar:
Hook form persist causing bugs yet again 😢 I want to refactor the whole tasks flow so badly

So I digged into this a bit, and it seems the error started happening since the introduction of this line https://github.com/LiteFarmOrg/LiteFarm/commit/7bc8b0771229f1103e5efa2ee82ff5cdbb7eb156#diff-c22350b9014536a1f8e281f83643f0fa173de34b5197b63499e686f61bb6b28dR76 (which is the one causing the error showed in the console). Before, the `task_type_name` variable was being calculated inside onSubmit and now it's outside of it in the render function on the component, so it seems this is indeed caused by the persisted form data not being defined on the first render for some reason like Duncan pointed out. Adding optional chaining seems to fix the problem

`let task_type_name = persistedFormData?.taskType?.task_translation_key?.toLowerCase();`

and I think that'd be a cleaner fix, thoughts? Would be great for persisted form data to not be a headache in the future though but that will require a ton more work 😞

antsgar:
Oh never mind, I forgot the optional chaining was part of the PR as well 😂 and no, it looks like I can't submit the form. I do hate the `yield delay` bit but I'll live with it because we honestly have to get rid of this whole saga anyway, lol

antsgar (approved):

PR #3617, opened 2024-12-12
LF-4612c (2): Adjust margin with cell className

SayakaOno:
**Description**

Adjust margin using cell className

### DIFF
--- a/packages/webapp/src/containers/Animals/Inventory/index.tsx
+++ b/packages/webapp/src/containers/Animals/Inventory/index.tsx
@@ -308,6 +308,7 @@ export default function AnimalInventory({
             photoUrl={d.photo_url}
             subtext={isDesktop ? null : `${d.type} / ${d.breed}`}
             highlightedText={d.batch ? d.count : null}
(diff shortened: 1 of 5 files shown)


packages/webapp/src/containers/Animals/Inventory/index.tsx
```diff
@@ -308,6 +308,7 @@ export default function AnimalInventory({
             photoUrl={d.photo_url}
             subtext={isDesktop ? null : `${d.type} / ${d.breed}`}
             highlightedText={d.batch ? d.count : null}
+            className={view === View.TASK_SUMMARY ? '' : styles.cellAfterCheckbox}
```
SayakaOno, line 311: Could you double check this condition? I'm not so confident!

kathyavini (approved):
Thank you! 🙏 It's muuuuch better than checking for the animal id in `<Table />`!!!

packages/webapp/src/containers/Animals/Inventory/index.tsx
kathyavini: That should be right, now that the worker view was adjusted!

PR #3619, opened 2024-12-13
Update Worker.Consent.md

DChayanam:
**Description**
Worker consent form updated with information on data collected, contact information

Jira link:

**Type of change**

content change only

**How Has This Been Tested?**

content change only

**Checklist:**


kathyavini (approved):
For the early risers, the consent needs to updated for the sake of [these links](https://www.litefarm.org/data-policy), and not necessarily on App.

If that ends up still being the case tomorrow, perhaps we could discuss pointing the Wix links to integration? That would make future updates more painless.

If we do have to update (patch) main, the `CONSENT_VERSION` should also be bumped at that time, and perhaps a patch branch could be made tomorrow 🙏

kathyavini (approved):

SayakaOno (approved):

PR #3643, opened 2025-01-13
LF-4618 Fix remove error entry when click X on number input

Tbrid:
**Description**

Bug: error value in animals count input not restored to default value when clicking "X". 
Cause: default value not being used and logic of restoring default is wrong.
Fix: Change default value logic to be using prop default value instead of current value in the field and add default value prop to field input

Jira link: https://lite-farm.atlassian.net/browse/LF-4618

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/Animals/AddAnimalsFormCard/AddAnimalsFormCard.tsx
+++ b/packages/webapp/src/components/Animals/AddAnimalsFormCard/AddAnimalsFormCard.tsx
@@ -128,6 +128,7 @@ export default function AddAnimalsFormCard({
           label={t('common:COUNT')}
           className={styles.countInput}
           allowDecimal={false}
+          defaultValue={1}
           showStepper
           rules={{
             required: {
--- a/packages/webapp/src/components/Form/NumberInput/index.tsx
+++ b/packages/webapp/src/components/Form/NumberInput/index.tsx
@@ -60,8 +60,10 @@ export default function NumberInput<T extends FieldValues>({
   ...props
 }: NumberInputProps<T>) {
   const { field, fieldState, formState } = useController({ name, control, rules, defaultValue });
+  console.log(field.value);
+  console.log(defaultValue);
   const { inputProps, reset, numericValue, increment, decrement } = useNumberInput({
-    initialValue: field.value || get(formState.defaultValues, name) || defaultValue,
+    initialValue: defaultValue || field.value || get(formState.defaultValues, name),
     allowDecimal,
     decimalDigits,
     locale,


Duncan-Brain (changes requested):
Hi @Tbrid I think the defaultValue was placed as the last option was for the case where one arrives at the form with pre-filled values. 

So one example is when you navigate to the next page of the form and then you go back in the form flow:

https://github.com/user-attachments/assets/30df010a-7e1e-4d25-8748-12a577499d81

One other thing about this field surprises me: it does not validate when blurred. It validates when another field is validated which makes sense. But it seems like it should also be validating on blur.

Duncan-Brain (dismissed):
Thanks looking good to me !

packages/webapp/src/components/Form/NumberInput/index.tsx
```diff
@@ -86,7 +86,7 @@ export default function NumberInput<T extends FieldValues>({
       {...inputProps}
       className={className}
       error={fieldState.error?.message}
-      onResetIconClick={reset}
+      onResetIconClick={() => reset(defaultValue)}
```
SayakaOno, line 89: `initialValue` is `field.value || get(formState.defaultValues, name) || defaultValue`, so I don't think this would work unless `defaultValue` is passed to this component.
Can we think of a way to use the correct `initialValue` instead?

Tbrid: I am not sure what you mean? The ```defaultValue``` is being passed as a prop.

SayakaOno: Sorry, the issue is that `initialValue` is set as `field.value || get(formState.defaultValues, name) || defaultValue` (line 64), but `reset()` sets the value to `defaultValue`.

https://github.com/LiteFarmOrg/LiteFarm/blob/0f8a2784fe9b8685384ecf44f62b02be9285bc0b/packages/webapp/src/components/Form/NumberInput/index.tsx#L63-L65


When resetting, we expect the input to display `initialValue`, but if `field.value` or `get(formState.defaultValues, name)` is used, `reset()` won’t work as expected.

`reset: () => update(initialValue ?? NaN)` in `useNumberInput` seems correct, but the problem is that `initialValue` changes. Can you think of a way to ensure `reset()` always works when `field.value` or `get(formState.defaultValues, name)` is used as `initialValue`?

Tbrid: Sorry I was caught up with my own stuff lately, I just did a test with commenting out defaultValue and so initialValue will be using the other params, and reset still works, so I think there isn't an issue here.

SayakaOno: Thank you for jumping back in — really appreciate it!

Hope this example helps clarify the issue a bit more.
Here’s a screen recording of the behaviour:

https://github.com/user-attachments/assets/90ccb6ef-0691-4cd0-ae3a-8dc172db6b92

In the `QuantityApplicationRate` component, we're passing `100` as a `defaultValue`, but the actual value that's used comes from `react-hook-form`'s value (set in the grandparent component) — in this case, it's `80` (`field.value` in line 64 in this snippet above).

So when resetting, we should restore `field.value` (`80`), not `100`.

That said, I’m not entirely sure whether it should reset or just clear — the latter makes more sense to me. I’ll check with the team to confirm what’s intended.

For reference, here's the temporary change I made for the demo:
Screenshot 2025-04-16 at 10 04 06 AM

SayakaOno: @Tbrid 
We discussed the behaviour and confirmed that clicking the clear button should clear the input, not restore the initial value.
Could you please revert the changes to `reset` for now and use `clear` instead? 🙏 

The `reset` function still needs to be fixed, so I’ve created a separate bug [ticket](https://lite-farm.atlassian.net/browse/LF-4798) for it.

packages/webapp/src/components/Form/NumberInput/useNumberInput.ts
```diff
@@ -199,7 +199,7 @@ export default function useNumberInput({
   return {
     numericValue,
     inputProps,
-    reset: () => update(initialValue ?? NaN),
```
SayakaOno, line 202: It would be appreciated if you could add a comment to this line, like: `// TODO: Fix before use (LF-4798)`

packages/webapp/src/components/Animals/AddAnimalsFormCard/AddAnimalsFormCard.tsx
```diff
@@ -128,6 +128,7 @@ export default function AddAnimalsFormCard({
           label={t('common:COUNT')}
           className={styles.countInput}
           allowDecimal={false}
+          defaultValue={1}
```
SayakaOno, line 131: Thank you for the changes! 🙏 
It looks like the default values are already set in `animalBasicsDefaultValues` in `AddAnimalBasics`. Is there any other reason we’d need this?

Tbrid: Removed.

SayakaOno: Thank you so much!

SayakaOno:
@Tbrid I'm sorry for the delay in merging this PR. It will be included in the upcoming release, which will be out soon. Thanks again for your help! 🙏

PR #3653, opened 2025-01-15
LF-4626 Automatically detect field for sensor array

kathyavini:
**Description**

The main content of this PR is `geoUtils.ts`. It contains two functions:

1. `getAreaLocationsContainingPoint` -- Given a point and an array of polygon locations (e.g. the output of a locations selector), return those whose polygons (`grid_points`) contain the point being queried. On the current screens this is the one that would be used from the list of sensors page or an individual sensor's page
2. `getPointLocationsWithinPolygon` -- Given a polygon (location object with `grid_points`) and a an array of point locations (e.g. the return from sensorSelector), return those whose points are contained within the queried polygon. On the current screens this is one to call from the Field Technology tab of a given location.

The second file (the little hook) is just some example code that can be discarded entirely OR adapted to our needs based on the need of the calling component -- I pictured this as the easiest place to configure the desired selectors for the locations we'll query (particularly if this will be called against the same locations multiple times / from different components) but it's also fine for the calling component to call one of the util functions directly.

To use the example hooks, you can call like this from within `<SensorDetail>`:

```js
const areasContainingSensor = useFarmAreasContainingPoint(sensorInfo?.point);
```

And like this from within, e.g. `<PureLocationDetailLayout />` -- there isn't a great component for this, but this is a stand in for the field technology tab until that is built

```js
const containedSensors = useFarmSensorsContainedWithinArea(persistedFormData?.grid_points);
```

The returned value is the same format as using a location selector like e.g. `animalLocationSelector`

### Why turf.js over maps.geometry.poly.containsLocation()
We only use the Google Maps functions (including the other use in our codebase of `.containsLocation()` within the context of a rendered map component, handled by google-map-react. As far as I understand it (although I'm realizing now we don't have screens yet for the new "confirm location" functionality so I'm not even sure this will remain true 😬 !!) we want to be able to detect a location to populate a multi-select without having to render a map.

We could create an independent API call to Google Maps outside of google-map-react to use `.containsLocation()` BUT there is really no reason to do that when we can do this independently. turf is a [tiny package](https://bundlephobia.com/package/turf-size@1.1.1) to begin with, that also lets us selectively install the functionality we need (here only [@turf/boolean-point-in-polygon](https://turfjs.org/docs/api/booleanPointInPolygon) and the helpers to create the point and polygon features)

Jira link: https://lite-farm.atlassian.net/browse/LF-4626

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

I logged the values in the example code above for some different farm map setups; nothing exhaustive though.

**Checklist:**


Duncan-Brain (approved):
Nice ! love the use of non-google utils.

antsgar (approved):
Love this!! Good call using Turf instead of Google Maps 👏 

Not because I want to include them now, but do you think there'd be a way to add unit tests for these utilities?

kathyavini:
> Not because I want to include them now, but do you think there'd be a way to add unit tests for these utilities?

@antsgar Hmm, I think they should be fairly trivial to unit test (would just require some fake polygons and points). There isn't isn't too much logic in the files beyond the package call, though, do you think it's still worth testing (as we don't really want to test the package)?

antsgar:
@kathyavini they're pretty lean but, you never know, sometimes the simplest chunks of code lead to someone making a mistake just because they think the code is too simple for that to happen 😂 I don't think we need to add tests right now though or that it's a blocker to have them

kathyavini:
> code lead to someone making a mistake just because they think the code is too simple for that to happen

I believe it; I could definitely be that someone 😂

PR #3660, opened 2025-01-22
Lf 4629 [nice to have] implement map footer redesign

Duncan-Brain:
**Description**

Updates the footer styles + icons for map.

Notes:
- Added a new button theme styles type but it feels like Button belongs in core components not Form. And maybe there is a better name than 'location-menu' taken from figma.
- Considered adding "selected" option to match "active" state on button instead of custom styles.

Jira link: [LF-4629](https://lite-farm.atlassian.net/browse/LF-4629)

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

[LF-4629]: https://lite-farm.atlassian.net/browse/LF-4629?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


SayakaOno (commented):
I haven't reviewed all files yet, but we need to display the filter and download buttons for non-admins 😄 
 
Screenshot 2025-01-24 at 10 54 18 AM

Duncan-Brain:
Thanks @SayakaOno I misread that conditional render! That makes a lot of sense 😂

packages/webapp/src/components/Map/Footer/MapNavigationButtons.tsx
```diff
@@ -0,0 +1,91 @@
...
+    <div className={styles.container}>
+      <Button
+        color="location-menu"
+        data-cy="map-addFeature"
+        className={clsx(styles.button, showAddDrawer && styles.selected)}
+        id="mapFirstStep"
+        onClick={onClickAdd}
+        sm
+      >
+        <div className={styles.buttonText}>
```
SayakaOno: I feel like I've being doing this too 😅, but this has to be `span` since `button` is inline element.

packages/webapp/src/components/Map/Footer/styles.module.scss
```diff
@@ -2,69 +2,97 @@
...
-  background-color: white;
-  cursor: pointer;
+  border-radius: 8px;
+  box-shadow: 0px 2px 4px 0px rgba(43, 48, 58, 0.1);
 
-  .svg {
-    fill: var(--grey600);
+  .button {
+    // Uneven button sizes if one string much longer than others
+    width: 34%;
```
SayakaOno, line 12: I used to do `33.33%`, but `calc(100% / 3)` seems to be the best solution nowadays...

packages/webapp/src/components/Map/Footer/styles.module.scss
```diff
@@ -2,69 +2,97 @@
...
+    .circle {
+      background-color: var(--White);
+    }
+  }
+  
+  button:first-of-type {
+    border-radius: 8px 0 0 8px;
+
+    @include xs-breakpoint {
+      border-radius: 0 0 0 8px;
```
SayakaOno, line 43: Could we set `0` and let devices round the corners naturally?

Duncan-Brain:
Duncans attention to detail strikes again -- should be fixed!

packages/webapp/src/components/Map/Footer/styles.module.scss
```diff
@@ -2,69 +2,97 @@
...
-  font-weight: 600;
-  color: var(--grey900);
+.label {
+  margin-top: -4px;
+  font-weight: 400;
+  // Menu item text too long
+  overflow: hidden;
+  text-overflow: ellipsis;
+  // Awkward multi-line text on small screens
+  line-height: normal;
```
SayakaOno, line 86: I'm not sure if these are working as expected, is the label supposed to be truncated?

SayakaOno (commented):
I left comments, but it looks great!
I think the button height should be 64px, and we may need custom styling for the mobile view...?
Screenshot 2025-01-24 at 12 41 16 PM

packages/webapp/src/components/Map/Footer/MapNavigationButtons.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+import { useTranslation } from 'react-i18next';
+import clsx from 'clsx';
+import Button from '../../Form/Button';
+import styles from './styles.module.scss';
+import { ReactComponent as AddLogo } from '../../../assets/images/map/addSkew.svg';
+import { ReactComponent as FilterLogo } from '../../../assets/images/map/filter-funnel.svg';
+import { ReactComponent as ExportLogo } from '../../../assets/images/map/download.svg';
+
+export interface FormNavigationButtonsProps {
```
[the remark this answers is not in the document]
Duncan-Brain: Thanks about name. And thanks about boolean I was seeing the following and thought autofill from typescript package was correct for some reason.

Screenshot 2025-01-24 at 8 40 32 PM

Screenshot 2025-01-24 at 8 40 58 PM

packages/webapp/src/components/Map/Footer/styles.module.scss
Duncan-Brain: Makes sense!

packages/webapp/src/components/Map/Footer/styles.module.scss
Duncan-Brain: Yeah I think that makes sense - I copied from figma.

packages/webapp/src/components/Map/Footer/MapNavigationButtons.tsx
Duncan-Brain: button display value can be overridden right?  https://www.shecodes.io/athena/7267-is-button-an-in-line-element-in-html#:~:text=In%20HTML%20the%20element,block%2Dlevel%20element%20using%20CSS.

I think this Button component uses `.btn{ display:flex }`. 

Maybe I don't understand it.

packages/webapp/src/components/Map/Footer/styles.module.scss
Duncan-Brain: It was working I think I took out a `max-width` by accident. Thanks for spotting!

Duncan-Brain: Should be fixed

packages/webapp/src/components/Map/Footer/styles.module.scss
Duncan-Brain: Changed it to 0!

packages/webapp/src/components/Map/Footer/styles.module.scss
Duncan-Brain: Done

packages/webapp/src/components/Map/Footer/MapNavigationButtons.tsx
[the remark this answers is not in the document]
Duncan-Brain: updated 🙏

Duncan-Brain:
@SayakaOno thanks for review! 

I am hesitant to set the height to 64px and since the width is normally fixed and would want to show all words if possible. 

I chose to flex on height for label text because other languages might be much longer and I wanted to avoid truncating unless absolutely necessary. I added those lines for the long single word case.
Screenshot 2025-01-24 at 8 56 31 PM

Flex button item width with fixed menu width was also considered but if one of the text is much longer than the others it hogs the menu and squishes the others.

If you have any advice for me about this case, or how I would use the 64px height I would definitely appreciate it!

packages/webapp/src/components/Map/Footer/MapNavigationButtons.tsx
antsgar: Regardless of the display value, per the HTML spec buttons should only contain phrasing content elements https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button -- think spans and imgs, there's a full list here https://developer.mozilla.org/en-US/docs/Web/HTML/Content_categories#phrasing_content

antsgar:
I don't think the button needs to have a fixed 64px height per se but it's the vertical margins that are off -- they are 4px currently and should be 10px per the Figma designs. When those margins are updated, the button height ends up being pretty close to 64px in English where there's no text wrapping.

antsgar (commented):
Added a couple of comments but it looks great ❤️

packages/webapp/src/components/Form/Button/index.tsx
```diff
@@ -18,7 +18,15 @@ import styles from './button.module.scss';
 import clsx from 'clsx';
 
 export type ButtonProps = {
-  color?: 'primary' | 'secondary' | 'secondary-2' | 'secondary-cta' | 'warning' | 'error' | 'none';
+  color?:
+    | 'primary'
+    | 'secondary'
+    | 'secondary-2'
+    | 'secondary-cta'
+    | 'location-menu'
```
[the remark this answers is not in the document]
Duncan-Brain: The way we handle button styles -- It feels like it is both easier and less code to create a under-used color theme than a one-off new component that copies the styles of Button with just one style. I wanted to use the standardized button theming.

Duncan-Brain:
So we discussed with Loic today to remove the label from the mobile view and to truncate with an ellipsis if the text is too long -- no wrapping.

So that is updated. I bumped up the vertical padding to 8px (Loics magic padding margin number) curious though how you got 10px, clicking on the icon and text showed no margin/padding and this is what I was seeing for both:
Screenshot 2025-01-27 at 7 49 29 PM

packages/webapp/src/components/Map/Footer/MapNavigationButtons.tsx
Duncan-Brain: should be fixed

packages/webapp/src/components/Form/Button/index.tsx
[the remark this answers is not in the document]
antsgar: Could we name it something more generic in that case? I see the color scheme used is mostly the accent color, so maybe `accent` would be proper?

antsgar:
@Duncan-Brain re how I got the 10 px, I don't know if I ever shared this, but on Figma if you click on an element and press the Option key and then hover over other elements, you get the distance from that element to anything else. I can't take a screenshot because I don't know how to take on while keeping the effect 😂 but here's the tutorial from FIgma https://help.figma.com/hc/en-us/articles/360039956974-Measure-distances-between-layers

Duncan-Brain:
@antsgar thanks for the link will remember to look for missing spacing next time.

@SayakaOno I see the 64px on the individual button you were talking about too now! I deleted it originally... and it would have maybe avoided the spacing issue Anto also noticed. Wondering if you were saying there are missing css properties that are on figma?

Screenshot 2025-01-28 at 9 36 01 AM

I typically try to remove as much Figma css as possible because it can be from parent or globals or it is not following html spec (notice the `gap: -4px` which is not possible) . So I try to just use it as a guideline.

packages/webapp/src/components/Map/Footer/styles.module.scss
```diff
@@ -2,69 +2,106 @@
...
-  outline: none;
+
+.label {
+  margin-top: -4px;
+  font-weight: 400;
+  // Menu item text too long
+  max-width: 100%;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
```
SayakaOno, line 91: You can use @mixin truncateText() next time!

SayakaOno (approved):
Looks great!

> Wondering if you were saying there are missing css properties that are on figma?

No, I wasn't. I agree with your approach, we should have as little CSS as possible!

antsgar (approved):

packages/webapp/src/components/Map/Footer/styles.module.scss
Duncan-Brain: Oh thanks!  I didn't know about that one! thanks 🙏

PR #3663, opened 2025-01-23
LF-4625 Readability Refactors on Ensemble controller code

kathyavini:
**Description**

(As mentioned in tech daily Jan 23rd) These readability refactors on the Ensemble portion of the controller code were done as part of the ticket scope of the original controller ticket. The changes are limited to using a constant where there was a hardcoded string, refactoring out a few `.reduce()` functions for readability, and clarifying a few return statements that were set up misleadingly. A lot of the Ensemble code (probably whole add flow, maybe webhook function as well) is going to be removed in our "remove old sensors code" task, but I wanted to merge the refactors because:
- if we are ever going to refer to that old code when expanding sensors, I think it should help us understand what was done in the original flow
- I have been using the changes (actually it might be just the constant 😂) in my subsequent work on Ensemble controllers so it would be useful to have it in

Removing CSV is not going to be part of this ticket; that should happen in the cleanup/removal task.

Jira link: https://lite-farm.atlassian.net/browse/LF-4625

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/api/src/util/ensemble.js
+++ b/packages/api/src/util/ensemble.js
@@ -96,6 +96,7 @@ async function bulkSensorClaim(accessToken, organizationId, esids) {
     data: { esids },
   };
 
+  // partial or complete failures (at least some esids failed to claim)
   const onError = (error) => {
     if (error.response?.data && error.response?.status) {
       return { ...error.response.data, status: error.response.status };
@@ -104,6 +105,7 @@ async function bulkSensorClaim(accessToken, organizationId, esids) {
     }
   };
 
+  // full success (all esids successfully claimed)
   const onResponse = (response) => {
     return {
       success: esids,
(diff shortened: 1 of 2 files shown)


packages/api/src/util/ensemble.js
```diff
@@ -111,27 +133,27 @@ async function registerOrganizationWebhook(farmId, organizationId, accessToken)
...
+  };
+  const onError = (error) => {
+    console.log(error);
+    throw new Error('Failed to register webhook with ESCI');
+  };
+  const onResponse = async (response) => {
+    await FarmExternalIntegrationsModel.updateWebhookId(farmId, response.data.id);
+    return { ...response.data, status: response.status };
+  };
+  await ensembleAPICall(accessToken, axiosObject, onError, onResponse);
```
Duncan-Brain, line 156: While unused the return statement gave the option to see result of webhook call.

Duncan-Brain (approved):
I am good with this since if you think it will help, we are deleting it all mostly is my understanding though.

reduce() I believe is much faster for large volumes, and small reduce functions are fairly readable (I like (acc, cv) over (previous, current) from other author though ;) But filter + map is also readable if this is for looking back.

SayakaOno (approved):
Looks good!

I don’t think reduce() is faster for large datasets... I feel like discussions around reduce VS map/forEach often come up, let’s confirm this sometime 😂

Duncan-Brain:
@SayakaOno I am more interested in readability before premature optimization more just a personal note on small reduce functions and finding them fairly readable. I also like filter + map.

Not authoritative reading I did lately for 
-  [reduce vs filter+map](https://www.davideasaf.com/mapfilter-vs-reduce/)
- [Ops/sec comparison](https://leanylabs.com/blog/js-forEach-map-reduce-vs-for-for_of/)

PR #3667, opened 2025-01-28
LF-3787 The common name for a crop should be displayed on the Crop Detailed page once changes are made

den4ik1203:
…are made

**Description**

The common name for a crop should be displayed on the Crop Detailed page once changes are made

Jira link: https://lite-farm.atlassian.net/browse/LF-3787

**Type of change**

**How Has This Been Tested?**

- [X] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/Crop/CropHeader.jsx
+++ b/packages/webapp/src/components/Crop/CropHeader.jsx
@@ -38,7 +38,7 @@ function CropHeader({ variety, onBackClick }) {
         <div className={styles.headerTitleContainer} onClick={onBackClick}>
           <Back style={{ verticalAlign: 'text-bottom' }} />
           <Title className={clsx(styles.headerTitle, styles.textOverFlowBehaviour)}>
-            {t(`crop:${crop_translation_key}`)}
+            {t(`crop:${crop_variety_name}`)}
           </Title>
         </div>
         <div className={styles.headerAttributesContainer}>


packages/webapp/src/components/Crop/CropHeader.jsx
```diff
@@ -38,7 +38,7 @@ function CropHeader({ variety, onBackClick }) {
         <div className={styles.headerTitleContainer} onClick={onBackClick}>
           <Back style={{ verticalAlign: 'text-bottom' }} />
           <Title className={clsx(styles.headerTitle, styles.textOverFlowBehaviour)}>
-            {t(`crop:${crop_translation_key}`)}
+            {t(`crop:${crop_variety_name}`)}
```
kathyavini, line 41: This definitely works to implement the Jira-described solution! 👍 

However, it's just slightly misleading. What it is doing, is calling the translation function `t()` and saying "look for the translation in the `crop` file that corresponds to the key `{{crop_variety_name}}`". Crop variety name is going to be a user-input string (or by default, a translated string corresponding to the crop key) so it definitely doesn't exist as a key in that file.

Instead, I recommend using here `{crop_variety_name}` directly, without the unnecessary call to i18n.

Since i18n will anyway default to displaying the key if it doesn't exist, the end result in the UI will be the same, but it's clearer because it's not suggesting that we have a translation for `crop:${crop_variety_name}` when we don't.

kathyavini (commented):
This is a very interesting little UX issue which I didn't know about until taking a look at this ticket, thanks @den4ik1203!

It looks like there is some fundamental UX issue in these forms, because the Jira ticket description and the linked bug both describe "editing the crop common name" when the form item in question is actually the crop _variety_ common name.

I'm therefore not sure the right answer is showing the crop variety in the header (the use of the scientific name -- and also the name "Crop Header" for this component makes me think it was meant to stay fixed at crop) -- but I think we can take this to dev-design tomorrow; it will be a interesting discussion point there.

Screenshot 2025-01-28 at 1 44 07 PM

PR #3671, opened 2025-01-29
Lf 4694 create get endpoint to check if an ensemble addon exists b

Duncan-Brain:
**Description**

Adds a get endpoint for addons. Returns 200 if an addon exists with addon_partner_id or if not specified. Return 404 if there is no addon present on farm.

Jira link: [LF-4694](https://lite-farm.atlassian.net/browse/LF-4694)

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain): Postman check

**Checklist:**

[LF-4694]: https://lite-farm.atlassian.net/browse/LF-4694?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/api/src/controllers/farmAddonController.js
+++ b/packages/api/src/controllers/farmAddonController.js
@@ -1,0 +1,24 @@
+import FarmAddonModel from '../models/farmAddonModel.js';
+
+const farmAddonController = {
+  getFarmAddon() {
+    return async (req, res) => {
+      try {
+        const { farm_id } = req.headers;
+        const { addon_id } = req.query;
+        const rows = await FarmAddonModel.query().where({ farm_id, addon_id });
+        if (!rows.length) {
+          return res.sendStatus(404);
+        }
+        return res.sendStatus(200);
+      } catch (error) {
+        console.error(error);
+        return res.status(500).json({
+          error,
+        });
+      }
+    };
+  },
+};
+
+export default farmAddonController;
--- a/packages/api/src/controllers/farmAddonController.js
+++ b/packages/api/src/controllers/farmAddonController.js
@@ -56,7 +56,10 @@ const farmAddonController = {
         if (!rows.length) {
           return res.sendStatus(404);
         }
-        return res.sendStatus(200);
+        const result = rows.map(({ addon_partner_id, org_uuid }) => {
+          return { addon_partner_id, org_uuid };
+        });
+        return res.status(200).send(result);
       } catch (error) {
         console.error(error);
         return res.status(500).json({
--- a/packages/api/src/controllers/farmAddonController.js
+++ b/packages/api/src/controllers/farmAddonController.js
@@ -17,31 +17,33 @@ import FarmAddonModel from '../models/farmAddonModel.js';
 import { getValidEnsembleOrg } from '../util/ensemble.js';
 
 const farmAddonController = {
-  async addFarmAddon(req, res) {
-    const { farm_id } = req.headers;
-    const { addon_partner_id, org_uuid } = req.body;
+  addFarmAddon() {
+    return async (req, res) => {
+      const { farm_id } = req.headers;
+      const { addon_partner_id, org_uuid } = req.body;
 
-    try {
-      const organisation = await getValidEnsembleOrg(org_uuid);
+      try {
+        const organisation = await getValidEnsembleOrg(org_uuid);
 
-      if (!organisation) {
-        return res.status(404).send('Organisation not found');
+        if (!organisation) {
+          return res.status(404).send('Organisation not found');
+        }
+
+        await FarmAddonModel.upsertFarmAddon({
+          farm_id,
+          addon_partner_id,
+          org_uuid,
+          org_pk: organisation.pk,
+        });
+
+        return res.status(200).send();
+      } catch (error) {
+        console.log(error);
+        return res.status(400).json({
+          error,
+        });
       }
-
-      await FarmAddonModel.upsertFarmAddon({
-        farm_id,
-        addon_partner_id,
-        org_uuid,
-        org_pk: organisation.pk,
-      });
-
-      return res.status(200).send();
-    } catch (error) {
-      console.log(error);
-      return res.status(400).json({
-        error,
-      });
-    }
+    };
   },
   getFarmAddon() {
     return async (req, res) => {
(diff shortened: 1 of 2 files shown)


kathyavini:
@Duncan-Brain yup I just made those same two files on my POST branch 😂 But it might still work out.

Duncan-Brain:
Updated to allow `addon_partner_id` not be specified - same response style.

@kathyavini I have never took notice of it before but we used two different function styles and I wanted to align them here.

I am thinking about updating to the style I copied -- I think I got lucky on which template file I used but I am solely basing my intent to align them on GPT response:

Screenshot 2025-01-31 at 9 24 56 AM

Any other thoughts before I go with the higher order function way?

kathyavini:
@Duncan-Brain I have absolutely no preference when it comes to function styles 😂 `addFarmAddon()` used to live in the sensor controller before the new route and model is the only reason it's the way it is. Please feel free to refactor 👍 🙏

packages/webapp/src/routes/index.jsx
```diff
@@ -54,85 +54,87 @@ const Prices = React.lazy(() => import('../containers/Insights/Prices'));
 const ExpiredTokenScreen = React.lazy(() => import('../containers/ExpiredTokenScreen'));
```
kathyavini, line 54: Oh no the pre-commit linting diff returns 🤣 We really have to work that out.

But also I can't figure out why is there a diff on webapp routes in this PR? What's the actual diff?

Duncan-Brain: Its on the merge commit... so maybe merge commits count as touching files? There is no diff from me... Linting or merge process must be undoing or redoing same thing as Sayakas ... I think I did just redo a `pnpm i` also so maybe there was an update to the linter rule? I have zero personal vscode plugins like prettier installed so it wouldn't be that.

kathyavini: Okay, I guess it's getting committed, then! 😂 

I'll just ping @antsgar because it looks like it was [her merge commit](https://github.com/LiteFarmOrg/LiteFarm/commits/eb191cdce93ecd294362c4b4f05742bb931becf5/packages/webapp/src/routes/index.jsx) that commited the current state. Anto you weren't at our tech daily where we looked at this based on Sayaka's PR with the same diff, but the ESLint rule running on commit is making the diff on the right, while prettier linting (format on save) according to our `.prettierrc` does the diff on the left. So if you work on this file you get the left diff, and when you run the pre-commit hook it will transform to the right right 😝 Is it possible your pre-commit lint was off at one point doing the unused animals bit?

Anyway I think it would better if prettier + eslint weren't at odds in this way -- I think there should be some way to make prettier eslint aware (or vice versa) that we aren't implementing properly but in any case let's commit and sorry @SayakaOno I made you revert this earlier 😅

SayakaOno: Oh, thank you for finding that commit! I'm making some changes to the file again, and I'll leave it this time! 😂

packages/api/src/controllers/farmAddonController.js
```diff
@@ -17,31 +17,53 @@ import FarmAddonModel from '../models/farmAddonModel.js';
...
+    };
+  },
+  getFarmAddon() {
+    return async (req, res) => {
+      try {
+        const { farm_id } = req.headers;
+        const { addon_partner_id } = req.query;
+        const rows = await FarmAddonModel.query()
+          .where({ farm_id, addon_partner_id })
+          .skipUndefined();
```
kathyavini, line 55: This is a nice inclusion and should probably be our default!

kathyavini (commented):
It looks good!

I think we also need to get the `org_uuid` back though so the [component](https://www.figma.com/design/eMffC0h3JPlWtzNttxJpil/Litefarm-Updated?node-id=27096-63008&m=dev) can display it.

packages/webapp/src/routes/index.jsx
antsgar: Oh no, how annoying 😓 I always get the diff on the right no matter what, even if I change something on this file and format on save

Duncan-Brain:
Okay, I wasn't aware Loic decided on a design to definitely include org uuid yet. And thought we were doing the status thing.

How would you like the response @kathyavini and @SayakaOno 

id agnostic (like default animal type and breed)  - if we go this route we can consider sending `name` of `addon_partner` instead of id: 

```
// req includes addon_partner_id
["dsfdsf-adfsadf-asdfsf-adfsd", "dsfasfdsa-asdfsf-asdfdsa"]

// req does not include addon_partner_id
{
"Ensemble Scientific": ["dsfdsf-adfsadf-asdfsf-adfsd", "dsfasfdsa-asdfsf-asdfdsa"],
"No Integrating Partner": ["dsfdsf-adfsadf-asdfsf-adfsd", "dsfasfdsa-asdfsf-asdfdsa"]
}
```

or using using table ids

```
// req includes addon_partner_id
["dsfdsf-adfsadf-asdfsf-adfsd", "dsfasfdsa-asdfsf-asdfdsa"]

// req does not include addon_partner_id
{
"1": ["dsfdsf-adfsadf-asdfsf-adfsd", "dsfasfdsa-asdfsf-asdfdsa"],
"0": ["dsfdsf-adfsadf-asdfsf-adfsd", "dsfasfdsa-asdfsf-asdfdsa"]
}
```

or plain rows
```
// req includes addon_partner_id
[{org_uuid: "dsfdsf-adfsadf-asdfsf-adfsd"}, {org_uuid: "dsfdsf-adfsadf-asdfsf-adfsd"}]

// req does not include addon_partner_id
[{addon_partner_id: 1,  org_uuid: "dsfdsf-adfsadf-asdfsf-adfsd"}, {addon_partner_id: 0, org_uuid: "dsfdsf-adfsadf-asdfsf-adfsd"}]
```

My preference would be the first to continue our efforts to keep frontend id agnostic and send the name ("translation key") instead.

kathyavini:
It was discussed at Tech Daily today (Feb 4th) but for our records, it was decided we like option (3) above -- array of objects which correspond to the actual record shape 🙏

Duncan-Brain (commented):
Okay it should be updated! We were so into formatting into entities on backend that I thought for sure we were sticking with it. My new favorite is option 3.

antsgar (approved):

kathyavini (approved):
Looks great. I'll merge as soon as the ~integration TypeScript fix is merged~ backend Docker container fix is merged (sorry for the moving target 🙇)

PR #3673, opened 2025-01-30
LF-4690b Fix Ensemble API call bug on token refresh and return descriptive errors on API call failure

kathyavini:
**Description**

Thank you so much @SayakaOno uncovering both of these issues!

Two fixes are in this PR:

1. I failed to spot a parameter that should have been deleted in the Ensemble API call refactor that will make your (first) call to Ensemble fail if you don't have valid access tokens in your `addon_partner` table. **Once the access tokens expired, this same error became [visible on beta as well](https://litefarm.sentry.io/issues/6260073802/?project=4505387020648448)**!
3. This bug highlighted that errors were not being propagated correctly to the controller from the Ensemble service calls. For the two new Ensemble endpoints, now in addition to the 404 from before (no organisations), there is the possibility of getting 500 errors if Ensemble auth fails or an otherwise non-specified error occurs when calling the Ensemble API (I'll describe testing below) 

Jira link: none

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

You can trigger the two kinds of errors by:

1. Auth error: clear your access + refresh tokens and mess up your Ensemble keys in `.env` a little. Should get back 500 / "Failed to authenticate with Ensemble"
2. Unspecified error on endpoint: There are a lot of ways to trigger but the fastest is probably to turn off your wifi! Or you can adjust the endpoint code so Ensemble server returns 404. Expected error: 500 / "Unable to fetch ESCI organisations"

**Checklist:**

### DIFF
--- a/packages/api/src/util/ensemble.js
+++ b/packages/api/src/util/ensemble.js
@@ -70,7 +70,7 @@ const ENSEMBLE_UNITS_MAPPING = {
 const getValidEnsembleOrg = async (org_uuid) => {
   const allRegisteredOrganisations = await getEnsembleOrganisations();
 
-  const organisation = allRegisteredOrganisations.find(({ uuid }) => uuid === org_uuid);
+  const organisation = allRegisteredOrganisations?.find(({ uuid }) => uuid === org_uuid);
 
   return organisation;
 };
--- a/packages/api/src/util/ensemble.js
+++ b/packages/api/src/util/ensemble.js
@@ -92,7 +92,7 @@ const getEnsembleSensors = async (farm_id) => {
 
   const devices = await getOrganisationDevices(farmEnsembleAddon.org_pk);
 
-  if (!devices.length) {
+  if (!devices?.length) {
     return { sensors: [], sensor_arrays: [] };
   }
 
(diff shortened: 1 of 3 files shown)


packages/api/src/util/ensemble.js
```diff
@@ -456,7 +462,7 @@ async function fetchAccessToken() {
 async function refreshAndRecall(axiosObject, onError, onResponse, retries) {
   const result = await refreshTokens();
   if (!result?.access || !result?.refresh) return result;
-  return ensembleAPICall(result.access, axiosObject, onError, onResponse, retries - 1);
```
kathyavini, line 459: This is the cause of the errors. There is one less parameter expected by ensembleAPICall now.

SayakaOno (commented):
Thank you Joyce! The fixes look good!

On a related note, removing tokens from the table is enough to trigger a `500` error.

It seems like the `refreshTokens` function isn't handling the error correctly... could you double check??
```
if (isAuthError(error)) {
  return await authenticateToGetTokens();
}
```

kathyavini:
@SayakaOno I get a 200 after clearing my tokens! Are your Ensemble `.env` variables definitely correct?

SayakaOno (approved):

SayakaOno:
@kathyavini That's it... 😅 Sorry, it's working! Thank you for confirming!

antsgar (approved):

PR #3675, opened 2025-01-31
LF-4691: UI to connect to ensemble

SayakaOno:
**Description**

Implement the first screen of the add sensor flow.

- Add `Partners` component (to be renamed to `Addons` or something similar later)
   - The position of the cross icon for clearing error is off - this will be fixed in another ticket.
- Update `PostSensor` container 
   - API integration (POST farmAddon, GET sensors)
   - Remove white background colour from the form (Loïc confirmed!)
- Add translations and colours (Includes translations and colours for LF-4693 and 4696 🙏)
- Make `Loading` component fixed relative to viewport (The file is currently nested within ContextForm, but I plan to extract it in a future PR. Keeping it for now to avoid a bigger diff)
- Implement ability to hide `Previous` button in `WithStepperProgressBar`
- Add `isCompactSideMenu` prop to `POST_SENSOR_URL` route
- Update `Map` component
   - Navigate to the Add sensors form when "Sensor" is selected
   - Remove sensor upload related code
- Update `ContextForm` to accept `formMode` prop ("onBlur", "onChange" etc.)
- ~~Delete unused files and code for bulk sensor upload~~

Jira link: https://lite-farm.atlassian.net/browse/LF-4691

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/webapp/src/components/Sensor/v2/Partners.tsx
```diff
@@ -0,0 +1,98 @@
...
+  hasActiveConnection: {
+    esci: boolean;
+  };
+};
+
+const validateUuidFormat = (value: string, errorMessage: string) => {
+  return uuidValidate(value) || errorMessage;
+};
+
+const PARTNERS = [{ name: 'Ensemble scientific', url: 'www.esci.io', logoPath: EsciLogo }];
```
SayakaOno, line 40: Do we need 'Ensemble scientific' to be translated? I would find it strange, but considering Hannah's name was translated, should this be too?

packages/webapp/src/containers/LocationDetails/PointDetails/SensorDetail/v2/PostSensor.tsx
```diff
@@ -16,51 +16,59 @@ import { useTranslation } from 'react-i18next';
...
   };
 
   const getFormSteps = () => [
     {
-      FormContent: () => <div>Partner selection view</div>,
-      onContinueAction: linkOrganizationId,
+      FormContent: () => <Partners />,
+      onContinueAction: linkEsci,
     },
     { FormContent: () => <div>ESCI devices view</div> },
```
Duncan-Brain, line 67: Is this temporary? I don't think ESCI devices view should be a step in the form. I see it looks that way in figma... but I am not sure its appropriate. If we are making a pattern similar to AnimalsSummary we should maybe reuse that instead and redirect to this page.

Duncan-Brain (approved):
Looking good! 

Left a couple of inline questions ..nothing too important I don't think.

Empty scrollbar is present:
Screenshot 2025-02-04 at 12 49 25 PM

I am not sure if Loading should be under `/ContextForm` -- it seems like it should be for all of app or replace existing Loading animation if we are redesigning a new one.

Could consider creating a `/Addon` directory and placing Partners cards under it since may be shared between farm settings, add sensors and probably other addons like layers and pivots.

packages/webapp/src/components/Sensor/v2/Partners.tsx
Duncan-Brain: I don't think so. Companies normally decide their own brand translation in other languages, but maybe for phonetic translations? If we do maybe we should change name to translation key in backend?

packages/webapp/src/components/Form/ContextForm/Loading.tsx
```diff
@@ -13,25 +13,52 @@
...
-    <div className={styles.loadingScreen}>
-      <div>
-        <Spinner />
+    <FloatingContainer isCompactSideMenu={isCompactSideMenu} distanceFromBottom={verticalMargin}>
+      <div
+        className={clsx(
+          styles.loadingScreen,
+          isCompactSideMenu ? styles.withCompactSideMenu : styles.withExpandedSideMenu,
+        )}
+        style={style}
```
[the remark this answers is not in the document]
Duncan-Brain: Okay, I didn't see the props being used by any parents yet so thought maybe it was just for testing. Would a className not be able to override variables too?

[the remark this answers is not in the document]
SayakaOno: Ah I see! It should be possible, but we'd have to do `isCompactSideMenu ? styles.withCompactSideMenu : styles.withExpandedSideMenu` in the parent component and write this much of CSS for a className, so using `style` seems more intuitive and simpler.
```
.className {
  height: calc(100vh - var(--global-navbar-height) - var(--vertical-margin) * 2);

  &.withCompactSideMenu {
    width: calc(100vw - var(--global-compact-side-menu-width) - var(--horizontal-margin) * 2);
  }
  &.withExpandedSideMenu {
    width: calc(100vw - var(--global-side-menu-width) - var(--horizontal-margin) * 2);
  }

  @include xs-breakpoint {
    &.withCompactSideMenu,
    &.withExpandedSideMenu {
      width: calc(100vw - var(--mobile-margin) * 2);
      height: calc(100vh - var(--global-navbar-height) - var(--mobile-margin) * 2);
      margin: var(--mobile-margin);
    }
  }
}
```

SayakaOno:
Updates:
- The sensors list is not part of the form anymore.
- Update `WithStepperProgressBar` to:
   - accept showLoading instead of onContinueAction in steps
   - accept onAfterSave prop
   - avoid page transition blocking after form is saved
   - adjust [stories](http://localhost:6006/?path=/story/components-contextform--stepper-success-form-with-loading)
- Fix capitalization in 'Ensemble Scientific'
- Fix padding styling (it was weirdly scrollable)
- Hide unnecessary scrollbar
- PARTNERS constant cleanup

Thank you for re-reviewing! 🙏

packages/webapp/src/util/siteMapConstants.ts
```diff
@@ -92,3 +92,6 @@ export const createCompleteTaskUrl = (id: string | number, hasAnimals: boolean):
 // Maps
 export const MAP_URL = '/map';
 export const POST_SENSOR_URL = '/create_location/sensor';
+
+// Sensors
+export const SENSORS = '/sensors';
```
SayakaOno, line 97: 1. `/sensors?partner_id=1`
2. `/addon/1/sensors`

I initially presented the first option to the team, but if we prioritize the addon over sensors as Duncan mentioned, we could probably go with something like the second one? Any thoughts?

antsgar: Both would be good! The first one also adds flexibility if we ever decide we want to have a screen showing all sensors for all manufacturers, which would be `/sensors`, but Loïc said we wouldn't want to do that so maybe irrelevant.

antsgar (approved):
Sorry it took a while for me to review this! This looks soooo good, the loading screen with the animation is on point 🔥 it's so clean! Loving it ❤️

packages/webapp/src/util/siteMapConstants.ts
Duncan-Brain: Yeah this is tough, I agree both are good. You could choose either. I just want to keep `/map` out of it I think. It might make sense to eventually reach these multiple ways so I think you can choose!

So many thoughts -- but are they useful? haha

Just trying to think about how a farmer might set up an addon. Especially if that addon does multiple things like ESCI -- they provide us with both irrigation prescriptions and sensors. So I can see how navigating from `/farm_setting` to `/addon` makes sense. (option 2)

But I also understand that eventually (in the distant future) might be need for a `/farm_equipment/` module where we would move access to `/sensors` or `/irrigation_equipment` and then configuring or viewing there. (option 1)

Same with `/Insights/irrigation_prescriptions` etc.

Since for some reason we are adding an esci addon (with both prescriptions and sensors) with the "add sensors" button it might make sense either way 🤷

Duncan-Brain (approved):
Looking great I love how ContextForm is improving over time!

packages/webapp/src/util/siteMapConstants.ts
SayakaOno: Thank you for reviewing and sharing your thoughts!
I was considering it from both the addon and sensors perspectives as Duncan mentioned, but I couldn't reach a conclusion, it's tough!
That said, if we were to support sensors from non-partners in the future, option 1 would make more sense, so I'll go with that this time!

PR #3680, opened 2025-02-05
LF-4718 (2) Repair Deployment of Webapp

kathyavini:
**Description**

This PR has 2 parts:

1. I bumped the webapp TypeScript version to match the backend, which was set on ^5.6.2. This will have to be done at some point (frontend was on 4.9.5) so I figured might as well do it now and try to grab all the type errors -- although hopefully at this point they are exhausted!
2. Fixes for type errors that block `pnpm build` with a fresh `node_modules` install (pnpm-lock deleted / regenerated). There were only two files that needed to be touched for my build to complete

Jira link: https://lite-farm.atlassian.net/browse/LF-4718

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

Reinstalled `node_modules` as described above (without pnpm-lock). `tsc` passed in both pre-commit hook and pnpm build.

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/Animals/AddAnimalsFormCard/AnimalSelect.tsx
+++ b/packages/webapp/src/components/Animals/AddAnimalsFormCard/AnimalSelect.tsx
@@ -50,7 +50,8 @@ export function AnimalTypeSelect<T extends FieldValues>({
         control={control}
         rules={{
           required: { value: true, message: t('common:REQUIRED') },
-          validate: hookFormSelectOptionMaxLength as (selectedOption: any) => string | boolean,
+          // @ts-ignore
+          validate: hookFormSelectOptionMaxLength,
         }}
         render={({ field: { onChange, value } }) => (
           <CreatableSelect
@@ -97,7 +98,8 @@ export function AnimalBreedSelect<T extends FieldValues>({
         name={name}
         control={control}
         rules={{
-          validate: hookFormSelectOptionMaxLength as (selectedOption: any) => string | boolean,
+          // @ts-ignore
+          validate: hookFormSelectOptionMaxLength,
         }}
         render={({ field: { onChange, value } }) => (
           <CreatableSelect
--- a/packages/webapp/src/components/Animals/AddAnimalsFormCard/AnimalSelect.tsx
+++ b/packages/webapp/src/components/Animals/AddAnimalsFormCard/AnimalSelect.tsx
@@ -50,13 +50,13 @@ export function AnimalTypeSelect<T extends FieldValues>({
         control={control}
         rules={{
           required: { value: true, message: t('common:REQUIRED') },
-          validate: hookFormSelectOptionMaxLength,
+          validate: hookFormSelectOptionMaxLength as (selectedOption: any) => string | boolean,
         }}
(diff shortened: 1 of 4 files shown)


packages/webapp/src/components/Animals/AddAnimalsFormCard/AnimalSelect.tsx
```diff
@@ -50,13 +50,13 @@ export function AnimalTypeSelect<T extends FieldValues>({
         control={control}
         rules={{
           required: { value: true, message: t('common:REQUIRED') },
-          validate: hookFormSelectOptionMaxLength,
+          validate: hookFormSelectOptionMaxLength as (selectedOption: any) => string | boolean,
```
Duncan-Brain, line 53: I am still trying to understand TS but I think I am still scared of type assertions. So without really knowing in lieu of typing the hook form validation file -  I like ts-ignore. Not a blocker tho.

Duncan-Brain (approved):
Nice typescript matching 👍

packages/webapp/src/components/Animals/AddAnimalsFormCard/AnimalSelect.tsx
kathyavini: That's totally fine with me! I guess both circumvent typing in this file, but I like how `// @ts-ignore` has an implicit "todo" quality 🙂

Duncan-Brain (approved):

packages/webapp/src/components/Animals/AddAnimalsFormCard/AnimalSelect.tsx
navDhammu: Random comment - but you can use `@ts-expect-error` in place of `@ts-ignore`. It does the same thing, but it has a small benefit in that it'll give an error if its being used unnecessarily. Meaning that if the following line after `ts-expect-error` doesn't have a type error, it'll point that out by giving an error. Whereas ts-ignore will just keep ignoring it forever, even if for example someone fixes it but forgets to remove the ts-ignore.

antsgar (approved):

packages/webapp/src/components/Animals/AddAnimalsFormCard/AnimalSelect.tsx
kathyavini: Oops I also just figured out the fix for the type error a minute after merge 😂 

It was not a mere JS->TS issue, but the React Hook Form validate function wants it second argument (if supplied) to be FormState. I'll open a quick PR with the fix so we don't have to ts-ignore on the same validator later. No rush though and not really related to webapp deployment (WHICH JUST SUCCEEDED!! 🎉)

kathyavini: @navDhammu ooooh nice, I think I might have even seen one of those no longer relevant `//@ts-ignore` in this very file! I'll definitely start using `@ts-expect-error`

Duncan-Brain: Very cool Nav thanks!

PR #3682, opened 2025-02-06
LF-4722 Backend container crashes due to missing server.js file

kathyavini:
**Description**

This PR moves the backend's tsc (`npm run build`) step into the Dockerfile (image creation).

What is happening now:
1. Backend image is built according to Dockerfile, and all assets (ts) are copied
2. `npm run start:prod` is run once the image is complete. This is supposed to run `tsc` and output js. The output js (server.js) is then run
3. `server.js` file can't be located, and the container crashes (see Jira ticket). I don't know if `/dist` was output because it's not part of the image and the container is crashed at this point so I can't `exec -it` it!

What I would like to do is run tsc within the Dockerfile so the `dist/` file is included in the image

It's a diagnostic change because I want to inspect the `/dist` folder using the built image, but there are also a few Stack Overflow code examples that show this pattern so it might be preferable anyway
- [Can not build typescript on Dockerfile](https://stackoverflow.com/a/61205563)
- [How to compile typescript in Dockerfile](https://stackoverflow.com/a/52298481)

Anyway, whatever the issue is, it's specific to Docker as `npm run start:prod` works fine locally to build, create `/dist` and run `server.js`.

Jira link: https://lite-farm.atlassian.net/browse/LF-4722

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

Haven't really tested, I don't think I have an up to date local Docker setup at the moment...

**Checklist:**

### DIFF
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -13,7 +13,7 @@
     "test-end2end": "jest endToEnd",
     "jest-w": "jest --watch",
     "start": "set NODE_ENV=development&& node --import @swc-node/register/esm-register src/server.ts",
-    "start:prod": "npm run build && node dist/api/src/server.js",
+    "start:prod": "node dist/api/src/server.js",
     "build": "rimraf dist && tsc -p src/tsconfig.json",
     "dev": "set NODE_ENV=development && node --watch --import @swc-node/register/esm-register src/server.ts",
     "debug-email-template": "DEBUG=email-templates npm run dev",
--- a/packages/api/prod.Dockerfile
+++ b/packages/api/prod.Dockerfile
@@ -10,4 +10,6 @@ COPY ./api/ /usr/src/app/
 
 COPY ./shared/ /usr/src/shared/
 
+RUN npm run build
+
 CMD npm run migrate:dev:db && npm run start:prod || echo // `date` >> crashlog.js 


Duncan-Brain:
I don't think typescript is installed, it is a dev dependency. Your links show a few examples. I think also explain why it works locally -- could try trashing local packages and local dist to test.

packages/api/package.json
```diff
@@ -13,7 +13,7 @@
     "test-end2end": "jest endToEnd",
     "jest-w": "jest --watch",
     "start": "set NODE_ENV=development&& node --import @swc-node/register/esm-register src/server.ts",
-    "start:prod": "npm run build && node dist/api/src/server.js",
+    "start:prod": "node dist/api/src/server.js",
```
Duncan-Brain, line 16: should we rename to just `start` while we are here ;)

antsgar (approved):

kathyavini:
> I don't think typescript is installed, it is a dev dependency.

@Duncan-Brain honesty I was surprised it worked. Unless dev dependencies work differently than I thought, I think it may be the line

`NODE_ENV: integration`

in the docker-compose.beta.yml file, in which case we could expect a failure at that same point in prod 🤔 And as you mentioned with the typescript-related dependencies beyond typescript itself, it might be a bit of a messy failure. I think I'll make a ticket but leave it for the time being? A contributor with interest in Docker might want to pick it up and test.

Duncan-Brain:
@kathyavini could be that line yes maybe. We should also use `npm ci` on prod, could be that too.

At least we found the typo/path thing. We should prob change docker instead of npm script but its all good for now. Hopefully nothing else relies on `/api` path to that end.

kathyavini:
> Hopefully nothing else relies on /api path to that end.

@Duncan-Brain it's funny, I had the exact opposite thought. The Docker container has ALWAYS been on `/app` so I would be more weary of changing the dockerfile and introducing an issue than the `start:prod` script, which isn't used anywhere else. But probably the reality is that all the filepaths are relative and either would have worked!

PR #3685, opened 2025-02-10
LF-4693 Display existing Ensemble addon

kathyavini:
**Description**

This PR is the display of the active Ensemble connection in the Add Sensors page (thank you to @SayakaOno for already doing all of this!!) and the Farm Settings page.

"Disconnecting" the Addon will trigger the warning modal, but it will not actually disconnect -- there is a separate user story for the disconnection feature.

Notes:
- There is some scope for refactoring a more general "Addon Card" pattern within PureFarmAddons, but I thought it was premature to attempt this without knowing what future add-ons will look like (e.g. will they always have an organisation ID? Will they always have an internal link to an overview page)?
- The [Figma](https://www.figma.com/design/eMffC0h3JPlWtzNttxJpil/Litefarm-Updated?node-id=27080-147382&m=dev) "Disconnect ESCI" button is green. But that's the active state of that variant! So I used the variant as defined; i.e. it will be green when you click on it. Update: confirmed with Loïc that the active state being displayed in Figma was in error and this is the standard variant
- Within Farm Settings, I left the text of the link to the page that will eventually show all Ensemble sensors consistent with what is shown in Add Sensors (rather than what is on Figma); I thought we landed on not changing it? Because it's within the Ensemble Addon Card? But in any case if it is changed I think it would make more sense for those links to match (be changed in conjunction). This can be double-checked on Wednesday. Update: checked, and both strings have been changed to match the Farm Settings version

Jira link: https://lite-farm.atlassian.net/browse/LF-4693

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

There


packages/webapp/src/components/Profile/Addons/index.tsx
```diff
@@ -0,0 +1,152 @@
...
+};
+
+const PureFarmAddons = ({
+  hasActiveConnection,
+  organizationIds,
+  onDisconnect,
+}: PureFarmAddonsProps) => {
+  const { t } = useTranslation(['translation', 'common']);
+
+  useEffect(() => {
```
kathyavini, line 47: Despite the pattern shown in the React Router v4/v5 docs to be functional to create the hashed URL:

```
to={{ pathname: '/farm', hash: '#esci-addon' }}>
```

it didn't actually route to that component. Has anyone has gotten this working without a`useEffect()` by chance?

packages/webapp/src/components/Table/TableV2.jsx
```diff
@@ -163,7 +165,7 @@ export default function TableV2(props) {
 
   return (
     <Box sx={{ width: '100%' }}>
-      <TableContainer sx={{ maxHeight }}>
+      <TableContainer sx={{ maxHeight }} className={tableContainerClass}>
```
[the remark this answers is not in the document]
SayakaOno: Sorry about this!! 😱 Those changes were supposed to be included in my next PR, so I dropped the commit from my previous branch! I forgot it was also in this branch... Feel free to leave them in, sorry again! 🙏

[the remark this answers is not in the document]
kathyavini: Oh no worries then 😌 Thank you for explaining the mystery!

packages/webapp/src/components/Profile/Addons/index.tsx
SayakaOno: It doesn't work even if I directly enter the URL in the address bar, so I think the fixed header or sidebar is preventing the behaviour 😢
But the implemented behaviour looks beautiful!

packages/webapp/src/containers/Profile/Farm/Addons/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ */
+import PureFarmAddons from '../../../../components/Profile/Addons';
+import { useGetFarmAddonQuery } from '../../../../store/api/apiSlice';
+import { PARTNERS } from '../../../LocationDetails/PointDetails/SensorDetail/v2/constants';
+
+const FarmAddons = () => {
+  const { isSuccess: hasESciConnection, data: ESciData } = useGetFarmAddonQuery(
```
SayakaOno, line 20: Is the naming `ESciData` a known convention? For consistency, I think we should use something like `esciDataArray`...

SayakaOno (commented):
Looks great! Just left one comment about naming :)

packages/webapp/src/containers/Profile/Farm/Addons/index.tsx
kathyavini: Ah thank you! I actually opened a thread in UX-UI about this on Friday but I'm not sure it was interpreted correctly. Let me ask directly in the Ensemble channel as well.

I have seen ESci before and it is the abbreviation that Ensemble itself has used for e.g. their old Postman API docs -- they call that `ESci Cloud`. I highly doubt that they would ever abbreviate ESCI because it just doesn't make sense as an acronym.

But as far as this PR, I meant to not include it so we can stay consistent across our code until we confirm officially, so I'll remove it!

SayakaOno (dismissed):

packages/webapp/src/containers/Profile/Farm/Addons/index.tsx
SayakaOno: Very interesting, I didn't know any style guide addressed these cases!
Google's guide would be easier to follow (no need to overthink, right?), but I'd like to write "KPI" rather than "Kpi" 😂
Honestly, I wouldn't mind if they were mixed, so... if anyone else has a preference, I'd follow that!

kathyavini:
@SayakaOno I COMPLETELY forgot about that! Are they already "hideable" or will it take some manipulation of `<ContextForm />` as well? I think it's definitely in the scope of this PR but if you have an idea about how to implement it already...! 🙏 ❤️

packages/webapp/src/containers/LocationDetails/PointDetails/SensorDetail/v2/Partners.tsx
```diff
@@ -13,11 +13,27 @@
...
-const Partners = () => {
-  // TODO: LF-4693 GET /farm_addon?integrating_partner_id=1
+const Partners = ({
+  setIsEditing,
+}: {
+  setIsEditing: React.Dispatch<React.SetStateAction<boolean>>;
+}) => {
+  const { isSuccess: hasEsciConnection } = useGetFarmAddonQuery(
+    `?addon_partner_id=${PARTNERS.ESCI.id}`,
+  );
```
kathyavini, line 26: The "correct" way to control when this query re-runs is with tag invalidation, but I kind of want to add a permanent `{ refetchOnMountOrArgChange: true },`  here and to farm settings (which I use in development anyway) and nip all the concurrency issues in the bud before they can even arise. It just seems like the kind of data we want always up-to-date regardless of the user who has initiated the connection/disconnection 🤔

Thoughts?

SayakaOno: That makes sense, but it also makes me wonder if there’s a possibility where no addon exists, yet the user still sees sensors or vice versa…

kathyavini: Oh yes, definitely because the GET `sensors/` is also RTK Query cached. So this is probably something we should think about at the end of the epic as a single unit in terms of timing of re-fetch and concurrency.

SayakaOno (approved):
Looks good to me!

antsgar (dismissed):
Looks great! ❤️ love it

packages/webapp/src/components/Profile/Addons/index.tsx
antsgar: It looks like React Router v4 doesn't support the anchor links, I guess it just adds the hash to the URL but does nothing with it -- someone actually developed an external library for it, but it seems it pretty much does this same thing, so this works perfect.

antsgar:
Oh one comment that I forgot to add, I wonder if we should place the `FarmAddons` component outside of the `Profile` folder, since it's meant to be used outside of Settings too?

packages/webapp/src/components/Profile/Addons/index.tsx
```diff
@@ -0,0 +1,152 @@
...
+  hasActiveConnection: {
+    esci: boolean;
+  };
+  organizationIds: {
+    esci: string;
+  };
+  onDisconnect: {
+    esci: () => void;
+  };
+};
```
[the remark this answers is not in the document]
kathyavini: I really like this! In what sense would we have to do it throughout? Like in both Partner and FarmAddons? I think that's okay...!

packages/webapp/public/locales/en/translation.json
```diff
@@ -279,6 +279,19 @@
...
   },
+  "ADDON": {
+    "DISCONNECT_MODAL": {
+      "CONFIRM": "Disconnect",
+      "CANCEL": "Don't disconnect"
+    },
+    "ESCI": {
+      "MODAL_INFO": "You will not be able to access your ESCI sensors data until you reconnect",
+      "MODAL_TITLE": "Are you sure you want to disconnect your ESCI sensors?",
+      "DISCONNECT_ESCI": "Disconnect ESCI",
```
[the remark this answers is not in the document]
kathyavini: Sayaka said she was making a separate PR for this (in the sensor list PR's [comment](https://github.com/LiteFarmOrg/LiteFarm/pull/3686#issuecomment-2654814707)) where I figured it would just be done as one swoop

kathyavini:
> Oh one comment that I forgot to add, I wonder if we should place the FarmAddons component outside of the Profile folder, since it's meant to be used outside of Settings too?

@antsgar it's not currently scoped to be used outside of the Farm Profile page 🤔 But is that what we had said at some point? That there will be a dedicated Addons page at some point?

packages/webapp/src/components/Profile/Addons/index.tsx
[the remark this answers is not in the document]
antsgar: I thought maybe there were going to be a bunch of places where when typing we'd have to make the call to make the types flexible like this or to "hardcode" ESci as the only option, it might not affect that many types though

antsgar:
@kathyavini don't mind me, for a second I thought FarmAddons was being used both in the profile page and in the add sensors page, but it's not! It's perfectly placed as is 👍

antsgar (approved):

SayakaOno (approved):

PR #3687, opened 2025-02-11
Lf 4703 nice to have add base properties to farm addon

Duncan-Brain:
**Description**

This PR adds base properties to the `farm_addon` object. 

Also necessitates the addition of an `id` primary key to distinguish between deleted records.

Notes:
- bumped up in my queue because it will really affect how DELETE endpoint works
- Partial unique index constraint (PUI) used to guarantee maximum 1 single non deleted entry for a composite key. Use of this constraint is an indication to rethink the architecture. PUI is specifically difficult when working with multiple records at a time - like `soil_amendment_task_products`.  Looking back soil amendment task products should not have base properties - it should use hard delete -- so I forget why I added base properties there maybe I was still learning. Here it makes sense to have base properties on `farm_addon`, so the main re-think is about if multiple addons with the same partner should be allowed -- and I think the answer is: in the future we may do that so constraining it for now is ok and can be removed later. We will also not be manipulating multiple records at a time so the drawback is less.
- I might have some questions about the CRUD record cycle in the DELETE endpoint PR

Jira link: https://lite-farm.atlassian.net/browse/LF-4703

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain): Tested in Postman

**Checklist:**

### DIFF
--- a/packages/api/db/migration/20250203144322_add_base_properties_to_farm_addon.js
+++ b/packages/api/db/migration/20250203144322_add_base_properties_to_farm_addon.js
@@ -15,7 +15,7 @@
 
 export const up = async function (knex) {
   await knex.schema.alterTable('farm_addon', (t) => {
-    // Knex apparently does not rename indexes aliong with table names
+    // Knex apparently does not rename indexes along with table names
     t.dropPrimary('farm_external_integration_pkey');
   });
   await knex.schema.alterTable('farm_addon', (t) => {
@@ -48,6 +48,7 @@ export const up = async function (knex) {
 };
 
 export const down = async function (knex) {
+  await knex('farm_addon').where({ deleted: true }).del();
   await knex.schema.alterTable('farm_addon', (t) => {
     t.dropIndex(['farm_id', 'addon_partner_id'], 'farm_addon_uniqueness_composite');
     t.dropPrimary();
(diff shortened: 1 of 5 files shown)


kathyavini (commented):
It looks pretty good to me!

In what sense does the PUI indicate we need to rethink architecture? I didn't quite follow that part of the PR description but I think I would be interested in hearing about that in tech daily -- maybe Thursday or tomorrow if the dev-design is fast?

The only thing missing that I could spot is that `farmAdonModel.getOrganisationIds` (this is called by `getEnsembleSensors`) will also need a `.whereNotDeleted()` with this change or you'll keep getting back sensors even after deleting the `farm_addon` record.

Duncan-Brain:
@kathyavini 

Just that its a one-off thing to need to use so its good to second guess it. Happy to explain the two times I have used it and why I regret the previous use of it -- but think its fine here for now.

Thanks for finding will fix!

kathyavini (approved):
Oh yeah, that's a good call to keep the old functions compatible until the point at which we remove them! 👍

antsgar (approved):
Looks good! ☺️ And I honestly don't dislike the constraint, I think it's a good solution

PR #3694, opened 2025-02-20
LF-4738 Delete Task" Button Misalignment in Custom Task Screen

den4ik1203:
**Description**

This update fixes a bug that was causing [briefly describe the issue, e.g., "the delete button to be misaligned in the footer section"].

Changes Made:
Adjusted the CSS layout to properly align the delete button to the left.
Updated the flexbox properties in .footerContainer to ensure correct spacing.
Refactored the JSX structure for better readability and maintainability.
Reason for the Fix:
The delete button was not aligning as expected due to incorrect justify-content and margin rules. This fix ensures a consistent UI layout.

Jira link:
https://lite-farm.atlassian.net/browse/LF-4738
**Type of change**

**How Has This Been Tested?**

- [x] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/Task/TaskReadOnly/index.jsx
+++ b/packages/webapp/src/components/Task/TaskReadOnly/index.jsx
@@ -462,23 +462,25 @@ export default function PureTaskReadOnly({
       )}
 
       {isAdmin && isCurrent && !isDeleting && (
-        <IconLink
-          className={styles.deleteText}
-          style={{ color: 'var(--grey600)' }}
-          icon={
-            <TrashIcon
-              style={{
-                fill: 'var(--grey600)',
-                stroke: 'var(--grey600)',
-                transform: 'translate(0px, 6px)',
-              }}
-            />
-          }
-          onClick={preDelete}
-          isIconClickable
-        >
-          {t('TASK.DELETE.DELETE_TASK')}
-        </IconLink>
+        <div className={styles.pageContainer}>
+          <div className={styles.mainContent}></div>
+
+          <div className={styles.footerContainer}>
+            <IconLink
+              className={styles.deleteText}
+              onClick={preDelete}
+              isIconClickable
+              icon={<TrashIcon className={styles.trashIcon} />}
+            >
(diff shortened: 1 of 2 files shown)


SayakaOno (changes requested):
Could you remove invisible div and buttons?
These changes affect other task types, so it would be best to focus on fixing just what’s broken.

Duncan-Brain (changes requested):
Hi @den4ik1203 thanks for working on this!

I think Sayaka is right you have added extra buttons, these should be removed.

At least for me I am seeing that the fix is not correct for custom tasks with no location specified it just moves closer than before:
Screenshot 2025-02-21 at 9 47 54 AM

Some things to think about:
- It seems like the map fills up available space and pushes the delete button down in most cases you can see this as i delete normal task inputs in video
- If we want to make sure the delete icon is with buttons, have you considered grouping the delete button with form buttons?
- there is a lot of css there - do you need them all? Ideally as little css as possible is added (I think your instinct is right about min-height and it should not be used):



https://github.com/user-attachments/assets/e1f8f468-492a-431e-be0a-4f1b62a0d232

Duncan-Brain:
Hey @den4ik1203 it is looking closer to what I might expect!

Two things: 
1. I am seeing th mobile buttons seem to have changed:
Screenshot 2025-02-25 at 2 28 47 PM
Screenshot 2025-02-25 at 2 28 29 PM

2. We want to make sure that the delete button renders according to previous conditions only, and the other buttons render according to previous conditions only. And it is also nice that we do not duplicate conditions if there are shared.

In the code for the delete buttons the previous condition for rendering was:

` isAdmin && isCurrent && !isDeleting` -> render delete button 

In the previous code the condition for rendering the bottom buttons was:

`canCompleteTask || canAbandonTask) && isCurrent` -> render buttons

The new code and condition for rendering delete button is:
`canCompleteTask || canAbandonTask) && isCurrent .... && isAdmin && isCurrent && !isDeleting` -> render delete button

It might be okay but it is good to double check if there are cases where it is not rendering when it should because of extra conditions.

PR #3692, opened 2025-02-18
Lf 4700 create delete endpoint for farm addon

Duncan-Brain:
**Description**

Creates a DELETE endpoint for `farm_addon` and aligns the resource lifecycle for a farm_addon based on discussion in Jira. Will add tests for all of `farm_addon` in a follow up PR. 

Notes:
- modifies `hasFarmAccess()` to check that a farm owns a resource when the table does not name it's `id` property with `${resource}_id` and instead uses plain `id` property.
- Updates GET to return `id` necessary for deletion.
- Updates POST to insert or return 409 if resource already exists

Jira link: [LF-4700](https://lite-farm.atlassian.net/browse/LF-4700)

**Type of change**

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

**How Has This Been Tested?**

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [ ] Passes test case
- [ ] UI components visually reviewed on desktop view
- [ ] UI components visually reviewed on mobile view
- [x] Other (please explain): 
   - Add a `farm_addon`, the try adding again -> 409 conflict
   - Get `farm_addon` then delete farm addon -> success
   - Add a `farm_addon` again -> success

**Checklist:**

- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] The precommit and linting ran successfully
- [ ] I have added or updated language tags for text that's part of the UI
- [ ] I have added "MISSING" for all new language tags to languages I don't speak
- [ ] I have added [the GNU General Public License](https://lite-farm.atlassian.net/l/cp/BT0Dd7WW) to all new files

[LF-4700]: https://lite-farm.atlassian.net/browse/LF-4700?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3696, opened 2025-02-20
LF-4701 Call DELETE endpoint when clicking disconnect from Farm Addon

kathyavini:
**Description**

This PR connects the delete endpoint from @Duncan-Brain's PR #3692 to the frontend of app, wrapping up the disconnect user story and the whole FarmAddon arc (🎉!)

There weren't any snackbars on this flow in Figma so I didn't include them, but have been wondering about that. I think it would be okay to exclude because there is no 'success' snackbar for adding an addon (you are routed to the sensor list instead) and with the disconnect confirmation modal it would be a bit of a 'noisy' screen with two popups in a row. But it also kind of depends on how much it is supposed to feel like deleting another entity (e.g. task, animal, transaction) which always snackbar 🤔  I think I'm okay with the snackbar-less UX here, but what do you guys think? Should we check with Loïc?

Jira link: https://lite-farm.atlassian.net/browse/LF-4701

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/containers/Profile/Farm/Addons/index.tsx
+++ b/packages/webapp/src/containers/Profile/Farm/Addons/index.tsx
@@ -12,6 +12,9 @@
  *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
  *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
  */
+import { useDispatch } from 'react-redux';
+import { useTranslation } from 'react-i18next';
+import { enqueueErrorSnackbar, enqueueSuccessSnackbar } from '../../../Snackbar/snackbarSlice';
 import PureFarmAddons from '../../../../components/Profile/Addons';
 import { useGetFarmAddonQuery, useDeleteFarmAddonMutation } from '../../../../store/api/apiSlice';
 import { PARTNERS } from '../../../AddSensors/constants';
@@ -33,10 +36,20 @@ const FarmAddons = () => {
 
   const [deleteFarmAddon] = useDeleteFarmAddonMutation();
 
+  const dispatch = useDispatch();
+  const { t } = useTranslation(['message']);
+
+  const handleDisconnect = async (addonId: number) => {
+    try {
+      await deleteFarmAddon(addonId).unwrap();
+      dispatch(enqueueSuccessSnackbar(t('FARM_ADDON.SUCCESS_DISCONNECT_ADDON')));
+    } catch (error) {
+      dispatch(enqueueErrorSnackbar(t('FARM_ADDON.FAILED_DISCONNECT_ADDON')));
+    }
+  };
+
   const onDisconnect = {
-    esci: () => {
-      deleteFarmAddon(esciData?.id);
-    },
+    esci: () => handleDisconnect(esciData?.id),
   };
 
   const hasAnyActiveConnection = Object.values(hasActiveConnection).some(Boolean);
--- a/packages/webapp/src/containers/Profile/Farm/Addons/index.tsx
+++ b/packages/webapp/src/containers/Profile/Farm/Addons/index.tsx
@@ -13,7 +13,7 @@
  *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
  */
 import PureFarmAddons from '../../../../components/Profile/Addons';
-import { useGetFarmAddonQuery } from '../../../../store/api/apiSlice';
+import { useGetFarmAddonQuery, useDeleteFarmAddonMutation } from '../../../../store/api/apiSlice';
 import { PARTNERS } from '../../../AddSensors/constants';
 
 const FarmAddons = () => {
@@ -31,8 +31,12 @@ const FarmAddons = () => {
     esci: esciData?.org_uuid,
   };
 
+  const [deleteFarmAddon] = useDeleteFarmAddonMutation();
+
   const onDisconnect = {
-    esci: () => {}, // TODO: LF-4701
+    esci: () => {
+      deleteFarmAddon(esciData?.id);
+    },
   };
 
   const hasAnyActiveConnection = Object.values(hasActiveConnection).some(Boolean);
--- a/packages/webapp/src/containers/Profile/Farm/Addons/index.tsx
+++ b/packages/webapp/src/containers/Profile/Farm/Addons/index.tsx
@@ -20,15 +20,15 @@ import { useGetFarmAddonQuery, useDeleteFarmAddonMutation } from '../../../../st
 import { PARTNERS } from '../../../AddSensors/constants';
 
 const FarmAddons = () => {
-  const { isSuccess: hasEsciConnection, data: esciDataArray } = useGetFarmAddonQuery(
+  const { data: esciDataArray = [] } = useGetFarmAddonQuery(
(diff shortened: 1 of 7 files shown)


Duncan-Brain:
If we are moving away from snackbars it would be nice to know the specific situations or if all. I would otherwise assume Loic just forgot to add it in to the designs and we follow our normal pattern (even for success).

packages/webapp/src/store/api/apiSlice.ts
```diff
@@ -261,6 +261,13 @@ export const api = createApi({
       query: (param = '') => `${farmAddonUrl}${param}`,
       providesTags: ['FarmAddon'],
     }),
+    deleteFarmAddon: build.mutation<FarmAddon[], number>({
```
SayakaOno, line 264: I think `FarmAddon[]` should be `void`.

I was making sure the type (`FarmAddon[]`) is the return type here.
Screenshot 2025-02-21 at 10 05 03 AM

And found that the result contains an error object.
Screenshot 2025-02-21 at 10 06 50 AM

Not related to this PR, but the correct response status seems to be `204`(no content) instead of `200`... (in `deleteFarmAddon` in the API)

SayakaOno (commented):
I probably wouldn't have noticed that snackbars aren't implemented if you hadn't mentioned it, but I feel like having feedback for the user's actions is nice 🤔

packages/webapp/src/store/api/apiSlice.ts
Duncan-Brain: Looking to change to 204 in api tests PR!

kathyavini: Sorry, should have read the controller more carefully 🙇

kathyavini:
I have made the following changes to this branch:
- added snackbars
- fixed the delete mutation return type (thank you @SayakaOno!)
- alphabetised some keys by running `pnpm i18n` now that we are able to (again, thank you Sayaka!!)

And then in looking at more closely at fail conditions around `farmAddon`:
- conditionally invalidated tags, see https://stackoverflow.com/a/75952409 -- actually I didn't know the default behaviour was to invalidate even if the mutation errors; that doesn't seem like what we want
- altered the farmAddon GET to return an empty array!! Actually the 404 had been bugging me the whole time since we had _just_ decided on the other pattern (200 + empty array) in [this recent conversation here](https://github.com/LiteFarmOrg/LiteFarm/pull/3665#discussion_r1928970950). Using the `success`/`error` distinction as had been done originally here meant API error = farm addon disappears from frontend. It seems more reasonable that the API failing should have no effect on the visibility of the addon -- only a true "no addons exist" response should clear the existing cache.

Thank you for your comments and if you could kindly please re-review 🙏

SayakaOno (dismissed):
Looks great!
Nice find on the `invalidatesTags`, and returning `[]` instead of an error feels right ☺️

One thing I noticed is that the button's focused(?) style affects the text at smaller browser sizes. I think we could adjust it when we have time!

https://github.com/user-attachments/assets/338296bb-0f89-4d5b-8cff-e684024b87e9

packages/webapp/src/containers/Profile/Farm/Addons/index.tsx
```diff
@@ -12,27 +12,44 @@
...
+import { useTranslation } from 'react-i18next';
+import { enqueueErrorSnackbar, enqueueSuccessSnackbar } from '../../../Snackbar/snackbarSlice';
 import PureFarmAddons from '../../../../components/Profile/Addons';
-import { useGetFarmAddonQuery } from '../../../../store/api/apiSlice';
+import { useGetFarmAddonQuery, useDeleteFarmAddonMutation } from '../../../../store/api/apiSlice';
 import { PARTNERS } from '../../../AddSensors/constants';
 
 const FarmAddons = () => {
-  const { isSuccess: hasEsciConnection, data: esciDataArray } = useGetFarmAddonQuery(
+  const { data: esciDataArray = [] } = useGetFarmAddonQuery(
```
Duncan-Brain, line 23: Instead of providing a default value should we be using the `isLoading` to determine hasActiveConnection (and therefore if we should render) ?

Duncan-Brain (approved):
Nice, I also asked Anto last week if I could change the 404 in the PR where she scoped it! https://lite-farm.atlassian.net/browse/LF-4694

I added one comment for you to consider --  also feel free to update the delete return status to 204 if you want to to make the snackbar and store work properly on disconnect. I have added it in another PR but QA will probably flag it here first.

packages/webapp/src/containers/Profile/Farm/Addons/index.tsx
kathyavini: I'm not sure if I follow because `isLoading` can be false (loaded) and `hasActiveConnection` also false (because there are no active farm addons). However there IS a little blip not here but in `<Partners />` that I should probably have fixed with `isLoading`!

I would like to clean up that blip along with @SayakaOno's button issue too -- I hate those little visual things 🤣 So I'll make a follow up PR to prevent scope creep on this one.

> also feel free to update the delete return status to 204 if you want to to make the snackbar and store work properly on disconnect. I have added it in another PR but QA will probably flag it here first.

In what sense are you finding the snackbars to still not work properly such that QA would flag? I don't think RTK Query does anything different with a 200 vs 204 and they are working properly for me 🤔

kathyavini:
> Nice, I also asked Anto last week if I could change the 404 in the PR where she scoped it! https://lite-farm.atlassian.net/browse/LF-4694

❤️ I think dev team is now completely united in our dislike of the 404 pattern 😂

packages/webapp/src/containers/Profile/Farm/Addons/index.tsx
Duncan-Brain: In the sense if it rendered using the [] it would be incorrect -- possible blip somewhere ;) 

> In what sense are you finding the snackbars to still not work properly such that QA would flag? I don't think RTK Query does anything different with a 200 vs 204 and they are working properly for me 🤔

Do you not receive a error snackbar when disconnecting? Backend works correctly but my error snackbar shows because it cant parse the "O" in "OK" lol from the 200 response -- I thought that was what Sayaka was seeing also in her comment. I have to relog to see updates on frontend. The 204 fixes it.

SayakaOno (approved):

packages/webapp/src/containers/Profile/Farm/Addons/index.tsx
kathyavini: Alright, I've changed the response code but let us pocket this as a fun mystery to someday solve.

@antsgar I am looping you in too because this is something fun and wacky. With no differences that we can detect between our `@reduxjs/toolkit` versions (or anything else that seemed potentially relevant), Duncan and Sayaka get an error on the `.unwrap()` of a `.sendStatus(200)` from the controller -- and for me it works just fine.

Any of these solve the issue: `.sendStatus(204)`, `.status(200).send()` or `responseHandler: "text"` in apiSlice (from [this StackOverflow](https://stackoverflow.com/a/73480249)) but the question is why the inconsistency between our machines 😄 Same branch, same packages.

PR #3697, opened 2025-02-25
LF-4743 "Completion notes" instead of "Abandonment notes" is displayed once the user tries to abandon the crop plan.

den4ik1203:
**Description**

Fix completion notes vs abandonment notes display issue

Jira link:
https://lite-farm.atlassian.net/browse/LF-4743
**Type of change**

**How Has This Been Tested?**

- [x] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/Crop/CompleteManamgenentPlan/PureCompleteManagementPlan.jsx
+++ b/packages/webapp/src/components/Crop/CompleteManamgenentPlan/PureCompleteManagementPlan.jsx
@@ -48,7 +48,7 @@ export function PureCompleteManagementPlan({
   const DATE = isAbandonPage ? 'abandon_date' : 'complete_date';
 
   const RATING = 'rating';
-  const NOTES = 'complete_notes';
+  const NOTES = isAbandonPage ? 'abandon_notes' : 'complete_notes';
   const {
     register,
     handleSubmit,
@@ -180,7 +180,11 @@ export function PureCompleteManagementPlan({
       />
       <InputAutoSize
         style={{ marginBottom: '40px' }}
-        label={t('MANAGEMENT_PLAN.COMPLETION_NOTES')}
+        label={
+          isAbandonPage
+            ? t('MANAGEMENT_PLAN.COMPLETE_PLAN.ABANDON_NOTES')
+            : t('MANAGEMENT_PLAN.COMPLETION_NOTES')
+        }
         hookFormRegister={register(NOTES, {
           maxLength: { value: 10000, message: t('MANAGEMENT_PLAN.COMPLETE_PLAN.NOTES_CHAR_LIMIT') },
         })}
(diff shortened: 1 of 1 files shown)


kathyavini:
I noticed that when testing this branch locally, the abandonment notes are not being saved. I think it would be a great learning opportunity for you to figure out why this is happening.

Here are a few questions and hints to help you investigate:

- What is the purpose of the `NOTES` constant?
- Are the form values being correctly submitted? (e.g. what does the network request look like when you save the data)
- This data is saved to the `management_plan` table so I would recommend taking a look at that as well

Keep up the good work and happy debugging!

Screenshot 2025-02-25 at 3 44 53 PM

kathyavini (approved):
Great job figuring out the issue!

Duncan-Brain:
So I can maybe approve... but in my opinion we have to look one step further and ask others opinion of the following architecture:

Until now the `completion_notes` column has only (intentionally) been used for completion notes (thus the name). If we change what the column is used for ... that can be dangerous if someone is using the column for research data -- especially without renaming the column.

Looking at `task` table there is BOTH `completion_notes` and `abandonment_notes`.

There is also `abandonment_reason` which one might argue currently also holds some "abandonment notes".

I might argue one of these two things is true:
- we should only use `abandonment_reason` for both task and management_plan
- we should add the column `abandonment_reason` to the management plan table

@kathyavini  any opinions?

den4ik1203:
@Duncan-Brain @kathyavini Slack is working again

Duncan-Brain (approved):
Thanks for your input! I mis-typed the second point which is fixed.

What I am saying is I believe the bug was that the completion notes box was ever added to the abandonment flow -(I agree it has been consistently buggy).

@kathyavini what do you think about a low low priority tech debt bug to align this behaviour. For data integrity and I think you can migrate it to a new architecture by saying `if(abandon_date) -> move completion_notes to abandon_notes` .

kathyavini:
> what do you think about a low low priority tech debt bug to align this behaviour

@Duncan-Brain sounds reasonable to me! Maybe the `other_abandonment_reason` behaviour could be aligned at the same time too

Duncan-Brain:
@den4ik1203 would you be willing to make the Jira ticket for this? And do you have any opinion before I merge your PR ;)

den4ik1203:
> @den4ik1203 would you be willing to make the Jira ticket for this? And do you have any opinion before I merge your PR ;)

In my personal understanding this situation is not quite correct. I can create this ticket.
Just want to confirm, the purpose of the ticket is to create the new column in the table - "Abandon_Notes". Or should I add something else?

Duncan-Brain:
@den4ik1203  So the purpose of the ticket it to "have a consistent architecture in `task`, `management_plan` (other tables?) with regards to abandon reason and abandon notes" (whoever completes the ticket should decide or discuss a new or best solution).

You can also add some identified solutions:
- Making a new column and migrating data that is currently stored in another column
- Using reason instead of notes
- other unidentified solutions

PR #3698, opened 2025-02-25
LF-4698 Create endpoints to get sensor array readings from Ensemble

kathyavini:
**Description**

I really procrastinated on wrapping up this endpoint because I thought writing the API test would be a pain -- and it _was_, but only until I dropped trying to mock `ensembleAPICall` and just mocked axios. I did not write tests to cover all the other sensor endpoints -- just the one added in this PR -- but the setup should be extensible in that direction if that feels like a priority.

This PR includes:
- the route configured as described [in Jira](https://lite-farm.atlassian.net/browse/LF-4698)
- functions to call Ensemble for the sensor data and format it as described in Jira
- basic validation for the query parameters on the route
- basic test case to cover the formatting functionality. I did not elaborate on e.g. mocking the different query params because it did not seem valuable. Please let me know if you think a certain kind of test case is missing though 🙏 
- types for RTK Query (all taken from the tables Morten shared with us in Slack)
- an RTK Query endpoint for frontend (to please be modified as needed)

All the new files are in TS 🎉  That was pretty fun! Jest can run ts files right out of the box by the way so that was a pleasant surprise -- no config adjusts or compiling necessary there. Probably Nav set it up like that for us ☺️ 

Use this endpoint in a component like this:

```ts
const { data: readingsData } = useGetSensorReadingsQuery(
  {
    esids: 'LSZDWX,WV2JHV',
    startTime: '2025-02-20T19:19:31-08:00', // ISO 8601 in UTC or local (both work)
    endTime: '2025-02-22T19:19:31-08:00',
    truncPeriod: 'day', // enforced 'day' | 'minute' | 'hour' (by types on frontend, validation on back)
  },
  {
    refetchOnMountOrArgChange: true, // refetch when time query params change
  },
);
 ```
 
 or in Postman:
 
 ```http
GET /sensor/readings?esids={{esids}}&startTime={{start_time_ISO}}&truncPeriod={{trunc_period}}
```

Jira link: https://lite-farm.atlassian.net/browse/LF-4698

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

Tested with code above.

**Checklist:**


packages/api/tests/mock.factories.js
```diff
@@ -604,15 +604,18 @@ async function crop_management_planFactory(
   ]);
   const [{ management_plan_id }] = managementPlan;
   const { needs_transplant } = cropManagementPlan;
-  needs_transplant &&
-    (await insertPlantingMethod({
+
```
kathyavini, line 607: eslint on the backend has been brutal since the version bump 😓 Now precommit doesn't accept this `&&` syntax that had previously been fine, and I think the `const`/`let` below was automatically corrected.

Duncan-Brain: Yah it is pretty bad haha I think I fixed this same thing lol.

You could turn off eslint for the whole file if you find it disruptive.

kathyavini: I _think_ that should be it for this file now, but I will keep `/* eslint-disable */` in my pocked 😂 

Probably we will have some cleaning up to do whenever we touch an old file on the backend. Maybe it's for the best.

SayakaOno (commented):
Comments and JSDocs were extremely useful, and thank you for the frontend setup as always and for figuring out how to work with TS and mocked axios! I'll refer to your code next time I work on the backend 😁 
The API has been working great! Thank you for incorporating my requests ❤️

packages/api/src/util/ensemble.js
```diff
@@ -246,6 +247,109 @@ const enrichWithMockData = (
...
+ * @async
+ */
+const getEnsembleSensorReadings = async ({ farm_id, esids, startTime, endTime, truncPeriod }) => {
+  const { id: EnsemblePartnerId } = await AddonPartnerModel.getPartnerId(ENSEMBLE_BRAND);
+
+  const farmEnsembleAddon = await FarmAddonModel.getOrganisationIds(farm_id, EnsemblePartnerId);
+
+  if (!farmEnsembleAddon) {
+    return [];
+  }
```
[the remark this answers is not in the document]
kathyavini: Hmmm, I was thinking in the context of the overall sensor controller, we would eventually end up going through all the partners and concatenate the data, in which case Ensemble data not being present wouldn't be a reason to error (we'd want to just continue to the next partner).

But it's a bit nonsensical because the whole thing is Ensemble dependent (we even require esids 😂) so for the time being there definitely should be an error. But maybe not in that function but in the validation middleware? Like how in `checkFarmAddon()` there is this:

```js
if (addon_partner_id !== EnsemblePartnerId) {
  return res.status(400).send('Only Ensemble Scientific is supported');
}
```

Could similarly do: 

```ts
res.status(404).send('Farm does not have active Ensemble Scientific addon');
```

but remove it / alter it alongside the other param checks if we ever actually expand to other partners? Would that make sense?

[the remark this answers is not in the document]
SayakaOno: Ah! So you were considering the case where a sensor array includes Ensemble and other sensors, which I totally forgot about 🙄
For now, I think going with the temporary solution and modifying it later makes sense, but returning `[]` would also work. The difference would be whether the user sees an error snackbar or not. (I think this mainly matters when trying to trigger the concurrency issue, so it wouldn't make much difference in most cases?) Should we wait for input from others?
Handling uncertainty has been quite a pain...😂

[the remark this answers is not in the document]
kathyavini: I would like the snackbar!! Even if it's just for concurrency 😂 

So I added the check to middleware (because all of that validation would need to be re-written together anyway if we move off of only Ensemble); it's a bit more code and queries that way though so I'm happy to be overruled if you or others prefer it in the Ensemble file. Anything is good by me but I do really the option of showing a snackbar 🙏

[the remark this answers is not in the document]
SayakaOno: Okay 😂  The addition to the middleware looks good to me, thank you!!
Would it be safe to remove the if block `if (!farmEnsembleAddon)` here?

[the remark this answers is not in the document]
kathyavini: Well the reason I used the middleware was I was hoping this file could work as-is even when the other partners are added, so only validation middleware had to be updated (so removing it now is fine but then it would have to go back in later after the other partners are added).

But now the whole idea of not touching this file feels completely fussy to me now anyway so I'm just moving the 404 here! It's a a lot less code change 🤣😜

SayakaOno (approved):

packages/api/src/controllers/sensorController.js
```diff
@@ -126,6 +127,29 @@ const sensorController = {
...
+        esids,
+        startTime,
+        endTime,
+        truncPeriod,
+      });
+
+      return res.status(200).send(data);
+    } catch (error) {
+      console.error(error);
+      const status = error.status || error.code || 400;
```
antsgar, line 146: Should the default status code be 500 instead of 400?

packages/api/src/util/ensemble.js
```diff
@@ -45,8 +47,8 @@ if (process.env.NODE_ENV === 'integration') {
   baseUrl = 'http://localhost:' + process.env.PORT;
 }
 
-//Known aliases for units from ensemble mapping to convert-units representation
-const ENSEMBLE_UNITS_MAPPING = {
+// Known aliases for units from ensemble mapping to convert-units representation as needed by the old sensor readings controller for incoming webhook data
+const ENSEMBLE_UNITS_MAPPING_WEBHOOK = {
```
antsgar, line 51: This is only used in the "old" code, right? Perhaps we could make that a bit more explicit in the naming by prefacing with "deprecated", "legacy", "v1" or something of the sort

antsgar (approved):
Lovely! ❤️

packages/api/src/controllers/sensorController.js
kathyavini: Yeah I think I might accidentally be the last one propagating the this 400 pattern which I picked up from our codebase even though I didn't agree with it 😂  I'll adjust here and in FarmAddon which has the same code bit.

kathyavini:
@antsgar sorry I read your comments but then they slipped my mind! I agree with both and I'll make those adjusts at the same time I do the cleanup / adjust of the mock to prep the sensors release. I'll merge this branch as is, to clean up Sayaka's diff instead of causing a conflict on it! 🙏

PR #3700, opened 2025-02-27
Farm Addon & Partner visual adjustments

kathyavini:
**Description**

This is the PR to address visual issues described in #3696 that weren't strictly related to the delete functionality. Addressed:

1. The focus > hover button layout shift at narrow desktop widths [shown here](https://github.com/LiteFarmOrg/LiteFarm/pull/3696#pullrequestreview-2645957264) (thank you @SayakaOno for spotting this)
2. This flash of content / blip on `<Partners />` which I'll demostrate here with 4x cpu throttling for clarity -- both the text input and the form buttons are showing momentarily if the connection is active and the RTK query cache is missing (e.g. on first login).  Within the PR code, `useEffect()` address the buttons while the conditional rendering is for the text input flash. Both make use of `isLoading` which it was my mistake to not include originally!

https://github.com/user-attachments/assets/cbbeef51-981e-4252-a21a-ebe5eedb3997

The solution for the flash of buttons also fixes the console warning: `Cannot update a component ('AddSensor') while rendering a different component ('Partners').`:

Screenshot 2025-02-27 at 10 34 39 AM

Thank you @Duncan-Brain for suggesting a `useEffect()` yesterday although maybe today you weren't into it 😂 (I think it's a good suggestion though!)

Jira link: none

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/Form/Button/button.module.scss
+++ b/packages/webapp/src/components/Form/Button/button.module.scss
@@ -59,6 +59,7 @@
   --active-background: var(--Colors-Secondary-Secondary-green-100);
   --active-border: 1px solid var(--Colors-Primary-Primary-teal-600);
   --active-color: var(--Colors-Primary-Primary-teal-600);
+  --focus-border: 1px solid transparent;
 }
 
 .btn.secondary-cta {
--- a/packages/webapp/src/containers/AddSensors/Partners.tsx
+++ b/packages/webapp/src/containers/AddSensors/Partners.tsx
@@ -12,6 +12,7 @@
  *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
  *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
  */
+import { useEffect } from 'react';
 import PurePartners from '../../components/AddSensors/Partners';
 import { useGetFarmAddonQuery } from '../../store/api/apiSlice';
 import { PARTNERS } from './constants';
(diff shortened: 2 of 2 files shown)


SayakaOno (approved):
Nice fixes, thank you! ❤️

Duncan-Brain (approved):
I think your fixes work well! And useState is good!... I just was trying to figure if using setState in the top level function body was "wrong" but I couldn't find anything. 

Also I am maybe confused about how the button styling works. Is it supposed to be active after first click moving forward? Or should the active state clear? The same thing happens with the form buttons in addAnimals.

https://github.com/user-attachments/assets/7430a3f6-552a-4a86-b6a9-0e5aad3e5183

kathyavini:
> Also I am maybe confused about how the button styling works. Is it supposed to be active after first click moving forward? Or should the active state clear? The same thing happens with the form buttons in addAnimals.

@Duncan-Brain it's not an active state, it's a focus state. And yup it's correct for that to stay applied until you tab or click elsewhere -- you should see the same behaviour if you tab to or long-click _any_ button in app.

https://github.com/user-attachments/assets/9f8d0580-ca08-4e9a-90f1-23ed369e42f8

We aren't the ones applying those `:focus` states (that's the browser) so I can only assume it's done correctly, although maybe it's the way it persists through the modal opening and then closing (even if you click on modal buttons) that feels un-intuitive? And most of the time (form buttons as you mentioned an exception) you're triggering a page change away from the button on click so you don't see this.

Duncan-Brain:
@kathyavini Yeah its the modal and form navigation transitions where I sort of expected the focus state to reset/unfocus/blur whatever haha. Not familiar with native browser behaviour but wondering if we missed something.

https://github.com/user-attachments/assets/64982ecd-25e2-490f-bed9-d3618ffae7ba

kathyavini:
@Duncan-Brain for that particular transition in your video maybe it's the fact that the inner text of the button changes but in fact it's the same button (hasn't remounted) so it keeps focus state which feels particularly wrong? Although yeah I guess even if the text didn't change it should feel like a new "next" button 🤔 Let's give it a quick view in tech daily with Anto next week, but for a three step form like that it's a one-line fix with a remounting key on that right-hand button : `key={isFinalStep ? 'save-button' : 'next-button'}`.

PR #3705, opened 2025-03-04
Lf 4747 create sensor readings kpi component

Duncan-Brain:
**Description**

~~Waiting on final approval from Loic for multiple parameters styling see link in Jira comments~~

- Tiles: This creates [KPI tiles](http://localhost:6006/?path=/docs/components-tile-sensorkpitiles--docs) for use with sensors arrays and individual sensor metrics. 

- Icon: Adds the ruler Icon.

- Layout: It also creates[ a reusable layout](http://localhost:6006/?path=/docs/components-layout-bentolayout--docs) for composition of tiles. 
  - It copies the currently named "bento" style of sensor list overview stats -- let me know if there is a different preferred name. 
  - It is technically a drop in replacement for the top level div of overview stats

Notes
- Storybook: I had some issues with storybook loading css correctly -- if you have issues a refresh or restart should do the trick.

Jira link: [LF-4747](https://lite-farm.atlassian.net/browse/LF-4747)

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

[LF-4747]: https://lite-farm.atlassian.net/browse/LF-4747?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/webapp/src/components/Tile/sensorKPI.module.scss
```diff
@@ -0,0 +1,106 @@
...
+}
+
+.measurement {
+  display: block;
+}
+
+.topDetails {
+  display: flex;
+  gap: 8px;
+  display: flex;
```
SayakaOno, line 72: This is redundant ;)

SayakaOno (commented):
I left a few comments, but it looks great! Thank you!

packages/webapp/src/components/Tile/SensorKPI.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+export interface MeasurementProps extends TMeasurement {
+  rest?: React.HTMLAttributes<HTMLDivElement>;
+}
+
+const discriminatorIconName: { [key: string]: IconName } = {
+  depth_elevation: 'RULER',
+};
+
+const Measurement = ({ measurement, value, unit, ...rest }: TMeasurement) => (
```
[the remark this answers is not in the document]
Duncan-Brain: Yeah I think so, that is weird haha

packages/webapp/src/components/Tile/SensorKPI.tsx
```diff
@@ -0,0 +1,92 @@
...
+  sensor: {
+    id: string | number;
+    status: StatusIndicatorPillProps;
+  };
+  discriminator: TMeasurement;
+  measurements: TMeasurement[];
+  colorHex?: string;
+}
+
+export interface MeasurementProps extends TMeasurement {
```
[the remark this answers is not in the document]
Duncan-Brain: 🙏

packages/webapp/src/components/Tile/SensorReadingKPI.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+import styles from './sensorKPI.module.scss';
+
+export type TMeasurement = {
+  measurement: string;
+  value: string | number;
+  unit: string;
+};
+
+export interface SensorReadingKPIprops extends TMeasurement, React.HTMLAttributes<HTMLDivElement> {
```
SayakaOno, line 24: I forgot to ask, how do you decide whether to use an interface or a type? I thought you preferred interfaces, but you used a type here initially, so I'm curious!

SayakaOno (dismissed):

packages/webapp/src/components/Tile/sensorKPI.module.scss
SayakaOno, line 1: I just remembered I had one more question! Did you copy the layout scss file to create this one? I believe the layout scss is an exception, so we should follow the convention and name this styles.module.scss.

packages/webapp/src/components/Tile/SensorReadingKPI.tsx
Duncan-Brain: I think I read this a while ago.. start with type always unless interface is useful.

https://www.totaltypescript.com/type-vs-interface-which-should-you-use

packages/webapp/src/components/Tile/sensorKPI.module.scss
Duncan-Brain: My understanding is that the convention is to group css where overlap/reuse might occur and the naming does not really matter.

I think rather than merge this css file into into the other I would then be inclined to create a `/SensorTiles` folder. I could do that if you think that makes more sense.... But I might regret not combining the two tiles into one `index.jsx` at that point.

What do you think?

SayakaOno: Ah, sorry 🙏, let me clarify it. I was only suggesting renaming this file to `styles.module.scss`. I was assumed it was named `sensorKPI.module.scss` because you duplicated the layout scss file and had to rename it. I don't have any concerns about the file structure or anything else!

packages/webapp/src/components/Tile/SensorReadingKPI.tsx
SayakaOno: Interesting! Thank you for the link :)

packages/webapp/src/components/Tile/sensorKPI.module.scss
Duncan-Brain: So there is already a `styles.module.scss` in this folder. I thought it would be better to keep it separate from the other file that is why I have a different name, I did not copy layout scss.

SayakaOno: Ah, okay! Sorry, I thought you had created a new folder. I should have given a more specific name than `Tile`, but I didn't expect we'd add another component there.

Would it make sense to do one of the following?

- Create `/SensorTiles` and place it directly in /components.
- Create `/SensorTiles` along with another folder to hold the existing files.

Duncan-Brain: I am not sure... we are kind of getting into questions about component library organization from @antsgar 's RFC, so I would be interested in her opinion.

To me.. I find our components completely impossible to find anything lol. I personally see value in storing similar components all in one folder:
- `/Tile/DataTile.tsx` .. with the value being:
  - if you need to edit or refactor a "Tile" of some kind they are all easy to find. 
  - if you think you can abstract shared behaviour to maintain less code, it is all in the same area to easily compare.
  
But I also see that these can be pretty module specific (displaying generic data:`SensorReadingKPI.tsx`, and displaying sensor data `SensorKPI.tsx`) so maybe it makes sense to start searching for components from the point of view of the module:
- `/Map/Sensor/Tile/SensorKPITile.tsx`with the value being
  - it reflects the part of the app you are in
  
 So I am not sure about the best way for our case but we definitely have a big ole mix of things right now..haha

kathyavini (commented):
Very interesting PR -- a ton of creative CSS solutions here! `<BentoLayout>` as a wrapping component reminds me of a proper component library, very neat ☺️ 

@SayakaOno since you've already approved and are the one using this component in your views, please go head and merge 🙏 

The only code that really needs to be fixed I think is the RegEx since it doesn't work (if that 3 to 6 digit hex functionality is something the charts need) but even that I don't think is necessarily blocking.

packages/webapp/src/components/Tile/sensorKPI.module.scss
```diff
@@ -0,0 +1,105 @@
...
+  align-items: center;
+  @include svgColorFill(var(--color));
+  font-size: 14px;
+  font-weight: 400;
+  line-height: 30px; /* 214.286% */
+}
+
+.icon {
+  padding: 0px;
+  background-color: transparent;
```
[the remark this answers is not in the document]
Duncan-Brain: @kathyavini Did you try restarting Storybook? I had the same issue -- and I have it all the time where classes where I think I should override look funny sometimes but not others it is so flakey!  Even for the classes I did over-specify in this PR I could not for the life of me figure out why it was necessary, I literally have never had to over specify before.

I think there is an issue with vite not being able to compile properly on hot reloads. So by extension storybook-vite I think has the same issue.

packages/webapp/src/components/Tile/SensorKPI.tsx
```diff
@@ -0,0 +1,90 @@
...
+  colorHex = '#000000',
+  ...rest
+}: SensorKPIprops) {
+  const { measurement, value, unit } = discriminator;
+  const { status, id } = sensor;
+  // Ensure a 6 digit hex plus hash, if less than 7 digits force it to a 3 digit color code and turn that into 6 digit.
+  const color =
+    colorHex.length < 7
+      ? `${colorHex.slice(0, 4).replaceAll('#([0-9a-fA-F])([0-9a-fA-F])([0-9a-fA-F])', '#$1$1$2$2$3$3')}`
+      : `${colorHex.slice(0, 7)}`;
```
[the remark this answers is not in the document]
Duncan-Brain: Its a good point, I probably did not need to add this... and I totally forgot to check it 😅 ! Will totally update with your fix and suggestions ...or delete if that is preferred.

I added it because I was thinking about the graph line colors (and matching tile) and how we only had 4 colors (6 colors now)! That is too finite an amount of colors for the possible graphs I think that could happen ([see metos graphs](https://metos.ca/software-tools/actionable-solutions/water-management/))....  From that followed that to my knowledge we don't have a "random color generator" --- and so to be flexible in this component for whatever is come up with I thought it would be helpful to have options for the hex color code options thinking that rgb to hex is easy so dealing with one type of color format seemed to make sense. 

I think I thought the 3-character hex might have an easy "pattern" for nice or differentiated colors so maybe it would be used instead of the 6-char I am used to.

I could delete it and it would probably be fine. It might even better belong with the random color generator code if that ever exists. I didn't think it through.

packages/webapp/src/components/Layout/layout.module.scss
```diff
@@ -53,4 +54,24 @@
...
+
+.bento {
+  display: flex;
+  gap: var(--bentoGap);
+  flex-wrap: wrap;
+  
+  & > div {
+    flex: 1 1 calc((100% - var(--bentoGap) * (var(--bentoColumns) - 1)) / var(--bentoColumns));
+  }
+  // This is the default behaviour in LiteFarm - should that change pass as classname
```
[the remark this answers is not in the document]
Duncan-Brain: I think it is the same yah.. "pass as classname prop"?

I think what I am trying to say here is that flex boxes like this are responsive by nature. In `<OverviewStats />` we have chosen to ignore the inherent responsiveness dictated by the content overflow to follow design. Since it is odd to bypass that behaviour (since it will lead to the same result as this css once the content overflows) I made a comment noting that and a reminder of what to do if we change our default behaviour stance. 

The reminder is probably not needed, but I think the note on overriding responsive behaviour is valuable but could be maybe improved.

packages/webapp/src/components/Tile/sensorKPI.module.scss
```diff
@@ -0,0 +1,105 @@
...
+  padding: 8px 12px;
+  flex-direction: column;
+  flex: 1 0 0;
+  @include fontFamily();
+  font-style: normal;
+}
+
+.gap8px {
+  gap: 8px;
+}
```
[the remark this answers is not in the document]
Duncan-Brain: Fixed 🙏

packages/webapp/src/components/Tile/sensorKPI.module.scss
```diff
@@ -0,0 +1,105 @@
...
+ *
+ *  LiteFarm is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ */
+
+@import '@assets/mixin.scss';
+
+.sensorReadingKpi {
```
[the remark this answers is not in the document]
Duncan-Brain: Thanks 🙏

packages/webapp/src/components/Tile/SensorReadingKPI.tsx
```diff
@@ -0,0 +1,56 @@
...
+  measurement,
+  value,
+  unit,
+  colorHex = '#000000',
+  ...rest
+}: SensorReadingKPIprops) {
+  // Ensure a 6 digit hex plus hash, if less than 7 digits force it to a 3 digit color code and turn that into 6 digit.
+  const color =
+    colorHex.length < 7
+      ? `${colorHex.slice(0, 4).replaceAll('#([0-9a-fA-F])([0-9a-fA-F])([0-9a-fA-F])', '#$1$1$2$2$3$3')}`
```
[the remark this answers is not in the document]
Duncan-Brain: see above

packages/webapp/src/components/Tile/SensorKPI.tsx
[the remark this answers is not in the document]
Duncan-Brain: I decided to delete the validation from this component -- I think I should trust our own components and let the parent format. I have replaced it with JsDoc to help the developer!

packages/webapp/src/components/Layout/layout.module.scss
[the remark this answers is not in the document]
Duncan-Brain: Updated the text

Duncan-Brain:
Thanks for reviewing! I did try to make it like MUI -- I am now trying to understand when best to use "composition" (children prop) and make better components.

I updated the component to take validation of color out of it -- that can go wherever color is decided.

I updated stories since I did not realize the ability to edit parameters was negatively affected in the previous story and not possible. This is a better story but my typescript is a bit off. Maybe some future me can fix it but it is better to have a useful story than perfect types.

packages/webapp/src/components/Tile/sensorKPI.module.scss
Duncan-Brain: Currently no, but there could be shared css for a person looking to do that. And the "Tiles" parent could potentially use these tiles too in the future -- it just is not currently in use.

I personally like the idea of a subfolder just to keep all the similar abstractable code in one place. But if you think it should be moved up I can do that!

packages/webapp/src/components/Layout/layout.module.scss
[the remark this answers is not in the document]
kathyavini: It's clearer with the new text thank you. It's because we have the pattern `clsx(className, styles.bento)` when it comes to styles that I was confused whether you wanted that and had forgotten to set up an actual prop called `className`.

packages/webapp/src/stories/Tile/SensorKPITiles.stories.tsx
```diff
@@ -15,20 +15,43 @@
...
 import { componentDecorators } from '../Pages/config/Decorators';
 import { Status } from '../../components/StatusIndicatorPill';
-import { ReactNode } from 'react';
 
 const colors = ['#0669E1', '#E8A700', '#266F68', '#D02620', '#8F26F0', '#AA5F04'];
 
-type TemplateProps = {
-  children: ReactNode;
-};
+interface TemplateProps {
```
kathyavini, line 27: If you wanted to clean it up it's just the conflation of the two kinds of props that is causing most of your TS errors. You can separate them out and TS will be happy:

```ts
type SensorReadingTemplate = {
  tileType: 'sensorReadingKpi';
  kpiProps: SensorReadingKPIprops[];
};

type SensorKpiTemplate = {
  tileType: 'sensorKpi';
  kpiProps: SensorKPIprops[];
};

type TemplateProps = SensorReadingTemplate | SensorKpiTemplate;
```

packages/webapp/src/stories/Tile/SensorKPITiles.stories.tsx
```diff
@@ -15,20 +15,43 @@
...
+}
 
-const Template = ({ children }: TemplateProps) => {
-  return <BentoLayout>{children}</BentoLayout>;
+const Template = ({ tileType, kpiProps }: TemplateProps) => {
+  if (tileType == 'sensorKpi') {
+    return (
+      <BentoLayout>
+        {/* @ts-expect-error */}
+        {kpiProps.map((props: SensorReadingKPIprops) => {
```
kathyavini, line 37: This one should be SensorKPI props

kathyavini (approved):
Looks good!! 😍 

> but it is better to have a useful story than perfect types.

💯 for sure! Your typescript is only a smidge off anyway. Left a few optional comments for now or later if you do want to remove the ts-expect-errors 🙏

packages/webapp/src/stories/Tile/SensorKPITiles.stories.tsx
Duncan-Brain: Niiice that is 100 percent what I was looking for 🙏  thanks so much ! One of these days TS will click haha

packages/webapp/src/stories/Tile/SensorKPITiles.stories.tsx
Duncan-Brain: 🙏

kathyavini (approved):

SayakaOno (approved):

packages/webapp/src/components/Tile/sensorKPI.module.scss
antsgar: I'm good with that!

antsgar (approved):

PR #2963, opened 2023-11-09
LF-3796, LF-3765 fix finances crashes

antsgar:
**Description**

- Fixed crash in Finances dashboard that happened when switching farms. This was due to the sales and expenses for the new farm being fetched, but not the related entities such as revenue types, expense types and crop varieties.
- Since crop varieties are now used in the dashboard to display revenue transactions, added fetching them to the `useEffect` in the finances screen as well.
- Added several undefined checks for revenue and expense types and crop varieties to prevent crashing if above solutions fail, as well as default empty arrays for entities in the useTransactions hook.
- Added resetting date and types filters when loading finances dashboard. This will reset the filtering when switching farms, but will also add intended behavior of resetting when navigating away from the section on the same farm.

Jira link:

**Type of change**

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

**How Has This Been Tested?**

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [ ] Passes test case
- [ ] UI components visually reviewed on desktop view
- [ ] UI components visually reviewed on mobile view
- [] Other (please explain)

**Checklist:**

- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [x] The precommit and linting ran successfully
- [ ] I have added or updated language tags for text that's part of the UI
- [ ] I have added "MISSING" for all new language tags to languages I don't speak
- [ ] I have added [the GNU General Public License](https://lite-farm.atlassian.net/l/cp/BT0Dd7WW) to all new files

PR #3708, opened 2025-03-06
LF-4746(1) FinancesDateRangeSelector minor refactors and bugfixes

kathyavini:
**Description**

This PR is a preparation PR for the creation of the `<SensorsDateRangeSelector />` filter + hook.

It prepares the existing `Finances/DateRangeSelector` implementation for re-use in a general way by some renaming and conversion to TypeScript. In looking at this, I found a few bugs in Finances and fixed them. _Most of the code changes are just renames! That is the only reason so many files are touched._

Changes in this PR:
- I have reanamed the import of the `Finances/DateRangeSelector` component to `FinancesDateRangeSelector` throughout the app. It was sometimes imported as such and sometimes as `<DateRangeSelector />` which made code search more difficult. We also have a different component `components/DateRangeSelector` so I wanted to avoid the naming conflict.
- the hook `useDateRangeSelector.js` imports and uses the action and selector for the Finances Redux store. To make this dependency clear I have renamed it to `useFinancesDateRange.ts` and moved it into `Finances/`. It was previously co-located with `components/DateRangeSelector` but that component is redux-store agnostic
- both of these files have been converted to TS (hopefully with zero modification of their logic)
- The buggy behaviour of custom date range (how it resets and the date picker loads as open each time you navigate to a different page in finances) was due to the `.onChange()` handler of the base `<DateRangeSelector />` being called upon first load by a call to `setSelectedDateRangeOption`.
  - I'll leave a note in the code in case I did not understand the use of this correctly, but I _believe_ the functionality will be correct if setSelectedDateRangeOption is just not called on the first render.
  - There are a few Jira tickets that refer to this issue and I have linked two below
- Loading the transaction list used to include a reset of the date range back to "Year as Date". I don't know if we have strictly ever classified this as a bug, so I started [a thread in `#ux_ui`](https://litefarm.slack.com/archives/CAH7PMYE4/p1741297440094579) about it. To me it felt quite buggy and I have removed it. At some point a proper implementation of resetting when leaving finances can be added.
- one more bug fix related to the Redux filter data (see Jira links and conversation below)

Jira links:

Bug tickets related to the custom range behaviour (the reset + open picker)
- https://lite-farm.atlassian.net/browse/LF-4515
- https://lite-farm.atlassian.net/browse/LF-4200

Bug tickets related to incomplete reset of Redux state as spotted by @SayakaOno 🙏 
- https://lite-farm.atlassian.net/browse/LF-3958

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/webapp/src/components/DateRangeSelector/index.jsx
```diff
@@ -67,11 +67,18 @@ export default function DateRangeSelector({
     }
   }, [isValid, isCustomDatePickerOpen]);
 
+  const isInitialRender = useRef(true);
```
kathyavini, line 70: This hook was added in this PR:
- https://github.com/LiteFarmOrg/LiteFarm/pull/2963

to solve a crash we used to see in finances, but I think we all (mostly myself as reviewer on that PR 😅) missed that it also triggered buggy behaviour on custom date range (clearing and opening the picker on every navigation within finances).

The bug happens because the ReactSelect's `.onChange()` is called after this useEffect runs on the initial render, and that opens the date picker and clears the value. I believe that the initial render should be safely covered by the `defaultValue` prop on the ReactSelect, which is why I think it's okay to skip it, but @antsgar please let me know if you remember more context about how this contributed to the crash and if I just reverted something important! 🙏 🙇

packages/webapp/src/containers/Finances/saga.js
```diff
@@ -516,7 +514,6 @@ export function* fetchAllDataSaga() {
   ]);
   yield put(setSelectedExpenseTypes([]));
   yield put(resetTransactionsFilter());
-  yield put(setDateRange({ option: dateRangeOptions.YEAR_TO_DATE }));
```
kathyavini, line 519: This is the cause of the Finances date range resetting to Year to Date each time transactions list is loaded. I have moved the same call to `clearOldFarmStateSaga()` below since it was the between farm behaviour I believe we were finding problematic.

SayakaOno (commented):
This feels amazing! I was expecting it would need more changes to fix the behaviour, but the fixes seem effective and powerful! ❤️ 

The one thing I couldn't spot just from reviewing the code is that the custom range persists when switching farms. (set a custom range -> switch farms (no need to go to finances, though I did in the recording)-> switch back to the original farm and go to finances) Are you able to reproduce it?

https://github.com/user-attachments/assets/e76a09b4-f26c-494b-ab74-2455c81cd392

kathyavini:
@SayakaOno thank you so much for spotting that!! ❤️ 🙏 It wasn't even just the custom dates, but everything other than the `option` was maintained (i.e. the string and computed dates were out of alignment too 😬)... I was foiled by old-style Redux and didn't realize the objects were being merged instead of replaced by that reducer.

packages/webapp/src/components/DateRangeSelector/types.ts
```diff
@@ -0,0 +1,40 @@
...
+  LAST_MONTH = 'last_month',
+  CUSTOM = 'custom_range',
+}
+
+interface CustomRange {
+  startDate?: string | Moment;
+  endDate?: string | Moment;
+}
+
+export interface DateRangeSelection {
```
kathyavini, line 35: Unrelated to anything else in the PR, do you have any suggestions for this type? 😅 I'm not loving the name I picked here -- but is `DateRange` too unspecific? Is `DateRangeData` better? It's the shape of the date range data that the hooks and components act on.

kathyavini:
With apologies for the scope creep, I found another bug in the finance date range filters that seemed an easy fix so I have added it as well:

[Custom date range filter from finances dashboard isn't correctly carried into export date range filter](https://lite-farm.atlassian.net/browse/LF-4200)

Actually logically speaking to me the original `useEffect()` seemed sound so I'm going to spend a bit more time trying to figure out why the solution works and I'll update my comment here when I do 😆 I thought it was stale component state from the time the `<Report />` component first rendered (and thus the solution) but it actually seems to be a resetting to default value even after the useEffect has run. The solution still works but I would like to spend more time figuring out why.

packages/webapp/src/components/DateRangeSelector/types.ts
SayakaOno: I think I usually go with data or info, but both data and state sound good to me!

SayakaOno (dismissed):

packages/webapp/src/components/DateRangeSelector/types.ts
kathyavini: Thank you, I'll update all to `DateRangeData` in the sensors PR 🙏

packages/webapp/src/components/DateRangeSelector/index.jsx
kathyavini: Update Mar 12: in tech daily we decided to remove this `useEffect()` entirely.

kathyavini:
I'm embarrassed to admit how long it took me to debug why the dates were not setting properly within the `onClickAway()` handler 😅 It turns out two sequential calls to `changeDateMethod` don't work as excepted since the second call will use stale state for the parameter it's not concerned with. Therefore I have made the following changes:

- updated `changeDateMethod` to set the full range at once, which is appropriate to how it's now being triggered
- removed the clearing of the date range upon opening the picker
  - this is not essential for the timing fix, but it seems more in the spirit of the timing we are going for (where changes are only committed to the store once dates are selected and the picker closed)
  - in the context of charts it will mean that opening the custom picker will not trigger an API call
- removed the `useEffect()` as discussed today in tech daily

Thank you in advance for your re-reviews 🙏

SayakaOno (approved):
❤️

antsgar:
Looks good, thank you for all the improvements to the component! ❤️

antsgar (approved):

PR #3709, opened 2025-03-08
LF-4746 (2) Create date range filter for sensor readings

kathyavini:
Note that this is branched off of and diffed to LF-4746(1):
- #3708 

**Description**

### Rationale

The goals of this approach were as follows:
1. To maintain the functionality of the date range selector in Finances completely unchanged
2. To make it as easy as possible to re-use all this the date time logic (i.e. the calls to @SayakaOno's excellent DateRange utility class which handles turning semantic strings into dates) in other parts of the app
3. To maintain the calling component's pattern of hook + date range selector:

```ts
const { startDate, endDate } = useFinancesDateRange({ weekStartDate: SUNDAY }); 
// ^ use the hook's returned values in the component's computations
// ...  
<FinancesDateRangeSelector className={styles.dateRangeSelector} /> 
// ^ render the visual selector component in the same file
```
  
In regards to (3), I did maintain the pattern but it's worth noting that it's implemented slightly differently in Sensors:
- the hook ALSO initiates and holds the date range state component state (we don't need Redux here)
- since the hook has initialized the state, the component needs to receive it as props:

```tsx
const { startDate, endDate, dateRange, updateDateRange } = useSensorsDateRange({
// ^ as before startDate and endDate are the ones to use in the component
  weekStartDate: SUNDAY,
});

// startDate and endDate are returned as ISO strings so useable directly in a query:
const { data: readingsData } = useGetSensorReadingsQuery(
  {
    esids: 'LSZDWX,WV2JHV',
    startTime: startDate,
    endTime: endDate,
    truncPeriod: 'day',
  },
  {
    skip: !startDate || !endDate,
    refetchOnMountOrArgChange: true,
  },
);

// dateRange and updateDateRange need to be passed to the selector component
  <SensorsDateRangeSelector
    dateRange={dateRange}
    updateDateRange={updateDateRange}
  />
```

### Code changes
- Takes the `FinancesDateRangeSelector` and `useFinancesDateRange` (hook) and extracts two new components `SmartDateRangeSelector` and `useSmartDateRange` which have no reference to any particular store
  - naming suggestions very welcome on these
  - note that `FinancesDateRangeSelector`  accepted props for overriding Redux state use by passing a `value` and `onChange` (if they are supplied, the Redux state is untouched; we used this in the Finance Report modal)
    - to make this a little bit clearer, `SmartDateRangeSelector` only takes one source of state and one state setter, and the calling component should specify which source of state should be used
- The Finance component and hook are rewritten as lightweight wrappers around the above
- Finally, a sensor component and hook have also been written as lightweight wrappers around the above

The idea of this pattern is for as many `<Something>DateRangeSelector` and `use<Something>DateRange` hooks to be created as needed for future views in app, where each wrapper just defines and passes the correct state (Redux or local) to the underlying functions, and sets the right default date range for this part of app.

Very interested to get your thoughts on this structure though, because I was a bit hesitant to create yet another layer / wrapper between the component and `components/DateTimeSelector` as the file structure already felt a bit confusing to me when I was originally reading through this part of the code. Thank you in advance for your feedback 🙏 

Jira link: https://lite-farm.atlassian.net/browse/LF-4746

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

You can use the code above and the sensor readings API PR #3698 to test how the date ranges function.

**Checklist:**


SayakaOno (approved):
Looks great!
I think it would be clean enough even if we used `SmartDateRangeSelector` and `useSmartDateRange` directly in parent components!

kathyavini:
> I think it would be clean enough even if we used SmartDateRangeSelector and useSmartDateRange directly in parent components!

@SayakaOno yeah `SensorDateRangeSelector` in particular ended up SO thin I also wondered if it was pointless 😂 I kept it mainly because, as a pattern, "make new hook + selector file" seemed easiest to replicate going forward without having to think much about it in each use case (and assuming that at some point we would need Redux / persistence across multiple views again).

But if you feel like your chart component would work better just using the base components feel free to use them that way!

antsgar:
This is awesome and I really love the approach of extracting the code that was previously used for Finances and making it generic enough to be reused, as well as the idea of splitting the code into a selector component and a hook for each specific use case 🙌 it's all so clean!

The only thing that I'm wondering is, do you see a scenario where we would use the base component, that is, `DateRangeSelector`, directly instead of using the `SmartDateRangeSelector` layer? I'm wondering if we need that base layer or if the two components should actually be combined, and have the "smart" layer be the base layer that can then be reused on different views

kathyavini:
@antsgar I was wondering about that too initially. No, I can't really think of a situation where it would make sense to have the `<DateRangeSelector />` used in isolation, although I suppose you could potentially wrap it in a different kind of `<SmartDateRangeSelector />` if you really needed to use it with differently shaped Redux store data.

But at the same time I think the separation of concerns is correct, since the smart layer does all date computations and is tied to data store shape, while the base layer only handles the selection itself (i.e. it's a huge complicated "input"!). Without the separation I would worry about the component becoming too long to be readable 🤔 

Do you think there is a way to adjust the naming such that it's more clear what to use, but still maintaining the separation of concerns? I agree a component that can't be used in isolation is not the best default import from that folder!

antsgar:
@kathyavini that's totally reasonable! Maybe what is now the `SmartDateRangeSelector` file could be renamed to `index` and be the default import, and we could rename the current `DateRangeSelector` to something like `DateRangeInput`?

packages/webapp/src/stories/DateRangeInput/DateRangeInput.stories.jsx
```diff
@@ -89,7 +89,7 @@ export const WithDefaultOption = {
 
     const backButton = await canvas.findByText('back');
     expect(backButton).toBeInTheDocument();
-    expect(backButton).toHaveStyle('color: #d4dae3');
```
kathyavini, line 92: I don't think this has any relationship to the recent changes, but I noticed the interaction test was failing for me. The original color is `--grey400` but the passing color is `--grey500`. Both indicate disabled but maybe at some point the back button was not a TextButton and had a different disabled style?

(Just speculating, did not actually dig into any git history for this one since the fix was simple).

kathyavini:
Only change (other than a tiny Storybook adjust) is naming as follows:

- SmartDateRangeSelector > DateRangeSelector (main export from the folder)
- useSmartDateRange > useDateRange
- previous DateRangeSelector (the Redux-agnostic pure component wth Storybook) > now DateRangeInput

Thank you again for re-reviewing 🙏

SayakaOno (approved):

antsgar (approved):

PR #3714, opened 2025-03-14
LF-4699(1): Display sensor readings on the UI

SayakaOno:
**Description**

- Add `LineChart` component and [stories](http://localhost:6006/?path=/docs/components-chart-linechart--docs).
  - Supports both timely data and general data.
  - The default hover behaviour ([example](https://master--63da8268a0da9970db6992aa.chromatic.com/?path=/story/examples-tooltip--setting-tooltip-index)) of displaying tooltips and dots for all x-axis points is overridden using state.
  - Dots are displayed when the data length is 12 or fewer, to ensure tooltips are easily accessible, especially on mobile.
- Add `getTicks` function and tests (`pnpm test chartUtils`)
  - Returns monthly ticks if the range spans over two or more days at the start of a month; otherwise, returns daily ticks.
  - Limits the number of ticks to 7. 

Jira link: https://lite-farm.atlassian.net/browse/LF-4699

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/webapp/src/components/Charts/LineChart.tsx
```diff
@@ -0,0 +1,221 @@
...
+          <dl className={styles.tooltip} style={{ '--tooltipColor': color } as React.CSSProperties}>
+            <dt>{xAxisData}</dt>
+            <dd>{value}</dd>
+          </dl>
+        );
+      }
+    }
+
+    return null;
+  };
```
antsgar, line 125: Could we move this component outside of `LineChart` and pass any props needed from LineChart to it? It's best not to declare a component inside another one, because it will be re-created on every render. I also think it's harder to figure out which data the component needs to function when it's declared inside another one and uses some of its props/state

antsgar (commented):
The charts look amazing! :heart: I left a couple of small comments, but this looks great. I assume there'll be a follow up PR for showing these on the sensor detail view, correct?

packages/webapp/src/assets/theme/index.tsx
```diff
@@ -51,6 +51,12 @@ export const colors = {
   grey1: '#333333',
   brown700: '#AA5F04',
   brown900: '#7E4C0E',
+  '--Colors-Accent---singles-Blue-full': '#0669e1',
+  '--Colors-Accent---singles-Red-full': '#d02620',
+  '--Colors-Accent-Accent-yellow-600': '#e8a700',
+  '--Colors-Primary-Primary-teal-700': '#266f68',
+  '--Colors-Accent---singles-Purple-full': '#8f26f0',
+  '--Colors-Accent---singles-Brown-full': '#aa5f04',
```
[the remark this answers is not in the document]
SayakaOno: The existing colours were defined before Loïc joined, and the variable names match the CSS variables. Also, the chart colours are already used in different components.
That said, I don't think we'll be using the colors constant in the future... can we just go with `chartBlue`, `chartRed` etc.?

SayakaOno (commented):
Thank you Anto for reviewing!
Yes, my next PR will be for the sensor detail with the KPI, date range selector and charts!

antsgar:
@SayakaOno yeah I think that'd be okay, even if it's redundant with colors used elsewhere!

SayakaOno:
I've updated the keys, and they are much cleaner now 😄 
Thank you for re-reviewing 🙏

antsgar (approved):

packages/webapp/src/tests/chartUtils.test.js
```diff
@@ -0,0 +1,280 @@
...
+          expectedTicks: ['2025-06-30'],
+        },
+        {
+          input: ['2025-06-30', '2025-07-01'],
+          expectedTicks: ['2025-06-30', '2025-07-01'],
+        },
+      ];
+
+      testCases.forEach(({ input: [startDate, endDate], expectedTicks }) => {
+        const ticks = getTicks(startDate, endDate, 'day');
```
kathyavini, line 62: Is the `'day'` parameter from a previous version of the utility function? It looks like `getTicks()` takes an options object now as third parameter so I think the passed `'day'` is just ignored.

kathyavini (approved):
Fanstatic! As always I so love your organized tests and utils ❤️

This is a totally subjective thing, but do you like the tooltip animating in from the left side of the chart? Sometimes when hovering the lines on the right side, the travel across the chart seems a little extreme 😂  Does that animation seem to impact performance at all in the context of real data?

packages/webapp/src/tests/chartUtils.test.js
SayakaOno: You're right! I really appreciate your attention to detail 🙏❤️ I'll correct it in my next PR!

PR #3710, opened 2025-03-11
LF-4704 Update cancel flow modal

kathyavini:
**Description**

Small but sensors release-scoped ticket. This implements the changes to the cancel modal such that it no longer shows a title with an interpolated `{{flow}}` but rather has the same message in all cases.

In the PR:
- update of the modal as described in Jira ticket
- removal of the prop from the associated form components
- removal of the now unused `{{flow}}` strings from components + translations

Note: I tried to save files without formatting, but pre-commit ESLint did lint some of those touched files.

Jira link: https://lite-farm.atlassian.net/browse/LF-4704

(Also addresses: https://lite-farm.atlassian.net/browse/LF-4394)

**Type of change**

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] UI components visually reviewed on desktop view
- [x] UI components visually reviewed on mobile view
- [ ] Other (please explain)

**Checklist:**

- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [x] The precommit and linting ran successfully
- [ ] I have added or updated language tags for text that's part of the UI
- [ ] I have added "MISSING" for all new language tags to languages I don't speak
- [ ] I have added [the GNU General Public License](https://lite-farm.atlassian.net/l/cp/BT0Dd7WW) to all new files

PR #3715, opened 2025-03-14
LF-4627 [Nice to have] Implement Add to Map redesign

kathyavini:
**Description**

This PR updates the `<Drawer />` component to allow a side drawer opening from the left, and uses this variant in the "Add to Map" and "Filter Map" modals.

### Changes to `<Drawer />`
- Adds the property (`sideDrawerDirection`) and styles that control the left side drawer variant. [Storybook here](http://localhost:6006/?path=/story/components-drawer--left-side-drawer).
- Renders all off-screen drawers [inert](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inert). I was horrified to realize the other week when looking at our keyboard navigation that the feedback drawer components are gaining focus even when off screen! The same was true for the hamburger menu when un-opened at mobile widths.
  - TypeScript React required a type for this which I added (source documented in the file)
- slight reconfiguration of styles so that view-specific properties like `min-width` are passed by the component of that view and not in the base Drawer
  - this also required a new styling class since some styles (e.g. distance from the top of screen) should not be shared between bottom and side drawers

### Changes to `<MapDrawer />`
In addition to making it a left-opening side drawer with no backdrop:
- Updated the + icon to the new (+ Add) style
- Added a hover state
- Updated the section headers

### Changes not implemented (I think these warrant a different ticket)
- Updates to "Filter Map" not shared with "Add to Map", e.g. the icons -- I think this was designed in conjunction with the "add custom filters to map" `ilot` project
- Making map filters draggable/re-arrangeable. As above, and also a sizeable task in itself that deserves its own scope + ticket

Jira link: https://lite-farm.atlassian.net/browse/LF-4627

Side note: sorry for the lack of Jira numbers on the the commit messages. I committed for a whole day and a half (!) to integration before realizing my mistake, and then cherry-picked them over. This was not before I tried to `git rebase` them over though, and instead completely trashed my local integration (and never successfully moved the commits) 😝 I am _really_ not good with rebase!

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/webapp/src/components/PageTitle/CancelButton.jsx
```diff
@@ -27,7 +27,7 @@ export function CancelButton({ onCancel, showConfirmCancelModal, setShowConfirmC
   );
 }
 
-CancelButton.PropTypes = {
```
kathyavini, line 30: Completely unrelated but this was my typo from #3710 in two files that was causing a console warning:

Screenshot 2025-03-14 at 3 50 19 PM

packages/webapp/src/containers/Map/index.jsx
```diff
@@ -111,6 +111,7 @@ export default function Map({ history, isCompactSideMenu }) {
     }
     return () => {
       dispatch(canShowSuccessHeader(false));
+      dispatch(setMapAddDrawerHide(farm_id));
```
kathyavini, line 114: This hides the Add to Map Drawer when unmounting Map (i.e. when navigating away). There was no previous need to do so since the backdrop of the modal was basically a click away listener that would close if you clicked on, e.g., a menu.

Actually -- and I think this is a lowkey bug -- you can see on beta that if you start the "add area" action, which hides the backdrop and modal, and then navigate to a different part of app, the modal is re-opened when you return to Map because the 'showModal' redux state was never touched. I think resetting that state when leaving Map makes more sense!

SayakaOno (dismissed):
Looks great!

packages/webapp/src/components/MapDrawer/styles.module.scss
```diff
@@ -34,19 +34,49 @@
...
+}
+
+// Drawer
+.sideDrawerContainer {
+  top: calc(
+    var(--global-navbar-height) + var(--global-map-header-height) - 1px /* sm gap without adjust */
+  );
+  max-height: calc(100vh - var(--global-navbar-height) - var(--global-map-header-height));
+  min-width: 432px;
+  border-radius: 0px 4px 4px 0px;
```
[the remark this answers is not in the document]
kathyavini: Thank you for catching that! Maybe it's worth asking Loïc if he really wants different border radiuses on just this drawer, but I want to fix that specificity either way, since the 10px default was meant to be overridable 🙏

kathyavini:
I would like to discuss the header shadow on scroll tomorrow (on Figma [here](https://www.figma.com/design/eMffC0h3JPlWtzNttxJpil?node-id=22939-34838#1174167299)) and will draft until that is decided on 🙏

kathyavini:
I have reverted the shadow-on-scroll commit from this branch, and noted that commit down in a [new ticket for that feature](https://lite-farm.atlassian.net/browse/LF-4775) if someone with more frontend skills than me wants to take it up in the future).

For now on this branch I have just changed all the border radiuses to 4px as Loïc requested today, and I previously [rearranged the SCSS](https://github.com/LiteFarmOrg/LiteFarm/pull/3715/commits/ba2de86d92490e158d30bba9e75c4c48ef55fb30) a little so that only drawer directional properties (the slide in/up) are over-specified, to fix the issue above where the passed classes were not overriding default styles.

Thank you in advance for re-review 🙏

SayakaOno (approved):

packages/webapp/src/components/Drawer/index.tsx
```diff
@@ -33,13 +33,16 @@ interface DrawerProps {
   buttonGroup?: React.ReactNode;
   fullHeight?: boolean;
   desktopVariant?: DesktopDrawerVariants;
+  sideDrawerDirection?: 'left' | 'right';
```
antsgar, line 36: This only affects the styles applied in desktop, right? Should this be `desktopSideDrawerDirection` instead?

packages/webapp/src/components/Drawer/index.tsx
```diff
@@ -58,11 +61,15 @@ const Drawer = ({
...
   fullHeight,
   desktopVariant = DesktopDrawerVariants.MODAL,
+  sideDrawerDirection = 'right',
+  isCompactSideMenu,
   addBackdrop = true,
 }: DrawerProps) => {
   const theme = useTheme();
   const isDesktop = useMediaQuery(theme.breakpoints.up('sm'));
 
+  const isSideDrawer = isDesktop && desktopVariant === DesktopDrawerVariants.SIDE_DRAWER;
```
antsgar, line 71: Same here, maybe it'd be more explicit as `isDesktopSideDrawer`?

antsgar (dismissed):
It looks so beautiful! Thank you for thinking about accessibility ❤️

packages/webapp/src/components/Drawer/index.tsx
kathyavini: @antsgar Correct, because mobile has no variants! I'm definitely on board with rename but I just want to make sure it's not suggesting there is something like a `mobileSideDrawer` -- do you think that's clear enough from the props if this change is made?

antsgar: Maybe `sideDrawerDesktopVariantDirection` would be clearer? Although perhaps too long 🤔

It might also be good to make the typing more explicit so that setting the direction to left or right is only applicable if the right desktop variant is selected. Something like this could work

```
type CommonDrawerProps = {
  title: NonNullable<string | React.ReactNode>;
  isOpen: boolean;
  onClose: () => void;
  children: React.ReactNode;
  buttonGroup?: React.ReactNode;
  fullHeight?: boolean;
  isCompactSideMenu?: boolean; // only needed for left side drawer placement
  addBackdrop?: boolean;
  classes?: {
    modal?: string;
    drawerBackdrop?: string;
    drawerHeader?: string;
    drawerContent?: string;
    drawerContainer?: string; // applied to all drawers
    sideDrawerContainer?: string; // side drawer only
  };
}

type DrawerProps = CommonDrawerProps & ({
  desktopVariant?: DesktopDrawerVariants.DRAWER | DesktopDrawerVariants.MODAL;
  sideDrawerDirection: never;
} | {
  desktopVariant: DesktopDrawerVariants.SIDE_DRAWER;
  sideDrawerDirection: 'left' | 'right';
});
```

and then a TS error would show up if you try to set up the direction but forgot to set up the right variant.

packages/webapp/src/components/Drawer/index.tsx
kathyavini: I guess my original question was a bit unclear here -- I was worried if the need to add `desktop` to the sideDrawer props and booleans means something is fundamentally unclear that variants (`DesktopDrawerVariants`) cannot apply to mobile that also needs fixing. But maybe the prop name is fix enough? I'll add those first regardless of whether the types can be fixed up!

antsgar: I'm not sure, perhaps in the future every desktop variant can have its own props and we can pass them in an object that's called something like `desktopVariantProps` or such? I think that'd be clearer and more flexible, but probably overkill at this point

packages/webapp/src/components/Drawer/index.tsx
antsgar: Does it work like this?

```
type DrawerProps = CommonDrawerProps & ({
  desktopVariant?: DesktopDrawerVariants.DRAWER | DesktopDrawerVariants.MODAL;
  sideDrawerDirection?: never;
} | {
  desktopVariant: DesktopDrawerVariants.SIDE_DRAWER;
  sideDrawerDirection?: 'left' | 'right';
})
```

I'm getting an error if I try to pass a direction without a variant, and no error if I pass neither.

kathyavini: So strange! I don't get an error if I omit a variant -- only if I intentionally set the wrong one. I thought the default setting to `DesktopDrawerVariants.MODAL` had something to do with it? 🤔 This could also be fixed by requiring a variant, too, but most of the app uses the default (and is in JSX anyway)

Screenshot 2025-03-26 at 11 01 50 AM

In any case I still like the union types as a form of documentation, so I'll put them in!

packages/webapp/src/components/Drawer/index.tsx
kathyavini: That's a really nice idea actually! I'll go for the union types for now since they are most of what we need, but I think that would have been better.

kathyavini:
Sorry neglected GitHub a bit while thinking about the ESci endpoint, but props and types have been updated a bit -- sorry for invalidating reviews AGAIN 🙏

SayakaOno (approved):

antsgar (approved):

PR #3720, opened 2025-03-25
Backend typescript improvements

navdm:
This improves backend typescript imports in 2 ways:

**1. Allows using absolute imports**
 In relative imports sometimes you get unreadable imports like `import controller from '../../../src/controllers/something.js'`. Changing the location of an imported module means changing all the imports as well. With absolute imports it can be simplified to: `import controller from '@/controllers/something.js'`. This is achieved using the `paths` property of tsconfig which allows setting up absolute imports.

**2. Allows importing typescript files with the .ts extension**
Previously when using tsc to emit javascript code, it wasn't possible to have imports with .ts extensions. But recently in typescript 5.7 they added support for [Path Rewriting](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-7.html#path-rewriting-for-relative-paths). This means that its possible to use `ts` extensions and tsc will rewrite them to `js`, but its limited to relative imports and won't work with absolute imports. 

So in order in to have both of these features together (1 and 2 above), I used esbuild to bundle the output into a single file (could also use some other bundler like parcel). Esbuild has support for absolute imports and is able to resolve modules in respect to the tsconfig file, the final output is one big minified file that can be directly run using `node dist/server.js`. The `--enable-source-maps` allows easy debugging incase an error happens in production. 

**Some other benefits:**
- No need for the second tsconfig in src/tsconfig.json - it was there just for the tsc build.
- Easy migration to native typescript support in node.js - Node.js introduced experimental TS support, and according to somewhere I read in the documentation it will be released soon. If LiteFarm were to decide to migrate over to that, all imports must specify the exact file extension for the module they're importing. This PR would help keep the codebase in a state where switching over to the built-in Node.js support would be very easy.

### DIFF
--- a/packages/api/tests/sensor.test.ts
+++ b/packages/api/tests/sensor.test.ts
@@ -19,11 +19,11 @@ import chaiHttp from 'chai-http';
 chai.use(chaiHttp);
 
 import server from '../src/server.js';
-import knex from '../src/util/knex.js';
+import knex from '@/util/knex.js';
 import { tableCleanup } from './testEnvironment.js';
 
 jest.mock('jsdom');
-jest.mock('../src/middleware/acl/checkJwt.js', () =>
+jest.mock('@/middleware/acl/checkJwt.js', () =>
   jest.fn((req, _res, next) => {
     req.auth = {};
     req.auth.user_id = req.get('user_id');
@@ -37,7 +37,7 @@ const mockedAxios = axios as jest.Mocked<typeof axios>;
 
 import mocks from './mock.factories.js';
 import { Response } from 'superagent';
-import { ENSEMBLE_BRAND } from '../src/util/ensemble.js';
+import { ENSEMBLE_BRAND } from '@/util/ensemble.js';
 import { mockedFormattedReadingsData, mockedEnsembleReadingsData } from './utils/sensorMockData.js';
 
 interface User {


kathyavini (commented):
Thanks so much for looking into all of this!

I am 100% on board with this:

> This PR would help keep the codebase in a state where switching over to the built-in Node.js support would be very easy.

And I think being able to import with `.ts` would make it possible to add type declarations to JS files if necessary, too, which might come in handy.

As for absolute imports, I've never personally had any issues on the backend -- the file structure is so predictable and flat (compared to the mess that is the frontend) that I don't think I've even once run into a pathing issue when moving code around.

So I'm not sure I would want to bundle and minify server.ts _just_ for absolute imports, but I'm not against it, as long as long as  `--sourcemaps` means our Sentry issue contexts are still correct with full code (e.g. as below) and this would be the main thing I would want to check on beta after merging:

Screenshot 2025-04-02 at 3 26 52 PM


It does still feel a little bit weird to bundle and minify code that's not sent over the network, but if Sentry checks out, I can't think of any particular reasons why it should be problematic. Curious what @antsgar thinks, though!

navDhammu:
@kathyavini Yeah its not necessary to bundle the server side code, and its only really there for the absolute imports. The only benefit I can think of is that it uses less space and is faster to start the server but the difference is very negligible. If we don't care about the absolute imports then `tsc` with the new path rewriting feature in TS v5.7 would work just fine.

navDhammu:
@antsgar I did come across similar plugins like that one, didn't try them out though. But I think as @kathyavini said the backend structure is mostly flat so absolute imports might not be that important. Being able to import with `.ts` is more important in my opinion, and TypeScript now has built-in support for that - we just need to adjust the tsconfig a bit. Please let me know what the team decides and if I should create a new PR with the tsconfig changes and close this one.

kathyavini:
Ok thanks @navDhammu! We did get a chance to discuss this in tech daily and we're pretty happy with relative pathing on the backend (the imports are already working as we like), so if you don't mind setting up just the `.ts` importing change we'd love that, thank you 🙏

PR #3724, opened 2025-03-26
LF-4756(2): Strings for wind direction

SayakaOno:
**Description**

Add strings for wind direction. (I ran `pnpm i18n`, which sorted the file.)
They can be nested under `SENSOR.READING`, but since there is a possibility we'll use them elsewhere (or not?), I placed them at the top level.

Jira link: https://lite-farm.atlassian.net/browse/LF-4756

**Type of change**

### DIFF
--- a/packages/webapp/public/locales/en/translation.json
+++ b/packages/webapp/public/locales/en/translation.json
@@ -617,6 +617,24 @@
     "TUESDAY": "T",
     "WEDNESDAY": "W"
   },
+  "DIRECTION": {
+    "E": "E",
+    "ENE": "ENE",
+    "ESE": "ESE",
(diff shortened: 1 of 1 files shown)


Duncan-Brain (approved):

kathyavini (approved):
Thank you! I think I'll drop the translators an image in CrowdIn context to also make it clear what WSW etc are (in case they haven't seen those three letter abbreviations before)... maybe this one? Does it look good to you?

https://dev.qweather.com/en/docs/resource/wind-info/#16-wind-directions

SayakaOno:
Thank you @kathyavini!! 🙏 I referred to the same one, it should be good!

PR #3726, opened 2025-03-28
[WIP] LF-4699 (2): Display sensor readings on the UI

SayakaOno:
WIP!

**Description**

- 

Jira link: https://lite-farm.atlassian.net/browse/LF-4699

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/webapp/src/containers/SensorReadings/v2/LatestReadings/useLatestReadings.ts
```diff
@@ -0,0 +1,109 @@
...
+  update: () => Promise<void>;
+} {
+  const [latestReadings, setLatestReadings] = useState<SensorReadings[]>([]);
+
+  const [triggerGetSensorReadings, { isLoading, isFetching }] = useLazyGetSensorReadingsQuery();
+
+  const latestReadingTime = getLatestReadingTime(latestReadings);
+
+  const getLatestReadings = async (startTime?: Date) => {
+    let adjustedStartTime = startTime;
```
SayakaOno, line 53: As of Apr 14, 2025, it looks like Ensemble hasn’t added any data since Apr 9. To see something, try replacing this line with:
`let adjustedStartTime = new Date(2025, 3, 9);`

packages/webapp/src/containers/SensorReadings/v2/LatestReadings/useLatestReadings.ts
```diff
@@ -0,0 +1,109 @@
+/*
```
antsgar, line 1: There seem to be redundant API calls when loading the screen between this hook and the one used by the chart, `useFormattedSensorReadings` -- when loading the view I see 3 different calls to the readings endpoint. Since the chart already fetches the readings for the last week by default, could the results from that be reused to show the latest reading?

Additionally I think it's a bit odd that if there's a reading that's really old, you can view it when setting for example "Year to date", but the latest reading KPI says "No readings found" -- in reality there is a latest reading, it's just old. Is there a way we could fetch the actual latest reading from Ensemble? Or alternatively if this is the expected UX, should we update the copy to reflect somehow that "No readings found" actually means not found in the last X amount of time?

antsgar (commented):
Awesome work, so many details to take into account and you were so thorough with it. The screenshots along the way when reviewing the PR also really helped. I love how the readings view looks and feels and the code is super neat as usual as well.

I noticed a bug that's not strictly related to the work here so perhaps should be a different ticket, let me know. When clicking on Add sensor on the map, and then on "See Ensemble sensor list", and then clicking on "See on map" for a sensor or array, clicking on the sensor or array doesn't take me to the readings view as it happens when I click directly on it on the map.

packages/webapp/src/containers/SensorReadings/v2/Charts/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+  const { t } = useTranslation();
+  const theme = useTheme();
+  const isCompactView = useMediaQuery(theme.breakpoints.down('sm'));
+  const language = getLanguageFromLocalStorage();
+
+  const isSensorArray = 'sensorColorMap' in props;
+  const truncPeriod = getTruncPeriod(new Date(startDate), new Date(endDate))!;
+  const supportedReadingTypes = isSensorArray
+    ? SENSOR_ARRAY_CHART_PARAMS
+    : SENSOR_READING_TYPES[sensors[0].name];
```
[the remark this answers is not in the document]
SayakaOno: Thank you for the feedback, I’ll think this through!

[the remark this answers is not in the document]
SayakaOno: I've removed the check for whether the sensor type is an array — now the parent component is responsible for providing the correct props like colours and reading types based on the sensor type.
This should help avoid incorrect assumptions or potential errors.
Let me know if there's a better way to handle it!

SayakaOno (commented):
@antsgar Thank you so much for the feedback!
We might have assumed that the sensor icons in the map preview shouldn't be clickable. I'll check with Loïc to confirm what's expected.

packages/webapp/src/containers/SensorReadings/v2/LatestReadings/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+export type StandaloneSensorProps = {
+  type: SensorType.SENSOR;
+};
+
+export type LatestReadingsProps = { sensors: Sensor[] } & (
+  | SensorArrayProps
+  | StandaloneSensorProps
+);
+
+function isGeneralSensor(sensor: Sensor): sensor is GeneralSensor {
```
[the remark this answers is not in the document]
antsgar: Perhaps we could reflect exactly that and have this function be called something like `isNotWeatherStationSensor`? And then we wouldn't need an extra `GeneralSensor` type

packages/webapp/src/containers/SensorReadings/v2/LatestReadings/useLatestReadings.ts
antsgar: That makes sense to me, I'd keep it as is! Thank you for clarifying ☺️ 

One more thing that I was wondering about and forgot to ask in my previous review -- I'm curious about the reason we prefer or need to trigger the queries to fetch the data programmatically, rather than using the regular RTK query hooks and passing state variables for the date filters as params to the hooks?

SayakaOno: The first reason was to make the two-step approach possible. The second reason is that I was aiming for a single result to work with — whether it's from the initial load or after clicking "Update readings".

Initially, we’d use:
```
const { data } = useGetSensorReadingsQuery();
```

If I'm not mistaken, for refetching on button click, we’d need:
```
const [triggerGetSensorReadings, { data: refetchedData }] = useLazyGetSensorReadingsQuery();
```

Then we could do something like:
```
return { latestReadings: refetchedData || data, ... }
```

I had originally planned to handle the formatting inside this hook, so working with just one result seemed simpler. But now that I look at it, using both results wouldn't be as bad as I initially thought...

Would `refetchedData || data` be the right way to go about it?

(The message is now "No recent sensor readings (past 2 hours)")

SayakaOno:
We checked with Loïc this morning about the sensor icon. It's been decided that it should navigate to the sensor readings page. Here's the [ticket](https://lite-farm.atlassian.net/browse/LF-4801). (Thank you Joyce!)

packages/webapp/src/containers/SensorReadings/v2/LatestReadings/useLatestReadings.ts
antsgar: I understand! There regular `useGetSensorReadingsQuery` hook should return a `refetch` function in the result (https://redux-toolkit.js.org/rtk-query/api/created-api/hooks#usequery-signature), perhaps that could be used to trigger the programmatic fetching when user clicks on Update?

packages/webapp/src/containers/SensorReadings/v2/LatestReadings/index.tsx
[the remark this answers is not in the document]
antsgar: Ah that seems alright then! Maybe add a comment on top of the GeneralSensor type to clarify it refers to non weather station sensors?

packages/webapp/src/containers/SensorReadings/v2/Charts/index.tsx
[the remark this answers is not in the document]
antsgar: It looks good, although I think part of the problem remains -- I added a comment for it in the SensorReadings index file

packages/webapp/src/containers/SensorReadings/v2/LatestReadings/useLatestReadings.ts
antsgar: Oh and thankk you for fixing the message ☺️

packages/webapp/src/containers/SensorReadings/v2/index.tsx
```diff
@@ -58,6 +64,37 @@ const filterSensors = (id: string, type: SensorType, sensors?: Sensor[]): Sensor
...
+  if (type === SensorType.SENSOR_ARRAY) {
+    return {
+      readingTypes: SENSOR_ARRAY_CHART_PARAMS,
+      getSensorColorMap: () => generateSensorArrayColorMap(sensors),
+    };
+  }
+
+  return {
+    readingTypes: SENSOR_READING_TYPES[sensors[0].name],
+    getSensorColorMap: getSensorColorMapFunc(sensors[0].external_id),
```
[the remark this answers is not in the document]
SayakaOno: The length of `sensors` won't be greater than 1 for `SensorType.SENSOR`, since `external_id` is unique. The filtering logic ensures this:
```
const filterSensors = (id: string, type: SensorType, sensors?: Sensor[]): Sensor[] | undefined => {
  return type === SensorType.SENSOR_ARRAY
    ? sensors
        ?.filter(({ sensor_array_id }) => sensor_array_id == id)
        ?.sort((a, b) => a.depth - b.depth)
    : sensors?.filter(({ external_id }) => external_id === id);
};
```
So for standalone sensors, we’ll always get either one sensor or none.

That said, I understand it might still be unclear from the consumer’s perspective. Would it be clearer if I explicitly did something like:
```
sensors = [sensors.find(({ external_id }) => external_id === id)];
```
Or, as you suggested, passed:
```
sensors={type === SensorType.SENSOR ? sensors[0] : sensors}
```
What do you think?

packages/webapp/src/containers/SensorReadings/v2/LatestReadings/useLatestReadings.ts
SayakaOno: It should be possible now, let me try. Thank you!

packages/webapp/src/containers/SensorReadings/v2/LatestReadings/index.tsx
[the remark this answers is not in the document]
SayakaOno: Sorry for the confusion, thank you!

packages/webapp/src/containers/SensorReadings/v2/LatestReadings/useLatestReadings.ts
SayakaOno: Actually, this seems more complicated...
The second call in `setInitialLatestReadings` would need to happen in `useEffect`, and `refetch` seems to be designed to be called with the same params, so passing a different `startTime` would add more complexity.
I'll leave it as is this time!

packages/webapp/src/containers/SensorReadings/v2/index.tsx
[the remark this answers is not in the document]
antsgar: I missed that portion of the logic, that makes sense, thank you! I'd leave as is and maybe add a comment somewhere if you think it would help

antsgar (dismissed):

packages/webapp/src/containers/SensorReadings/v2/index.tsx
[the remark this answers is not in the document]
SayakaOno: Sounds good, I'll add a comment later when addressing Joyce's comments. Thank you!

packages/webapp/src/containers/SensorReadings/v2/utils.ts
```diff
@@ -0,0 +1,235 @@
...
+  if (timeOffset === 0) {
+    return dateTime;
+  }
+
+  const isAheadUTC = timeOffset < 0;
+
+  return new Date(utcYear, utcMonth, utcDate + (isAheadUTC ? 1 : 0)).getTime() / 1000;
+};
+
+export const getAdjustHourlyDateTimeFunc = (
```
kathyavini, line 99: Could this one get an explanatory comment the way that the day one does?

Do you happen to remember if the hourly value is incorrectly set to the UTC hour but the averaged time window is actually based on the local hour (in parallel to how the day one is incorrectly set to the UTC midnight but the value actually represents local midnight)? I was about to double-check the inputs and outputs but since ESci went down, I can't 😅

kathyavini (approved):
@SayakaOno wow, this is so complex (sooooooo many things to adjust and take into account!!) but you still kept it so neat and modular 😯🙇 I don't think I would change anything at all, I just had a few optional requests for comments and naming.

Just to make sure my understanding is correct, the `adjustDailyDateTime` function is necessary because:

1. We send midnight in our local (as UTC) to ESci
2. They send back midnight _of the day_ that corresponds to the above, but with a time range of data that is actually 24 hours from the timepoint in (1) -- I think I did read Morten confirming this in Slack
3. Function re-aligns data and timepoint

If so, it's crazy that (2) is happening at all, and it sounds buggy enough that I'm wondering if Morten may fix this at the same time he fixes the end time business. It's a good solution for a messy situation!

packages/webapp/src/containers/SensorReadings/v2/utils.ts
SayakaOno: I've added comments to these three functions! 

If I remember correctly, we were getting the start of the hour (0 minute) in UTC with the data for the local hour window. I was assuming that was the intended behaviour though!

SayakaOno (commented):
@kathyavini 
Thank you so much for the suggestions! I'd been working on this for so long that I wasn't sure anymore what was clear and what wasn't, so your feedback was really helpful! Could you double-check if the changes are aligned with what you expected?

And yes, your understanding of `adjustDailyDateTime` is correct! I'm not sure if Morten would be willing to fix it, but it would definitely be nice if we could eventually remove those adjustment functions!

kathyavini (approved):

antsgar (approved):

PR #3729, opened 2025-04-03
Lf 4281 migrate current weather api call to backend

gursimran-singh:
**Description**

This PR introduces a new endpoint to provide weather information for the farm. Previously, the frontend directly called the OpenWeather API, exposing API keys. Now, all logic is handled securely behind this endpoint. Additionally, this PR updates the frontend to use the correct RTK Query endpoint for fetching weather data.

Jira link:https://lite-farm.atlassian.net/browse/LF-4281

**Type of change**

**How Has This Been Tested?**
Login with your credentials and select farm. On landing page, weather information is shown.

- [x] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/api/src/util/weatherService.ts
```diff
@@ -0,0 +1,66 @@
+/*
```
antsgar, line 1: I forgot to mention this in another PR that also creates a service, but could we move services to their own folder?

packages/api/src/util/weatherService.ts
```diff
@@ -0,0 +1,66 @@
...
+
+export const weatherService = {
+  async getWeather({ lat, lon, units = 'metric' }: WeatherParams): Promise<WeatherData> {
+    try {
+      const url = `${openWeatherAPI}?units=${units}&lat=${lat}&lon=${lon}&appid=${OPEN_WEATHER_APP_ID}&lang=en&cnt=1`;
+      const response = await fetch(url);
+
+      if (!response.ok) {
+        throw new Error(`HTTP error! Status: ${response.status}`);
+      }
```
antsgar, line 46: The error response contains other useful information on top of the status, so it might be good to include more details in this error https://openweathermap.org/api/one-call-3#errors

antsgar (commented):
Left a few suggestions but looks really good overall, thank you so much again for helping improve this ☺️

packages/api/src/util/weatherService.ts
gursimran-singh: @antsgar  Sure, let me know the folder path and i will change accordingly.

antsgar: `api/src/services` should be good!

packages/api/src/controllers/weatherController.ts
```diff
@@ -0,0 +1,47 @@
...
+
+const weatherController = {
+  async getWeather(req: Request, res: Response) {
+    try {
+      const farm_id = req.headers['farm_id'];
+      const row = await baseController.getIndividual(FarmModel, farm_id);
+
+      if (!row.length) {
+        return res.sendStatus(404);
+      }
```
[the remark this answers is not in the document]
gursimran-singh: This function `getFarmByID` also uses `await baseController.getIndividual(FarmModel, id);` so the result would be same. I have updated the code to use array de-structure rather than index to make it cleaner.

packages/webapp/src/containers/WeatherBoard/index.jsx
```diff
@@ -1,40 +1,20 @@
...
-  useEffect(() => {
-    if (currentTime - data?.lastUpdated > two_hours) {
-      refetch();
-    }
-  }, []);
   const { t } = useTranslation();
-  const { tempUnit, speedUnit } = utils.getUnits(measurement);
+  const { tempUnit, speedUnit } = utils.getUnits(data?.measurement);
   const formattedForecast = {
     humidity: `${t('WEATHER.HUMIDITY')}: ${data?.humidity}`,
```
SayakaOno: Could you add the unit `%`? 🙏

gursimran-singh: done

SayakaOno:
Thank you @gursimran-singh!
Moving the "Weather" tag under `FARM_TAGS` makes sense to me. What do you think @kathyavini?

For the date field, I guess `date: data.dt + data.timezone` was incorrect? Anyway, the new code looks good, thank you!

packages/api/src/services/weatherService.ts
```diff
@@ -0,0 +1,66 @@
...
+}
+
+const OPEN_WEATHER_APP_ID = credentials.OPEN_WEATHER_APP_ID;
+const openWeatherAPI = endpoints.openWeatherAPI;
+
+export const weatherService = {
+  async getWeather({ lat, lon, units = 'metric' }: WeatherParams): Promise<WeatherData> {
+    try {
+      const url = `${openWeatherAPI}?units=${units}&lat=${lat}&lon=${lon}&appid=${OPEN_WEATHER_APP_ID}&lang=en&cnt=1`;
+      const response = await fetch(url);
```
kathyavini: Can we use `axios` here and not native `fetch()`? Axios is already used throughout the backend, so that will keep the the error and response shapes consistent with the existing patterns.

And it will also clean up the error handling a little, as with `axios` this is not necessary (an axios call will already throw on non-success codes)

```ts
if (!response.ok) {
  throw new Error(`HTTP error: ${JSON.stringify(response)}`);
}
```

packages/api/src/services/weatherService.ts
```diff
@@ -0,0 +1,66 @@
...
+        icon: data.weather[0].icon,
+        date: data.dt,
+        temp: Math.round(data.main.temp),
+        wind: data.wind.speed,
+        city: data.name,
+        measurement: units,
+      };
+    } catch (error) {
+      console.error('Error fetching weather data:', error);
+      throw new Error('Failed to fetch weather data');
```
kathyavini: As Anto mentioned, we want to get the informative [Open Weather API error response](https://openweathermap.org/api/one-call-3#errors) viewable -- which to me would be all the way to the frontend because checking the logs on the server console would be quite an ordeal (oh I forgot that we also have Sentry alongside our logging! But still nice not to have to track down / match issue). The current code just returns `'Failed to fetch weather data'` in the response.

If using `axios`, the error message in the documentation linked above will end up on `error.response.data` -- could we pass that along in the throw if it exists, with maybe a fallback to the generic axios error message?

I was thinking something like this in the catch block would preserve that richer data:

```ts
const axiosError = error as AxiosError;
throw Object.assign(new Error('Failed to fetch weather data'), {
  status: axiosError.response?.status,
  details: axiosError.response?.data ?? axiosError.message,
});
```

and then could just pass along in the controller:

```ts
} catch (error: unknown) {
  console.error(error);
  return res.status((error as HttpError).status || 500).json({ error });
}
```

packages/webapp/src/containers/WeatherBoard/index.jsx
```diff
@@ -1,40 +1,20 @@
...
-    units: { measurement },
-    language_preference: lang,
-    grid_points: { lat, lng: lon },
-  } = useSelector(userFarmSelector);
-  const {
-    data,
-    error,
-    isSuccess: loaded,
-    refetch,
-  } = useGetWeatherQuery({ units: measurement, lang, lat, lon });
```
kathyavini, line 23: One nice side-effect of your original code here was that the use of `userFarmSelector` meant we were automatically listening to changes on `userFarm` including updates on measurement (`metric`/`imperial`) and re-triggering the query based on that.

Now since measurement is queried and provided on the backend, we don't have that live update upon switching units, which we do need (it's probably the easiest way to verify units having been switched 😂)

I'm sure there's more than one way to manage this, but I think my favourite way might be using RTK query's built-in cache for query params? Then I think it will not re-fetch upon returning to original measurement, for instance:

```ts
// here in WeatherBoard
  const {
    units: { measurement },
  } = useSelector(userFarmSelector);

  const {
    data,
    error,
    isSuccess: loaded,
  } = useGetWeatherQuery({
    measurementSystem: measurement,
  });
```

and 

```ts
// weatherApi.ts
getWeather: build.query<WeatherData, { measurementSystem?: string }>({
  query: ({ measurementSystem }) => `${weatherUrl}`,
```

It could be managed with `useEffect()` + `refetch()` too, but I think this reads a little cleaner, and has nicer cache behaviour. What do you think?

kathyavini:
@gursimran-singh I re-read my response and I realized it could have been a lot clearer, but this was also exactly what I was asking for in terms of the pattern and functionality that I like in the `apiTags.ts` of the current PR!

> Move FarmTags and LibraryTags to apiTags. Combine them and export a flat array. 

So I think we're in total agreement there 👍 I also like the type safe `invalidateTags()` a lot! One thing I'd want to look at further is how the tags will be organized and accessed in each of the endpoints of the current `apiSlice` (so the farmTag _objects_ will also have to be exported, so the endpoints can refer to the correct tag)? I think it could definitely work, but I'd love to look at the full code for it all together as a unit!

I'll open another ticket for that refactor, and tag you on it in Jira. Could you put the refactor proposed above on that ticket number, after this feature is completed? I'd like to keep this ticket focused on just the code needed to move the Open Weather API to the backend.

packages/webapp/src/containers/WeatherBoard/index.jsx
gursimran-singh: `useSelector(userFarmSelector)` was useful earlier because the OpenWeather API call was handled on the frontend, and we needed to detect changes in the selected units to trigger a new API request and update the displayed values. Now that this logic has been moved to the backend, the units are retrieved from the database and updated when the form is saved. Therefore, I believe we can simply invalidate the weather tag upon saving the form if the units have changed. The only downside is caching, but user won't be changing units again and again. Let me know your views on this.

packages/api/src/services/weatherService.ts
gursimran-singh: Sure, I’ll take care of that. I just want to clarify how the error message should be shown — should we display `'Failed to fetch weather data'` in a popup and log the full error in the browser console, or should the message appear directly in the section where the weather data is displayed?

kathyavini: Oh, neither, neither! The above code change would be 100% sufficient just on it own. I mean we should be able to see the full error returned from the API in the Network tab; not that it has to be further handled on the frontend.

Screenshot 2025-10-10 at 3 59 55 PM

It's just that the current code only shows this in the Network tab, which is hard to debug on:

Screenshot 2025-10-10 at 4 01 25 PM

packages/webapp/src/containers/WeatherBoard/index.jsx
kathyavini: Invalidating the tag manually should also work, but I think using the query parameter is more idiomatic RTK Query? But as you said the user won't be changing units much so I agree not much in particular is gained from caching!

> Therefore, I believe we can simply invalidate the weather tag upon saving the form if the units have changed. 

I think using the selector is going to be cleaner than writing logic that checks/compares if specifically the units have changed when the farmSettings form is submitted, isn't it?

packages/api/src/services/weatherService.ts
gursimran-singh: done

packages/webapp/src/containers/WeatherBoard/index.jsx
gursimran-singh: done

PR #3731, opened 2025-04-04
LF-4768 Display IP Detail (1) / Display IP on Map

kathyavini:
**Description**

This is the first part of displaying irrigation prescription details for the IP view and the task view. It renders a pivot with radius based on a centerpoint (`lat`/`lng`) and radius. It also renders the underlying field [as in Figma here](https://www.figma.com/design/eMffC0h3JPlWtzNttxJpil/Litefarm-Updated?node-id=32620-88255&m=dev) and irrigation zones [as in Figma here](https://www.figma.com/design/eMffC0h3JPlWtzNttxJpil/Litefarm-Updated?node-id=31428-78859&m=dev).

Please see [Component Storybook here](http://localhost:6006/?path=/docs/components-irrigationprescriptionmapview--docs)

### Changes to LocationPicker (the map renderer) included in this ticket:
- Ability to draw a circle. This is included in the Google maps asset geometry but we hadn't used it yet
- Ability to overwrite several style properties by passing them in the location object. This includes z-Index and several other properties needed by the pivot + zones setup where colours needed to be changed dynamically based on component logic instead of based off location type. We're departing a bit from standard map objects here and I think I preferred the individual overwrite rather than expanding the map design system as a whole
- Ability to hide the label of an area by passing a `null` or omitting the "name" property. This will be used by the irrigation zones, which are unlabeled. Also added to circle in case a label is not desired in any future uses of circle
- Ability to disable hover and map controls
- Ability to disable the overlapping areas modal which would otherwise be triggered on click because the irrigation zones overlay the field
- Ability to configure the Google Maps property [gestureHandling](https://developers.google.com/maps/documentation/javascript/examples/interaction-cooperative) to a value other than 'greedy'.

In the end, I didn't feel comfortable disabling all pointer-events on this view because I don't fully trust the CSS transform that renders the map at the correct Zoom (I saw it bug out and Zoom way out once) and I'm worried about it rendering at a view that is of no use to the user. I'm trying `gestureHandling: 'cooperative'` as an intermediate between no pointer-events and greedy, and have disabled the other interactions. Feedback appreciated about this 🙏 

### Changes to Task Read Only's use of LocationPicker:
- Increased height and disabled hover now that this prop is available

#### General notes on components -- all optional refactors and so I did not pick up at this time
- This will be the third non-selecting/non-picking use of `<LocationPicker />` (the others being in `<TaskReadOnly />` and in the sensor-utilized `<LocationViewer />`)
- I wasn't quite sure what whether I wanted to refactor and use the `<LocationViewer />` newly created for sensors which includes a map-style header and renders all its objects as selected -- I could have changed both of those behaviours with props and made it more general but it was easier to use `<LocationPicker />` directly so I went that route. I could instead expand `<LocationViewer />` to make it compatible with all three read-only map views. Curious what others think.
- Otherwise, it might be better to rename `<LocationPicker />` to indicate it's also for viewing locations; it's a basic non-entity-creating map view at this point, regardless of whether locations need to be selected or not. It is also misnamed `singleLocationPicker` according to its folder, but it can render any number of locations.

Jira link: https://lite-farm.atlassian.net/browse/LF-4768

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/assets/colors.scss
+++ b/packages/webapp/src/assets/colors.scss
@@ -131,6 +131,7 @@
   --Colors-Accent---singles-Brown-dark: #633700;
   --Colors-Accent---singles-Purple-light: #f4e8ff;
   --Colors-Accent---singles-Purple-full: #8f26f0;
+  --Colors-Accent---singles-Purple-dark: #3a0070;
 
   --Form-focus: #89d1c7;
   --Box-bg: #f5fafa;
--- a/packages/webapp/src/components/IrrigationPrescription/IrrigationPrescriptionMapView.tsx
+++ b/packages/webapp/src/components/IrrigationPrescription/IrrigationPrescriptionMapView.tsx
@@ -19,7 +19,7 @@ import { useMaxZoom } from '../../containers/Map/useMaxZoom';
 import LocationPicker from '../LocationPicker/SingleLocationPicker';
 import { GestureHandling } from '../LocationPicker/SingleLocationPicker/types';
 import { Location } from '../../types';
-import { IRRIGATION_ZONE_COLOURS, EARTH_RADIUS } from './constants';
+import { IRRIGATION_ZONE_COLOURS, EARTH_RADIUS, BRIGHT_PIVOT_COLOUR } from './constants';
 
 interface IrrigationZonePolygon {
   grid_points: { lat: number; lng: number }[];
@@ -42,9 +42,12 @@ const IrrigationPrescriptionMapView = ({
 }: IrrigationPrescriptionMapViewProps) => {
   const { maxZoomRef, getMaxZoom } = useMaxZoom();
 
-  const vriZonesPresent = !!vriZones?.length;
-
-  const pivotMapObjects = createPivotMapObjects(pivotCenter, pivotRadius, vriZonesPresent);
+  const pivotMapObjects = createPivotMapObjects(
+    pivotCenter,
+    pivotRadius,
+    !!vriZones?.length,
+    (vriZones?.length || 0) > 3,
+  );
 
   const irrigationZoneMapObjects = vriZones ? createIrrigationZoneMapObjects(vriZones) : [];
 
@@ -72,6 +75,7 @@ const createPivotMapObjects = (
   center: { lat: number; lng: number },
   radius: number,
   vriZonesPresent: boolean,
+  moreThanThreeZones: boolean,
 ) => {
   const label = `${radius}m`; // TODO: units
 
@@ -82,6 +86,12 @@ const createPivotMapObjects = (
     radius,
     name: label,
     ...(vriZonesPresent ? { fillOpacity: 0 } : {}),
+    ...(moreThanThreeZones
+      ? {
+          markerColour: BRIGHT_PIVOT_COLOUR,
+          lineColour: BRIGHT_PIVOT_COLOUR,
+        }
+      : {}),
   };
 
   // Calculate the endpoint on the circle's circumference (eastward) [Source: Copilot]
@@ -95,6 +105,7 @@ const createPivotMapObjects = (
     line_points: [center, endpoint],
     width: 2,
     zIndex: 1, // render on top of pivot fill
+    ...(moreThanThreeZones ? { lineColour: BRIGHT_PIVOT_COLOUR } : {}),
   };
 
   return [pivot, pivotArm];
--- a/packages/webapp/src/components/IrrigationPrescription/IrrigationPrescriptionMapView.tsx
+++ b/packages/webapp/src/components/IrrigationPrescription/IrrigationPrescriptionMapView.tsx
@@ -15,10 +15,11 @@
 
 import clsx from 'clsx';
 import styles from './styles.module.scss';
+import { convert } from '../../util/convert-units/convert';
 import { useMaxZoom } from '../../containers/Map/useMaxZoom';
 import LocationPicker from '../LocationPicker/SingleLocationPicker';
 import { GestureHandling } from '../LocationPicker/SingleLocationPicker/types';
-import { Location } from '../../types';
+import { Location, System } from '../../types';
 import { IRRIGATION_ZONE_COLOURS, EARTH_RADIUS, BRIGHT_PIVOT_COLOUR } from './constants';
 
 interface IrrigationZonePolygon {
@@ -28,25 +29,28 @@ interface IrrigationZonePolygon {
 interface IrrigationPrescriptionMapViewProps {
   fieldLocation: Location;
   pivotCenter: { lat: number; lng: number };
-  pivotRadius: number; // in meters
+  pivotRadiusInMeters: number;
   className?: string;
   vriZones?: IrrigationZonePolygon[];
+  system?: System;
 }
 
 const IrrigationPrescriptionMapView = ({
   fieldLocation,
   pivotCenter,
-  pivotRadius,
+  pivotRadiusInMeters,
   className,
   vriZones,
+  system = 'metric',
 }: IrrigationPrescriptionMapViewProps) => {
   const { maxZoomRef, getMaxZoom } = useMaxZoom();
 
   const pivotMapObjects = createPivotMapObjects(
     pivotCenter,
-    pivotRadius,
+    pivotRadiusInMeters,
     !!vriZones?.length,
     (vriZones?.length || 0) > 3,
+    system,
   );
 
   const irrigationZoneMapObjects = vriZones ? createIrrigationZoneMapObjects(vriZones) : [];
@@ -76,8 +80,14 @@ const createPivotMapObjects = (
   radius: number,
   vriZonesPresent: boolean,
   moreThanThreeZones: boolean,
+  system: System,
 ) => {
-  const label = `${radius}m`; // TODO: units
+  const unit = system === 'imperial' ? 'ft' : 'm';
+
+  const labelRadius =
+    system === 'imperial' ? Math.round(convert(radius).from('m').to('ft')) : radius;
+
+  const label = `${labelRadius}${unit}`;
 
   const pivot = {
     type: 'pivot',
--- a/packages/webapp/src/components/IrrigationPrescription/IrrigationPrescriptionMapView.tsx
+++ b/packages/webapp/src/components/IrrigationPrescription/IrrigationPrescriptionMapView.tsx
@@ -17,6 +17,7 @@ import clsx from 'clsx';
 import styles from './styles.module.scss';
 import { useMaxZoom } from '../../containers/Map/useMaxZoom';
 import LocationPicker from '../LocationPicker/SingleLocationPicker';
+import { GestureHandling } from '../LocationPicker/SingleLocationPicker/types';
 import { Location } from '../../types';
 import { IRRIGATION_ZONE_COLOURS, EARTH_RADIUS } from './constants';
 
@@ -57,6 +58,9 @@ const IrrigationPrescriptionMapView = ({
         maxZoomRef={maxZoomRef}
         getMaxZoom={getMaxZoom}
         showControls={false}
+        disableOverlappingAreasModal={true}
+        disableHover={true}
+        gestureHandling={GestureHandling.COOPERATIVE}
       />
     </div>
   );
--- a/packages/webapp/src/components/IrrigationPrescription/IrrigationPrescriptionMapView.tsx
+++ b/packages/webapp/src/components/IrrigationPrescription/IrrigationPrescriptionMapView.tsx
@@ -18,8 +18,7 @@ import styles from './styles.module.scss';
 import { useMaxZoom } from '../../containers/Map/useMaxZoom';
 import LocationPicker from '../LocationPicker/SingleLocationPicker';
 import { Location } from '../../types';
-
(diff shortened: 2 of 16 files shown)


packages/webapp/src/assets/colors.scss
```diff
@@ -131,6 +131,7 @@
   --Colors-Accent---singles-Brown-dark: #633700;
   --Colors-Accent---singles-Purple-light: #f4e8ff;
   --Colors-Accent---singles-Purple-full: #8f26f0;
+  --Colors-Accent---singles-Purple-dark: #3a0070;
```
kathyavini, line 134: Actually not used in this ticket! The map objects can't really take CSS variables in this way so this hex was added directly to `containers/Map/styles.module.scss'. But for documentation of the design system I wanted to include it.

packages/webapp/src/components/LocationPicker/SingleLocationPicker/index.jsx
```diff
@@ -325,7 +353,7 @@ const LocationPicker = ({
         onGoogleApiLoaded={({ map, maps }) => handleGoogleMapApi(map, maps)}
         options={getMapOptions}
       />
-      {overlappedPositions.length > 1 && !isPinMode && (
+      {overlappedPositions.length > 1 && !isPinMode && !disableOverlappingAreasModal && (
```
kathyavini, line 356: This element, which is renders onClick if there are multiple areas overlapping. This will always be the case between fields and irrigation zones in the IP map view so I wanted to disable it.

Screenshot 2025-04-08 at 1 03 16 PM

packages/webapp/src/components/LocationPicker/SingleLocationPicker/index.jsx
```diff
@@ -325,7 +353,7 @@ const LocationPicker = ({
         onGoogleApiLoaded={({ map, maps }) => handleGoogleMapApi(map, maps)}
         options={getMapOptions}
       />
-      {overlappedPositions.length > 1 && !isPinMode && (
+      {showOverlappingAreasModal && overlappedPositions.length > 1 && !isPinMode && (
```
kathyavini, line 356: This element, which is renders onClick if there are multiple areas overlapping. This will always be the case between fields and irrigation zones in the IP map view so I wanted to disable it.

Screenshot 2025-04-08 at 1 03 16 PM

packages/webapp/src/components/LocationPicker/SingleLocationPicker/drawLocations.js
```diff
@@ -20,37 +22,42 @@ export const SELECTED_POLYGON_OPACITY = 1.0;
 export const DEFAULT_POLYGON_OPACITY = 0.5;
 export const HOVER_POLYGON_OPACITY = 0.8;
 
-export const drawCropLocation = (map, maps, mapBounds, location) => {
-  if (isNoFillArea(location.type)) return drawNoFillArea(map, maps, mapBounds, location);
-  if (isLine(location.type)) return drawLine(map, maps, mapBounds, location);
-  if (isArea(location.type)) return drawArea(map, maps, mapBounds, location);
-  if (isPoint(location.type)) return drawPoint(map, maps, mapBounds, location);
+export const drawCropLocation = (map, maps, mapBounds, location, disableHover = false) => {
```
antsgar, line 25: Should this be renamed to `drawLocation` instead since it's not strictly related to crops?

packages/webapp/src/components/LocationPicker/SingleLocationPicker/drawLocations.js
```diff
@@ -102,29 +109,87 @@ const drawArea = (map, maps, mapBounds, location) => {
...
+    map,
+  });
+
+  if (!disableHover) {
+    maps.event.addListener(circle, 'mouseover', function () {
+      if (this.fillOpacity !== SELECTED_POLYGON_OPACITY) {
+        this.setOptions({ fillOpacity: HOVER_POLYGON_OPACITY });
+      }
+    });
+    maps.event.addListener(circle, 'mouseout', function () {
```
antsgar, line 136: Should all of these event listeners be removed at some point?

packages/webapp/src/components/IrrigationPrescription/IrrigationPrescriptionMapView.tsx
```diff
@@ -0,0 +1,141 @@
...
+        disableHover={true}
+        gestureHandling={GestureHandling.COOPERATIVE}
+      />
+    </div>
+  );
+};
+
+export default IrrigationPrescriptionMapView;
+
+const createPivotMapObjects = (
```
antsgar, line 78: It's tough to gauge the performance by only viewing this component in Storybook so we can wait and see once it's being used in app, but I'm wondering if there's enough calculations in here to warrant wrapping this in a `useCallback` and the results of it in a `useMemo`

antsgar (approved):
Looks great and I'm really impressed at how quickly you got the hang of all of these map changes, I feel like that would have taken me forever. Nicely done! 🙌

Duncan-Brain (commented):
It is looking really good! I have left a few very minor inline comments, and I have some comments about the structure question you asked below but I am not requesting changes or anything like that.

> I wasn't quite sure what whether I wanted to refactor and use the <LocationViewer /> newly created for sensors which includes a map-style header and renders all its objects as selected -- I could have changed both of those behaviours with props and made it more general but it was easier to use <LocationPicker /> directly so I went that route. I could instead expand <LocationViewer /> to make it compatible with all three read-only map views. Curious what others think.

I think the reason that it feels so different from the other option like ` <LocationViewer />` is that you are forming the location objects pivot, pivot arm, and irrigation zones here. I think if these objects were formed more or less in the backend like sensors or any other location object then using `<LocationViewer />` with your new props improvements would be viable.

I am trying to think about the reusability of this component in irrigation task and looking to understand more of that part. In the location selection of a task we will typically show all possible selectable areas for the task, with the items selected from the previous screen if coming from a map or crop view. So in the irrigation task we will need to have logic that says: `if(comingFromIP) display <IrrigationPrescriptionMapView />` and that is the only way to get this view with a pivot right? Otherwise if coming from add task generally then it is not possible to reach this component or select a pivot/irrigation zone because they might want to select other locations?  Just curious about what limitations are envisioned and if I am missing a piece of the puzzle there.

I think what the above means is that I might have expected was that the `drawCircle` was also added to the `useMapAssetRenderer` as well and that to make pivots and zones available on all map viewers we might pretend we get the location data back exactly how we want it in a way that mimics a location selector or query hook, and set the drawing information for complex shapes in a `drawX` sort of way (complex shapes seem novel to LiteFarm as we have had mostly 1-1 shape-locations, but pivot = (circle + line) and fieldZones = (multiple polygons) are new to understand) .

Happy to approve but any extra info you can give me for the reusability part would be helpful! Thanks

packages/webapp/src/components/IrrigationPrescription/IrrigationPrescriptionMapView.tsx
kathyavini: I've been doing all my testing in App actually (with a fake container) and it seems okay, but I'd love to know how to quantify that!

packages/webapp/src/components/IrrigationPrescription/IrrigationPrescriptionMapView.tsx
```diff
@@ -0,0 +1,141 @@
...
+import clsx from 'clsx';
+import styles from './styles.module.scss';
+import { convert } from '../../util/convert-units/convert';
+import { useMaxZoom } from '../../containers/Map/useMaxZoom';
+import LocationPicker from '../LocationPicker/SingleLocationPicker';
+import { GestureHandling } from '../LocationPicker/SingleLocationPicker/types';
+import { Location, System } from '../../types';
+import { IRRIGATION_ZONE_COLOURS, EARTH_RADIUS, BRIGHT_PIVOT_COLOUR } from './constants';
+
+interface IrrigationZonePolygon {
```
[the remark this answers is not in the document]
kathyavini: This wasn't a very good name for that interface! It's not a reference to a location type, but to Irrigation Data that includes polygons (i.e. VRI data). I renamed it to `IrrigationPolygonData` in my follow-up PR [here](https://github.com/LiteFarmOrg/LiteFarm/pull/3738/files#diff-48390799d6887cc8d6113701cdb3799dff28845c823e804ad71d1ca26a652f19) where hopefully it's more clear what it means.

packages/webapp/src/components/IrrigationPrescription/IrrigationPrescriptionMapView.tsx
```diff
@@ -0,0 +1,141 @@
...
+    ...(vriZonesPresent ? { fillOpacity: 0 } : {}),
+    ...(moreThanThreeZones
+      ? {
+          markerColour: BRIGHT_PIVOT_COLOUR,
+          lineColour: BRIGHT_PIVOT_COLOUR,
+        }
+      : {}),
+  };
+
+  // Calculate the endpoint on the circle's circumference (eastward) [Source: Copilot]
```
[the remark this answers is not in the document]
Duncan-Brain: Okay I might prefer to have some source to reference, maybe it is possible to find a good reference. 

I am surprised that the turf package doesn't have something like this built in -- it looks like they only have a circular polygon creator that I can see.

packages/webapp/src/components/IrrigationPrescription/IrrigationPrescriptionMapView.tsx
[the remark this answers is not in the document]
kathyavini: I like the consistency of always using the same named type, but you don't think it obscures readability a bit? I find it harder to understand the shape if I have to click through a different file:

Screenshot 2025-04-15 at 12 38 36 PM

kathyavini:
@Duncan-Brain 

Okay, let me try to address your comments about reusability.

1. This one might be quick

> I think if these objects were formed more or less in the backend like sensors or any other location object then using `<LocationViewer />` with your new props improvements would be viable.

I'm wasn't talking about `<LocationViewer />` as a replacement for `<IrrigationPrescriptionMapView />`, I was talking about it as a replacement for `<LocationPicker />` within that component. Because I'm using LocationPicker with props that make it a "viewer" (as does readonly tasks) and I thought maybe all the components that "view" instead of "pick" should have a common base? But it's not that important if it wasn't bothering you -- as the author of `<LocationViewer />` I thought you in particular would have been the one bothered if anyone :)

2. Regarding task flow:
   
> So in the irrigation task we will need to have logic that says: `if(comingFromIP) display <IrrigationPrescriptionMapView />` and that is the only way to get this view with a pivot right?

Kind of. If you're talking about task creation I don't think there is a locations view at all in the flow from IP view --> create new task from the IP. The location selection step would be skipped in this flow.

If you mean after the task is created, then yes, I imagine that if there is an associated IP uuid on the task, the whole `<PureIrrigationPrescription />` component is shown (it's map + table both -- see the [follow-up PR](https://github.com/LiteFarmOrg/LiteFarm/pull/3738)) and no traditional locations map.

> Otherwise if coming from add task generally then it is not possible to reach this component or select a pivot/irrigation zone because they might want to select other locations?

Correct, not possible to reach. The pivots and zones being map objects were de-scoped, and as far as I know we have not scoped adding an IP to an irrigation task that is created with add task and not from the IP. So the standard add task flow has nothing to do with pivots or zones.

> I think what the above means is that I might have expected was that the `drawCircle` was also added to the `useMapAssetRenderer` as well and that to make pivots and zones available on all map viewers

Kind of same thing here -- I think this was originally the case, but then intentionally de-scoped. Pivots will not be map objects, nor are the zones. 

I think the fundamental difference in how we are thinking about it, is that you are putting a bit more emphasis on the fake "locations" created for this view, while I'm thinking of those objects as the visual representations of an Irrigation Prescription, which contains the sort of data that can be rendered on a map. I mean obviously they are both 😂  But for me, they are location-esque only because I wanted to piggy-back on our existing map drawing infrastructure rather than re-create that from scratch, and their identity as locations extends no further than this one component.

Does that address what you're asking? I didn't really understand this last part

> and set the drawing information for complex shapes in a drawX sort of way (complex shapes seem novel to LiteFarm as we have had mostly 1-1 shape-locations, but pivot = (circle + line) and fieldZones = (multiple polygons) are new to understand) .

packages/webapp/src/components/IrrigationPrescription/IrrigationPrescriptionMapView.tsx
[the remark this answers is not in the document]
Duncan-Brain: Yeah I get that, for me its not consistency so much as the type checking that guarantees compatibility between everything we want as important. I agree it would be nice if intellisense showed that info, sometimes I wonder if I am missing a setting to see more that way.

Duncan-Brain (approved):
@kathyavini Thanks for the explanation!

I think I did not process that it would not be possible to see this from following and creating a regular irrigation task flow. And so being able to use the fake location constructs in any map viewer seemed valuable since we will always be piggybacking.

I did also think the fakeLocations were valuable in the sense of they can contribute to current and future scope goals with maybe similar lines of code -- I think I missed that displaying pivots on the map was fully descoped, I thought we just decided to them like sensors from ESci. ... the last part was just acknowledging that things like colours and z index would probably belong on frontend and not part of the fake location.

It looks fantastic, cool to see our first circle "location" !

PR #3737, opened 2025-04-11
Lf 4767 fetch and display available i ps

Duncan-Brain:
**Description**

This PR Makes the table, and tab for the irrigation route. It also mocks the data which will be removed and put in the backend in LF-4765.

Notes:
- I have separated out the Table Cell types to a different PR. 
- noticed the PR called for 24 hours refresh on prescriptions, not quite sure what was expected there. but am planning on putting a date range on the endpoint that should cover it. If we want a real-time effect I can try polling(https://redux-toolkit.js.org/rtk-query/usage/polling) but I didnt get the sense that was what was asked for?

Jira link: [LF-4767](https://lite-farm.atlassian.net/browse/LF-4767)

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

[LF-4767]: https://lite-farm.atlassian.net/browse/LF-4767?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kathyavini:
> Is IP attached to pivot or to field?
Do we filter IP's based on pivot gridpoints? Or does esci store our location id and send back to us to filter by?

@Duncan-Brain it's attached to the field. We do indeed send them the `location_id` and have asked them to associate the IP with it. The `location_id`  is what I would expect to filter on for this table. Pivot is never a LiteFarm entity was my understanding.

Duncan-Brain:
Thanks @kathyavini that makes sense! Definitely pivot won't be a LiteFarm entity but I am thinking about if any details are pivot dependant than a prescription would have a few different foreign keys. And pondering if we are only showing a few prescriptions if a second endpoint is effective for us. I won't think too hard though and just make the assumption for now that we get exactly what we want in separate calls.

packages/webapp/src/store/api/apiSlice.ts
```diff
@@ -290,6 +294,23 @@ export const api = createApi({
       invalidatesTags: (_result, error) =>
         error ? [] : ['FarmAddon', 'Sensors', 'SensorReadings'],
     }),
+    getIrrigationPrescriptions: build.query<IrrigationPrescription[], void>({
+      query: () => `${irrigationPrescriptionsUrl}`,
+      async onQueryStarted(_id, { dispatch, queryFulfilled }) {
+        try {
+          // Replacement for invalidates tags 'Tasks'
```
Duncan-Brain, line 301: Since it kind of requires up to date tasks I did something in the api query that is a bit different to avoid concurrency issues, please check that out and tell me if you think it is similar to invalidating tags.

packages/webapp/src/components/Table/Cell/CellTypes/RightChevronLink.tsx
```diff
@@ -14,18 +14,20 @@
...
 export type RightChevronLinkProps = {
   path: string;
 };
 
 const RightChevronLink = ({ path }: RightChevronLinkProps) => {
   return (
-    <IconButton color="info" className={styles.rightChevronLink} onClick={() => history.push(path)}>
-      <BsChevronRight />
-    </IconButton>
+    <Link to={{ pathname: path }}>
```
Duncan-Brain, line 26: ChevronLink cell type has been updated to use react router Link which has some nice functionality like the hover that shows your target destination in the bottom left.

packages/webapp/src/store/api/apiSlice.ts
```diff
@@ -290,6 +294,23 @@ export const api = createApi({
...
+          dispatch({ type: 'getTasksSaga' });
+          await queryFulfilled;
+        } catch (err) {
+          dispatch({
+            type: 'snackbarReducer/enqueueErrorSnackbar',
+            payload: i18n.t('message:FARM_ADDON.IRRIGATION_PRESCRIPTION.FAILED_TO_GET'),
+          });
+        }
+      },
+      keepUnusedDataFor: 60 * 60 * 24 * 365, // 1 year
```
antsgar, line 311: Why are we overriding this?

packages/webapp/src/containers/LocationDetails/LocationIrrigation/useIrrigationPrescriptions.ts
```diff
@@ -0,0 +1,82 @@
...
+    task_id: undefined,
+  },
+];
+
+export default function useIrrigationPrescriptions(location?: Location) {
+  if (!location) {
+    return [];
+  }
+
+  const { data, error, isFetching } = useGetIrrigationPrescriptionsQuery();
```
antsgar, line 50: The query will actually need to be called (or pre-fetched) as soon as the app loads, since the GET endpoint for IPs will be taking care of generating the notification if there's a new IP -- this should still happen even if the user doesn't specifically navigate to this view.

antsgar (commented):
Left a few comments! For the "real time" behavior, the reason we need it is to show the notification to the user as soon as the IP is available. Ultimately this is not very optimal and it'd be cleaner to have them let us know when there's a new IP, but this is what we landed on for simplicity of implementation for now.

We could do polling, or what I was thinking was something much more barebones to avoid making so many requests when we only expect one new IP per day. I thought we might be able to roughly calculate the time that should ellapse between the timestamp for the last IP in the response and the moment where we should be getting the next IP, and then just setting a timeout to trigger a new request after that time -- at that point we could start polling until we get the new one. But honestly perhaps just polling would be much easier, so I'm totally okay with that.

antsgar:
Two additional comments re styling
- There's a big gap in the table between the header and the contents which looks off, not sure where it's coming from.
- On mobile, the table headers disappear, is that expected? I think this table might be hard to understand without a header.

packages/webapp/src/store/api/apiSlice.ts
Duncan-Brain: Yeah agreed, but there is no Tasks tag currently. And that would be a big refactor I think

packages/webapp/src/store/api/apiSlice.ts
Duncan-Brain: I used it wrong here actually because I use the data directly in the component vs app level, but if I move the query up to the app level I think it is there so that the data does not get purged. I asked Joyce about its original implementation here:

https://github.com/LiteFarmOrg/LiteFarm/pull/3665#discussion_r1929031820

packages/webapp/src/containers/LocationDetails/LocationIrrigation/useIrrigationPrescriptions.ts
Duncan-Brain: I can add it there too if you think I should, I might be now confused about what this PR should cover and what 
Ticket LF # 4762 (avoiding linking for jira auto) should cover.

Duncan-Brain:
@antsgar Thanks for the review, I have a few questions

1. 24 hr expiry + notifications: Generally, what should this PR cover and what should the notifications story cover? ( LF # 4762 ) I can add polling and a second query top level but it doesn't seem specifically related to this view.
  - Aren't notifications stored in the db and created on the backend? Maybe I misunderstand how it works but it seems like they should really be created and sent from there rather than setting up a "frontend cron/timer".
  - Polling seems easy enough that we could maybe consider it, it is to our own backend so it would be easy enough maybe to compare against the notification table
  
2. Should I ask if the mobile view header in figma was removed in error?
Screenshot 2025-04-28 at 10 10 20 AM


Will take a look into the heading gap and element type!

packages/webapp/src/components/LocationIrrigation/index.tsx
```diff
@@ -0,0 +1,139 @@
...
+  return (
+    <Layout className={layoutStyles.paperContainer}>
+      <PageTitle title={location.name} onGoBack={() => history.push('/map')} />
+      <RouterTab
+        classes={{ container: { margin: '30px 0 26px 0' } }}
+        history={history}
+        tabs={routerTabs}
+        variant={Variant.UNDERLINE}
+      />
+      <div className={styles.subtitle}>{t('IRRIGATION_PRESCRIPTION.IRRIGATION_PRESCRIPTIONS')}</div>
```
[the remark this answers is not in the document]
Duncan-Brain: should be fixed!

Duncan-Brain:
Big gap is fixed - it is a special thing that was added for the animals TableV2 design but agreed doesn't make sense when we update those table styles to look differently.

packages/webapp/src/containers/LocationDetails/LocationIrrigation/useIrrigationPrescriptions.ts
antsgar: The notification ticket covers modifying the backend method to check if the notification exists and if not, create one, but the call to the endpoint itself when the app loads from the UI belongs in the ticket related to this PR according to the first item

> Fetch IPs by calling GET /irrigation_prescriptions when app loads.
> If 24 hs have elapsed since last prescription timestamp, IPs should be re-fetched.

So at this point there shouldn't be any notifications happening, but the endpoint call should be happening at the right time.

antsgar:
@Duncan-Brain to add a bit more regarding the notification stuff -- it's the BE endpoint that fetches the IP list that will generate the notification once it gets the results from Ensemble, but we need to trigger a call to that endpoint from the UI at the right time to make sure we generate that notification in a timely manner. Let me know if that clarifies it! Alternatively I suppose we could have a second cronjob in the BE polling for the IPs and checking if a new notification needs to be generated.

Re the table header stuff, I couldn't find a Figma view for mobile but if that's the way it's in Figma then yeah, maybe I'd double check with Loïc if that's right just in case.

packages/webapp/src/containers/LocationDetails/LocationIrrigation/useIrrigationPrescriptions.ts
```diff
@@ -0,0 +1,82 @@
...
+  const tasks = useSelector(tasksSelector);
+
+  // TODO: refactor once mocked data is no longer needed
+  const irrigationPrescriptions = isFetching
+    ? []
```
kathyavini, line 55: I see the TODO so I'm not sure how much of this is destined for the final code, but this is causing a horrific flickering when you switch tabs due to Irrigation tab appearing then disappearing then appearing again.

Even in the final version I don't think you need need to track or (definitely not) set data empty based on `isFetching` like this -- pretty sure that would cause a flicker whenever you have a cache. Definitely worth a check, but without `selectFromResult` I don't think you need  to track `isFetching` at all.

```ts
// eventual code -- this should be sufficient I think
 const { data = [] } = useGetIrrigationPrescriptionsQuery();
 ```

For now, this would cure that really gnarly tab layout shift flicker on every tab change:

```ts
// with mock this is enough
  const { data } = useGetIrrigationPrescriptionsQuery();
...
  const irrigationPrescriptions = data ?? getMockData(location, tasks);
 ```

File: packages/webapp/src/store/api/apiSlice.ts (hunk 3 above)
Comment (kathyavini) [reply packages/webapp/src/store/api/apiSlice.ts:311]: Sorry! Totally never responded to that thread.

I think not purging is only necessary if the query hook isn't also called in the component that shows the data, whether or not it's initially called at login. In the case of IP I believe it needs to be called in both though, because the query hook is also your selector for the list.

This was specifically for the sensors in the map view since, IIRC, the way that map objects are built, it wasn't possible to call the sensor query hook it in the map component and have it update or persist correctly.

**Comment by Duncan-Brain** [comment]
@antsgar Okay thanks for the extra info -- I might need some extra help and direction with this then as I think it is a new pattern. 

Option A:

So I have a [slightly messy] function that I do not know where to place it (I put it in /containers/Home to test it). There is not really a top level authenticated component that I can think of that exists on every page that would "persist" a function relying on the latest call. On top of this architecture-wise, it feels like it could get cluttered and maybe inefficient if more such subscriptions were added to the top level. So I cannot explain exactly why yet but it is giving me "code-smell".

```
  const { data: ipData, refetch: refetchIrrigationPrescriptions, isFetchingIps, errorIps } = useGetIrrigationPrescriptionsQuery();
  
  // Delete me from here
  const locations = useSelector(locationsSelector);
  // TODO: refactor once mocked data is no longer needed
  const irrigationPrescriptions = isFetchingIps
    ? []
    : errorIps || !ipData
    ? getMockData(locations.find((l) => !l.deleted), [], true)
    : ipData;

  // Delete me to here

  const ONE_DAY_IN_MS = 1000 * 60 * 60 * 24;
  useEffect(() => {
    let timeoutId;
    if (irrigationPrescriptions && irrigationPrescriptions.length) {
      const mostPastDate = irrigationPrescriptions.reduce((acc, cv) => {
        return new Date(acc) < new Date(cv.recommended_start_datetime) ? acc : cv.recommended_start_datetime;
      }, irrigationPrescriptions[0].recommended_start_datetime)
      const timeouttime = (new Date(mostPastDate).valueOf() + ONE_DAY_IN_MS) - Date.now(); 
      timeoutId = setTimeout(function() {
        refetchIrrigationPrescriptions();
      }, timeouttime);
    }
    return () => clearTimeout(timeoutId);
  }, [irrigationPrescriptions]);
```

Option B:

Using that polling interval, but I still don't know where to put it or what a good interval is, something like : 

```
  const { data: ipData, refetch: refetchIrrigationPrescriptions, isFetchingIps, errorIps } = useGetIrrigationPrescriptionsQuery({pollingInterval: 1000*60*30}); // every 30 mins
```

Option C:

I believe the getAlert() saga polls the backend for new notifications, so if we kept it on the backend the notifications would handle itself and we could use the `irrigationPrescriptions` query only where needed in the code.

Other options?

Other questions:
- I didn't realize I should be adding notifications on the GET endpoint ticket I have started - if one is generated daily -- that could be a lot of notifications so I will need to ask Loic about that maybe before it can be completed.

File: packages/webapp/src/store/api/apiSlice.ts (hunk 3 above)
Comment (Duncan-Brain) [reply packages/webapp/src/store/api/apiSlice.ts:311]: I think you did respond! 

Yeah okay let me try removing it then thanks @kathyavini ! I think I understand now that because the mock selectors using `[api.endpoint.getSensors()]` or something like that, they don't call the hook but instead rely on the home page calling it.

Did you consider other top level components to put it on by chance?

Comment (kathyavini) [reply packages/webapp/src/store/api/apiSlice.ts:311]: > Did you consider other top level components to put it on by chance?

Do you mean like higher up the component tree to Navigation? I don't think it sounds like a good pattern to call the query hook on every page change to make sure the data is refreshed -- compared to that I would prefer the non-purging.

> they don't call the hook but instead rely on the home page calling it.

I think so, yeah. I'm still so completely sketched out by those "selectors" and mad at RTK Query for not giving clearer instructions in their docs on how to integrate the two parts of their _very own library_ 😅

**Comment by antsgar** [comment]
@Duncan-Brain your comment about how we poll for notifications pointed me in the right direction and I noticed something I had missed -- we already have a cronjob on the BE generating notifications for tasks that are due each day or unassigned, so we just have to reuse that same cronjob and add a new method to handle the IP notifications. Sorry about the confusion, I still have to learn a lot about notifications not having worked on that portion of the code before! Having that in mind we can scratch anything related to prefetching the data from this ticket, and I'll update the notification ticket to reflect the required changes.

As for the daily notifications which will be a lot of them, we did discuss that with Divya and Loïc (I think you were on vacation at the time!) and although we all agreed it is going to be quite cluttering we decided on moving forward with it anyway. Ideally on the long run the notifications would be grouped to avoid that problem.

File: packages/webapp/src/containers/LocationDetails/LocationIrrigation/useIrrigationPrescriptions.ts (hunk 6 above)
Comment (Duncan-Brain) [reply packages/webapp/src/containers/LocationDetails/LocationIrrigation/useIrrigationPrescriptions.ts]: Temporary conditional yeah. But thanks it got weird here I guess and I created the situation I was trying to avoid haha. I didn't want to render `[ ]` because of the tab showing only if they exist. But I really didn't do well with that goal haha 😂

File: packages/webapp/src/store/api/apiSlice.ts (hunk 3 above)
Comment (Duncan-Brain) [reply packages/webapp/src/store/api/apiSlice.ts]: > I don't think it sounds like a good pattern to call the query hook on every page change to make sure the data is refreshed -- compared to that I would prefer the non-purging.

Agreed, I ask because now that I understand better I have noticed that on multiple occasions I go to the map and I don't have any sensors and I think it is because I skip the `Home` page somehow. So it is not necessarily reliable unless put at the top level (since its react hopefully it would only render once unless refreshed) but yeah ideally just spots we use the data for.

Yeah it would be nice to have another way to store those I was also pretty sketched when I was pretty sure I was directly editing the store for them somehow during 4702, thanks for looking at that with me too.

**Comment by Duncan-Brain** [comment]
@antsgar thanks for the info, I have not worked on notifications yet either. This option sounds like it will be cleaner.

I am going to add that to the PR description of LF # 4765 that notifications should also be created on that endpoint

File: packages/webapp/src/containers/LocationDetails/LocationIrrigation/useIrrigationPrescriptions.ts (hunk 4 above)
Comment (Duncan-Brain) [reply packages/webapp/src/containers/LocationDetails/LocationIrrigation/useIrrigationPrescriptions.ts]: I think this is resolved that we will be instead u sing the existing notification subscription and handling on BE. So use of the data where it is needed is good.

File: packages/webapp/src/store/api/apiSlice.ts (hunk 3 above)
Comment (kathyavini) [reply packages/webapp/src/store/api/apiSlice.ts]: > I have noticed that on multiple occasions I go to the map and I don't have any sensors and I think it is because I skip the Home page somehow

@Duncan-Brain I saw that recently (only) and chalked that up to the sensor endpoint being down. Can you reproduce it somewhat consistently with the endpoint up?

Lifting it up the component tree is not the only option, there is a also a [prefetch](https://redux-toolkit.js.org/rtk-query/usage/prefetching#prefetching-without-hooks) command I could try in `locationSlice` or maybe even another `initiate()` like used in the saga. I don't like the query on `Home` very much conceptually.

**Comment by antsgar** [comment]
@Duncan-Brain actually no, if we go with the solution I described there'd be a cronjob generating the notification for IPs, the endpoint wouldn't have to do anything except fetch the list of IPs (so I undid you change in the ticket 😅)

**Review by antsgar — DISMISSED** [review]

File: packages/webapp/src/store/api/apiSlice.ts (hunk 1 above)
Comment (antsgar) [reply packages/webapp/src/store/api/apiSlice.ts:301]: Ohhh right! Okay this makes sense then, could you expand the comment a little bit and explain how this should be replaced once Tasks is migrated?

File: packages/webapp/src/store/api/apiSlice.ts (hunk 3 above)
Comment (Duncan-Brain) [reply packages/webapp/src/store/api/apiSlice.ts]: It was before the esci endpoint went down but no I do not know how to reproduce it -- thats why I havent bothered to bring it up. Fine for now I think but just seeing if you had expereinced it ever also.

File: packages/webapp/src/store/api/apiSlice.ts [hunk 7]
```diff
@@ -290,6 +294,22 @@ export const api = createApi({
...
     }),
+    getIrrigationPrescriptions: build.query<IrrigationPrescription[], void>({
+      query: () => `${irrigationPrescriptionsUrl}`,
+      async onQueryStarted(_id, { dispatch, queryFulfilled }) {
+        try {
+          // TODO: Once tasks is migrated to rtk use invalidatesTags instead of onQueryStarted'
+          dispatch({ type: 'getTasksSaga' });
+          await queryFulfilled;
+        } catch (err) {
+          dispatch({
```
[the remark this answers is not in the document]
antsgar: @kathyavini @Duncan-Brain oh I was wondering why I got a double snackbar sometimes, but totally missed this, so thank you so much for pointing it out!

So a couple of things:
- I actually don't think triggering a snackbar here is the way to go conceptually, even beyond this particular scenario. The snackbar is a presentational/visual change that shows up on the UI based on the error result returned by the call to the API -- triggering it from the component that consumes the result seems more appropriate and in line with the patterns shown in the RTK Query docs. Alternatively, RTK Query shows the use of middlewares for generic errors https://redux-toolkit.js.org/rtk-query/usage/error-handling#handling-errors-at-a-macro-level, so perhaps that's something we could implement on a global level to show the snackbar any time any query fails, if we wanted to.
- We shouldn't be re-fetching the IPs every time we switch tabs even if the query has failed -- we only need to fetch them once when the whole field view loads to figure out whether to display the tab or not. We also shouldn't really be rendering the Tabs on the component for each view (just like we don't have an instance of the navigation bar called on each single view of the app, we only have one). The composition of the React components is not right here, but I think fixing that would require significant refactoring, so not something we can tackle in this PR.
- I wonder if the problem could be solved by encapsulating both query calls (IPs + tasks) inside a `queryFn` parameter (https://redux-toolkit.js.org/rtk-query/api/createApi#queryfn) instead of putting the IPs one in `query` and the `tasks` one in `onQueryStarted`. If both of them go on `queryFn`, I assume we could wrap them both in a try/catch and return an error if any of them fail. I still think the actual display of the snackbar should be handled by the components if an error is returned.

packages/webapp/src/components/LocationIrrigation/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+              label={t(`TASK.STATUS.${data.task.status.toUpperCase()}`)}
+              taskId={data.task.task_id}
+            />
+          ) : (
+            <Cell kind={CellKind.PLAIN} text={'--'} />
+          ),
+      },
+      {
+        id: 'irrigationPrescriptionLink',
+        label: undefined,
```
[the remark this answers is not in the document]
Duncan-Brain: Thank you yes forgot that placeholder in there.

packages/webapp/src/components/LocationIrrigation/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+import clsx from 'clsx';
+import { Location } from '../../types';
+import { LocationTab } from '../../containers/LocationDetails/types';
+
+type Tab = {};
+
+type LocationIrrigationProps = {
+  location: Location;
+  history: History;
+  match: match;
```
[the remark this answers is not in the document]
Duncan-Brain: Thank you and the container!

packages/webapp/src/components/LocationIrrigation/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+import { History } from 'history';
+import { match } from 'react-router-dom';
+import { IrrigationPrescription } from '../../store/api/types';
+import Cell from '../Table/Cell';
+import { partnerEntities } from '../../containers/AddSensors/constants';
+import clsx from 'clsx';
+import { Location } from '../../types';
+import { LocationTab } from '../../containers/LocationDetails/types';
+
+type Tab = {};
```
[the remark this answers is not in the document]
Duncan-Brain: Woops, nope I think I just forgot it in thanks!

packages/webapp/src/components/Table/Cell/CellTypes/RightChevronLink.tsx
Duncan-Brain: Uhm not toooo hard...the onClick behaviour has to be reformatted to the Link way.  I originally copied the other list item type and specialized it so I would love to see the two list item types merged into one generic or at least match for that too.

packages/webapp/src/components/LocationIrrigation/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+        className: styles.tableCell,
+        format: (data) => (
+          <Cell kind={CellKind.RIGHT_CHEVRON_LINK} path={`/irrigation_prescription/${data.id}`} />
+        ),
+      },
+    ];
+  };
+
+  return (
+    <Layout className={layoutStyles.paperContainer}>
```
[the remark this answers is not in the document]
Duncan-Brain: Sure!

packages/webapp/src/store/api/types/index.ts
```diff
@@ -317,3 +317,10 @@ export interface SensorReadings {
   unit: SensorReadingTypeUnits;
   readings: SensorDatapoint[];
 }
+export interface IrrigationPrescription {
+  id: string;
+  some_location_id: string; // from Location 'location_id', or Pivot 'id'
```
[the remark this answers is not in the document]
Duncan-Brain: Thanks yes missed that mock placeholder.

packages/webapp/src/containers/LocationDetails/LocationIrrigation/useIrrigationPrescriptions.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  // TODO: refactor once mocked data is no longer needed
+  const irrigationPrescriptions = data ?? getMockData(location, tasks);
+
+  let filteredIrrigationPrescriptionsWithTask: LocationIrrigationPrescription[] = [];
+  if (location && location.grid_points) {
+    filteredIrrigationPrescriptionsWithTask = irrigationPrescriptions
+      .filter(
+        // return matching plans for this location
+        (plan: IrrigationPrescription) => plan.some_location_id === location.location_id,
```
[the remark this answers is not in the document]
Duncan-Brain: Yes thanks will just desructure

packages/webapp/src/containers/LocationDetails/LocationIrrigation/useIrrigationPrescriptions.ts
```diff
@@ -0,0 +1,78 @@
...
+      )
+      .map(({ task_id, ...irrigationPrescription }) => {
+        // map the plan to a new object that includes the full task object
```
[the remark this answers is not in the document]
Duncan-Brain: I think it just seemed easier to me. Maybe I also thought the hook might be reused with other IP views needing other task data or even other iterations of design of this same element. 

The field technology tab was actually annoying with map too because the location object was nested and passed multiple times so tracking down all instances and adding extra fields was a pain.

All this to say I could do it if you want!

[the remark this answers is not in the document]
kathyavini: This is probably the second most important change to me after the snackbars 😅 

If you inspect the `task` object you're re-creating by spread here, it's a huge nested thing which, in the current code (and I think everyone is in agreement refactoring the tab structure is not desired or feasible) you're passing back in every call to every tab.

Looking at the feature as designed now, I don't think any other components will ever consume this particular hook with the IP list. But even if they do, much better to be explicit about what's needed now (i.e. replace `...task` with   `task_id`) and let it grow if it has to later.

> The field technology tab was actually annoying with map too because the location object was nested and passed multiple times so tracking down all instances and adding extra fields was a pain.

Didn't quite follow this, but if you have to add extra fields (which again, with the current feature scope I don't think will happen), it would still just be in this one place.

packages/webapp/src/store/api/apiSlice.ts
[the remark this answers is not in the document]
kathyavini: Just from my perspective, 1-yes, 2-for me, with the snackbar gone I'm not concerned about the cache on the error. Not sure I understood @antsgar's third bullet point though so I'll let her respond to that 😂, 3-yeah, as you said, it's the pattern we have been doing so far with sensors. I also at one point wanted to return a `[]` for the same reason as you described above. But I think it's okay to add that if/when more partners exist.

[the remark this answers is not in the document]
antsgar: Returning [] from the API endpoint and only logging the ESci error seems like a good compromise at this point since we can't quickly solve the fact that we're querying Ensemble so many times. I think theoretically it would be fine to show a snackbar due to an Ensemble error, or to somehow let the user know that there's a problem rather than only showing "No IPs available" -- as a user, if I saw the empty state and was expecting some IPs to be there I'd be left wondering what's going on, while an error would be more transparent. But ultimately I think no snackbar at all is preferable to multiple snackbars -- especially since only a few users would actually be getting IPs, and if we leave it as is everyone would be getting multiple snackbars if ESci is down which seems really annoying.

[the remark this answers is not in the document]
Duncan-Brain: @antsgar If no IP's are present -- there is no router tab so it is not possible get to the empty state by clicking around (without exploiting a bug that is pre-existing to enter the URL directly) . 

Thats why it is fired on the other tabs. The first is to check for existence to determine if the tab should render, then for the the actual data on the tab (query is coded twice in that tab, but only actually queried once because of natural caching behaviour). 

This was Loic's design that new tabs going forward are not shown if no data exists.  If we always showed the tab, we would not have to check for existence and only query the data in the one place -- but then we would indeed need that "No IP's available state.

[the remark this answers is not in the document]
Duncan-Brain: Note the backend, if I haven't added it already, will include a check to see if FarmAddon exists, and return empty `[]` if it does not so that users wont see the error who do not have esci set up.

packages/webapp/src/containers/LocationDetails/LocationIrrigation/useIrrigationPrescriptions.ts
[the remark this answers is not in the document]
Duncan-Brain: done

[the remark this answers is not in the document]
kathyavini: Beautiful! The comment also needs updating.

packages/webapp/src/store/api/apiSlice.ts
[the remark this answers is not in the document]
Duncan-Brain: @antsgar @kathyavini  Okay I have added a console error. I might ask if we can push this through as it is blocking my other PR. 

For the multiple calls I think we can get a solution but I need to populate data object (successful response) in the query. So I would like to try it on the backend rather than build out the mocking in the apiSlice.

I think using some combination of the `keepUnusedDataFor` and/or `skip` option in the irrigation query in the `useLocationRouterTabs` hook. 

This is the example I am seeing that might do it but it required cached data which doesn't exist because it is erroring (error does not count as cached result I guess): 
```
const selectUserResult = api.endpoints.getUser.select(id);
const cached = useSelector(selectUserResult);
const shouldSkip = !!cached?.data;

const { data, error, isLoading } = useGetUserQuery(id, {
  skip: shouldSkip,
});
```

packages/webapp/src/components/LocationIrrigation/index.tsx
```diff
@@ -115,7 +109,7 @@ export default function PureLocationIrrigation({
   };
 
   return (
-    <Layout className={layoutStyles.paperContainer}>
+    <CardLayout className={layoutStyles.paperContainer}>
```
kathyavini, line 112: `<CardLayout />` doesn't need the classname -- `.paperContainer` is its base style so you're applying those styles twice here.

packages/webapp/src/store/api/apiSlice.ts
```diff
@@ -301,11 +301,10 @@ export const api = createApi({
...
           dispatch({ type: 'getTasksSaga' });
           await queryFulfilled;
-        } catch (err) {
-          dispatch({
-            type: 'snackbarReducer/enqueueErrorSnackbar',
-            payload: i18n.t('message:FARM_ADDON.IRRIGATION_PRESCRIPTION.FAILED_TO_GET'),
-          });
+        } catch (error: unknown) {
+          // getTasksSaga has its own try/catch block, this error handler will not catch that one
+          // @ts-expect-error - error type not definable
```
kathyavini, line 306: If you don't want to silence all type checking you could cast as an object that may or may not have error proprty:

```ts
  const err = error as { error?: unknown };

  console.error(
    'GET: Irrigation Prescriptions',
    err.error ?? err
  );
  ```

kathyavini (approved):

packages/webapp/src/store/api/apiSlice.ts
Duncan-Brain: I don't know if its considered type checking to assert a type either. I try to avoid 'as', but sometimes it is indeed useful.

The original error I see: https://stackoverflow.com/questions/69021040/why-catch-clause-variable-type-annotation-must-be-any

Seems like the proper way is a type guard but I did not want to make one for the `FetchBaseQueryError` rtk. I tried a few things but decided it wasn't worth the effort.

packages/webapp/src/components/LocationIrrigation/index.tsx
Duncan-Brain: 🙏

packages/webapp/src/store/api/apiSlice.ts
kathyavini: I'm not 100% sure, but I would think type casting is the lesser of two evils, since `@ts-expect-error` will silence all type checking across the board, which goes beyond what the casting is doing with that single property.

I also wouldn't write a type narrowing guard just for this.

kathyavini (approved):

packages/webapp/src/store/api/apiSlice.ts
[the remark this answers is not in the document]
antsgar: Don't worry about the extra calls for now, we can improve on that later on, it's okay! Re your comment, I totally forgot for a second that the tabs would actually be hidden so there'd be no empty state. That's okay, I think all the error handling flow will need a second iteration down the line.

antsgar (approved):

PR #3743, opened 2025-04-22
LF-4172 Create new Time component

kathyavini:
**Description**

The "Recommended Schedule" section of the [Irrigation Prescription view](https://www.figma.com/design/eMffC0h3JPlWtzNttxJpil/Litefarm-Updated?node-id=31428-78899&m=dev) has a new "time" input, alongside a matching "date" component

This PR:
- implements those two components
  - uses [MUI X components](https://mui.com/x/react-date-pickers/time-picker/) as had previously been discussed
  - styled to match our other `<Input />` in terms of colours, label, etc.
- fixes our MUI theme which had an incorrect use of `contrastText` (details below)
  - also added a style override to one component that needed adjusting (the base MUI component being used there was `AppBar`)
- includes the `<LocationalizationProvider />` and component logic needed to show dates in a form that should be identical to using the native HTML `type={date}`/`type={time}`
  - this is done by creating the actual native Intl.DateTime format and reverse engineering a format string from it
- Adds Storybook stories. Stories are _tremendously_ redundant, but I think we want to keep these as separate component exports, so I'm not sure there is a way around this (more than one component per Storybook story is maybe possible, but would be messy to implement)
  - [DateInput Story](http://localhost:6006/?path=/docs/components-form-dateinput--docs)
  - [TimeInput Story](http://localhost:6006/?path=/docs/components-form-timeinput--docs)

This PR does NOT:
- work out any React Hook Form integration. A new ticket was created for this. It is not necessary within the scope of Irrigation Prescription

Jira link: https://lite-farm.atlassian.net/browse/LF-4172

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/webapp/src/components/Navigation/TopMenu/TopMenu.jsx
```diff
@@ -266,7 +266,7 @@ const TopMenu = ({
 
   return (
     showNav && (
-      <AppBar position="sticky" className={styles.appBar}>
+      <AppBar position="sticky" className={styles.appBar} color="default">
```
kathyavini, line 269: If no colour is specified (and nothing is overridden in CSS), this creates a header of `palette.primary.contrastText` against `palette.primary.main`:

Screenshot 2025-04-17 at 4 23 47 PM

Since the background colour was overridden off of dark teal, this creates this situation, and might be how `contrastText` ended up as dark grey?

Screenshot 2025-04-17 at 3 58 36 PM


One solution would be to override on the text itself, but adding `color="default"` uses the primary text colour directly (against I think the default backgound? but we override background-colour anyway) so this is equivalent.

packages/webapp/src/components/Navigation/styles.module.scss
```diff
@@ -38,8 +38,10 @@
   }
 }
 
+// useSectionHeaders.tsx
 .text {
   @include xs-breakpoint {
     order: 2;
   }
+  color: var(--grey900);
```
kathyavini, line 46: Override color of top bar header text without using `color="default"` prop on AppBar (please see outdated comment in PR body 🙏)

packages/webapp/src/components/Inputs/DateTime/styles.module.scss
```diff
@@ -0,0 +1,53 @@
...
+ *  (at your option) any later version.
+ *
+ *  LiteFarm is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ */
+
+// styles taken from the <Input /> component
+// !important was necessary as the MUI-applied classes have quite high specificity
```
antsgar, line 17: We could either increase the specificity of the selectors here, or use MUI's `sx` prop to customize what we want to override -- important is almost never the answer unless truly as a last desperate resort 😅

antsgar (commented):
The pickers look so cool! ❤️ I don't remember if we talked about this already, but did we confirm with Loïc whether the goal is to eventually migrate other date pickers in the app to this new version?

Duncan-Brain (dismissed):
Looks great! 

I looked around for other mui components maybe impacted and couldnt find any. 

I also was interested in the label and for attribute moving forward -- its nice because I think it normally behaves to select the input when the label is clicked.

Since it doesn't work with forms yet should we have a `/readonly` folder or something along those lines until that gets addressed? Just thinking maybe it should be commented or highlighted somehow.

Just wondering too did you mean it does not need to be addressed in GET IP? it will be necessary for irrigation task right?

packages/webapp/src/components/Inputs/DateTime/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+  label,
+  ...labelProps
+}: DateTimePickerProps) => {
+  const [isDatePickerOpen, setIsDatePickerOpen] = useState(false);
+
+  const { dateFormat } = getLocaleDateTimeFormats();
+
+  return (
+    <div className={className}>
+      {label && <InputBaseLabel label={label} {...labelProps} />}
```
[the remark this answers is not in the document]
kathyavini: Unfortunately the so-called `<Label />` component here and throughout app is not an HTML `<label />` under the hood -- it's an `<h5>` so it can't take a `for`. I briefly looked into using a real `<label>` but wasn't sure it was the right time to refactor -- there are some incompatible props.

Do you think refactoring `<Label />` should be scoped to this PR?

packages/webapp/src/components/Inputs/DateTime/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+}: DateTimePickerProps) => {
+  const [isTimePickerOpen, setIsTimePickerOpen] = useState(false);
+
+  const { timeFormat, uses12HourFormat } = getLocaleDateTimeFormats();
+
+  return (
+    <div className={className}>
+      {label && <InputBaseLabel label={label} {...labelProps} />}
+      <MuiXTimePicker
+        defaultValue={moment(date)}
```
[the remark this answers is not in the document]
kathyavini: The MUI X component can use also take `date-fns` and `dayjs` adapters (and one other as well, I think). I just stuck with the library we already have installed, but should we migrate off of moment, should be easy to change.

Duncan-Brain (approved):
Cool, thanks I wasn't sure how it would work within the task flow and the task form due_date but it sounds like they are separate. 

And yea I get there is a new jira ticket for it. Just thinking since it may be a while before it is updated and if there is a chance without highlighting its limitations via folder or comments that someone could mistakenly use it before then.

packages/webapp/src/components/Inputs/DateTime/styles.module.scss
antsgar: It's okay! Much better than important, thanks for updating ☺️

packages/webapp/src/components/Inputs/DateTime/index.tsx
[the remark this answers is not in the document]
antsgar: Oh no 😞 that's pretty bad! I don't think that refactor should be scoped within this PR. Do you need everything that's included in the label component for these pickers for IPs though, or perhaps we could straight up use a `<label>` tag here which would eventually be the precursor to a new label component, instead of using `InputBaseLabel`?

antsgar (approved):

PR #3746, opened 2025-04-24
LF-4768 (3) Display IP Detail - IP Detail Container

kathyavini:
**Description**

This completes the irrigation prescription view by adding the container and situating the previous components (map view, table, time / date components) within it.

You can test the container in app and see the VRI view by supplying any fake IP primary key:

```
http://localhost:3000/irrigation_prescription/78734
```

To see the URI view though, specifically use this fake route:

```
http://localhost:3000/irrigation_prescription/uri_pk
```

IP Notes:
- mock data is partially pulled from Storybook (personally I'm a bit curious to know if the storybook folder is ending up in the beta build!) and partially defined right in the component... ultimately it will all come from the backend and the type should be moved to the RTK Query config
- @SayakaOno you probably have defined the fake weather data yourself elsewhere so please feel free to delete that part and configure as you need for your component 🙏
- I completely ignored the normal use of container and component by putting SCSS right into this file and calling only the inner map + table the "pure" component. If you're bothered by this though, let me know!

(Updated) Design Notes:
- `<Layout className={layoutStyles.paperContainer}>` has been replaced with a simpler `<CardLayout />` (without footer section) as discussed on Monday Apr 28 and applied to all the new views which had been using `layoutStyles.paperContainer`
-  these card styles have been matched up between animals and all the new views. They should all have the same border, border-shadow, padding, and border-radius now as confirmed with Loïc on April 24 and for the IP view on Tues Apr 29. They will all only apply past max-width of the mainContent column.
  - PageTitle header has been updated to the new style, which a grey chevron instead of a green one. I updated both PageTitle components (some Finance components use the v1 version). **THIS TITLE WILL GET ADDITIONAL DESIGN ADJUSTS** but probably not on this PR

Jira link: https://lite-farm.atlassian.net/browse/LF-4768

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/Layout/layout.module.scss
+++ b/packages/webapp/src/components/Layout/layout.module.scss
@@ -12,7 +12,7 @@
  *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
  *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
  */
- @import '@assets/mixin.scss';
+@import '@assets/mixin.scss';
 
 .container {
   width: 100%;
@@ -60,7 +60,7 @@
   display: flex;
   gap: var(--bentoGap);
   flex-wrap: wrap;
-  
+
   & > div {
     flex: 1 1 calc((100% - var(--bentoGap) * (var(--bentoColumns) - 1)) / var(--bentoColumns));
   }
@@ -80,9 +80,13 @@
 .paperContainer {
   background-color: white;
   margin: 16px 0 0 0;
-  padding: 16px 8px 8px 8px;
+  padding: 16px;
   border-radius: 4px;
 
-  /* simple-box-shadow */
+  border: 1px solid var(--Colors-Primary-Primary-teal-50);
   box-shadow: 0px 1px 0px 0px rgba(0, 0, 0, 0.05);
+
+  @include xs-breakpoint {
+    margin: 0;
+  }
 }
--- a/packages/webapp/src/components/Layout/layout.module.scss
+++ b/packages/webapp/src/components/Layout/layout.module.scss
@@ -79,14 +79,13 @@
 
 .paperContainer {
   background-color: white;
-  margin: 16px 0 0 0;
   padding: 16px;
-  border-radius: 4px;
 
-  border: 1px solid var(--Colors-Primary-Primary-teal-50);
-  box-shadow: 0px 1px 0px 0px rgba(0, 0, 0, 0.05);
-
-  @include xs-breakpoint {
-    margin: 0;
+  // card only past the width of mainColumn's max-content
+  @include lg-breakpoint {
+    margin: 16px auto;
+    border-radius: 4px;
+    border: 1px solid var(--Colors-Primary-Primary-teal-50);
+    box-shadow: 0px 1px 0px 0px rgba(0, 0, 0, 0.05);
   }
 }
--- a/packages/webapp/src/components/LocationDetailLayout/PureLocationDetailLayout.jsx
+++ b/packages/webapp/src/components/LocationDetailLayout/PureLocationDetailLayout.jsx
@@ -12,6 +12,7 @@ import useLocationRouterTabs from '../../containers/LocationDetails/useLocationR
 import { useSelector } from 'react-redux';
 import { locationByIdSelector } from '../../containers/locationSlice';
 import { Variant } from '../RouterTab/Tab';
+import Layout from '../Layout';
 import layoutStyles from '../Layout/layout.module.scss';
 
 export function PureLocationDetailLayout({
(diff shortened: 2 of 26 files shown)


packages/webapp/src/components/Layout/layout.module.scss
```diff
@@ -80,9 +80,13 @@
...
   border-radius: 4px;
 
-  /* simple-box-shadow */
+  border: 1px solid var(--Colors-Primary-Primary-teal-50);
   box-shadow: 0px 1px 0px 0px rgba(0, 0, 0, 0.05);
-}
\ No newline at end of file
+
+  @include xs-breakpoint {
+    margin: 0;
```
kathyavini, line 90: To avoid this situation on mobile view:

mobile-gap

packages/webapp/src/components/LocationDetailLayout/PureLocationDetailLayout.jsx
```diff
@@ -98,55 +99,45 @@ export function PureLocationDetailLayout({
   }, [locationCategory]);
 
   return (
-    <FormProvider {...formMethods}>
-      <Form
-        buttonGroup={
-          <LocationButtons
-            disabled={disabled}
+    <Layout className={layoutStyles.paperContainer}>
```
kathyavini, line 102: I really dislike the way that `<Layout />` is now serving double duty as both the standard "card container" component now AND the element that configures the inner layout of forms. This isn't the first case of a `<Layout />` wrapping a `<Form />` (there is one in edit (old) sensor view as well), but I did not want to undertake any refactor of Layout now.

With the Form-defined buttons in this view, I need to use `<Layout />` as a wrapper to get the card view border + box shadow to wrap the whole card (and not stop before the buttons) -- see image (1).

The original code solution (with the classes + `hasWhiteBackground` prop) actually had a box shadow in this location _that was covered up by a div_ (this is what the `hasWhiteBackground` prop does -- it covers up the border area with a div rendered on top, aahhhhh 🫨).

The result was that on integration there is no shadow at all on this "card" which is very visually disturbing (image 2):

Image 1: `layoutStyles.paperContainer` passed to form

inner-layout-issue

Image 2: current beta/integration (no visible shadow on card):
no-shadow-issue

As a side effect this code change will fix that focus border bus Denis showed us the other week.

antsgar: This seems like a pretty messed up styling situation that we've got ourselves into 😅 we'll need to get rid of that div with the linear gradient at the very least at some point. But beyond that, I think it's a bit off that we're wrapping a bunch of white cards inside a white card? I don't see that happening in any of the other views. When we talked about adapting this view to work with a green background, I thought we'd have the title and all of the "floating elements" be on top of the background. Come to think of it, in the screen shown in the browser, I think it's odd that the title of the page is inside the card as well (not odd per se, but odd in comparison with our other views). Is this something we've considered, or could it be covered in the next dev design meeting?

antsgar: Oh also the big card looks weird if you downsize the screen, and even weirder in mobile. There's this tiny bit of green background showing up at the bottom, and the card borders and border radius are visible. My recommendation would be to get rid of the outside card, if Loïc agrees with it!

antsgar (commented):
I left a comment about the card styling, but otherwise this is great! Good job mocking the data and thank you for providing really clear instructions for how to test it

packages/webapp/src/containers/IrrigationPrescription/index.tsx
```diff
@@ -0,0 +1,152 @@
...
+        cumultative_rainfall_unit: 'mm',
+        et_rate: 2,
+        et_rate_unit: 'mm/h',
+        weather_icon_code: '02d',
+      },
+    },
+    estimated_time: 2,
+    estimated_time_unit: 'h',
+    estimated_water_consumption: 100,
+    estimated_water_consumption_unit: 'l',
```
SayakaOno, line 90: These aren't included in the type — just to confirm, they'll be supplied by the backend, correct?

packages/webapp/src/components/LocationDetailLayout/PureLocationDetailLayout.jsx
kathyavini: @antsgar could you take a look at the IP view now? I haven't adjusted the other views or the `<Layout />` logic yet (the green on the bottom is the footer from that component so I'll have to change that up in the tab views) but I was hoping you could check this one first to see if this is the way we want to go. This is just a straight card, no card within cards, and it loses its card styling under max-width now.

I also tried the scheduling as a card and the table+map as a card, but since the view was designed against white on Figma, that just looked a bit off. I can whip up a screenshot in a moment for comparison when you talk to Loïc about it.

packages/webapp/src/containers/IrrigationPrescription/index.tsx
kathyavini: Oh! Thank you! Yes, supplied by the backend and sorry I confused myself by referring to the types file I was prepping for Morten. Yes, should definitely be in the frontend types, I'll fix that.

SayakaOno: Thank you!!
One more question, would it be safe to assume we'll always get `weather_forecast`? If so, I'll remove the `?` from the type!

kathyavini: Yes, should be safe!  Thank you for removing 🙏

SayakaOno (approved):

packages/webapp/src/components/LocationDetailLayout/PureLocationDetailLayout.jsx
antsgar: @kathyavini we reviewed it with Loïc today and we both liked the way it looks currently on this branch 👍 let's keep it that way! Only two minor observations, there's a green background on the "Based on this data" div that should be white, and I think it'd be good to make the font size consistent for the titles in the nav bar -- note the title in the Animals section has a much smaller font size

kathyavini: @antsgar let me check with Loïc about those two tomorrow.

As I wrote above ☝️, the problem with the animal headers is that they are the same size as the first subheading, which I think looks incorrect, especially at mobile.

The green background on that section is [accurate to Figma](https://www.figma.com/design/eMffC0h3JPlWtzNttxJpil/Litefarm-Updated?node-id=31428-78859&m=dev); did Loïc say he wants that changed? As much as I dislike the green now that there is padding (before the green was against the white background), I don't think it looks right to have that section running into the table header either 🤔

It feels like a lot of these things need to be designed from the start with the correct parameters like background, header, and not adjusted afterwards as they all kind of relate to each other 🫤 

Screenshot 2025-04-29 at 3 24 34 PM

Screenshot 2025-04-29 at 3 28 53 PM

antsgar: @kathyavini my guess is the green background on the div in Figma is just an oversight, but yeah you can definitely bring it up to Loïc tomorrow! Although personally I think the green is odd there and would just make the call to change it to white -- for small things like this I think it's 100% okay to make a judgment call as long as the screen overall looks as it was intended. Same with the heading, I agree with you it makes more sense for the font to be bigger, in that case I'd just adjust the one in Animals to get consistency ☺️

kathyavini: @antsgar it wasn't a reluctance to make a judgement call, so much as I tried the larger font on animals initially and didn't like it (in my opinion it ruined the intended layout of _that_ page) -- I think it gets down to when devs are starting to have different opinions about how something should look, it should go back to design. In this case, Loïc actually wanted the larger font gone app-wide, and will re-design the page headers appropriately.

In the case of the "Based on this data" and its background colour, Duncan's preference to have the whole section gone has won out!

kathyavini:
As discussed today Apr 30 in dev-design:

IP view:
- "Recommended schedule" and "Based on this data" have both been removed
- Page header with back button has been restored

Across map views (but ultimately to be the new pattern across app)
- Sensors, IP, and all location type details have been specified as belonging to "Map" -- the top menu title area has been adjusted accordingly
- Mobile title 'Map' is not required for any of these views
- Page level title has been retained _but will get a new design_ -- Loïc finds the 20px way too large

Loïc was saying that being able to interact would be helpful for making a final call on the headers, so I was hoping we could get this much to beta (follow up PR to adjust headers further) if it looks accurate to what we discussed today @SayakaOno @Duncan-Brain 🙏

SayakaOno (approved):
Looks great, thank you!

packages/webapp/src/components/LocationDetailLayout/PureLocationDetailLayout.jsx
antsgar: Yeah it's true, we don't always agree on which judgment call to make 😂 to be fair though I'm only expressing my opinion but I wouldn't block merging the PR because of any of these points, the card in card situation was the only one I felt more strongly about.

antsgar (approved):

PR #3752, opened 2025-05-01
LF-4770(2): Display flow to approve an IP and create a task from it

SayakaOno:
**Description**

- Refactor `FormNavigationButtons` to support custom button content ("< Back" and "Approve >")
   - [story](http://localhost:6006/?path=/story/components-formnavigationbuttons--with-custom-button-content)
- Add `FormNavigationButtons` to `IrrigationPrescription` container
- Add `management_plan_id` to `IrrigationPrescription` type
- Update `useApproveIrrigationPrescription` hook
   - Replace hard-coded `irrigationPrescription` with `irrigationPrescription` parameter
   - Adjust property names
- Update routes to pass `isCompactSideMenu` to `IrrigationPrescription`

Here’s the ticket for finalizing estimated water usage in the irrigation task: [LF-4810](https://lite-farm.atlassian.net/browse/LF-4810)

For testing, please run:
```
git fetch
git cherry-pick 8b7166441e073d91f9fe511f65bc291e04601687 --no-commit
```
This will add:
- Links to VRI and URI irrigation prescriptions on the home page (needed for the back and cancel buttons to work).
- A `location_id` on your farm (without it, a mocked one is used, which causes an error on the irrigation task creation form).

Jira link: https://lite-farm.atlassian.net/browse/LF-4770

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

[LF-4810]: https://lite-farm.atlassian.net/browse/LF-4810?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/webapp/src/components/Form/FormNavigationButtons/index.tsx
```diff
@@ -70,14 +74,15 @@ const FormNavigationButtons = ({
...
         >
-          {isFinalStep ? (
-            t('common:SAVE')
-          ) : (
-            <>
-              <span>{t('common:NEXT')}</span>
-              <ChevronRight />
-            </>
-          )}
+          {continueButtonContent ||
```
antsgar, line 77: In this case it's irrelevant because there's only one step, but for complete flexibility and to avoid confusion should we allow customization of both the "Next" button and the "Save" button, with separate "content" props?

antsgar (dismissed):
Works great, thank you for the detail on how to test! 🙌

kathyavini (commented):
Looks good, and I like the more flexible form buttons now!

I had two question that were mostly related to my not implementing my bit correctly...

1. For the margin I left on the IP container card for the form buttons, I used a value from Animals, not realizing that the additional text ("By approving you'll create an irrigation task") here in IP makes these form buttons taller. Could the margin be bumped up so there is no overlap with the bottom of the table?

Screenshot 2025-05-02 at 3 20 39 PM


2. My mocking solution (with fixed `location_id`) wasn't very good for the task creation part, sorry! Is there a reason we couldn't just commit your changes to the container in the cherry-pick commit, so that task creation succeeds on beta as well? I think it might help out with testing the controller and the downstream tickets (readonly task view) etc too. Maybe just passing the farm's `location_id` to the hook (to preserve the map)? I promise to remember to undo it when I move the mock to backend 😂 

```ts
  const onApprove = useApproveIrrigationPrescription(history, {
    ...irrigationPrescription,
    location_id: tempLocationId,
  });
 ```

SayakaOno:
I've addressed Joyce's comments (Thank you Joyce! ❤️ ), and I plan to open a follow-up PR to update `BACK` to `GO_BACK`.

I also added the condition to show the approve button, as discussed this morning.
I noticed that irrigation tasks were not being updated in the Redux store when switching farms, so I replaced `adapter.upsertMany` with `adapter.setAll`. 

@antsgar Would it be a good idea to apply this fix to all task types as well - maybe in a separate PR?

antsgar:
@SayakaOno yes, that sounds like a bug so let's do it for all tasks in a separate PR! Thank you for catching that ❤️

antsgar (approved):

kathyavini (dismissed):
Wait so our task-type specific entities were getting stale and more bloated on every farm change? 😳 And we didn't notice because we usually just use the `taskSelector` and not the specific ones? That's a bit nuts! Really good find.

antsgar (approved):

PR #3754, opened 2025-05-02
LF-4769 Modify task creation to accept IP id and update status on Ensemble

kathyavini:
**Description**

One decision we have to make with this part of the feature is how to relate the API call to Ensemble (to inform them that the IP is approved) to the main flow of LiteFarm task creation. As I understand it, the two main options are:

1. Block on it, such that if the request doesn't go through to Ensemble successfully, the task creation is rolled back and the frontend receives an API error
2. Treat it as a non-blocking side-effect, such that the task is created regardless of whether this goes through to Ensemble or not. In this case an error would only be logged (in our case, to Sentry) but the user would not be notified via UI

I have opted for (2) for the following reasons:
- From LiteFarm's perspective, the entire task creation flow is complete regardless of whether this update goes through; we have all the details we need to show the full task including the IP (only the id, and not the approval status, has any relevance for showing it on the task view)
- we can always double-check the approval status we get from Ensemble later against our own records (= the external id on the `irrigation_task` table)
- the way Morten described this, it sounded more like approval status was a nice-to-have and not a dependency in their model

For these reasons I have put the call in a new function called _after_ the success response is sent to the client. This is a new pattern and @antsgar I took a guess that this might be a services thing and not a controller thing and situated the new function in `/services` -- let me know if that was right though; I wasn't sure!

Jira link: https://lite-farm.atlassian.net/browse/LF-4769

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

Tested locally running `task.test.js` file, which is how I ended up with the 500ms pause to give the controller time to hit the mock... this might need adjusting on GitHub actions though.

Also tested with in app with the IP hook (initial version) and some data from the IP container.

**Checklist:**


packages/api/src/middleware/validation/checkTask.js
```diff
@@ -39,13 +39,8 @@ export function checkAbandonTask() {
...
       const { task_id } = req.params;
       const { user_id } = req.headers;
-      const {
-        abandonment_reason,
-        other_abandonment_reason,
-        happiness,
-        duration,
-        abandon_date,
-      } = req.body;
+      const { abandonment_reason, other_abandonment_reason, happiness, duration, abandon_date } =
```
kathyavini, line 42: Some linting got through in the end 😅 But should just be this line.

packages/api/src/services/taskActions.ts
```diff
@@ -0,0 +1,49 @@
...
+type TaskType = (typeof TASK_TYPES)[number];
+
+interface Task {
+  irrigation_task?: {
+    irrigation_prescription_external_id?: number;
+  };
+}
+
+/**
+ * Non‑blocking post‑response side effects; do not send HTTP responses here
```
antsgar, line 28: Since the method doesn't have access to the `res` object, it wouldn't be possible to send out an HTTP response, correct? I'd probably remove the second portion of the comment

packages/api/src/services/taskActions.ts
```diff
@@ -0,0 +1,49 @@
+/*
```
antsgar, line 1: I think we could keep this service more generic by calling it just `task`, eventually it could encapsulate a lot of the business logic currently living in the task controller

packages/api/src/services/taskActions.ts
```diff
@@ -0,0 +1,49 @@
...
+interface Task {
+  irrigation_task?: {
+    irrigation_prescription_external_id?: number;
+  };
+}
+
+/**
+ * Non‑blocking post‑response side effects; do not send HTTP responses here
+ */
+export async function triggerPostTaskCreatedActions(typeOfTask: TaskType, createdTask: Task) {
```
antsgar, line 30: Do you foresee us having to do similar "post-processing" for other task types? Otherwise I think the service function might be a bit overkill in this scenario -- the controller could call the Ensemble service directly, and the line where we fetch the `esciExternalId` could just be part of that service's `patchIrrigationPrescriptionApproval`. But if we do see this being reusable for other tasks then I could see value in this added layer.

packages/api/src/services/taskActions.ts
```diff
@@ -0,0 +1,49 @@
...
+  try {
+    switch (typeOfTask) {
+      case IRRIGATION_TASK:
+        {
+          const esciExternalId = createdTask.irrigation_task?.irrigation_prescription_external_id;
+
+          if (esciExternalId) {
+            await patchIrrigationPrescriptionApproval(esciExternalId);
```
antsgar, line 38: One thing to keep in mind here is that we're calling a service function from another service function -- nothing wrong about it, we just have to be careful not to call this one from the other function and create a circular dependency (probably unlikely due to what the purpose of these two functions in particular is).

antsgar (commented):
Looks great to me, and good call making the Ensemble patch non-blocking -- I agree with all of your points!

packages/api/src/services/taskActions.ts
kathyavini: I was afraid that down the line, someone might pass an additional parameter from controller into this function in order to do this 😅 Is it unlikely?

packages/api/src/services/taskActions.ts
kathyavini: I'm not sure if we would do it again for other task types or not. I think it would depend on how future partner integrations go?

In extracting it, I was more thinking that the `taskController` is already quite bloated (in fact, should be slimmed), and that it seemed like a distinct responsibility from what a controller is meant to handle? But I wasn't quite sure about that last part! It's still dealing with an HTTP request, just not with creating an HTTP response 🤔

packages/api/src/services/taskActions.ts
kathyavini: Yeah the `taskController` is really full of stuff 😬 I'll rename it for now, but it sounds like maybe for this purpose we will hold off on creating the services file instead?

packages/api/src/services/taskActions.ts
antsgar: Oh that makes sense! You never know so good call

antsgar (approved):

packages/api/src/util/ensembleService.ts
```diff
@@ -147,3 +147,28 @@ function selectCropData(managementPlans: ManagementPlan[]) {
...
+    const onError = (error: AxiosError) => {
+      const status = error.response?.status || 500;
+      const errorDetail = error.message ? `: ${error.message}` : '';
+      const message = `Error patching approval status with ESci${errorDetail}`;
+      throw customError(message, status);
+    };
+
+    await ensembleAPICall(axiosObject, onError);
+  } catch (error) {
+    console.log(error);
```
Duncan-Brain, line 171: do we need this console.log too?

Duncan-Brain (approved):
Looks great! Tests all pass for me. Left one small comment, feel free to merge if you think its not useful or can be done in another pr!

packages/api/src/util/ensembleService.ts
kathyavini: I think the outer should log (and thereby send to Sentry), but I don't think it hurts -- best case makes Sentry capture more likely?

PR #3760, opened 2025-05-06
LF-4818: Fix get task type tasks success reducers

SayakaOno:
**Description**

While working on https://github.com/LiteFarmOrg/LiteFarm/pull/3752, I noticed that when switching farms, task type tasks were not being updated in the Redux store as expected. The reducer used `upsertMany`, which merges tasks instead of replacing them. As a result, tasks from the previous farm were still present after switching.

This caused a bug when checking if any irrigation task had an IP - it was being evaluated against tasks from a different farm as well.

I fixed the `getIrrigationTasksSuccess` reducer in the PR above. This PR applies the same fix to the remaining task types.

Jira link: https://lite-farm.atlassian.net/browse/LF-4818

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/containers/Task/saga.js
+++ b/packages/webapp/src/containers/Task/saga.js
@@ -350,19 +350,20 @@ const taskTypeActionMap = {
   },
   PLANT_TASK: {
     success: (tasks) =>
-      call(getPlantingTasksAndPlantingManagementPlansSuccessSaga, { payload: tasks }),
+      call(getPlantingTasksAndPlantingManagementPlansSuccessSaga, { payload: { tasks } }),
     getAllSuccess: (tasks) =>
-      call(getPlantingTasksAndPlantingManagementPlansSuccessSaga, { payload: tasks, getAll: true }),
+      call(getPlantingTasksAndPlantingManagementPlansSuccessSaga, {
+        payload: { tasks, getAll: true },
+      }),
     fail: onLoadingPlantTaskFail,
     completeUrl: (id) => createBeforeCompleteTaskUrl(id),
   },
   TRANSPLANT_TASK: {
     success: (tasks) =>
-      call(getTransplantTasksAndPlantingManagementPlansSuccessSaga, { payload: tasks }),
+      call(getTransplantTasksAndPlantingManagementPlansSuccessSaga, { payload: { tasks } }),
     getAllSuccess: (tasks) =>
       call(getTransplantTasksAndPlantingManagementPlansSuccessSaga, {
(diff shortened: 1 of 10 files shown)


Duncan-Brain:
@SayakaOno any chance you could describe the steps to reproduce in a jira ticket? It sounds interesting

SayakaOno:
@Duncan-Brain Here you go! https://lite-farm.atlassian.net/browse/LF-4818
Let me know if you have any questions, thank you :)

antsgar (approved):

Duncan-Brain:
@SayakaOno 

When I look at the store it looks to me like the old redux pattern is to store multiple farms data side by side in the entities list. I am seeing almost all the entities having multiple farms. 

Our new RTK pattern seems to not be set up to filter that way. And I am guessing we have some slices that are in between too as bugs or some third/fourth pattern.

Just wondering if we should change the old behaviour if we are migrating it to rtk or update any new/broken selectors to the old pattern. This is the pattern I am seeing in crop variety for example (its probably done a million different ways lol):

```
export const cropVarietiesSelector = createSelector(
  [cropVarietyEntitiesSelector, loginSelector],
  (cropVarietyEntities, { farm_id }) => {
    return Object.values(cropVarietyEntities).filter(
      (cropVariety) => cropVariety.farm_id === farm_id,
    );
  },
);

```

SayakaOno:
@Duncan-Brain I don't see any issues with storing data only for the active farm, and I'm in favour of adopting the new pattern when migrating. (Now that I think about it, the old pattern was designed with offline support in mind...?)
Currently, I think only management plans and tasks follow the "new" pattern.

Duncan-Brain (changes requested):
@SayakaOno 

I took a second look and the `upsertMany` is important for cases where only 1 task or management plan is being added otherwise `setAll` replaces all tasks with 1 task. For example I am seeing a crash when adding a new task. 

I think because the `getTasksSuccessSaga` is used in createTask, createManagementPlan and repeatManagementPlan sagas that a crash would occur in each of these flows.

SayakaOno:
@Duncan-Brain Thank you so much for catching that! I was completely blindsided.

The issue should be resolved now, thank you for re-reviewing 🙏

Duncan-Brain:
No more crash 👍

Duncan-Brain (approved):

antsgar (approved):
Thank you for fixing!!

PR #3762, opened 2025-05-06
LF-4771 Create notification when new IP is available

kathyavini:
**Description**

This adds irrigation prescription notifications, run daily at 5am _summer time,_ otherwise 4am, to the other notifications (= tasks due today, and weekly unassigned tasks, if applicable) that are run daily by our jobs scheduler.

**The PR includes:**
- Slight refactor of scheduler to make it extensible
  - I haven't changed any of the core logic, but only wrapped it in a loop so that we can add our 5am job (and in the future, more times and jobs if desired). Previously 6am + these the two original notification types were hardcoded
- The new API endpoint (route + controller) that is called by scheduler
  - the timeNotification API endpoints are a bit unique in that they are designed to be called ONLY via scheduler and not from the webapp
  - they have their own authentication including distinct headers and JWT so I don't think they can be called by Postman. I will describe testing below
- API tests for the new endpoint
- mock functions in the `ensembleService.ts` file that are named with "mock" and can be deleted entirely when the GET endpoint from #3755 is complete
  - In the tests I have mocked axios itself and I don't believe any changes should be necessary to keep the tests passing once the endpoint is complete

### Notes on the existing Time Notification Controller methods:
In looking at this flow I have fixed one small bug in the Daily Task Due Notifications, and added a test case for it. That should address this Jira ticket:  https://lite-farm.atlassian.net/browse/LF-3673. Source of bug is indicated below.

I'm not ready to definitively say it's an error/bug, but I did want to remark that there are two different patterns already in place for the Weekly Unassigned Tasks and the Daily Task Due notifications regarding the `notification` and `nofication_user` tables/models:

- Weekly Unassigned Tasks: 1) creates one record in the `notification` table, and 2) notifies all admin users (creates _n_ records in the `notification_user` table) that correspond to that one notification
- Daily Task Due: 1) Creates one record in the `notication` table per record created in the `notification_user` table. This is beause the Daily Task Due loops through users and calls the `.notify()` method of the model once per user. At first I thought it was a bug, but it may also be intentional in the sense that these notifications (has tasks due) in particular are unique per user?

In any case, for IP there I have followed the Weekly Unassigned Tasks pattern. So one `notification` record is created, and _n_ `notification_user` records will map to that one notification. There is nothing user-specific about the irrigation prescription notification so this seemed correct.

Jira link: https://lite-farm.atlassian.net/browse/LF-4771

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

You can add the environment variable:

```
MOCK_TIMER=true
```

to `packages/api/.env` to use a very neat pre-existing testing functionality of the scheduler. Normally the scheduler runs on an hourly cronjob, but with that change, it will run every minute. This means if you run the scheduler in the background and wait at least 24 minutes, every farm you have will get its daily notifications.

Since the scheduler is shared with the export server (locally and in production); in dev use `docker compose up export` to run it. You can leave it running in the background while it makes its loop through all time zones.

If you're feeling wildly impatient though, search for this line:

```js
  driverQueue.add({}, { repeat: { cron: `${mockTimer ? '*' : '0'} * * * *` } });
```
and change it to

```js
  driverQueue.add({}, { repeat: { cron: `${mockTimer ? '* *' : '0'} * * * *` } });
```

to run the cronjob once every second. There will be a lot of console output though; be ready to stop it.

_**Actually I don't particularly recommend this**_ as it's easy to accidentally flood your bull queue and have the jobs still running for quite a while. I did have to manually clear my queue from Redis at one point after doing this.

**Checklist:**


packages/api/src/controllers/timeNotificationController.js
```diff
@@ -83,7 +85,7 @@ const timeNotificationController = {
         );
 
         if (hasTasksDueToday) {
-          await sendDailyDueTodayTaskNotification(farm_id, user_id);
+          await sendDailyDueTodayTaskNotification(farm_id, user_id, isDayLaterThanUtc);
```
kathyavini, line 88: I'm pretty sure is the source of the bug here: https://lite-farm.atlassian.net/browse/LF-3673

Or more generally, the fact that the task filtering on "Take me there" from the notification list is not correct for UTC+ farms (gave it a quick test on my Japan demo farm). There is logic based on this argument but it was never passed.

packages/api/src/middleware/validation/checkTask.js
```diff
@@ -304,11 +304,14 @@ async function checkIrrigationTask(req) {
...
   if (esciExternalId) {
+    const { farm_id } = req.headers;
     const existing = await TaskModel.query()
+      .whereNotDeleted()
       .joinRelated('irrigation_task')
+      .join('location', 'irrigation_task.location_id', 'location.location_id')
       .select('task.*')
       .where('irrigation_task.irrigation_prescription_external_id', esciExternalId)
-      .whereNotDeleted()
+      .andWhere('location.farm_id', farm_id)
```
kathyavini, line 314: This check wasn't originally limited by farm, and in production I don't think it would be possible for one IP to be generated for multiple farms, as they have a `location_id` property, but strictly speaking this does match the check the frontend makes to hide the buttons so maybe even beyond our testing phase it's fine to keep like this.

SayakaOno (dismissed):
Great job figuring this out -- I just looked at the code for the first time and learned a lot! 😵 From what I saw, it looks like the time a notification is sent depends on the time of year the farm was added, since `utc_offset` is set when the farm is created. Do you happen to know if that's right?

I hadn’t run the export server for a long time, and it took a bit of effort to get it working again 😢, but I successfully received notifications!

kathyavini:
>  From what I saw, it looks like the time a notification is sent depends on the time of year the farm was added, since utc_offset is set when the farm is created. Do you happen to know if that's right?

Ahhh timezones! 😭 I hadn't worried about it, but I do think we need to make an adjust.

The `utc_offset` value in the database is only storing the [rawOffset](https://developers.google.com/maps/documentation/timezone/requests-timezone#TimeZoneResponse) from the Google Timezone API, so it is not taking into account daylight savings; when the farm was created is therefore irrelevant, which I think is probably good. At least it is consistent. I think that what this means, though, is that the notification scheduler is off for those farms if/when the farms are on daylight savings. I wish it meant that the notification comes an hour early (5am instead of 6am instead in the summer) but instead I guess it would be an hour late (7am).

Since we never stored the `dstOffset` from the timezone API and we're not querying the timezone API again (and I'm pretty sure we don't want to do this for every single farm every single day), I might prefer the very low-effort solution of just pushing our notifications earlier by an hour? In the winter, or in non-daylight savings zones, they would come at 5am and 4am respectively, but an hour earlier for a start of day notification doesn't sound bad to me. What do you think @SayakaOno and @antsgar?

packages/api/src/controllers/timeNotificationController.js
```diff
@@ -95,13 +97,65 @@ const timeNotificationController = {
...
+  async postDailyNewIrrigationPrescriptions(req, res) {
+    const { farm_id } = req.params;
+    const { isDayLaterThanUtc } = req.body;
+    try {
+      // TODO: Use real function after LF-4765 is merged
+      const farmIrrigationPrescriptions = await mockGetFarmIrrigationPrescriptions(farm_id);
+
+      let notificationsSent = 0;
+
+      for (const prescription of farmIrrigationPrescriptions) {
```
antsgar, line 116: Should we just check the last IP here instead of looping through them? If this runs daily could we assume we'd have generated the notification for other prescriptions on previous runs -- and even if by some mistake we hadn't, the older IPs probably wouldn't be relevant anyway if there's a newer one?

packages/api/src/controllers/timeNotificationController.js
antsgar: Good catch! It seems that bug has been happening since forever 😱

packages/api/src/controllers/timeNotificationController.js
```diff
@@ -152,4 +207,38 @@ async function sendDailyDueTodayTaskNotification(farmId, userId, isDayLaterThanU
...
+ * @async
+ */
+async function sendDailyNewIrrigationPrescriptionNotification(
+  farmId,
+  userIds,
+  isDayLaterThanUtc,
+  irrigation_prescription_id,
+) {
+  const today = new Date();
+  if (isDayLaterThanUtc) today.setDate(today.getDate() + 1);
```
antsgar, line 224: NIT but I prefer to avoid inline ifs as they're a bit hard to read for me personally

antsgar (commented):
Awesome work! 🙌 it's really interesting to see a PR for a portion of the codebase that we almost never work on, really refreshing ☺️

packages/api/src/controllers/timeNotificationController.js
kathyavini: Okay, so the first time this runs for that farm, we want to generate just one notification? I guess I was expecting we would want two, so that every unseen IP in the table list has one notification corresponding to it.

But in this case (the first run) there would be just one notification for tomorrow's IP?

packages/api/src/controllers/timeNotificationController.js
kathyavini: I have the same preference! This was following the other two controllers, but I'll go ahead and change it in all three.

kathyavini:
@antsgar did you also have a thought to the daylight savings issue that Sayaka brought up and the proposed solution [above](https://github.com/LiteFarmOrg/LiteFarm/pull/3762#issuecomment-2867634885)?

Duncan-Brain:
Sorry for piping in.. but I saw farm dates +timezones being talked about! Which I also recently looked into thinking I had to do some parts ip notification fetching. 

My understanding is that @SayakaOno is correct. Depending on the time of the year a farm is created, the same farm could have two different utc_offset values. 

Also the Api call is non-blocking. Which means if the google api, or we hit our api limit at creation it could be `null`. (I was seeing errors locally a few years ago when my google API key was not set to work with the Timezones Api in google cloud console)

I think your solution @kathyavini is the best and fastest we have for now to make it work with what we have given the unreliable `utc_offset`. 

As an alternate, and to take a step towards proper farm dates, it might not be too too hard to populate the db farm table with the timezone name "America/Toronto" and use that. The proposed `dayjs` replacement for `moment` uses that: https://day.js.org/docs/en/timezone/timezone.

kathyavini:
> My understanding is that @SayakaOno is correct. Depending on the time of the year a farm is created, the same farm could have two different utc_offset values.

@Duncan-Brain on what basis are you proposing this? The function that populates `utc_offset` is `getUTCOffsetFromGridPoints()`. The return object from the timezone API looks like this:

```js
{
  dstOffset: 3600,
  rawOffset: -28800,
  status: "OK",
  timeZoneId: "America/Vancouver",
  timeZoneName: "Pacific Daylight Time",
}
```

And we're only storing the `rawOffset`. This IS the daylight-savings unaware timezone that corresponds to the timezone name you proposed above. `dstOffset` is the daylight savings adjustment. I linked the Google Timezone API docs about it in my comment above. Did you find something to suggest otherwise?

Duncan-Brain:
@kathyavini I have seen the docs yes, but it wasn't clear to me reading what rawOffset return reading it. I can read it your way now! It says:

> The offset from UTC (in seconds) for the given location. This does not take into effect daylight savings.

I was reading it as : `UTC.now()  - LocalTime.now()` ... dst was not determined. But I see what you are saying now that it reads. `UTC.now()  - LocalTimeUTC.now()`.. not adjusted for daylight savings. Sorry it was subtle to me! 

In any case it can be `null` in some cases! If it is not happening in prod then that is great!

Still moral of the story is I like your solution for now especially since Chile and Nova Scotia are in the same timezone but probably have opposite DST schedules. So no harm in doing it earlier to cover both cases.

packages/api/src/controllers/timeNotificationController.js
antsgar: I think they'll generate one IP each time our cronjob triggers, so if we trigger the cronjob for the first time today, they'll generate one IP for tomorrow and that's what we're getting on the first call here

antsgar:
@kathyavini sorry I missed answering to the timezones question! Lowest effort solution is okay here, I don't think the exact timing is as important as making sure that when they first login for the day they see the notification

kathyavini:
Okay, the timing for all notifications pushed back by one hour to land at our desired time if/when the farm is on summer time, per discussion 🙏  The cronjob will not need adjusting as it was already calculated for daylight time in Alberta (MDT).

And the controller + tests have been updated so only the last (of the two expected) IPs generates a notification.

SayakaOno (approved):
@kathyavini Thank you for looking into the documentation! The changes look great!

I was seeing these errors:
```
npm error code ENOTEMPTY
npm error syscall rename
npm error path /packages/api/node_modules/eslint-config-prettier
npm error dest /packages/api/node_modules/.eslint-config-prettier-VNN3FcXW
npm error errno -39
npm error ENOTEMPTY: directory not empty, rename '/packages/api/node_modules/eslint-config-prettier' -> '/packages/api/node_modules/.eslint-config-prettier-VNN3FcXW'
exited with code 217
```

And removing the volume fixed it!

antsgar (approved):

PR #3765, opened 2025-05-07
LF-4773: Cancel and back buttons are not working on the add sensors flow

SayakaOno:
**Description**

We were passing `isEditing` and `setIsEditing` to `<ContextForm />` to hide the form navigation buttons. However, when `setIsEditing` is passed, `ContextForm` skips calling `history.back()` on cancel (`onCancel`).

This PR adds a new `hideFormNavigationButtons` prop on `WithStepperProgressBar` to control button visibility.

Jira link: https://lite-farm.atlassian.net/browse/LF-4773

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/Form/ContextForm/WithStepperProgressBar.tsx
+++ b/packages/webapp/src/components/Form/ContextForm/WithStepperProgressBar.tsx
@@ -69,6 +69,11 @@ interface WithStepperProgressBarProps {
   onAfterSave?: (values: FieldValues) => void;
 }
 
+/**
+ * Only provide setIsEditing when the form needs to toggle to read-only mode (e.g., SingleAnimalView).
+ * If setIsEditing is provided, ContextForm will call setIsEditing(false) instead of history.back()
+ * when the user cancels or goes back from step 0
+ */
 export const WithStepperProgressBar = ({
   children,
   history,
--- a/packages/webapp/src/components/Form/ContextForm/WithStepperProgressBar.tsx
+++ b/packages/webapp/src/components/Form/ContextForm/WithStepperProgressBar.tsx
@@ -67,6 +67,7 @@ interface WithStepperProgressBarProps {
   showPreviousButton?: boolean;
   showLoading?: boolean;
   onAfterSave?: (values: FieldValues) => void;
+  hideFormNavigationButtons?: boolean;
 }
 
 export const WithStepperProgressBar = ({
@@ -95,6 +96,7 @@ export const WithStepperProgressBar = ({
   showPreviousButton = true,
   showLoading,
   onAfterSave,
+  hideFormNavigationButtons,
 }: WithStepperProgressBarProps) => {
   const [transition, setTransition] = useState<{ unblock?: () => void; retry?: () => void }>({
     unblock: undefined,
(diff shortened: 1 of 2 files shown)


kathyavini (approved):
Thank you so much  🙏 

> However, when setIsEditing is passed, ContextForm skips calling history.back() on cancel (onCancel)

Ack, despite having apparently [co-authored the commit](https://github.com/LiteFarmOrg/LiteFarm/pull/3489/commits/a93713c6eb29c58a55e30e67a6f67a4bcd9e0a4f) that implemented it (!) I did not remember that this relationship existed. Outside of the single animal form, it's not very intuitive, is it? 😓

For my own sake, I would probably enjoy a comment on the `setIsEditing` prop, but curious if you don't think it's necessary. I was of course the bug-causer, so I'm biased -- it could just be me 😅 

```ts
  /**
   * If provided, setIsEditing(false) is used instead of history.back()
   * when the user cancels or goes back from step 0
   */
```

SayakaOno:
@kathyavini 
Thank you! I find the props a bit confusing too, and I’d like to prevent them from being passed when the form isn’t like the single animal form.
How does this sound?
```
Only provide isEditing and setIsEditing when the form needs to toggle to read-only mode
(e.g., AddAnimals). If setIsEditing is provided, ContextForm will call setIsEditing(false)
instead of history.back() when the user cancels or goes back from step 0
```

kathyavini:
@SayakaOno okay excellent yes👌. I would have actually been tempted to use `isEditing` on its own without the setter actually so I think I needed that addition. I think it's just Single Animal View though, and not Add Animals, that has readonly/editing states.

Not for the comment, but then just to make sure I'm myself 100% clear on the difference between using `isEditing` and `hideFormNavigationButtons` (because in the context of the form itself, I think both booleans are only used to control `shouldShowFormNavigationButtons`):
- `isEditing` -- pass in a pair with its setter to toggle readonly/edit states. Passed to child components to act on that state (i.e. toggle readonly mode on inputs)
- `hideFormNavigationButtons` -- not passed to form children, just set from the parent, to... hide buttons. So the default choice when not following the Single Animal View pattern

antsgar:
Oh wow this took me a bit to follow as well 😂 so many different use cases for the form. Wouldn't it make sense though to only pass `isEditing` without the setter? To me that means, we're telling the form whether it's in editing or read-only mode so it knows whether to show the buttons or not, but by not passing we don't allow the form to toggle itself in or our of that state -- which makes sense, because in this case the state is controlled by the result of the addon query.

SayakaOno:
I guess I didn't fully understand `isEditing` and `setIsEditing`... Thank you for stopping me from adding the unnecessary prop Anto!
I removed `hideFormNavigationButtons`, and it should be good now!

Could you double check the change? Thank you so much 🙏

kathyavini (approved):
This is exactly what I thinking and I just assumed I was missing something about these props myself 😅

They really are confusing! But I'm glad we have reached consensus 😁

antsgar (approved):
Looks good! ❤️

PR #3773, opened 2025-05-16
LF-4829 Create soil sample endpoints and update location controller

kathyavini:
**Description**

This is the backend for the new location type. It is based entirely off of following the existing pattern for (= code search on) gate / water valve.

Because the meaningful changes in most of these files are so small, I didn't open a separate linting PR. I'll indicate below what is the meaningful change in the one file where the linter made a huge diff.

Jira link: https://lite-farm.atlassian.net/browse/LF-4829

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

Tested in app with the frontend components I'll open shortly.

**Checklist:**

### DIFF
--- a/packages/api/src/models/locationModel.js
+++ b/packages/api/src/models/locationModel.js
@@ -218,19 +218,6 @@ class Location extends baseModel {
           from: 'location.location_id',
           to: 'soil_sample_location.location_id',
         },
-        task: {
-          modelClass: taskModel,
-          relation: Model.ManyToManyRelation,
-          join: {
-            from: 'location.location_id',
-            through: {
-              modelClass: locationTasksModel,
-              from: 'location_tasks.task_id',
-              to: 'location_tasks.location_id',
-            },
-            to: 'task.task_id',
-          },
-        },
       },
       sensor: {
         modelClass: sensorModel,
--- a/packages/api/src/models/locationModel.js
+++ b/packages/api/src/models/locationModel.js
@@ -31,6 +31,7 @@ import surfaceWaterModel from './surfaceWaterModel.js';
 import naturalAreaModel from './naturalAreaModel.js';
 import residenceModel from './residenceModel.js';
 import waterValveModel from './waterValveModel.js';
+import soilSampleLocationModel from './soilSampleLocationModel.js';
 import taskModel from './taskModel.js';
 import locationTasksModel from './locationTasksModel.js';
 import sensorModel from './sensorModel.js';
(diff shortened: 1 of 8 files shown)


packages/api/tests/location.test.js
```diff
@@ -80,6 +83,7 @@ const assetSpecificMock = {
   buffer_zone: () => ({}),
   gate: () => ({}),
   water_valve: mocks.fakeWaterValve,
+  soil_sample_location: () => ({}),
```
kathyavini, line 86: Only changes to the test file was adding to the structures above, which will add the new location to the loop through all locations.

No test content was changed at all so the following is all linting 👇

packages/api/src/models/locationModel.js
```diff
@@ -210,6 +211,27 @@ class Location extends baseModel {
...
+          join: {
+            from: 'location.location_id',
+            through: {
+              modelClass: locationTasksModel,
+              from: 'location_tasks.task_id',
+              to: 'location_tasks.location_id',
+            },
+            to: 'task.task_id',
+          },
+        },
```
antsgar, line 233: I didn't think this relationship would be present, because I was thinking the primary location for the soil sample task would be the field or area, and there would be a second table to store the sample points linked to the task (see the `soil_sample_task_location` in the diagram). On second thought though I'm not so sure anymore, I guess we could just link the task to both types of locations using the same table/relationship, even if in the UI it's two separate steps?

I also wonder why `water_valve` is the only location that I can see that currently includes this relation.

kathyavini: Oh! I didn't even see that this was unique to `water_valve`, thank you! 🙏 Then I'm really unsure what the relationship mapping is for and I'll just remove it.

Even if it did have to get added later, thinking about it now, I think the appropriate time would be in the task user-story anyway. I think for me, it will be easier to make the decision about the table relationship for those two locations in the context of that whole flow.

kathyavini: I just started reading the ticket breakdown for the next story and I guess the decision point about the task associating tables is already upon us!

Yeah, I think in general (with readonly view being a great example) it's going to be a lot easier to just query in one place / table / entity for the task-location relationships.

SayakaOno (approved):
Looks good to me!

packages/api/src/models/locationModel.js
antsgar: Alright let's go with that and see if we hit any roadblocks!

PR #3774, opened 2025-05-16
LF-4830 UI flow to create soil sample location

kathyavini:
This is based to the backend branch: #3773 

**Description**

This completes frontend of the "Add" flow for the new location type, Soil Sample Location. Once the location is created it cannot yet be viewed because the LocationDetailsRoutes are part of the next ticket ("Edit" flow). I'll open that PR momentarily.

_Most important note about the frontend (this PR and next):_ 
- I have made a fairly deliberate decision to not change _any_ implementation for this new location compared to the current locations and particularly from gate/watercourse
  - this is more or less a verbatim copy-paste + find/replace on one or other of those locations!
- This includes sticking with slices + saga AND no TypeScript migrations at this time.

This was both a practical decision (since we're pressed for time with this feature) AND I think it's probably better maintainability in the long run if we don't have any one-off locations following their own patterns that would make future maintence / migration that much more convoluted. I'm not sure I adore the table architecture for it, but locations is a fairly stable and even relatively extensible part of the frontend when sticking to the core pattern so I have continued with it.

Jira link: https://lite-farm.atlassian.net/browse/LF-4830

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


SayakaOno (dismissed):
Worked great!

antsgar (dismissed):
Looks great!

My only question is if there's a reason we're disabling the inputs for lat long? I'm thinking it could be useful to fine tune the coordinates on that second step if they don't look exactly right after placing the point on the map manually.

antsgar (approved):

kathyavini:
@antsgar we already discussed this a bit today synchronously but I'll just make a note here for the PR:
- readonly inputs were given the okay by Loïc based on the assumption this flow is only for farmers, and he didn't believe they were going to be refining locations based on coordinates
- if we don't end up doing a CSV-based flow for researchers to input locations, we do want to work out the additional steps to 1) make the map responsive to changes applied to lat/lng properties in the following step of location creation; no existing pattern for this one, and 2) make sure the controller can handle updating the point's properties on edit of an existing location

A follow-up ticket will be made depending on which of the above (CSV flow or editable `lat`/`lng` we go with).

PR #3776, opened 2025-05-20
LF-4832 Add database migration to support new task type

kathyavini:
**Description**

This is the migration for the new task type. The PR:
- Adds the task type to the `task_type` table (this holds custom types as well so the generated id will depend on the db environment)
- Creates the table for the new task type
- Assumes the task locations will be associated via the `location_tasks` table (i.e. in the normal way) for both field + soil sample locations
- Uses an array for the sampling depths. This array is implemented in `jsonb` which is our codebase pattern for other tables with array datatypes (e.g. `grid_points` of area or `variables` of notification)
  - Update May 21st: this will end up being an array of _objects_ (indicating `upperDepth` and `lowerDepth` of a depth range) but from the perspective of the database, the datatype and checks are unchanged from the original `number[]`
- Sampling depths array has checks on being non-empty and having the correct length
- `sampling_depths_unit` is required (despite no dropdown) as this is where we store whether the user was in `in` or `cm` when they input the value

Jira link: https://lite-farm.atlassian.net/browse/LF-4832

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

Database inspected and inserted in using the model (will open PR for that momentarily) to check constraints.

**Checklist:**

### DIFF
--- a/packages/api/db/migration/20250520185926_create_soil_sample_task_table.js
+++ b/packages/api/db/migration/20250520185926_create_soil_sample_task_table.js
@@ -27,6 +27,11 @@ export const up = async function (knex) {
       .comment('Report document if uploaded');
     table.integer('samples_per_location').notNullable();
     table.jsonb('sample_depths').notNullable();
+    table.enu('sample_depths_unit', ['cm', 'in']).notNullable().defaultTo('cm');
+    table
+      .enu('sampling_tool', ['SOIL_PROBE', 'AUGER', 'SPADE'])
+      .notNullable()
+      .defaultTo('SOIL_PROBE');
 
     table.check('jsonb_array_length(??) > 0', ['sample_depths'], 'sample_depths_not_empty_check');
 
--- a/packages/api/db/migration/20250520185926_create_soil_sample_task_table.js
+++ b/packages/api/db/migration/20250520185926_create_soil_sample_task_table.js
@@ -20,11 +20,6 @@
 export const up = async function (knex) {
   await knex.schema.createTable('soil_sample_task', (table) => {
     table.integer('task_id').references('task_id').inTable('task').primary();
-    table
-      .uuid('document_id')
-      .references('document_id')
-      .inTable('document')
-      .comment('Report document if uploaded');
     table.integer('samples_per_location').notNullable();
     table.jsonb('sample_depths').notNullable();
     table.enu('sample_depths_unit', ['cm', 'in']).notNullable().defaultTo('cm');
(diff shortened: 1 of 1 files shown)


Duncan-Brain (dismissed):
Works as designed !... left some comments on the Jira ticket.

Also I forget but do we store `unit` even if there is only one option generally?

kathyavini:
Thanks so much for bringing up unit @Duncan-Brain 🙏 🙏

Yes we definitely have to store it!! Old sensors was a good example and `furrow_hole_depth_unit` in `soil_amendment_task` is an _exact_ parallel (only cm or in displayed, and db stores which of those two it was when the depth was added)

packages/api/db/migration/20250520185926_create_soil_sample_task_table.js
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+export const up = async function (knex) {
+  await knex.schema.createTable('soil_sample_task', (table) => {
+    table.integer('task_id').references('task_id').inTable('task').primary();
+    table
+      .uuid('document_id')
```
antsgar, line 24: I added a comment in the Jira doc -- I think I heard it should be possible to upload multiple docs to the same task, in which case we'll need a separate table instead of a column (still pending confirmation)

antsgar (approved):

antsgar (approved):

packages/api/db/migration/20250520185926_create_soil_sample_task_table.js
```diff
@@ -27,6 +27,11 @@ export const up = async function (knex) {
       .comment('Report document if uploaded');
     table.integer('samples_per_location').notNullable();
     table.jsonb('sample_depths').notNullable();
+    table.enu('sample_depths_unit', ['cm', 'in']).notNullable().defaultTo('cm');
+    table
```
Duncan-Brain, line 26: @kathyavini Sorry I forget Loic's response to this -- we are making this mandatory and the default is soil probe? Seems kinda strange to me to force this. Anyways approving for your discretion!

Duncan-Brain (approved):

packages/api/db/migration/20250520185926_create_soil_sample_task_table.js
kathyavini: Correct, this was what he said at dev-design last week. Since we aren't going to show an `null` corresponding option in the frontend, I'm okay with the db looking like this, but I'm also generally happy to be looser at the database level to let the model and frontend enforce. You think it should be nullable @Duncan-Brain?

PR #3777, opened 2025-05-20
LF-4833 Add endpoint to create a soil sampling task

kathyavini:
**Description**

These are the backend changes needed to create (and fetch -- this is just adding the type to the GET `.withGraphFetched`) the new task type. It can be tested in Postman with the following request:

POST `/task/soil_sample_task`

```js
{
    "task_type_id": 34, // check against your own DB
    "due_date": "2025-05-20",
    "locations": [ // pull from own DB
        {
            "location_id": "59142bf8-d9d0-11ef-9bfe-92760ca15ed7"
        },
        {
            "location_id": "a5ee4414-3118-11f0-b67b-acde48001122"
        }
    ],
    "managementPlans": [],
    "soil_sample_task": {
        "samples_per_location": 2,
        "sample_depths": [
            {
                "from": 0,
                "to": 15
            },
            {
                "from": 15,
                "to": 30
            }
        ],
        "sampling_tool": "AUGER"
    },
    "notes": "Notes here",
    "assignee_user_id": null,
    "override_hourly_wage": false,
    "wage_at_moment": null
}
```

Notes:
- I have used the keys `from` and `to` for the depth ranges
  - Although `lowerDepth` and `upperDepth` would be more descriptive of what we expect these two values to mean, in the UI there is no actual indication that would stipulate the user fills them out in this way -- just two values listed one after the other
  - Therefore `from` and `to` felt like the closest approximation of what the user is actually inputting
- No particular middleware or checks on this task type beyond model validation, and therefore no special test cases
  - For testing, the new task type has just been added to the normal "create a task of type {{type}}" test loop
- One linting and one eslint error fix were also required by the pre-commit hook; should be self-evident

Jira link: https://lite-farm.atlassian.net/browse/LF-4833

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

Checked the db + the API-returned list of tasks after submitting the POST above in Postman

**Checklist:**


antsgar (approved):
Looks good! 🙌 you're working through these tickets so fast!!

I'm curious about your thoughts about the advantage of storing an array of objects for the depth ranges instead of an array of arrays? My first thought was that we could have each range stored as something like [0, 10]. The object seems easier to understand, although I'm not sure if there'd be any benefit to the array to save up space when storing or if it'd be faster to query from.

SayakaOno (approved):
Looks good! Thanks for providing the request body!

PR #3778, opened 2025-05-21
LF-4802 Finance pages crash when selecting a custom date range and closing the component without clicking away

navdm:
**Description**

Fixes the crash that results from the click away listener not firing and updating the redux state properly. Since the whole react select is a child of ClickAwayListener, clicking on the react select itself while the dropdown is open doesn't trigger the callback.

Not documented in the jira ticket but the app also crashed by immediately navigating to another finances page right after opening the custom date picker (while its open). So the click away solution didn't feel right, it also lacked accessibility like not able to use the keyboard for closing / navigating properly.

**Solution**
- Uses the Popover component from MUI for the custom date picker.
- Updating the redux state for date range happens on popover close. 
- Fully accessible by default.

**Potential drawbacks**
- Differs from the original styling, like not taking the full width of the react select. This is because the popover renders in a portal as a child of document.body so it can't inherit the width of the react select. It will likely require some dynamic solution if necessary.
- Storybook tests are failing and will need updating.

Jira link: https://lite-farm.atlassian.net/browse/LF-4802

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/DateRangeSelector/styles.module.scss
+++ b/packages/webapp/src/components/DateRangeSelector/styles.module.scss
@@ -49,16 +49,10 @@
 
 // CustomDateRangeSelector
 .customDateRangeSelector {
-  position: absolute;
-  top: 100%;
-  left: 2px;
-  width: calc(100% - 4px);
   height: 405px;
   padding: 6px 15px 15px 15px;
   background: var(--white, #fff);
   border-radius: 4px;
-  box-shadow: 0px 4px 4px 0px rgba(0, 0, 0, 0.25);
-  z-index: 1000; // should be over the carrousel in Finances
 }
 
 .buttons {
--- a/packages/webapp/src/components/DateRangeSelector/styles.module.scss
+++ b/packages/webapp/src/components/DateRangeSelector/styles.module.scss
@@ -47,10 +47,15 @@
   }
 }
 
+.popover {
+  margin-top: 4px;
+}
+
 // CustomDateRangeSelector
 .customDateRangeSelector {
   height: 405px;
-  padding: 6px 15px 15px 15px;
+  width: 360px;
+  padding: 16px;
   background: var(--white, #fff);
   border-radius: 4px;
 }
--- a/packages/webapp/src/components/DateRangeSelector/DateRangeInput.jsx
+++ b/packages/webapp/src/components/DateRangeSelector/DateRangeInput.jsx
@@ -26,7 +26,7 @@ import styles from './styles.module.scss';
 import { Popover } from '@mui/material';
 
 export default function DateRangeInput({
-  optionValue,
+  defaultDateRangeOptionValue,
   defaultCustomDateRange = {},
   onChangeDateRangeOption,
   placeholder,
@@ -35,7 +35,7 @@ export default function DateRangeInput({
 }) {
   const [isMenuOpen, setIsMenuOpen] = useState(false);
   const [isCustomDatePickerOpen, setIsCustomDatePickerOpen] = useState(false);
-  const isCustomOptionSelected = optionValue === rangeOptions.CUSTOM;
+  const [isCustomOptionSelected, setIsCustomOptionSelected] = useState(false);
 
   const { t } = useTranslation();
   const selectRef = useRef(null);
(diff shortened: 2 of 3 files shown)


packages/webapp/src/components/DateRangeSelector/styles.module.scss
```diff
@@ -49,16 +49,10 @@
 
 // CustomDateRangeSelector
 .customDateRangeSelector {
-  position: absolute;
-  top: 100%;
-  left: 2px;
-  width: calc(100% - 4px);
```
kathyavini, line 55: For some reason, on my mobile the date inputs are registering as having almost no width when they are empty so the whole component collapses down too:

mobile-inputs

Could we add maybe a min-width of 360px? That would do double-duty and match up the width of the select and the popover on desktop, which is our primary use case, as well as preventing the squish on mobile.

kathyavini (commented):
Thanks so much @navDhammu!!

I like the Popover-based solution and I think the width getting a little off is a small price to pay for being more accessible ☺️ What has gone wrong with the Storybook stories, exactly?

I had two small comments about styling below, but my main thought is about the crash itself. I inadvertently discovered (in the process of looking at the styling) it's still possible to generate the crash by reloading the page with the popover open.

It's not such a normal use case that a user would be reloading the page at that time (although probably still worth handling), but it did make me take a second look at why we are even getting to that crashing line when it's behind a check for `isValid`:

```js
if (data.value === rangeOptions.CUSTOM) {
  if (isValid) {
    formattedOption = [customFromDate, customToDate]
      .map((date) => date.format('YYYY-MM-DD'))
 ```
 
Since the popover hasn't been closed (or in the original code, the onClickAway hasn't been triggered), the dates are undefined and `isValid` should be `false`.

I think the reason is this line:

```js
 const isValid = !isCustomOptionSelected || !!(areValidDates && isValidRange);
 ```
 in combination with
 
 ```js
   const [isCustomOptionSelected, setIsCustomOptionSelected] = useState(false);
  ```
  
This means when the page is reloaded (or a new component instance is loaded by navigating to a different page, like in your description comment), `isCustomOptionSelected` resets to false (making `isValid` true) even though `data.value === rangeOptions.CUSTOM` is still true in the Redux store, and that code block above runs.
 
I'm all for the onClickAway() to onClose() adjust, but I would also recommend fixing that logic so there is no way to get into that erroneous state; I think it would just be grabbing the redux state instead of defaulting to false every time:

```js
  const [isCustomOptionSelected, setIsCustomOptionSelected] = useState(
    defaultDateRangeOptionValue === rangeOptions.CUSTOM,
  );
 ```
 
 Does that look right to you?

navDhammu:
@kathyavini Good catch on that bug! And you’re right that `isValid` is incorrectly being set to true even though the dates are undefined. 

For the fix - initializing `isCustomOptionSelected` to match the value in redux store the way you described looks correct, but I couldn't identify the need for even having `isCustomOptionSelected` as local state in the first place - it just duplicates the same value in redux store and local state. So instead I used the option value in the redux store to determine if the 'custom option' is selected, making the react select value controlled and removing the redundant local state. But please let me know if there was some requirement or edge case I might have missed. 

I also updated the styling and width to 360px, I used width over min-width because min-width was causing a bit of layout shift when the error message shows up.

Regarding the storybook tests - I didn't get a chance to look carefully as to why they're failing, but I think some other components tests were also failing so maybe better to have a separate ticket that identify can fix all the tests?

kathyavini:
@navDhammu I'm sorry for the slow response!

For now, could you keep the component uncontrolled? If I log `optionValue` on the latest version of the branch and select the custom option, I see the value jumping between `custom_range` and `year_to_date` -- it's something about the timing of the timing of `useEffect()` in conjunction with the initial invalid Redux state. No issues in the sensor readings which use component state, and no issues once valid date ranges have been selected. The small bug I see from that is that when you then open the report modal, the dates are set to 'Year to Date' instead of the selected range.

I trust it's probably not that hard to fix, but since we have a release coming up very soon, I would prefer to be conservative with the component right now. If you're still interested in this improvement though, a new ticket + PR to specifically to make `DateRangeInput` controlled would be super welcome, and we could revisit it post-release without the pressure of the release deadline 🙏

navDhammu:
@kathyavini I reverted it back to uncontrolled and defaulted `isCustomOptionSelected` to match the value in redux store.

kathyavini (approved):
Thanks so much @navDhammu!

PR #3785, opened 2025-05-29
return the response early and guarantee a response

Duncan-Brain:
**Description**

Saw this while working on documents - Small change to guarantee a response from the getTasksByFarmId endpoint.

Hopefully avoids endpoint hangups, open test handles.

Jira link: none

**Type of change**

**How Has This Been Tested?**

- [x] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/taskController.js
+++ b/packages/api/src/controllers/taskController.js
@@ -910,6 +910,10 @@ const taskController = {
           ]`,
         )
         .whereIn('task_id', taskIds);
+
+      if (!graphTasks.length) {
+        return res.status(200).send([]);
+      }
       const filteredTasks = graphTasks.map(removeNullTypes);
 
       /* Clean before returning to frontend */
@@ -925,9 +929,7 @@ const taskController = {
         task.animal_batches?.forEach(flattenInternalIdentifier);
       });
 
-      if (graphTasks) {
-        res.status(200).send(filteredTasks);
-      }
+      return res.status(200).send(filteredTasks);
     } catch (error) {
       console.log(error);
       return res.status(400).send({ error });


SayakaOno (commented):
I couldn’t spot anything problematic in the original code. What was the problem you were seeing?

packages/api/src/controllers/taskController.js
```diff
@@ -925,9 +929,7 @@ const taskController = {
         task.animal_batches?.forEach(flattenInternalIdentifier);
       });
 
-      if (graphTasks) {
```
Duncan-Brain, line 928: @SayakaOno 

if `!graphTasks` . What happens on frontend and backend?

I think maybe this happens:
frontend
- get request times out and does not receive a response => error.

backend
- nothing bad really we just run an extra query that _could_ fail (also unlikely) even though we know the result is `[]`

I agree it seems very unlikely to `!graphTasks` but then why put it in the conditional at all. My understanding it is bad not to return _something_ and important to make a get response reachable if all code prior is "successful".

This has no implications that I can see as the code is currently.

SayakaOno: When `graphTasks` is undefined in the original code, I think `graphTasks.map()` will fail and an error is thrown, so maybe it doesn't even reach `if (graphTasks)`...🤔?

I think `if (!graphTasks.length) {` needs `?` not to return an error.

Duncan-Brain: yes, presumably graphTasks.length will fail almost the same as graphTasks.map - I left everything the same except the earlier return and guaranteed response from get endpoint.

We an close if you don't think it remains essentially the same which was my goal. Its just a code cleanliness thing I saw..

SayakaOno: I'm still not sure if there's actually a concern about not getting a response with the original code, but I love early returns!

PR #3789, opened 2025-06-02
LF-4774 Remove mocked sensor data; set up valid defaults; check against real sensor data

kathyavini:
**Description**

This PR handles the changes necessary to get sensors 1) cleaned up for release, 2) compatible with the real sensors data now in the ground.

Backend changes:
- remove the location mock and make it a fallback default instead
  - i.e. sensors + profiles will be shown at farm centre if a location is not provided
- new Ensemble API call to the `profiles/` endpoint shared in Slack
  - this is now used to fetch sensor arrays rather than generating them from the individual sensors
  - (note: profile location has been provided right on that endpoint, and `profile_id` never ended up on the device endpoint!)
- cleanup to match properties to their actual keys; remove unused key `sensor_readings` which was no longer being returned anyway
- return the descriptive names Ensemble has set up on the sensors + profiles 

Frontend changes (minimal):
- harmonize types with key changes
- use the descriptive names in the ESci sensor list + readings view

Jira link: https://lite-farm.atlassian.net/browse/LF-4774

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

I'll message in Slack with the org_id that will return both profiles and individual sensors.

**Checklist:**


packages/api/src/util/ensemble.js
```diff
@@ -164,89 +173,38 @@ const mapDeviceToSensor = (device) => {
...
-  for (const sensor of sensors) {
-    if (Math.abs(sensor.latest_position.depth) < Math.abs(selectedSensor.latest_position.depth)) {
-      selectedSensor = sensor;
-    }
-  }
-
-  return {
-    lat: selectedSensor.latest_position.coordinates.lat,
-    lng: selectedSensor.latest_position.coordinates.lng,
+const enrichDeviceWithDefaultPosition = (device, grid_points) => {
```
Duncan-Brain, line 186: For the sake of moving sensors through faster I am looking at draft PRs.

It seems like a UI point to decide to place sensors with no current position at the farm center coordinate.

It reminds me of IP in the sense of "is it core data"? or is it a definition of an "offline" sensor? Wondering also if instead of overwriting these values we should even keep the `null` or whatever value and define it on the frontend like:

```
const hasCurrentPosition = device.latest_position ?  true : false;
const sensorPositionOnMap = hasCurrentPosition ? device.latest_position : farmCenterCoordinates;
```

kathyavini: Update from dev-design discussion:

1. Farm centre coordinates got the okay from Loïc as the default
2. The hope is to do as little re-architecting of the frontend as possible at this point (right before release) so I'm not really in favour of moving the defaults now, but in general I think that would be an architectural improvement + I like how it opens up the possibility of someday flagging the missing location property on the UI. Do you want to create a nice-to-have ticket for that @Duncan-Brain to record your thinking on that?

Duncan-Brain: Sure! Here is the ticket if you feel you want to add it to a comment: https://lite-farm.atlassian.net/browse/LF-4847

packages/api/src/util/ensemble.js
```diff
@@ -145,108 +156,71 @@ const mapDeviceToSensor = (device) => {
...
     last_seen: device.last_seen,
     point: {
-      lat: device.latest_position.coordinates.lat,
-      lng: device.latest_position.coordinates.lng,
+      lat: Number(device.latest_position.latitude),
+      lng: Number(device.latest_position.longitude),
     },
-    depth: device.latest_position.depth,
-    depth_unit: 'cm', // to be confirmed
+    depth: device.latest_position.vertical_position,
```
kathyavini, line 132: Every real device I've seen so far has a `sensor_environment` of `soil`, with a positive value here, or `sensor_environment = surface`, with a `null` vertical_position (which is fine to pass on to the frontend; it will display as a depth 0cm).

I believe from previous conversations with Morten that there should be a third environment (something like `air`?) that requires flipping the sign of this `vertical_position` value, but I haven't seen it. I'll check with Morten next time we meet, but I don't think it is going to be relevant to this project so I'm okay with not blocking on that confirmation.

packages/webapp/src/containers/SensorList/useGroupedSensors.ts
```diff
@@ -102,17 +104,14 @@ const formatSensorToGroup = (
...
-      if (!(name in acc)) {
-        acc[name] = 0;
-      }
-      acc[name] += 1;
-
-      return acc;
-    },
-    {} as Record<Sensor['name'], number>,
-  );
+  const summaryMap = sensors.reduce((acc, { name }) => {
```
kathyavini, line 107: Pre-commit lint; no change

packages/api/src/util/ensemble.js
```diff
@@ -99,41 +99,52 @@ const getEnsembleSensors = async (farm_id) => {
...
 
   const farm = await FarmModel.query().findById(farm_id);
   const farmCenterCoordinates = farm.grid_points;
 
+  const sensorArrays = [];
+  for (const system of orgSystems) {
+    const systemProfiles = await getOrganisationProfiles(farmEnsembleAddon.org_pk, system.pk);
+
+    const mappedProfiles = systemProfiles
+      .filter((profile) => profile.water_profile?.sensors)
```
SayakaOno: Should we check `length` just in case??

SayakaOno (approved):
Looks great! I left a comment, but if there's no issue, feel free to merge!

packages/api/src/util/ensemble.js
kathyavini: Yes, definitely needed; thank you! 🙏

Duncan-Brain (approved):

PR #3792, opened 2025-06-03
Lf 4836 (1) allow associating result document s with task on complete

Duncan-Brain:
**Description**

This is the first half and backend of the linking tasks to documents ticket.

Notes:
- entire tasks object should return with the task to bypass worker restrictions on viewing documents.
- not all create task endpoints use the checkTask middleware - requiring some duplication in the controller.
- does not assume number of documents provided
- documents are not deleted if the task is deleted (but I don't think you can delete a completed task anyways)
- documents can be deleted separately in the documents section, the link won't be broken but it will be filtered out on the task
- tests: did not add here - is it valuable? maybe more so if we set soil sample result properties on the backend?

Unresolved and adding to next ticket
- frontend UI
- ~~setting soil sample results document properties on frontend or backend? Specifically farm_id is required also.~~ Adding `farm_id` on backend
- ~~how many documents are created - one document per file, OR one document, multiple files -- leaning towards one document multiple files~~  multiple documents preferred as discussed in dev design

Jira link: [LF-4836](https://lite-farm.atlassian.net/browse/LF-4836)

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

[LF-4836]: https://lite-farm.atlassian.net/browse/LF-4836?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/api/src/middleware/validation/checkDocument.ts
```diff
@@ -0,0 +1,55 @@
+import { NextFunction, Request, Response } from 'express';
+import { customError, LiteFarmCustomError } from '../../util/customErrors.js';
+import { DocumentWithFiles } from '../../models/types.js';
+
+export function validateFilesLength(document: DocumentWithFiles) {
```
Duncan-Brain, line 20: Wanted to reuse this for tasks so made the middleware.

packages/api/src/models/taskModel.js
```diff
@@ -318,6 +341,8 @@ class TaskModel extends BaseModel {
       animal_movement_task: 'omit',
       managementPlans: 'omit',
       locations: 'edit',
+      documents: 'omit',
+      task_documents: 'omit',
```
Duncan-Brain: copy crop plan - this assumes the documents are not copied between copied tasks, this may only make sense for soil sample results.

packages/api/src/util/customErrors.ts
```diff
@@ -15,35 +15,58 @@
...
-  error.type = 'LiteFarmCustom';
-  return error;
+interface LiteFarmError extends Error {
+  type: 'LiteFarmCustom';
+  code: number;
+  body?: Record<string, unknown>;
+}
+
+// Extends the error class
+export class LiteFarmCustomError extends Error implements LiteFarmError {
```
Duncan-Brain, line 25: Typescript really wanted me to extend the class in order to type properly, that why this has changed its form. Tests all pass though so it does not affect anything else.

packages/api/db/migration/20250523133726_create_task_document_join_table.js
```diff
@@ -0,0 +1,35 @@
...
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ */
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+export const up = async function (knex) {
+  await knex.schema.createTable('task_document', (table) => {
```
Duncan-Brain, line 21: I technically changed the db table name requested. 

I am happy to follow our convention. But since it leans towards more-task-like than more-document-like. I think `parent_child` makes more sense when working in the db viewer because alphabetically you don't have to search around as much. 

`task_animals...` foloows this convention too. `location_tasks` may be the only one

Let me know if you hate!

kathyavini: What I hate is `location_tasks` (I DO keep searching in the wrong place in the table list for it 😂) but I guess in a sense it's acceptable because we have a view for "this location's tasks" in app. I wouldn't protest a rename either, though!

Here since it's definitely "this task's documents" (and no reverse view), I like what you picked 👍

packages/api/src/controllers/taskController.js
```diff
@@ -212,6 +216,28 @@ async function updateTaskWithCompletedData(
       return task;
     }
 
+    case 'soil_sample_task': {
+      const noInsert = [
+        ...nonModifiable.filter((asset) => !['documents', 'taskDocuments'].includes(asset)),
+        'soil_sample_task',
+      ];
+      console.log(data);
```
SayakaOno: `console.log`! 😁

packages/api/src/controllers/taskController.js
```diff
@@ -212,6 +216,28 @@ async function updateTaskWithCompletedData(
...
+      const task = await TaskModel.query(trx)
+        .context({ user_id })
+        .upsertGraph(
+          { task_id, ...data, ...wagePatchData },
+          {
+            noUpdate: nonModifiable,
+            noDelete: true,
+            noInsert,
+            relate: ['animals', 'animal_batches'],
+            unrelate: ['animals', 'animal_batches'],
```
SayakaOno: Are these correct??

packages/api/src/models/taskModel.js
```diff
@@ -247,6 +249,27 @@ class TaskModel extends BaseModel {
...
+        },
+      },
+      taskDocuments: {
+        modelClass: TaskDocumentModel,
+        relation: Model.HasManyRelation,
+        join: {
+          from: 'task.task_id',
+          to: 'task_document.task_id',
+        },
+      },
```
SayakaOno: This seems redundant to me... Wouldn't it work without this?

packages/api/src/models/taskModel.js
```diff
@@ -247,6 +249,27 @@ class TaskModel extends BaseModel {
           to: 'location_tasks.task_id',
         },
       },
+      documents: {
+        modelClass: DocumentModel,
+        relation: Model.ManyToManyRelation,
```
SayakaOno, line 254: `ManyToManyRelation` means multiple tasks can have the same document?

packages/api/src/controllers/taskController.js
Duncan-Brain: 😮  haha thanks

packages/api/src/controllers/taskController.js
Duncan-Brain: My logic was this: the default case use these relations -- even though most tasks do not relate to animals. `soil_amendment_task` is the only one that does not relate them. 

I went with what most tasks use to avoid problems haha. But maybe the frontend does not expect anything and I can remove. Do you think I should remove?

packages/api/src/models/taskModel.js
Duncan-Brain: Yes it should work without it. 
- Originally I was not going to return the full document but I forgot about worker permissions.
- locations_tasks did it so I was hoping to avoid problems again

I think I will remove it!

packages/api/src/models/taskModel.js
Duncan-Brain: Yes, I saw no reason to restrict it at this time, even though it cannot happen in this one way upload flow.

packages/api/src/controllers/taskController.js
```diff
@@ -212,6 +216,27 @@ async function updateTaskWithCompletedData(
       return task;
     }
 
+    case 'soil_sample_task': {
+      const noInsert = [
+        ...nonModifiable.filter((asset) => !['documents', 'taskDocuments'].includes(asset)),
```
SayakaOno: I think we can remove `taskDocuments` here too?

packages/api/src/controllers/taskController.js
SayakaOno: I thought you had copied them from another task type and just forgot to remove them 😂
I vote for removing them, it should be safe, and right now it kind of makes it look like a soil sample task can have animals. Or maybe it would be less confusing if `soil_amendment_task` had them too!

packages/api/src/controllers/taskController.js
```diff
@@ -1008,9 +1074,14 @@ const taskController = {
...
-  return ['createdByUser', 'updatedByUser', 'location', 'management_plan'].concat(
-    nonModifiableAssets,
-  );
+  return [
+    'createdByUser',
+    'updatedByUser',
+    'location',
+    'management_plan',
+    'documents',
+    'taskDocuments',
```
SayakaOno: Could you double check 'taskDocuments' here too?

packages/api/src/middleware/validation/checkDocument.ts
```diff
@@ -0,0 +1,70 @@
...
+  if (!Array.isArray(document.files) || document.files?.length <= 0) {
+    throw customError('Can not create document without file links');
+  }
+}
+
+export function checkCreateDocument() {
```
SayakaOno, line 26: Did you duplicate this function in case it needs to be different from the other one later?

packages/api/src/middleware/validation/checkTask.js
```diff
@@ -179,6 +182,8 @@ export function checkCreateTask(taskType) {
         return res.status(400).send('task type requires products');
       }
 
+      checkCreateTaskDocument(req);
```
SayakaOno: Should this be `checkCreateTaskDocument(req.body);`?

packages/api/src/middleware/validation/checkDocument.ts
```diff
@@ -0,0 +1,70 @@
...
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ */
+
+import { NextFunction, Request, Response } from 'express';
+import { customError, LiteFarmCustomError } from '../../util/customErrors.js';
+import { DocumentWithFiles } from '../../models/types.js';
+
+export function validateFilesLength(document: DocumentWithFiles) {
+  if (!Array.isArray(document.files) || document.files?.length <= 0) {
```
kathyavini: NIT: works fine, but reads strange since lengths can't be negative. Check for empty is usually `document.files?.length === 0` or `!document.files?.length`

packages/api/src/middleware/validation/checkTask.js
```diff
@@ -319,3 +324,27 @@ async function checkIrrigationTask(req) {
...
+
+export function checkCreateTaskDocument(task) {
+  if ('documents' in task && task.documents.length) {
+    throw customError('Documents not permitted on task creation');
+  }
+}
+
+export function checkCompleteTaskDocument(task, taskType) {
+  if ('documents' in task && task.documents.length) {
+    if ([...TASKS_WITH_DOCUMENTS].includes(taskType)) {
```
kathyavini:
```suggestion
    if (TASKS_WITH_DOCUMENTS.includes(taskType)) {
```
No need to create a new array here

SayakaOno (commented):
Sorry this took a while...!

What would be the expected request body?
Since we're just associating an existing document that already has files, I don’t think we need to check the file length when completing a task.

The request below worked for me after adding "documents" to `relate` in `upsertGraph`:
```
{
    "duration": 15,
    "happiness": null,
    "completion_notes": "",
    "complete_date": "2025-06-10",
    "documents": [
        { "document_id": "632d47ec-6883-11ee-868f-e66db4bef552" }
    ]
}
```
Do we need to include files here?

packages/api/src/middleware/validation/checkTask.js
```diff
@@ -319,3 +324,27 @@ async function checkIrrigationTask(req) {
...
+      checkTaskDocuments(task.documents);
+    } else {
+      throw customError('Documents not permitted on this task type');
+    }
+  }
+}
+
+function checkTaskDocuments(documents) {
+  if (documents.length) {
```
kathyavini: No need to check length a second time; it was already checked just above before calling `checkTaskDocuments`

```js
if ('documents' in task && task.documents.length)`
```

packages/api/src/controllers/taskController.js
Duncan-Brain: Ah silly me, I thought there was more but I searched vscode for task_documents haha 🙏

packages/api/src/middleware/validation/checkDocument.ts
Duncan-Brain: yes

packages/api/src/middleware/validation/checkTask.js
Duncan-Brain: Ya... woops! I should've made a test I guess haha

Duncan-Brain: fixed

packages/api/src/controllers/taskController.js
Duncan-Brain: 🙏

packages/api/src/controllers/taskController.js
Duncan-Brain: anyways removed them haha

packages/api/db/migration/20250523133726_create_task_document_join_table.js
```diff
@@ -0,0 +1,35 @@
+/*
```
kathyavini, line 1: Do we care about the order of `knex_migrations` records being the same beta and prod? Only for record keeping I might like to see this named with a later date than the last one we merged, but since the migrations themselves can be run in either order it's not a very strong preference.

kathyavini (commented):
It looks really good to me!

packages/api/src/middleware/validation/checkDocument.ts
kathyavini: I also wrote this down but Sayaka beat me to it! This seems a bit premature to have two 100% duplicated functions anticipating this. In the future it might be cleaner to send a parameter e.g. `checkDocument('create')` anyway -- or maybe it will never be necessary.

packages/api/src/middleware/validation/checkTask.js
```diff
@@ -125,6 +126,8 @@ export function checkCompleteTask(taskType) {
         await checkAnimalCompleteTask(req, taskType, task_id);
       }
 
+      checkCompleteTaskDocument(req, taskType);
```
SayakaOno: Should `req` be `req.body` here too?

packages/api/src/middleware/validation/checkDocument.ts
Duncan-Brain: Copied from previous middleware - deleted it and updated this.

packages/api/src/middleware/validation/checkTask.js
Duncan-Brain: 🙏

packages/api/src/middleware/validation/checkTask.js
Duncan-Brain: Oh yah true, I did not originally want this second function here but the complete task routes that don't use middleware threw me off. Thanks!

packages/api/db/migration/20250523133726_create_task_document_join_table.js
Duncan-Brain: I dont think so - I am sure we probably already have this scenario.

packages/api/src/middleware/validation/checkTask.js
Duncan-Brain: 🙇

packages/api/src/middleware/validation/checkDocument.ts
Duncan-Brain: maybe!

PR #3794, opened 2025-06-05
LF-4839 Add endpoint to complete a soil sampling task

kathyavini:
**Description**

This is the boilerplate (no unique checks) implementation of the complete endpoint for soil sample task. The frontend only needed a single change (addition of the details component into `TaskComplete/StepOne` so I have included that too.

With that change, the complete flow should be totally functional, and just needs the document part, which is a separate set of tickets.

Jira link: https://lite-farm.atlassian.net/browse/LF-4839

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/api/tests/task.test.js
+++ b/packages/api/tests/task.test.js
@@ -35,6 +35,7 @@ import {
   CROP_FAILURE,
   sampleNote,
   abandonTaskBody,
+  expectTaskCompletionFields,
 } from './utils/taskUtils.js';
 import { setupFarmEnvironment } from './utils/testDataSetup.js';
 import { connectFarmToEnsemble } from './utils/ensembleUtils.js';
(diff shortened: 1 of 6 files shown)


packages/api/tests/task.test.js
```diff
@@ -2561,6 +2562,74 @@ describe('Task tests', () => {
...
+          task_type_id,
+          owner_user_id: user_id,
+          assignee_user_id: user_id,
+        }),
+      );
+
+      // Create location_tasks record to associate with farm
+      await mocks.location_tasksFactory({
+        promisedTask: [{ task_id }],
+        promisedField: [{ location_id }],
```
kathyavini, line 2597: The only thing I considered on this route to make it non-boilerplate would be restricting the locations to soil sample locations, but it didn't really get scoped (i.e. it was never mentioned on any of the tickets, and would have belonged on POST as well).

But more importantly, I don't think we have a pattern of restricting what locations can be associated as far as `location_tasks` is concerned -- even though different task types DO use different selectors in their frontend location selection flows -- so I'm comfortable with letting all location types be allowable. Of course when submitted from the frontend, the only locations that end up here would be soil sample locations.

SayakaOno (approved):
Works great!

Screenshot 2025-06-05 at 2 46 45 PM
It looks like I had to add top padding to the detail component... I'll make sure to have it fixed later!

SayakaOno:
By the way, what are the changes to the old code (`usersFactory` and `returnUserFarms`) for??

kathyavini:
> By the way, what are the changes to the old code (usersFactory and returnUserFarms) for??

@SayakaOno ah sorry, I described that only in the commit message details, which is super hidden! That was a duplication of a code change from the IP GET branch and highlighted [there](https://github.com/LiteFarmOrg/LiteFarm/pull/3787#discussion_r2114661737); that missing async was the reason for the type error that magically appeared the day we fixed the `knex.js` import!

PR #3797, opened 2025-06-06
Migration failure delete task animal relationships too

Duncan-Brain:
**Description**

***My db crashed and in order to add this to beta we need to rollback the last migration on beta.

1. Adds two more missing task relationships, it is good to be complete and cover all cases since we don't know state of all LiteFarm developers local databases. Are there any more that we skipped because they were not on beta?

This is the case where we are deleting a task that is also related to an animal. (Custom task)

2. Fix migrate:testing:db
- testing database migration fails because there is an empty array being attempted to insert

Jira link:

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/api/db/migration/20250527173006_cleanup_previous_sensors_implementation.js
+++ b/packages/api/db/migration/20250527173006_cleanup_previous_sensors_implementation.js
@@ -180,9 +180,14 @@ export const up = async (knex) => {
     table.timestamp('created_at').defaultTo(knex.fn.now()).notNullable();
   });
 
-  await knex('migration_deletion_logs').insert(
-    rowsToLog.map((row) => ({ migration_name: 'cleanup_previous_sensors_implementation', ...row })),
-  );
+  if (rowsToLog.length) {
+    await knex('migration_deletion_logs').insert(
+      rowsToLog.map((row) => ({
+        migration_name: 'cleanup_previous_sensors_implementation',
+        ...row,
+      })),
+    );
+  }
 
   // Remove notifications for sensors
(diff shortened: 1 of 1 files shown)


kathyavini (approved):
I'll ultimately defer to @SayakaOno, but it looks good to me!

Thank you for finding that testing db migration error!!

packages/api/db/migration/20250527173006_cleanup_previous_sensors_implementation.js
```diff
@@ -40,6 +40,8 @@ const sensorRelatedTaskTypeTables = [
   'pest_control_task',
   'field_work_task',
   'cleaning_task',
+  'task_animal_relationship',
+  'task_animal_batch_relationship',
```
SayakaOno, line 44: To cleanup test data added from tools like postman, I think it should be enough to add the task types found in [taskModel](https://github.com/LiteFarmOrg/LiteFarm/blob/integration/packages/api/src/models/taskModel.js).

Duncan-Brain: These two are possible in custom task -- I am not sure we need to cover postman cases or not.

If something is possible on backend but not frontend it is probably a bug.

SayakaOno: Okay, then let's go with this!
I wish there were validation for location types based on task type.

PR #3800, opened 2025-06-07
Lf 4836 (2) allow associating result document s with task on complete 2

Duncan-Brain:
**Description**

This PR 
- extracts out the FileUploader that previously existed only on add/edit documents and makes some small updates to the logic.
- It then adds that extracted component to the complete step one flow.
- Updates the read only task for completed tasks
  - What if a private file is added to this document in the documents section, and now the worker can access it?

Notes:
- Updating the uploader and document tile in ux improvement specific ticket
- Documents tab did not seem to need an update from the task complete saga. It calls its own saga every time it seems.

A bug exists on beta that I am also troubleshooting (fixed it it was a bad react key - now possible error only on duplicate of same file name is that acceptable?):

https://github.com/user-attachments/assets/71663bf5-7d40-448f-a5e2-84b78d334488

The logic change also fixes a bug where navigating directly to the add_document page results in infinite loading:
Screenshot 2025-06-06 at 9 02 35 PM

Jira link:

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/webapp/src/containers/Task/TaskReadOnly/index.jsx
```diff
@@ -128,11 +128,10 @@ function TaskReadOnly({ history, match, location }) {
     files.push({ url: externalIrrigationPrescription.prescription.vriData.file_url });
   }
   if (task.documents?.length) {
-    files.push(
-      ...task.documents.reduce((acc, cv) => {
```
Duncan-Brain: I don't quite understand why... but this caused wacky behaviour:

Going to the readonly completed task and refreshing once or twice would result in the complete task form submission then the task would be overwritten missing the complete_date details.

packages/webapp/src/components/Documents/Add/index.jsx
```diff
@@ -110,23 +103,9 @@ function PureDocumentDetailView({
...
-  const onUpload = () => {
-    setShouldShowLoadingImage(true);
-  };
-  useEffect(() => {
-    uploadedFiles?.length && setShouldShowLoadingImage(false);
-  }, [uploadedFiles?.length]);
 
   const disabled = isEdit
-    ? !isValid || uploadedFiles?.length === 0 || !(isDirty || isFirstFileUpdateEnded)
+    ? !isValid || uploadedFiles?.length === 0 || !isDirty || isUploading
```
kathyavini: I think `isUploading` isn't enough to replace the role of `isFirstFileUpdateEnded` -- this line still needs some sort of a boolean that specifically becomes true after file upload, since `isDirty` doesn't seem to track that. Resulting bug is that the update button doesn't become active anymore after adding a new file on edit.

I don't know if's better done in this component instead, but my first thought was that your hook could keep a second state boolean like that (maybe `fileChangesMade`) and update it `onUpload` or `onUploadEnd`?

packages/webapp/src/components/FilePicker/useFilePickerUpload.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { deleteUploadedFile } from '../../containers/hooks/useHookFormPersist/hookFormPersistSlice';
+import { DocumentUploader } from '../../containers/Documents/DocumentUploader';
+import { MediaWithAuthentication } from '../../containers/MediaWithAuthentication';
+import { useState } from 'react';
+
+export type FilePickerFunctions = {
+  deleteImage: (url: URL) => void;
+  imageComponent: (props: any) => JSX.Element;
+  documentUploader: (props: any) => JSX.Element;
+  isUploading: boolean;
```
kathyavini: Small question here about types -- the naming `FilePickerFunctions` makes me think it should be matching the `filePickerFunctions` which are getting passed around, but since those are always created by destructuring with rest properties `{ isUploaded, ...filePickerFunctions }` shouldn't `isUploaded` boolean not be included? Just going by naming I wonder if the hook's actual return type should be a union type of the boolean and the `FilePickerFunctions`?

packages/webapp/src/containers/Task/TaskReadOnly/index.jsx
```diff
@@ -125,7 +125,13 @@ function TaskReadOnly({ history, match, location }) {
 
   let files = [];
   if (externalIrrigationPrescription?.prescription?.vriData?.file_url) {
-    files.push(externalIrrigationPrescription.prescription.vriData.file_url);
+    files.push({ url: externalIrrigationPrescription.prescription.vriData.file_url });
+  }
+  if (task.documents?.length) {
+    const documentFiles = structuredClone(task.documents).reduce((acc, cv) => {
+      return [...acc, ...cv.files];
```
kathyavini: I don't think you need to `structuredClone` if you are already creating new arrays on each iteration with spread. But even better for this use case would be a `.flatMap` instead of the reduce at all:

```js
const documentFiles = task.documents.flatMap((doc) => doc.files);
files.push(...documentFiles);
 ```

**Review by kathyavini — COMMENTED** [review]
This looks like an excellent extraction (and improvement) of the orignal document uploading 🔥 And the use in task looks super clean to me!!

(Didn't look into cause of this at all, just noticed) -- for me the uploaded documents (task and regular) are downloading now with a uuid instead of the uploaded filename. Is that a bug? Do you see that too?

Screenshot 2025-06-09 at 11 41 24 PM

File: packages/webapp/src/components/Documents/Add/index.jsx (hunk 2 above)
Comment (Duncan-Brain) [reply packages/webapp/src/components/Documents/Add/index.jsx]: Thanks for spotting, you are right!

File: packages/webapp/src/components/FilePicker/useFilePickerUpload.tsx (hunk 3 above)
Comment (Duncan-Brain) [reply packages/webapp/src/components/FilePicker/useFilePickerUpload.tsx]: Yep its bad naming, I was lazy, let me see what I can do

File: packages/webapp/src/containers/Task/TaskReadOnly/index.jsx (hunk 4 above)
Comment (Duncan-Brain) [reply packages/webapp/src/containers/Task/TaskReadOnly/index.jsx]: If I did not use structuredClone() I seemed to mutate the state data somehow ! I made a comment above about how it them submitted the completeTask form -- it seemed crazy ( I have no clue why this has happened to me twice now)

Good point about flatMap() I used it in my first try but I was maybe troubleshooting it it at the same time as my mutating bug, lets have a second go and see

**Comment by Duncan-Brain** [comment]
Yeah its okay, I would have liked to find time for the ui upgrade and other Filepicker behaviour upgrade too but oh well. It will at least force us to fix both at once in the future.

Strange about the file naming... let me check it out.

File: packages/webapp/src/containers/Task/TaskReadOnly/index.jsx (hunk 4 above)
Comment (Duncan-Brain) [reply packages/webapp/src/containers/Task/TaskReadOnly/index.jsx]: So weird I swear I tried something more basic and it went crazy haha -- 😅 . Anyways thanks 🙏  seems to work fine now with no clone don't know what the heck that was.

File: packages/webapp/src/components/Documents/Add/index.jsx (hunk 2 above)
Comment (Duncan-Brain) [reply packages/webapp/src/components/Documents/Add/index.jsx]: Since it is trying to log the dirty state for the files.. I tried to do just that. It should be an improvement vs tracking if a file was uploaded and no hook needed.

File: packages/webapp/src/components/FilePicker/useFilePickerUpload.tsx (hunk 3 above)
Comment (Duncan-Brain) [reply packages/webapp/src/components/FilePicker/useFilePickerUpload.tsx]: should be fixed

File: packages/webapp/src/components/FilePicker/index.tsx [hunk 5]
```diff
@@ -0,0 +1,85 @@
...
+  return (
+    <div className={styles.filePickerContainer}>
+      {uploadedFiles?.map(({ thumbnail_url, file_name, url }) => (
+        <ContainerWithIcon
+          icon={<TrashIcon />}
+          onIconClick={() => {
+            deleteImage(url);
+            onUploadEnd();
+          }}
+          key={file_name}
```
SayakaOno: We should use thumbnail_url or url in case the same file is uploaded again.

packages/webapp/src/containers/Documents/DocumentUploader/index.jsx
```diff
@@ -20,9 +20,12 @@ export function DocumentUploader({ style, linkstyle, onUpload, linkText, onUploa
       e.target.value = '';
     } else if (e?.target?.files?.[0]) {
       onUpload?.();
-      dispatch(uploadDocument({ file: e.target.files[0], onUploadEnd }));
+      dispatch(uploadDocument({ file: e.target.files[0] }));
       e.target.value = '';
     }
+    if (onUploadEnd) {
+      onUploadEnd();
```
SayakaOno: This runs before the `uploadDocument` has finished.

kathyavini:
> wondering if its just a minio config thing?

Ahh that must be it!! Not a document uuid but a minio uuid. Thank for checking 🙏

packages/webapp/src/components/Task/TaskReadOnly/index.jsx
```diff
@@ -588,5 +590,5 @@ const taskComponents = {
   HARVEST_TASK: (props) => <PureHarvestingTaskReadOnly {...props} />,
   IRRIGATION_TASK: (props) => <PureIrrigationTask {...props} />,
   MOVEMENT_TASK: (props) => <PureMovementTask disabled {...props} />,
-  SOIL_SAMPLE_TASK: (props) => <PureSoilSampleTask {...props} />,
+  SOIL_SAMPLE_TASK: (props) => <PureSoilSampleTask disabled {...props} />,
```
SayakaOno: I think `disabled` is included in the props (line 479), so any task types shouldn't need it...?

packages/webapp/src/stories/Pages/Document/MainDocument.stories.jsx
```diff
@@ -27,8 +27,8 @@ Primary.args = {
       },
     ],
   },
-  imageComponent: (props) => <img src={props.fileUrl} />,
-  fileDownloadComponent: (props) => <img src={props.fileUrl} />,
+  imageComponent: (props) => <img src={props.fileUrls[0]} />,
+  fileDownloadComponent: (props) => <AddLink {...props}>{props.linkText}</AddLink>,
```
SayakaOno: Was this intentional?
Screenshot 2025-06-10 at 11 46 33 AM

This does nothing but I think `<MediaWithAuthentication {...props} />` is better...

packages/webapp/src/components/Task/TaskComplete/StepOne.jsx
```diff
@@ -187,6 +205,21 @@ export default function PureCompleteStepOne({
...
         : null}
+      {taskType === 'SOIL_SAMPLE_TASK' && (
+        <div>
+          <Main style={{ marginBottom: '24px' }}>{t('TASK.DID_YOU_GET_RESULTS')}</Main>
+          <RadioGroup hookFormControl={control} required name={RESULTS_AVAILABLE} />
+          {resultsAvailable && (
+            <FilePicker
+              uploadedFiles={uploadedFiles}
+              linkText={t(`TASK.UPLOAD_LAB_DOCUMENT`)}
+              shouldShowLoadingImage={isUploading}
```
SayakaOno: I think the prop name is `showLoading`?

packages/webapp/src/containers/Documents/DocumentUploader/index.jsx
Duncan-Brain: Thanks! Silly I forgot dispatch was doing its own thing -- added back into saga but in a different spot.

packages/webapp/src/components/Task/TaskReadOnly/index.jsx
Duncan-Brain: yes true, was following the other pattern

packages/webapp/src/components/FilePicker/index.tsx
Duncan-Brain: Thanks!

packages/webapp/src/stories/Pages/Document/MainDocument.stories.jsx
Duncan-Brain: Thanks, I assumed that media with authentication and document uploader were hard to use in stories since original file was somewhat broken anyways. 

So I was fine with it being weird if it was slightly improved. This one was a mistake, I forgot that main view uses it differently. Using the actual components seems to make everything nice for me.

packages/webapp/src/components/Task/TaskComplete/StepOne.jsx
Duncan-Brain: 🙏

Duncan-Brain:
@SayakaOno thanks for reviewing, IN the pr where I added documents to task read only I forgot to add image preview prop. It is now added!

kathyavini (approved):
Awesome!!

kathyavini:
Shucks, [deploy failure](https://github.com/LiteFarmOrg/LiteFarm/actions/runs/15596857822/job/43929126972)... maybe on lodash import? Will take a look

PR #3803, opened 2025-06-11
LF-4842 Copy all non cjs, js, ts files to compiled dist folder

Duncan-Brain:
**Description**

This PR copies all files from src that are not .cjs, .js, and .ts to the dist folder since non-javascript are not compiled as part of tsc.

This hopefully fixes the no-emails issue where the files are not found. Not 100% sure on this but maybe 90% sure this is it.

Notes:
- `rsync` was added to dockerfile as an easy to understand utility, the find and copy was unreadable comparably and it seems like this is more standard for this now?
- only added the copy script to dockerfile (not `npm build`) because the copy path would be different `./dist/app/src/` vs `./dist/api/src/ so it will work locally with a tweak to that path.

Jira link: [LF-4842](https://lite-farm.atlassian.net/browse/LF-4842)

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain):

Ran the docker container locally by adding this to my docker compose:
```
backend:
    container_name: litefarm-api
    deploy:
      restart_policy:
        condition: on-failure
        max_attempts: 3
    build:
      context: ./packages/
      dockerfile: ./api/prod.Dockerfile
    volumes:
      - /etc/letsencrypt:/usr/src/letsencrypt
    ports:
      - "5000:5000"
      - "5432:5432"
    environment:
      NODE_ENV: development
      DEV_DATABASE_HOST: host.docker.internal
      DEV_DATABASE: pg-litefarm
      DEV_DATABASE_USER: **insert user**
      DEV_DATABASE_PASSWORD: **insert password**
```

**Checklist:**

[LF-4842]: https://lite-farm.atlassian.net/browse/LF-4842?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -15,7 +15,7 @@
     "start": "set NODE_ENV=development&& node --import @swc-node/register/esm-register src/server.ts",
     "start:prod": "node dist/app/src/server.js",
     "build": "rimraf dist && tsc -p src/tsconfig.json",
-    "cp-uncompiled-to-dist": "rsync -a --include='*' --exclude='*.cjs' --exclude='*.js' --exclude='*.ts' ./src/templates ./dist/app/src/templates && rsync -a --include='*' --exclude='*.cjs' --exclude='*.js' --exclude='*.ts' ../shared/locales ./dist/shared",
+    "cp-uncompiled-to-dist": "rsync -a --include='*' --exclude='*.cjs' --exclude='*.js' --exclude='*.ts' ./src/templates ./dist/app/src && rsync -a --include='*' --exclude='*.cjs' --exclude='*.js' --exclude='*.ts' ../shared/locales ./dist/shared",
     "dev": "set NODE_ENV=development && node --inspect=0.0.0.0:9230 --watch --import @swc-node/register/esm-register src/server.ts",
     "debug-email-template": "DEBUG=email-templates npm run dev",
     "debug-i18n": "DEBUG=i18n:* npm run dev",


navDhammu (commented):
This looks like a good working solution - by copying the non-js, ts files over to dist. My main concern is that it may include every type of file and not just the ones we need, like .env or other hidden files that may somehow cause unintended side-effects. 

Also, I noticed the dist output includes `/shared` and `/.knex` folders as well, so if there are any needed non-TS or JS files in those directories, they would also need to be copied. 

I'm thinking its better to just copy the exact files we need? One alternate and more long-term solution that I saw someone mention somewhere, (I think it was in this [discussion](https://github.com/microsoft/TypeScript/issues/30835)) is to create a dedicated `static` folder that will include all template files and other static assets needed by the backend in `/api/static`, and copy it over to `/dist/api/static` during build.

kathyavini:
Chatted a bit synchronously about this today, but updating here, too. I was looking at the files copied with `--dry-run`:

```bash
rsync -avi --dry-run --exclude='*.js' --exclude='*.ts' --exclude='*.cjs' ./src/ ./dist/app/src/
```

and personally I like the idea of only copying `templates/` as those are the only files we need -- it's a bit close to a `static/` folder anyway. Copying `/shared` also sounds good 👍 There aren't any files in there that are still needed, now that we've removed the old sensor flow, but I think we intend to start using that package more than we do so sounds like a good idea to set it up.

Duncan-Brain:
Thanks @navDhammu good points.

I will update to restrict just to `/templates` folder and `/shared/locales` for now. The silent error on this one made me want to just add everything and avoid further issues since 'src` does not have those extra files --  but maybe not all are necessary at this time.

Thanks for the conversation about static. That might be a good idea! But not sure it would cover all cases or not.

Thanks @kathyavini for the options for `dry-run` that is pretty cool!

navDhammu:
@Duncan-Brain  I noticed while reviewing that locally I'm not able run the backend server after building with tsc because its using `dist/app` as the entry point rather than `dist/api`, which I think was done to match the docker container's path. This felt a bit weird to have a script just so it works on docker, so I created a separate [PR](https://github.com/LiteFarmOrg/LiteFarm/pull/3809) to address it. This might allow you to do the file copying in the build script rather than just in docker.

packages/api/package.json
```diff
@@ -15,6 +15,7 @@
     "start": "set NODE_ENV=development&& node --import @swc-node/register/esm-register src/server.ts",
     "start:prod": "node dist/app/src/server.js",
     "build": "rimraf dist && tsc -p src/tsconfig.json",
+    "cp-uncompiled-to-dist": "rsync -a --include='*' --exclude='*.cjs' --exclude='*.js' --exclude='*.ts' ./src/templates ./dist/app/src/templates && rsync -a --include='*' --exclude='*.cjs' --exclude='*.js' --exclude='*.ts' ../shared/locales ./dist/shared",
```
kathyavini:
```suggestion
    "cp-uncompiled-to-dist": "rsync -a --include='*' --exclude='*.cjs' --exclude='*.js' --exclude='*.ts' ./src/templates ./dist/app/src && rsync -a --include='*' --exclude='*.cjs' --exclude='*.js' --exclude='*.ts' ../shared/locales ./dist/shared",
```

This is currently creating `/templates` within `/templates`:

Screenshot 2025-06-13 at 10 30 48 AM

The syntax used for `/shared/locales` in the second half is correct; or you could add trailing slashes to both `/templates/` (but I like matching syntax!)

Duncan-Brain: Thanks I tested the second one 😂 committed this suggestion from my phone cuz I cant test right now

PR #3805, opened 2025-06-11
LF-4857 Both selected and de-selected visual map icons overlay for soil samples

kathyavini:
**Description**

This was actually a bit painful to get perfect and there is the tiniest issue still 😅 -- but this is mostly good

This fixes the weird offset and two icons overlaying issue visible when selecting soil sample locations by:
- using the new icons provided by Loïc where all three states are the same shape (round)
- manually adjusting the width and height (_but not viewbox_) of the hover + active states to match the default -- exported from Figma the hover + active icons are 28px, while default is 26px. I have made them all 26px.
  - because the ratio of viewbox to width/height is slightly different from the exported file, the actual spade part of the icon is going to look like it has a slightly different stroke width, proportionally speaking, than on Figma. You can see this using the swipe view on the file diff
  - however in app, I found this difference to be neglible, compared to the 2px width/height difference, which was noticeable

This PR also:
- Bumps up the width/height of the sensor (non-array) icon to 26px so that all the circular icon point locations (= gate, water valve, soil sample location, sensor) have the same radius -- the minor difference between them was bothering me when looking at them side-by-side on one map
  - This particular combination is probably not going to be used by a lot of users though

Jira link: https://lite-farm.atlassian.net/browse/LF-4857

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


kathyavini:
@Duncan-Brain thanks so much for bringing this issue up! It turns out the jaggies at least can absolutely be fixed, and the issue was Figma exporting with `crispEdges` applied to the SVG. Here is even a [Figma feature request](https://forum.figma.com/suggest-a-feature-11/by-default-disable-shape-rendering-crispedges-in-svgs-12181) for this to be  fixed. We might be able to fix a few more of our SVGs with this in mind 🤔 

That's 1 of 2 mysteries down... randomly thicker borders still remain, though...

Screenshot 2025-06-12 at 1 42 03 PM

Duncan-Brain:
👍 nice find!

PR #3806, opened 2025-06-11
LF-4858: Data for the last day in the sensor reading date range is not included as expected

SayakaOno:
**Description**

Jira link: https://lite-farm.atlassian.net/browse/LF-4858

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/containers/SensorReadings/v2/Charts/useFormattedSensorReadings.ts
+++ b/packages/webapp/src/containers/SensorReadings/v2/Charts/useFormattedSensorReadings.ts
@@ -57,6 +57,13 @@ function isChartSupportedReadingType(
   return CHART_SUPPORTED_PARAMS.findIndex((param) => param === readingType) !== -1;
 }
 
+const getAdjustedEndTime = (endDate: string) => {
+  const dateObject = new Date(endDate);
+  dateObject.setDate(dateObject.getDate() + 1);
+
+  return new Date(dateObject).toISOString();
+};
+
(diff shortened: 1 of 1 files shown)


Duncan-Brain (approved):

kathyavini:
@Duncan-Brain

> Just trying to understand if this was an ESci "Adjustment" or not, since I didn't look at this originally. But I think this is not an esci adjustment, and expected behaviour of a date range right?

I guess not, yeah! Actually probably the issue was introduced at the same time we started translating the date range components dates (which are just dates, no times) to full ISO strings to send to Ensemble, which I totally failed to think about earlier. In finances I think they always stay as dates, and the filtering is based on dates, which is why these issues don't come up.

Regarding whether the endpoint is inclusive or not -- regardless of how we construct the end date, we don't control that; Ensemble is the one that filters on the dates. Sayaka's note above suggests for readings at least, probably they are using exclusive filtering on the end time?

> The Ensemble API returns the same data for end_time=2025-06-11T00:00:00.000Z and 2025-06-10T23:59:59.000Z

So we could go ahead and assume we can send half open sets but... honestly we don't know if they will keep consistent filtering logic between readings and IP (very different endpoints) so I wouldn't worry too much about it yet. We can take another look at IP dates before we finalize those.

kathyavini (approved):

Duncan-Brain:
@kathyavini there is probably no data available on that last second to actually test the difference and sounds good thanks!

PR #3813, opened 2025-06-13
LF-4875 Point locations icons are not centered at the point location

kathyavini:
**Description**

Please see details in [Slack thread](https://litefarm.slack.com/archives/CAH7PMYE4/p1749764219274949) and/or Jira ticket.

Our point location (water valve, gate, sensor, and now sensor array, and soil sample location) are all circular icons, but the default icon anchor for Google maps assumes a _pin_ location (i.e. the anchor point is in the bottom centre).

This PR:
- wraps calls to the point location icons in a function that adds an appropriate anchor point for the standard icon dimension (26px by 26px)
- harmonizes the dimensions of our last SVG icon (sensor array) that was not according to these dimensions

Note: farm centre and wild crops are actual pins, and are not part of the `icons[type]` structure that is being centered here, so they will continue to render as before.

Jira link: https://lite-farm.atlassian.net/browse/LF-4875

**Type of change**

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

**How Has This Been Tested?**

- [ ] Passes test case
- [x] UI components visually reviewed on desktop view
- [ ] UI components visually reviewed on mobile view
- [ ] Other (please explain)

**Checklist:**

- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [x] The precommit and linting ran successfully
- [ ] I have added or updated language tags for text that's part of the UI
- [ ] I have ordered translation keys alphabetically (optional: run `pnpm i18n` to help with this)
- [ ] I have added [the GNU General Public License](https://lite-farm.atlassian.net/l/cp/BT0Dd7WW) to all new files

PR #3818, opened 2025-06-17
LF-4875b Point locations icons are not centered at the point location

kathyavini:
**Description**

This is an alternative solution for the problem of the point icons being anchored at the bottom middle of the icon instead of the centre, and replaces the solution in #3813. See discussions in [Slack](https://litefarm.slack.com/archives/CAH7PMYE4/p1749764219274949) and the Jira ticket. This PR updates our existing (gate, water valve), and new (sensor, sensor array, soil sampling locations) point icons to pin-shaped icons.

Advantages of moving to pins throughout:
- being able to use the Google maps default icon anchor logic, which is built for pins
- more precise-looking points with no ambiguity about what location is being marked (particularly useful for soil sample locations, our first point locations which report their `lat` / `lng`, and sensors, which receive a `lat`/ `lng` from Ensemble)
- pins obscure the actual location being marked slightly less than a centred circle icon
- if existing point locations were placed based on the map based on the visible icon, the icon will stay in roughly the same place, which was not the case for the original solution of centering existing icons on their actual coordinates
  - Note: that icon will "read" differently though, as the old icons suggested a central anchor. I'm curious to know whether we will get user feedback about this
- (not related to moving to pins, but a nice side effect) this was an opportunity to harmonize our point icons throughout app, giving them consistent hover + active styling, and moving the png icons (gate, water valve) onto SVG with the others

Disadvantages:
- As mentioned above, although we will not be repositioning existing user icons, the icons "read" differently now (indicate a slightly different point in space), so users may find that the placement of their gates, water valves, etc. is not what they wanted. Curious to see if we receive user feedback on this
- (This is completely subjective) Replacing the circular icons with pins is a very different design language on the map -- maybe a little less friendly / more scientific this way? -- and it remains to be seen if the users who, e.g. like to print out their farm maps, prefer this design language to the original one. Again, this is hard to speculate on this without user feedback.

Jira link: https://lite-farm.atlassian.net/browse/LF-4875

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


Duncan-Brain (approved):
Looks great!

PR #3826, opened 2025-06-27
LF-4882 Update IP request POST to match Ensemble specifications

kathyavini:
TODO: with the go-ahead from Ensemble on Monday, the daily cron can be set to `shouldSend = true` and the ability to call with `allOrgs = true` can be removed from the API endpoint -- but it is preserved in this draft state to make it easier to test together on Monday.

**Description**

This updates the Ensemble POST request (Irrigation Prescription request) according to the revised specifications:

1. POST should be directed to an organisation-level endpoint with only that organisation's data in the request
2. Shape of the request should be just the location array; no more reference to org_uuids as keys
3. `crop_data` should contain at most one crop -- this has been done by selecting the crop management plan with the most recent seed date and sending only that one

Jira link: https://lite-farm.atlassian.net/browse/LF-4882

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

At the moment, this can be tested locally as follows on this endpoint, if you have farms locally connected to valid Ensemble orgs:

```
/irrigation_prescription_request?allOrgs=true&shouldSend=false
```

(I'll update once the `allOrgs` change is pushed)

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/irrigationPrescriptionRequestController.ts
+++ b/packages/api/src/controllers/irrigationPrescriptionRequestController.ts
@@ -14,7 +14,10 @@
  */
 
 import { Response } from 'express';
-import { getOrgLocationAndCropData, sendFieldAndCropDataToEsci } from '../util/ensembleService.js';
+import {
+  getOrgLocationAndCropData,
+  sendAllFieldAndCropDataToEsci,
+} from '../util/ensembleService.js';
 import { LiteFarmRequest } from '../types.js';
 
 interface HttpError extends Error {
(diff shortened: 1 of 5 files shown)


kathyavini:
@SayakaOno That's a good question! No, I think it merely a holdover from wanting to keep that param `allOrgs` on the direct API endpoint as long as possible -- even though with the last bit of code change, it's meaningless now anyway 😬

I'll have `isSchedulerRequest` control whether all farm data is included... could you check that and let me know if that looks like the cleanup you would have expected 🙏

PR #3827, opened 2025-07-02
LF-4883 (1) Update IP GET requests to match Ensemble specifications

kathyavini:
**Description**

This is all the changes necessary to show real Ensemble Irrigation Prescription data! 🙌 

With the changes that came in today, most everything is in a completed state. It's a pretty long PR of bits and pieces, but none of the changes should be particularly complex in terms of logic. Maybe the only part I would have some trepidation about is the handling of dates, since we have moved off all timestamps -- I would consider this the most likely candidate for introduced bugs!

### Here is the full list of changes, in roughly file order, to reference if needed during review:

**Backend**

Backend - controllers
- the daily IP notifications controller now 1) checks for Ensemble connection, and 2) filters by `farm_id`
  - checking for Ensemble connection has a new helper function which is now also used in `triggerPostTaskCreatedActions`
  
 Backend - middleware
- the IP datetime stamps are now date strings, so the `startTime` and `endTime` query parameters have been updated to be the same. _Ensemble has not implemented any time filtering yet, so I'm just assuming this is the format!_ I'll leave a note below to double-check

Backend - addonPartner.ts & task.ts
- there was an incorrectly defined error message that would have shown `[object Object]`
- `farm_id` filtering was added to the returned list GET
- `triggerPostTaskCreatedActions` was refactored to use the new helper function that checks for Ensemble connection

Backend - ensembleService.ts
- helper function for checking Ensemble connection without throwing errors was added
- GET requests are now targetted at the new organisation-specific endpoints (i.e. endpoints that contain `org_pk`)
- notification controller's mock functions were deleted; now daily IP notifications always call the real Ensemble list endpoint
  - note: `shouldSend` is not part of this flow, so this is never mocked
- pivot data is transformed into the format we were expecting

Backend - ensembleService.types.ts, generateMockPrescription.ts, mockAddonPartner.ts
- cleanup of some errors on our end around the type of `management_plan_id`
- reflecting the actual incoming pivot + system data correctly in our type structure
- mocks also updated to reflect these types
  - largest changes being the timestamp --> date string adjusts

Backend - tests
- adjusting for each of the above
- `location_id` added to mock data when not previously required, as we are filtering by `farm_id` using it

**Frontend**

Frontend - components & containers
- the map will crash (not render) without a centre coordinate, so a field polygon point is now the fallback, to handle `null` pivots
- `soil_moisture_deficit_unit` now used (displayed directly; no conversion)
- date helper functions replaced; moved from those expecting ISO string format to YYYY-MM-DD format
- fallback added for missing `recommended_start_date` in `taskData.due_date`, because this in particular does crash upon task creation
- `recommended_start_time` removed from the UI
- keys updated

Frontend - `store/api` & stories
- now requesting real instead of mock data
- sending start and date times in YYYY-MM-DD format to match the date property Ensemble is providing, instead of the 48hr timestamp range we had been sending originally
- estimated time removed from the RTK query type (the KPI / UI was already updated in the units PR)
- RTK query type updated to reflect updated structure
- `soil_moisture_deficit_unit` key added to all stories

That's it!

### The follow-up LF-4883 (2) should include these changes:

- if ESci is able to return `farm_id` on the list GET endpoint, we can filter directly on that property instead of using the model method `getFarmIdByLocationId()`
- I don't believe there is any date filtering on the list GET endpoint yet; we want to re-confirm that this is happening and the query parameter format
  - IF THEY DON'T END UP ADDING THIS, we will want to filter the GET list on our own to only show two prescriptions. This not added yet as it's a bit easier to debug the whole situation by seeing the full GET list return in our UI
- We should revisit error states (which was given its own user story but called nice-to-have) once everything is finalized (e.g. IP details not belonging to farm)

Jira link: https://lite-farm.atlassian.net/browse/LF-4883

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

Only newly generated (Jul 10th onwards) IPs will have all properties (pivot, system, etc). To generate such an IP, I recommend using 

```
http://localhost:5001/irrigation_prescription_request?shouldSend=true
```

while authenticated under each of your irrigation-system containing test farms. One of the pilot project organisations has two pivots within one irrigation system (!) To correctly get back IPs for that org, divide the satellite-visible field into two, [as shown in Slack here](https://litefarm.slack.com/archives/C02P3UR5988/p1752180888632659?thread_ts=1752171885.959889&cid=C02P3UR5988), so that each field contains only one pivot. I also noticed that I did not get an IP until I had set up crop data on each of those two fields.

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/IrrigationPrescription/IrrigationPrescriptionTable.tsx
+++ b/packages/webapp/src/components/IrrigationPrescription/IrrigationPrescriptionTable.tsx
@@ -61,7 +61,7 @@ export default function IrrigationPrescriptionTable({ data }: IrrigationPrescrip
         format: (d: IrrigationPrescriptionTableInfo) => (
           <Cell
             kind={CellKind.PLAIN}
-            text={`${d.soil_moisture_deficit}%`}
+            text={`${d.soil_moisture_deficit}${d.soil_moisture_deficit_unit}`}
             className={styles.tableText}
           />
         ),
--- a/packages/api/.env.default
+++ b/packages/api/.env.default
@@ -97,3 +97,15 @@ API_HOST=localhost
 OOO_MESSAGE_ENABLED=
 # Format YYYY/MM/DD
 OOO_END_DATE=
+
+
+# Debugging variables for working on specific features
+
+# Speed up the notifications timer (cycle through 1 hour every minute)
+# MOCK_TIMER=true
+
+# Specify which mock implementation to use for addon partners in development (GET irrigation prescription list)
+# MOCK_ADDON_PARTNER=ESCI
+
+# Mock Ensemble irrigation prescription details
+# USE_IP_MOCK_DETAILS=true
--- a/packages/api/db/migration/20250707200337_cleanup_legacy_ensemble_farm_addons.js
+++ b/packages/api/db/migration/20250707200337_cleanup_legacy_ensemble_farm_addons.js
@@ -1,0 +1,71 @@
+/*
+ *  Copyright 2025 LiteFarm.org
+ *  This file is part of LiteFarm.
+ *
+ *  LiteFarm is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  LiteFarm is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ */
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+export const up = async (knex) => {
+  // Rows to be inserted to 'migration_deletion_logs' table
+  // When rolling back, rows will be restored backwards
+  const rowsToLog = [];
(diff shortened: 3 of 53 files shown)


packages/webapp/src/components/IrrigationPrescription/IrrigationPrescriptionTable.tsx
```diff
@@ -61,7 +61,7 @@ export default function IrrigationPrescriptionTable({ data }: IrrigationPrescrip
         format: (d: IrrigationPrescriptionTableInfo) => (
           <Cell
             kind={CellKind.PLAIN}
-            text={`${d.soil_moisture_deficit}%`}
+            text={`${d.soil_moisture_deficit}${d.soil_moisture_deficit_unit}`}
```
kathyavini, line 64: I noticed that I never added converting application depth off of `mm` depending on system, and I'm not sure if it was an oversight on my part, or something we checked and got the go-ahead for! I think maybe the latter, though -- I think it's another case of an industry standard (and values too small to show nicely in `inches`).

In any case, this is matching, and so will not convert!

kathyavini:
@SayakaOno ahhh I see! Thank you, that's a great suggestion -- an `.env` variable seems like better practice for sure 🙏  I'll make that change and clean up the query param.

Just curious your opinion on documentation -- would you personally want to see this kind of debuging variable in `.env.default`? Or leave it since it's not really part of the default setup (kind of like the undocumented`MOCK_TIMER` used in the timed notification flows)?

SayakaOno:
@kathyavini I agree, and it wouldn't hurt! 😄 Thank you 🙏

PR #3829, opened 2025-07-07
LF-4810: Unit adjustments based on Ensemble's confirmation

SayakaOno:
**Description**

- ET rate - mm/24h
- Estimated water consumption - AF for both imperial and metric systems
- don’t need estimated time


kathyavini (approved):
@SayakaOno I gave it a quick read through and it looks good!

PR #3831, opened 2025-07-08
LF-4885 Handle pivot arcs

kathyavini:
**Description**

These are the updates needed to represent pivots that only traverse partial circles (i.e. pie slices, which Turf.js calls [sectors](https://turfjs.org/docs/api/sector)) in irrigation prescriptions. The code had to be updated in three ways to handle the `pivot.arc` object, listed below. If `pivot.arc` or start/end angle is omitted, the original code -- for a full circle -- will be used in each of these places:

1. The map view needs to render a sector instead of the full circle. The Google Maps circle geometry didn't include the ability to define sectors natively, so the super-useful Turf package has been used once again to create a polygon of lat/lng coordinates based on centre, radius, and angles. Map styles were updated for this polygon type to match the original pivot (purple outline + pivot arm, white fill, small purple label).
3. The pivot arm (also part of the map view) was updated so that it points to an angle within the sector (happens to be the midpoint) instead of due east, which might not fall within the sector. I'm a tiny bit ambivalent about this -- @SayakaOno what do you think? Is that a good default, visually? I couldn't really figure out how to solve the issue of the pivot label and the pivot arm sometimes overlapping depending on the angles selected.
4. The URI water calculation was updated to take into account start and end angles if they are provided.

The mock sent from the backend for URI (but not VRI) was updated, and both URI and VRI Storybook stories were added:
- [Partial angle VRI story](http://localhost:6006/?path=/story/components-irrigationprescription-mapview--partial-angle-pivot-vri)
- [Partial angle URI story](http://localhost:6006/?path=/story/components-irrigationprescription-mapview--partial-angle-pivot-uri)

There's also a little bit of context [in Slack here](https://litefarm.slack.com/archives/C02P3UR5988/p1751914539957019). Please also me know also @SayakaOno what you think of the question of clockwise (CW) vs counterclockwise (CCW) start and end angles 🙏 This code is CW, matching the example images in that Slack thread.

Jira link: https://lite-farm.atlassian.net/browse/LF-4885

**Type of change**

**How Has This Been Tested?**

- [x] Passes test case
- [ ] Other (please explain)

The API test was also mimally updated to include a sector of obvious proportions intead of a full circle.

**Checklist:**

### DIFF
--- a/packages/webapp/.storybook/state.js
+++ b/packages/webapp/.storybook/state.js
@@ -1467,6 +1467,12 @@ export default {
       loading: false,
       loaded: false,
     },
+    soilSampleLocationReducer: {
+      ids: [],
+      entities: {},
+      loading: false,
+      loaded: false,
+    },
     showedSpotlightReducer: {
       loaded: true,
       loading: false,
--- a/packages/webapp/src/components/IrrigationPrescription/index.tsx
+++ b/packages/webapp/src/components/IrrigationPrescription/index.tsx
@@ -24,6 +24,12 @@ interface CommonIrrigationPrescriptionProps {
   fieldLocation?: Location;
   pivotCenter: Point;
   pivotRadiusInMeters: number;
+  pivotArc?: {
+    start_angle: number;
+    end_angle: number;
+  };
+  vriData?: VriPrescriptionData[];
+  uriData?: UriPrescriptionData;
   system: System;
 }
 
@@ -43,6 +49,7 @@ const PureIrrigationPrescription = ({
   fieldLocation,
   pivotCenter,
   pivotRadiusInMeters,
+  pivotArc,
   vriData,
   uriData,
   system,
@@ -69,6 +76,7 @@ const PureIrrigationPrescription = ({
         fieldLocation={fieldLocation}
         pivotCenter={pivotCenter}
         pivotRadiusInMeters={pivotRadiusInMeters}
+        pivotArc={pivotArc}
         vriZones={sortedVriData}
         system={system}
       />
--- a/packages/webapp/src/components/IrrigationPrescription/index.tsx
+++ b/packages/webapp/src/components/IrrigationPrescription/index.tsx
@@ -28,8 +28,6 @@ interface CommonIrrigationPrescriptionProps {
     start_angle: number;
     end_angle: number;
   };
-  vriData?: VriPrescriptionData[];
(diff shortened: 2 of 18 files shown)


packages/webapp/.storybook/state.js
```diff
@@ -1467,6 +1467,12 @@ export default {
       loading: false,
       loaded: false,
     },
+    soilSampleLocationReducer: {
```
kathyavini, line 1470: I was completely mistaken about the use of this file; I thought it was optional and only limited to a few stories that explicitly imported it, but that wasn't so! In neglecting to add the new location here, which I should have done alongside the other code changes [in this commit](https://github.com/LiteFarmOrg/LiteFarm/pull/3774/commits/992c109fb6969dcb59d3e11b25d3d84ebb5f6e9f), I created this crash on the IP story which can be seen on integration:

Screenshot 2025-07-08 at 1 41 17 PM

(I belive the `useMapZoom()` hook in particular is what is crashing)

SayakaOno: It's great that you figured out what the problem was!

packages/webapp/src/components/IrrigationPrescription/index.tsx
```diff
@@ -24,6 +24,12 @@ interface CommonIrrigationPrescriptionProps {
   fieldLocation?: Location;
   pivotCenter: Point;
   pivotRadiusInMeters: number;
+  pivotArc?: {
+    start_angle: number;
+    end_angle: number;
+  };
+  vriData?: VriPrescriptionData[];
+  uriData?: UriPrescriptionData;
```
SayakaOno: I wonder if we can call this `PureIrrigationPrescriptionProps` now...?

kathyavini: Oh my goodness, I truly don't know how those got into that diff! 😳 I think it was an over-aggressive IDE suggestion that I tabbed to when adding the pivot arc... sorry! 😬 

I actually want them gone to restore the original union type -- would that work too? Or is the spirit of of your comment that do you prefer the looser but easier-to-read type? 😄

SayakaOno: The original one works well! I usually only use that kind of strict types when TS complains, so I instinctively thought the simpler one would be fine.

PR #3833, opened 2025-07-16
LF-4886 Users should get one IP notification per location, not per day

kathyavini:
**Description**

This PR fixes two bugs with IP notifications:

1. We had been returning only the latest IP _per farm_, when a given farm potentially has 2-3+ new IPs per day, depending on the number of pivot-containing locations defined.
2. (Noticed this when testing notifications, but it's an oversight affecting all IPs) IPs from deleted locations weren't being filtered out. That filtering has been added to the location model method that we have been using to identify + filter the correct farm. This means notifications, the list GET, and the details GET will all no longer return IPs if their listed location has been deleted.

This PR also includes updates to the test cases on IP notifications to reflect the above.

Jira link: https://lite-farm.atlassian.net/browse/LF-4886

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

Triggered the notifications flow by running the scheduler (`docker compose export`) with `MOCK_TIMER=true`.

Bumping this line to something a little higher than 5 will also make 4am Alberta time come faster -- I think I used `9`? But I haven't been precise, I do usually just let it run for a bit.

```js
let utcHour = mockTimer ? 5 : undefined; // mock starts 2 "hours" before day shifts
```

**Checklist:**

### DIFF
--- a/packages/api/tests/utils/testDataSetup.ts
+++ b/packages/api/tests/utils/testDataSetup.ts
@@ -125,3 +125,25 @@ export async function setupManagementPlans({ farm, field }: { farm: Farm; field:
     seedManagementPlan,
   };
 }
+
+/**
+ * Creates a field for an existing farm and returns the location with field data
+ */
+export async function createField(farm: Farm) {
+  const [location] = await mocks.locationFactory({ promisedFarm: Promise.resolve([farm]) });
+
+  await mocks.fieldFactory({
+    promisedLocation: Promise.resolve([location]),
+  });
+
+  const field = await LocationModel
+    /* @ts-expect-error don't know how to fix */
+    .query()
+    .context({ showHidden: true })
+    .whereNotDeleted()
+    .findById(location.location_id).withGraphFetched(`[
+      figure.[area], field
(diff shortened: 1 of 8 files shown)


packages/api/tests/utils/testDataSetup.ts
```diff
@@ -125,3 +125,25 @@ export async function setupManagementPlans({ farm, field }: { farm: Farm; field:
     seedManagementPlan,
   };
 }
+
+/**
+ * Creates a field for an existing farm and returns the location with field data
+ */
+export async function createField(farm: Farm) {
```
SayakaOno, line 120: Is this a copy of the code block in `setupFarmEnvironment`?? I'm wondering if we can replace that with this function!

kathyavini: @SayakaOno that is _exactly_ what that code block is 😳 Of course, what a good idea. Thank you!

SayakaOno (approved):
Thank you 😍💖

PR #3838, opened 2025-07-22
LF-4889 User data deletion and export scripts

kathyavini:
**Description**

So sorry @SayakaOno -- I didn't set out to give you 900 lines of SQL to review when I was thinking about the fastest way to implement deletion 🙇 I also had forgotten about export at that time. I don't think this is an ideal method, and I'll mention some future improvements at the end of this description, but I think for now improving this isn't a very high priority overall.

## Explanation of the four scripts

I would lightly recommend reading the scripts in execution (numbered) order.

```
db/user_data_requests/
├──common/
    ├── 00-scoped_table_lists.sql 
    └── 01-user_data_indices.sql
├── 02-export_user_data.sql 
└── 03-delete_user_data.sql
```

The logic behind the scripts is that for user export, basically we want to to loop through a pattern that looks like this:

```sql
  SELECT json_build_object(
    'userLog', (
      SELECT json_agg(ul)
      FROM "userLog" ul
      WHERE ul.user_id = target_user_id
    ),
    -- ...
```

where we select the correct records from each table using some identifier (here, `user_id` on the `userLog` table) and add them to some big jsonb object that will be our export

For deletion, it's very similar:

```sql
DELETE FROM "userLog" WHERE user_id = target_user_id;
```

Initially, in the first pass I (with AI help) wrote out all our tables and the logic needed to select the right record, and just repeated it all for both [delete](https://github.com/LiteFarmOrg/LiteFarm/blob/0ca8d89aeacbb7f40cc0ec2c8ed97cc71649243e/packages/api/db/sql/deleteUserData.sql) and [export](https://github.com/LiteFarmOrg/LiteFarm/blob/0ca8d89aeacbb7f40cc0ec2c8ed97cc71649243e/packages/api/db/sql/exportUserData.sql).

The problem was that there was no way of keeping the two files in sync or making sure that the exact same data was exported and deleted, and that was the initial impetus for the current solution. Here, the two files in `common/` identify the tables to search and the ids to search under, and these are exported as functions for use in both the export and delete loops.

- **00-scoped_table_lists.sql:** manually groups all the tables in LiteFarm (including legacy tables that had some sort of data on prod) according to the primary identifier (`task_id`, `location_id`, `farm_id`) that can be traced back to a farm or user. The order of tables is catered to deletion, so that the records can be removed top to bottom without triggering FK violations 
- **01-user_data_indices.sql:** this script handles finding the actual record keys that we can delete / export on. It's slightly ugly, which I guess you could say comes from LiteFarm's own messiness when it comes to tying records to farms (especially when it comes to tasks).

Export and delete are kind of a messy mix (sorry) between SQL loops and manual writing-out. In essence I was just trying to programatically perform the loops described above, but I'm not that good at SQL, and in the end both export and delete run through the different table types in `00-scoped_table_lists.sql` manually, rather than in a programmatic way. That makes them kind of long to read through, but I'm not very good at SQL and I think writing any more SQL functions would have made them even harder to reason about.

The upside of the current approach is that the only file that should have to be updated in the future as our schema changes is `00-scoped_table_lists.sql`, particularly if we are mindful (and I think we should be) of making sure any new tables we create can be mapped to `farm_id` without too many joins. You can see that some tables (`soil_amendment_task_products_purpose_relationship` for instance) don't have this characteristic and I think in general we might want to avoid that level of complication in the future?

## Using the four scripts

The scripts should be run on your db in order. I just used the SQL query feature of Postico and copy-pasted each script there (into its own query file):

1. Run the `00-scoped_table_lists.sql` to populate the db with the getter functions which return the table lists
2. Run `01-user_data_indices.sql` to populate the db with `get_user_data()` and its required variables

It's not necessary for the flow, but if you want to inspect the identifiers that get output from `get_user_data()`, you can do that with:

```sql
SELECT to_jsonb(get_user_data) AS user_data
FROM get_user_data('user_id_here');
```

3. Copy `02-export_user_data.sql`, populate the correct `user_id` at the top of the file, and run. To save the output, I have been just right-clicking on the data > Copy special > Copy JSON. You could also just expand the data and select all. Using Copy JSON does create an unnecessary top-level array and key "export", but they can both be deleted.

Note: [jsonb doesn't preserve key order](https://www.postgresql.org/docs/current/datatype-json.html) so the order of tables under farm / user come out completely random. To look at them myself I have been sorting the JSON keys alphabetically with a Node script -- not included in the PR -- but I don't know if that is really necessary for export purposes.

The final overall export structure will be like this:

```js
{
  "user_id": "<user_id_here>",
  "exported_at": "<timestamp>",
  "export_type": "all_user_data",
  "data": {
    "farms": [
      {
        /* farm_scoped tables */
        "location_tasks": [
          // ...
        ],
        "task_document": [
          // ...
        ]
        // ...
      }
      /* additional farms, if present */
    ],

    /* user_scoped tables */
    "supportTicket": [
      // ...
    ],
    // ...
    "userLog": [
      // ...
    ],
    "users": [
      // ...
    ]
  }
}

```

5. Copy `03-delete_user_data.sql`, populate the correct `user_id` at the top of the file, uncomment the safety comment (to prevent accidental running), and run. By default the `COMMIT` of the transaction is commented out so you can inspect the deletion or `ROLLBACK;`. Note that if you disconnect from the db in Postico, it will rollback, _but if you shut down Postico without disconnecting, all transactions are committed_ (I learned this the hard way!)

Deletion should not work on farms with multiple `userFarm` records -- it will error.

## Issues / Future Improvements
As mentioned above, this is wrapped up at this point because it's not a high priority and probably not worth much dev time. If we had more time, I wonder if it would have been better to either:
- Use the `information_schema.table_constraints` to grab the tables and foreign keys and do something more automated in place of the two scripts in `common/` (which would avoid having to update `00-scoped_table_lists.sql` each time we add tables)
  - to be clear, I don't expect us to keep this file updated in general, though -- I think it can be updated when a deletion request comes in
- Handle this via the API. Looking back I regret entirely that this was done in SQL; I think would have been much less painful to write and understand in JS/TS with the ORM + `.withGraphFetched()`. To do this as a controller function I think it would be necessary to create some sort of `admin` auth level in the API, but maybe we already have the precedent via the job scheduler? In any case, the one advantage of this method of SQL run directly in the db is that we are completely un-tied to release in order to perform these requests.

Not the most beautiful or readable solution, but it should tide us over for all our current data policy requests.

Jira link: https://lite-farm.atlassian.net/browse/LF-4889

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

Please see section above "Using the four scripts". I targetted a user account with a few single-user farms (you can also delete `userFarm` records to make an existing multi-user farm single-user) and ran the deletion without commit.

**Checklist:**

### DIFF
--- a/packages/api/db/user_data_requests/03-delete_user_data.sql
+++ b/packages/api/db/user_data_requests/03-delete_user_data.sql
@@ -50,15 +50,15 @@ user_data := get_user_data(target_user_id);
 
 FOREACH farm_data IN ARRAY user_data.farms LOOP
     -- prevent deletion of multi-user farms
-    SELECT COUNT(*) INTO farm_count
+    SELECT COUNT(*) INTO user_count
         FROM "userFarm"
         WHERE farm_id = farm_data.farm_id;
 
-        IF farm_count > 1 THEN
+        IF user_count > 1 THEN
         RAISE EXCEPTION
             'Cannot delete: farm % has % users',
             farm_data.farm_id,
-            farm_count;
+            user_count;
         END IF;
 
     FOR task_prod_tables IN SELECT * FROM get_task_product_tables() LOOP
--- a/packages/api/db/user_data_requests/03-delete_user_data.sql
+++ b/packages/api/db/user_data_requests/03-delete_user_data.sql
@@ -28,7 +28,7 @@ DO $$
 DECLARE
     target_user_id VARCHAR := 'user_id_here';
     user_data        user_data_collections;
-    farm_item        farm_data_collections;
+    farm_data        farm_data_collections;
     farm_count       INTEGER;
 
     task_tables          RECORD;
@@ -48,16 +48,16 @@ DECLARE
 BEGIN
 user_data := get_user_data(target_user_id);
 
-FOREACH farm_item IN ARRAY user_data.farms LOOP
+FOREACH farm_data IN ARRAY user_data.farms LOOP
     -- prevent deletion of multi-user farms
     SELECT COUNT(*) INTO farm_count
         FROM "userFarm"
-        WHERE farm_id = farm_item.farm_id;
+        WHERE farm_id = farm_data.farm_id;
 
         IF farm_count > 1 THEN
         RAISE EXCEPTION
             'Cannot delete: farm % has % users',
-            farm_item.farm_id,
+            farm_data.farm_id,
             farm_count;
         END IF;
 
@@ -66,7 +66,7 @@ FOREACH farm_item IN ARRAY user_data.farms LOOP
             'DELETE FROM %I WHERE task_products_id = ANY($1)',
             task_prod_tables.table_name
         )
-        USING farm_item.task_products_ids;
+        USING farm_data.task_products_ids;
     END LOOP;
 
     FOR task_tables IN SELECT * FROM get_task_tables() LOOP
@@ -74,7 +74,7 @@ FOREACH farm_item IN ARRAY user_data.farms LOOP
             'DELETE FROM %I WHERE task_id = ANY($1)',
             task_tables.table_name
         )
-        USING farm_item.task_ids;
+        USING farm_data.task_ids;
     END LOOP;
 
     FOR pmp_tables IN SELECT * FROM get_pmp_tables() LOOP
@@ -82,7 +82,7 @@ FOREACH farm_item IN ARRAY user_data.farms LOOP
             'DELETE FROM %I WHERE planting_management_plan_id = ANY($1)',
         pmp_tables.table_name
     )
-        USING farm_item.pmp_ids;
+        USING farm_data.pmp_ids;
     END LOOP;
 
     FOR management_plan_tables IN SELECT * FROM get_management_plan_tables() LOOP
@@ -90,7 +90,7 @@ FOREACH farm_item IN ARRAY user_data.farms LOOP
             'DELETE FROM %I WHERE management_plan_id = ANY($1)',
         management_plan_tables.table_name
     )
-        USING farm_item.management_plan_ids;
+        USING farm_data.management_plan_ids;
     END LOOP;
 
     FOR plan_repetition_tables IN SELECT * FROM get_mp_repetition_tables() LOOP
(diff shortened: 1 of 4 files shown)


packages/api/db/user_data_requests/03-delete_user_data.sql
```diff
@@ -0,0 +1,185 @@
...
+    farm_tables          RECORD;
+    created_by_tables    RECORD;
+    user_tables          RECORD;
+
+BEGIN
+user_data := get_user_data(target_user_id);
+
+FOREACH farm_data IN ARRAY user_data.farms LOOP
+    -- prevent deletion of multi-user farms
+    SELECT COUNT(*) INTO farm_count
```
SayakaOno: Should `farm_count` be `user_count`?

SayakaOno (commented):
Great job!!!
I'm not very good at SQL either, so I really appreciated that you didn’t use anything too fancy -- It was easy to follow. I'm not sure how you could've done it better!

I tested with my main user, but deleting the user failed. 
It looks like some tasks weren't deleted, and none of them had a record in `location_tasks`.
After running the delete script up to the user_tables deletion, I still had cleaning tasks without a location, and some plant and transplant tasks.

Please let me know if you need more details!

kathyavini:
@SayakaOno hmmmmmm, I dunno, I'm a bit worried 😅 A cleaning task without a `location_task` entry shouldn't be possible, correct? That has to be a testing data situation. But I'm nervous about the planting and transplant tasks. What do they look like in the frontend/in the response body from API? Anything unusual about them? We can also take a look in tech daily on Monday!

SayakaOno (approved):
@kathyavini Ah sorry, by "cleaning task without a location_task", I meant the ones created for wild crops!

Okay, I confirmed all the records were bad.
- The cleaning tasks didn't have records in `management_tasks`.
- The plant and transplant tasks didn't even have records in plant_task and transplant_task...!

After cleaning those up, the delete script ran successfully.
I'm so sorry for the false alarm! 🙏🙏🙏

PR #3839, opened 2025-07-23
LF-3134: Cleaning tasks not updating null values on task complete

SayakaOno:
**Description**

This PR fixes an issue where null values were not being added (e.g. product-related fields in cleaning and pest control tasks) to the DB on task completion. The major cause was the frontend not sending null values. The backend has also been updated to accept them.

Details:
- Update `cleaningTaskModel` and `pestControlTaskModel` to accept `null` for `product_quantity`.
- Add tests for task completion. (`npm run test /task.test`)
- Update `CleaningTask` and `PestControlTask` components to properly clear product info when removed.
- Update `hookFormPersistUnMount` (triggered during page transitions) not to remove `null` values for specified properties. (`propertiesToKeep` parameter is added to the `getCorrectedPayload` function)

NOTE:
- For irrigation tasks, there's currently no logic on either frontend or backend to remove irrelevant fields based on the selected method type. These fields are hidden in the UI, so I'll leave this as is.
- I plan to add tests for plant and transplant tasks later when I work on task re-completion tests. (I didn't have the energy left for that in this PR...)

Jira link: https://lite-farm.atlassian.net/browse/LF-3134

**Type of change**

**How Has This Been Tested?**

- [x] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/api/tests/utils/taskCompletionTestCases.js
+++ b/packages/api/tests/utils/taskCompletionTestCases.js
@@ -1,0 +1,271 @@
+/*
+ *  Copyright 2025 LiteFarm.org
+ *  This file is part of LiteFarm.
+ *
+ *  LiteFarm is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  LiteFarm is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ */
+
+import { faker } from '@faker-js/faker';
+import knex from '../../src/util/knex.js';
+import { setupSoilAmendmentTaskDependencies } from './testDataSetup.js';
+
+// field_work_task
+const fieldWorkName = faker.lorem.word();
+const fieldWorkTaskTestCases = {
+  newFieldWorkType: {
+    getFakeCompletionData: (initialData) => ({
+      field_work_task: {
+        task_id: initialData.task_id,
+        field_work_task_type: {
+          label: 'Other',
+          value: 'OTHER',
+          field_work_name: fieldWorkName,
+        },
+      },
+    }),
+    getExpectedData: async () => {
+      const { field_work_type_id } = await knex('field_work_type')
+        .where({ field_work_name: fieldWorkName })
+        .first();
+
+      return { field_work_type_id };
+    },
+  },
+};
+
+// pest_control_task
+const pestControlTaskInitialData = {
+  weight: 2,
+  weight_unit: 'kg',
+  pest_target: 'Aphids',
+  control_method: 'foliarSpray',
+};
+const pestControlTaskTestCases = {
+  changeToMethodWithoutProduct: {
+    getFakeCompletionData: (initialData) => ({
+      pest_control_task: {
+        task_id: initialData.task_id,
+        product_quantity: null,
+        product_quantity_unit: null,
+        pest_target: null,
+        control_method: 'biologicalControl',
+      },
+    }),
+    getExpectedData: async () => {
+      return {
+        weight: null,
+        weight_unit: null,
+        volume: null,
+        volume_unit: null,
+        pest_target: null,
+        control_method: 'biologicalControl',
+      };
+    },
+  },
+};
+
+// cleaning_task
+const cleaningTaskInitialData = {
+  cleaning_target: 'Benches',
+  agent_used: true,
+  water_usage: 25,
+  water_usage_unit: 'l',
+  volume: 500,
+  volume_unit: 'ml',
+};
+const cleaningTaskTestCases = {
+  nullOptionalFields: {
+    getFakeCompletionData: (initialData) => ({
+      cleaning_task: {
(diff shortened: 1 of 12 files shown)


packages/webapp/src/components/Task/CleaningTask/index.jsx
```diff
@@ -28,9 +28,9 @@ const PureCleaningTask = ({
 
   useEffect(() => {
     if (isCleaningAgentUsed === false && getValues('cleaning_task.product')) {
-      setValue('cleaning_task.product', null);
+      setValue('cleaning_task.product', undefined);
       setValue('cleaning_task.product_id', null);
-      setValue('cleaning_task.product_quantity', undefined);
+      setValue('cleaning_task.product_quantity', null);
```
SayakaOno, line 33: `cleaning_task.product` should not be sent to the backend, but `cleaning_task.product_quantity: null` should.

kathyavini:
@SayakaOno I'm not done reviewing the test files, but can I check on one part regarding the frontend object?

I was testing a cleaning task that had a product defined, and only changed the supplier field on complete. This causes an API crash for me with `ValidationError: product_translation_key: must be string` -- do you see that as well?

The original request looked like this for me on the product part:

```js
  "cleaning_task": {
    "product": {
      "product_id": 134,
      "name": "Vinegar",
      "supplier": "Updated supplier",
      "type": "cleaning_task",
      "farm_id": "f07b2536-4182-11f0-a420-acde48001122"
    },
    "task_id": 610,
    "product_id": 134,
```

and with the change to the kept `null` values it now looks like this

```js
  "cleaning_task": {
    "product": {
      "product_id": 134,
      "name": "Vinegar",
      "product_translation_key": null,
      "supplier": "Updated supplier",
      "on_permitted_substances_list": null,
      "type": "cleaning_task",
      "farm_id": "f07b2536-4182-11f0-a420-acde48001122",
      "soil_amendment_product": null
    },
    "task_id": 610,
    "product_id": 134,
 ```
 
The crash is because `createOrPatchProduct()` is missing a `try`/`catch` (which is an easy fix), but I wans't sure if the frontend was supposed to be sending those extra `nulls` within product or not 🤔

**Comment by SayakaOno** [comment]
@kathyavini
Thank you so much for catching that! I was expecting it might affect something...

I think sending a value only for properties that should be updated (including `null`) is the right approach.
I haven't looked into the backend yet, but if it's not expecting those properties, the request body should probably be formatted before the API call.

I'll update the model if that makes sense, but do you agree that we should format the request before the API call?
I'm wondering if updating `getCorrectedPayload` to preserve all nulls under `cleaning_task` was a bad idea.

**Comment by kathyavini** [comment]
@SayakaOno I don't know, I think I was initially wondering if the response body DID needed formatting because of the inclusion of the `soil_amendment_product`, which isn't on the product table, but I see that's because it's added to the graph and returned from the API on the product endpoint.

Seeing that actually what we're sending matches the shape of the product API response exactly makes me feel like it's okay to just leave it exactly as it is now -- seems simpler. Functionally, it seems to work perfectly fine with just the model change on that translation key property!

(However, I think I would like error handling -- probably the nicely formatted `handleObjectionError` -- added to the `createOrPatchProduct()`! I was really surprised when a model validation crashed the backend 🤣)

**Comment by SayakaOno** [comment]
@kathyavini Thank you for looking into it! I've updated the model and added a try catch block to `createOrPatchProduct`.

I'd appreciate a re-review, thank you! 🙏

File: packages/api/tests/utils/taskUtils.js [hunk 2]
```diff
@@ -240,3 +240,32 @@ export function expectTaskCompletionFields(
   expect(task.happiness).toBe(happiness);
   expect(task.completion_notes).toBe(completion_notes);
 }
+
+export const taskWithLocationFactory = async ({ userId, locationId, taskTypeId, farmId }) => {
```
kathyavini, line 244: Nice! I was thinking we needed this, but just went ahead and made another non-general case in #3840. Now I want to refactor `irrigationTaskGenerator()` to use this one, though!

Edit: done 🙂

packages/api/tests/utils/taskCompletionTestCases.js
```diff
@@ -0,0 +1,267 @@
...
+        irrigation_type_name: irrigationTypeName,
+      };
+    },
+  },
+  switchMeasuringType: {
+    initialData: irrigationTaskInitialData,
+    getFakeCompletionData: (initialData) => ({
+      irrigation_task: {
+        task_id: initialData.task_id,
+        irrigation_type_id: initialData.irrigation_type_id,
```
kathyavini, line 161: Okay in this case, the initial data IS just `null` though, right, for `irrigation_type_id` and `irrigation_type_name`? I guess `irrigation_type_name` being in the `irrigationTaskFakeCompletionData` threw me off a little about what is being tested here, since this also creates a new irrigation type, same as the other test case. I don't think it's wrong, but just wanted to make sure that was intended and I'm understanding correctly.

SayakaOno: I struggled a bit while writing tests for the irrigation task and probably got confused trying different values. I added `irrigation_type_name` as a workaround, but the test is passing without it...😓 
Sorry for the confusion, I’ll go through it again and clean things up. Thank you for catching the details!

SayakaOno:
Thank you Joyce!
The debugging behaviour you described feels like what happens when there's a defect in the test. I'm not sure if what I'm seeing is exactly the same, but the soil amendment task test might have an issue...
It would be great if you could show me what you saw when we have time in tech daily!

PR #3840, opened 2025-07-23
LF-4888 Modify task deletion to accept IP id and update task status on deletion

kathyavini:
**Description**

This PR adds patching an Ensemble IP as `approved: false` after we delete a task that had an `irrigation_prescription_external_id` associated with it, keeping the Ensemble state in sync with LiteFarm.

The PR includes:
- tiny update to `patchIrrigationPrescriptionApproval()` so we can use it for passing both kinds of patch
- the `triggerPostTaskDeletedActions()` side-effect function that mirrors `triggerPostTaskCreatedActions`
  - The logic is slightly more involved only because the deleted task (unlike the created task) is not returned with the whole graph of task type + task type-specific tables
  - this missing graph is also the reason for the slight modification of the `deleteTaskAndTaskProduct` model method, to return some task type information
- test case + one supporting utility function for it
- two minor bug fixes I'll indicate in-line. Neither are specific to this feature, but I noticed both while working on it

Jira link: https://lite-farm.atlassian.net/browse/LF-4888

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

The call to Ensemble is silent from the app perspective, but you can debugger on the `onError` of the patch request, or watch your API console to see the 400 error that (currently) results from this request, which happens to be

```json
{
    "detail": "Prescription already approved"
}
```

Ultimately we expect a 200 or 204 from this once they update their endpoint.

**Checklist:**

### DIFF
--- a/packages/api/tests/mock.factories.js
+++ b/packages/api/tests/mock.factories.js
@@ -1324,8 +1324,11 @@ function fakePesticide(defaultData = {}) {
 }
 
 function fakeTaskType(defaultData = {}) {
+  const task_name = faker.lorem.word();
+  const task_translation_key = defaultData.task_translation_key ?? task_name;
   return {
-    task_name: faker.lorem.word(),
+    task_name,
+    task_translation_key,
     ...defaultData,
   };
 }
--- a/packages/webapp/src/containers/Task/TaskReadOnly/index.jsx
+++ b/packages/webapp/src/containers/Task/TaskReadOnly/index.jsx
@@ -65,7 +65,7 @@ function TaskReadOnly({ history, match, location }) {
   if (externalIrrigationPrescription?.prescription?.vriData?.file_url) {
     files.push({ url: externalIrrigationPrescription.prescription.vriData.file_url });
   }
-  if (task.documents?.length) {
+  if (task?.documents?.length) {
     const documentFiles = task.documents.flatMap((doc) => doc.files);
     files.push(...documentFiles);
   }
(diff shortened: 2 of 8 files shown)


packages/api/tests/mock.factories.js
```diff
@@ -1324,8 +1324,11 @@ function fakePesticide(defaultData = {}) {
 }
 
 function fakeTaskType(defaultData = {}) {
+  const task_name = faker.lorem.word();
+  const task_translation_key = defaultData.task_translation_key ?? task_name;
   return {
-    task_name: faker.lorem.word(),
+    task_name,
+    task_translation_key,
```
kathyavini, line 1331: I don't believe this has caused any problems with the tests yet, but this resulted in mock custom task types having `null` for their task translation keys. In the actual application, no task types have a `null` value -- custom task types have a `task_translation_key` equivalent to their name.

packages/webapp/src/containers/Task/TaskReadOnly/index.jsx
```diff
@@ -65,7 +65,7 @@ function TaskReadOnly({ history, match, location }) {
   if (externalIrrigationPrescription?.prescription?.vriData?.file_url) {
     files.push({ url: externalIrrigationPrescription.prescription.vriData.file_url });
   }
-  if (task.documents?.length) {
+  if (task?.documents?.length) {
```
kathyavini, line 68: 100% unrelated, but since I was testing deletion I noticed this.

This line, which is on prod since last release 😬, occurs before the `unknown_record` redirect, and will cause a crash if you try to e.g. access a notification that points to a deleted task.

SayakaOno (approved):
Looks great!!

PR #3842, opened 2025-07-24
LF-4897 Make final adjustments on URI irrigation prescriptions: rounding, date filtering, and farm filtering

kathyavini:
**Description**

This PR includes two minor frontend adjustments:
- applying `roundToTwo()` on the data in the IP table (soil moisture content + application depth)
- showing only today + future IPs in the IP table using `selectFromResult` on the list GET

On the backend, the changes are also minor but across more files:
- updates to the types for the new properties (`system_id`, `system_name`, and `farm_id`) being given to us
- a minor adjustment to the filtering on `farm_id` and `location_id` on the list GET returns (used in notifications `useGetIrrigationPrescriptionsQuery`)

The new filtering is definitely better because it queries the db for `farm_id` and active (non-deleted) locations in one pass instead of within a loop per prescription... but I realized when looking at this that _asking Ensemble to return the `farm_id` was totally unnecessary_ 😬 We could always have just limited our search of locations to the controller-defined `farm_id`, regardless of whether this is a property Ensemble returned. In the end we don't even need to use the Ensemble property... oops!

Jira link: https://lite-farm.atlassian.net/browse/LF-4897

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/containers/LocationDetails/LocationIrrigation/useIrrigationPrescriptions.ts
+++ b/packages/webapp/src/containers/LocationDetails/LocationIrrigation/useIrrigationPrescriptions.ts
@@ -34,7 +34,9 @@ export default function useIrrigationPrescriptions(location?: Location) {
   const { prescriptionList = [] } = useGetIrrigationPrescriptionsQuery(undefined, {
     selectFromResult: ({ data }) => ({
       prescriptionList: data?.filter(
-        (prescription) => prescription.recommended_start_date >= today,
+        (prescription) =>
+          prescription.recommended_start_date >= today &&
+          prescription.location_id === location.location_id,
       ),
     }),
   });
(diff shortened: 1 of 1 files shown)


packages/webapp/src/containers/LocationDetails/LocationIrrigation/useIrrigationPrescriptions.ts
```diff
@@ -29,12 +29,20 @@ export default function useIrrigationPrescriptions(location?: Location) {
     return [];
   }
 
-  const { data = [] } = useGetIrrigationPrescriptionsQuery();
+  const today = new Date().toISOString().split('T')[0];
+
+  const { prescriptionList = [] } = useGetIrrigationPrescriptionsQuery(undefined, {
+    selectFromResult: ({ data }) => ({
+      prescriptionList: data?.filter(
+        (prescription) => prescription.recommended_start_date >= today,
```
SayakaOno: Could the filter on line 46 be combined with this one?
I'm still figuring out when it's better to filter inside the query vs outside...

kathyavini: Thank you!!

> I'm still figuring out when it's better to filter inside the query vs outside...

I think `selectFromResult` is the ideal place to filter. (Although granted I think in a component like this it doesn't matter _too_ much because I don't think the memoizing is super necessary?) But I've been totally guilty of not using it! I think you set up the other two we have in the codebase already 🙏  I want to try to use it more!

PR #3845, opened 2025-07-31
LF-4905: Cannot complete animal movement task

SayakaOno:
**Description**

Fix backend to accept empty string `""` for `other_purpose` when completing animal movement tasks.
- Update `AnimalMovementTaskPurposeRelationshipModel` to remove `other_purpose` `minLength` to allow `""`
- Fix `createMovementTaskForReqBody` function used for testing movement task completion
- Pre-commit flagged unused variables in the test file, so I removed them. The rest of the diffs are from the auto-formatting 🙏 

Jira link: https://lite-farm.atlassian.net/browse/LF-4905

**Type of change**

**How Has This Been Tested?**

- [x] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/api/tests/animalTask.test.js
+++ b/packages/api/tests/animalTask.test.js
@@ -17,12 +17,11 @@ import chai from 'chai';
 import chaiHttp from 'chai-http';
 
 chai.use(chaiHttp);
-import server from '../src/server.js';
 import knex from '../src/util/knex.js';
 
 jest.mock('jsdom');
 jest.mock('../src/middleware/acl/checkJwt.js', () =>
-  jest.fn((req, res, next) => {
+  jest.fn((req, _res, next) => {
     req.auth = {};
     req.auth.user_id = req.get('user_id');
     next();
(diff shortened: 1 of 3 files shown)


packages/api/tests/animalTask.test.js
```diff
@@ -68,15 +64,11 @@ const createFakeMovementTask = (purposeRelationships) => {
...
 
-const createMovementTaskForReqBody = (purposeIds, otherPurpose) => {
-  const animalMovementTask = {};
-
-  if (purposeIds) {
-    animalMovementTask.purpose_ids = purposeIds;
-  }
-  if (otherPurpose) {
-    animalMovementTask.other_purpose = otherPurpose;
-  }
```
SayakaOno, line 79: The test "should complete a movement task with deleting other_purpose" should have caught the bug. This function was treating `""` as `undefined`, dropping it; I updated it so it preserves both `null` and `""`.

kathyavini:
I don't think we actually want to preserve an empty field as `""` in the database the way it's sent over in the request now. There's not a huge amount of other purpose data on prod, but an un-filled "other purpose" has been always been `null` so far, and I think that's our preference anyway for skipped fields.

Could we instead use `checkAndTrimString()` on the `other_purpose` field within `formatAnimalMovementTaskForDB` instead of removing the min length from model?

SayakaOno:
Thank you Joyce, it's much cleaner now!! 🙏

PR #3848, opened 2025-08-01
LF-4910 Better naming convention to distinguish notifications

kathyavini:
**Description**

This PR adds location and start date into the title and body of Irrigation Prescription notifications. Notifications already have a property `variables` for values meant to be interpolated into the translation strings, so I used that.

Notes:
- The location name is fetched on the backend, which means it is stored on the notification record, and won't update if the location is subsequently edited. In other words, the notification title will refer to the name of the location at the time the notification was created. I couldn't decide if this was preferable UX to the location name updating, but since it matches how task assignee names are handled, I left it be
- I did not bother thinking up a solution for the existing notifications (which are only on local and beta, so not relevant to prod). Therefore, those will look as follows:

Screenshot 2025-07-31 at 6 26 55 PM

- CrowdIn will have to be updated, although if a language is not updated, the current notification strings will be shown in other languages, which is more or less fine. There is also one language (Punjabi) where I recommend using the current string anyway because the `Date.IntlDateTimeFormat` is busted in Punjabi. I will leave a comment in CrowdIn.
- Because this is a pre-release PR, I wouldn't normally have done any refactoring of the inline styles to SCSS, but `<PureNotificationCard />` was a _really_ egregious example of inline styling (so much inline styling!!) which made it hard to work with. So I did refactor to an SCSS module. If it helps for reading, the functional changes are only here: [styles](https://github.com/LiteFarmOrg/LiteFarm/commit/710d480f7248e0bf696e217beb2d7547b01ed2b9) and [component](https://github.com/LiteFarmOrg/LiteFarm/commit/f6f194052045f4e3cc6d55da7b2a613be5dacaa9)

Jira link: https://lite-farm.atlassian.net/browse/LF-4910

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

Tested by manually clearing today's notifications from db, and re-generating.

**Checklist:**

### DIFF
--- a/packages/api/src/models/locationModel.js
+++ b/packages/api/src/models/locationModel.js
@@ -315,17 +315,17 @@ class Location extends baseModel {
   }
 
   /**
-   * Return the name of a location by its id, with an empty‐string fallback.
-   *
-   * @param {string} location_id - The id of the location to look up.
-   * @param {Knex.Transaction} [trx] - Optional transaction object.
+   * Retrieves the name of a location by its id
+   * @param {string} location_id - The id of the location to look up
+   * @param {Knex.Transaction} [trx] - Optional transaction object
    * @static
    * @async
-   * @returns {Promise<{ name: string }>}
+   * @returns {Promise<string|null>}
+   *   Resolves to the location name if found, otherwise null
    */
   static async getLocationNameById(location_id, trx) {
-    const location = await Location.query(trx).findById(location_id).select('name').first();
-    return location || { name: '' };
+    const location = await Location.query(trx).findById(location_id).select('name');
+    return location?.name ?? null;
   }
 }
 
--- a/packages/api/src/controllers/timeNotificationController.js
+++ b/packages/api/src/controllers/timeNotificationController.js
@@ -257,7 +257,7 @@ async function sendDailyNewIrrigationPrescriptionNotification(
   }
   const todayStr = today.toISOString().split('T')[0];
 
-  const { name: locationName } = await LocationModel.getLocationNameById(location_id);
+  const locationName = await LocationModel.getLocationNameById(location_id);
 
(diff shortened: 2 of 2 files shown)


packages/api/src/models/locationModel.js
```diff
@@ -313,6 +313,20 @@ class Location extends baseModel {
...
+   *
+   * @param {string} location_id - The id of the location to look up.
+   * @param {Knex.Transaction} [trx] - Optional transaction object.
+   * @static
+   * @async
+   * @returns {Promise<{ name: string }>}
+   */
+  static async getLocationNameById(location_id, trx) {
+    const location = await Location.query(trx).findById(location_id).select('name').first();
+    return location || { name: '' };
```
SayakaOno: It seems like `first` is unnecessary since `findById` _"finds a single item by id"_ ([doc](https://vincit.github.io/objection.js/api/query-builder/find-methods.html#findbyid))

I'm just curious, is returning an object the usual pattern?

kathyavini: Thank you!! I thought that was a `.where()` clause still -- I wrote it two ways and it ended up a gross hybrid 🤮

As for the default object, I was try to be lazy about having to destructure it safely in the controller. But it's weird, isn't it? I'll return it normally!

Edit: or, maybe return just the string? I think I prefer that.

SayakaOno: Sorry, yes! Returning the string seemed simpler to me!

kathyavini:
> I noticed that the back button on the IP doesn’t work the first time when landing from the notification

@SayakaOno Uh oh, I see that too! And also weird back button behaviour for the notifications that lead to `unknown_record`

SayakaOno (approved):

PR #3857, opened 2025-08-11
LF-4812: Show button to re-trigger completion in read only task view

SayakaOno:
**Description**

Add `RevisionPrompt` component and show it in task read-only view when the task has been completed and the user can complete the task.

Jira link: https://lite-farm.atlassian.net/browse/LF-4812

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/public/locales/en/translation.json
+++ b/packages/webapp/public/locales/en/translation.json
@@ -1707,6 +1707,7 @@
     }
   },
   "REVISION_PROMPT": {
+    "FIX_IT_NOW": "Fix it now",
     "NOTICED_SOMETHING_OFF": "Noticed something off?",
     "UPDATE_THIS_TASK": "You can revisit and update the info you submitted for this task."
   },
--- a/packages/webapp/public/locales/en/translation.json
+++ b/packages/webapp/public/locales/en/translation.json
@@ -1706,6 +1706,10 @@
       "REVENUE_TYPE": "Revenue type"
     }
   },
+  "REVISION_PROMPT": {
+    "NOTICED_SOMETHING_OFF": "Noticed something off?",
+    "REVISIT_AND_UPDATE_THIS_ENTITY": "You can revisit and update the info you submitted for {{entity}}."
+  },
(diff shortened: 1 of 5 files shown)


packages/webapp/public/locales/en/translation.json
```diff
@@ -1706,6 +1706,10 @@
       "REVENUE_TYPE": "Revenue type"
     }
   },
+  "REVISION_PROMPT": {
+    "NOTICED_SOMETHING_OFF": "Noticed something off?",
+    "REVISIT_AND_UPDATE_THIS_ENTITY": "You can revisit and update the info you submitted for {{entity}}."
```
kathyavini: Ahhh, I love string interpolation as a component pattern, but we have four languages (`hi`, `pa`, `ml`, `de`)  where "this task" would have to take a different grammatical case when it's paired with "for" like in the sentence above. It could work if the "for" was paired with the entity noun, like:

```json
"...": "You can revisit and update the info you submitted {{for_entity}}."
```

along with

```json
"FOR_THIS_TASK": "for this task"
```

but since that makes the entity noun ("for this task") less reusable outside of this context, maybe it might be cleaner at that point to just keep the whole string as a prop?

kathyavini (commented):
Thank you for the sequential un-drafting; it makes it really easy to review ❤️ 

This looks great, and so super clean in the task; nice👌 Just left a comment not about the code, but the i18n strings.

packages/webapp/public/locales/en/translation.json
SayakaOno: Thank you!! I was hoping it would work 😅
I've updated this and also moved "FIX_IT_NOW" to this block.
Thank you for re-reviewing! 🙏

kathyavini: Oh, I didn't realize until you made the change, but I really like all the translations being within a single block like that 😁

kathyavini (approved):

PR #3859, opened 2025-08-11
LF-4817(2): Tests for revision date and revision author when re-completing a task

SayakaOno:
**Description**

Add tests for task re-completion and update task completion tests.

- mock.factories
   - Rename `activity` (task's old name) to `task`
   - Update `irrigation_taskFactory` to include `irrigation_type_id` by default
     - Although `null` is allowed, but it never actually becomes `null`
     - This change addresses Joyce's [comment](https://github.com/LiteFarmOrg/LiteFarm/pull/3839#discussion_r2237687757) on irrigation task completion tests and should help write cleaner tests
- task.test
   - replace the `for` loop in the task completion tests with `test.each` to fix strange debugger behaviour ([reference](https://github.com/LiteFarmOrg/LiteFarm/pull/3839#pullrequestreview-3064358642))
     - Test cases were formatted first; besides that, the only change is to make the last `expect` more flexible to handle objects
     - Add tests for task re-completion
- taskCompletionTestCases
   - add tests for soil sample tasks
   - add helper functions to create test cases for field and irrigation tasks to avoid reusing field work type and irrigation type names, which causes re-completion test failures
   - Add re-completion test cases

Jira link: https://lite-farm.atlassian.net/browse/LF-4817

**Type of change**

**How Has This Been Tested?**

- [x] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/api/tests/task.test.js
+++ b/packages/api/tests/task.test.js
@@ -40,10 +40,14 @@ import {
   completeTaskRequest as completeTaskRequestAsync,
   deleteTaskRequest as deleteTaskRequestAsync,
   taskWithLocationFactory,
+  commonTaskTypes,
 } from './utils/taskUtils.js';
 import { setupFarmEnvironment } from './utils/testDataSetup.js';
 import { connectFarmToEnsemble } from './utils/ensembleUtils.js';
-import { taskCompletionFieldUpdateTestCases } from './utils/taskCompletionTestCases.js';
+import {
+  taskCompletionFieldUpdateTestCases,
+  taskRecompletionTestCases,
+} from './utils/taskCompletionTestCases.js';
 
 describe('Task tests', () => {
   function assignTaskRequest({ user_id, farm_id }, data, task_id, callback) {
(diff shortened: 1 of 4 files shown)


packages/api/tests/task.test.js
```diff
@@ -3036,65 +3040,170 @@ describe('Task tests', () => {
...
-                locationId: location_id,
-                farmId: farm_id,
-              });
-
-              // Create a task-type-specific record (e.g. soil_amendment_task, cleaning_task, etc.)
-              const [initialTaskTypeDataInDB] = await mocks[`${taskType}Factory`](
-                { promisedTask: [{ task_id }] },
-                initialTaskTypeData,
-              );
+          const formattedTestCases = Object.entries(testCases).flatMap(
```
kathyavini, line 3043: Thank you for solving the debugger weirdness! This is great ❤️

packages/api/tests/task.test.js
```diff
@@ -3036,65 +3040,170 @@ describe('Task tests', () => {
...
+        const fakeRecompletionData = {
+          complete_date: faker.date.recent().toISOString().split('T')[0],
+          duration: Math.floor(Math.random() * 12 * 60) + 15,
+          happiness: Math.floor(Math.random() * 5) + 1,
+          completion_notes: faker.lorem.sentence(),
+        };
+
+        const fakeReqBody = {
+          ...fakeRecompletionData,
+          ...(getFakeTaskTypeCompletionData?.(
```
kathyavini, line 3176: In the context of this test, where both completion and recompletion occur, could `getFakeCompletionData` be renamed to something like `getFakeTaskTypeRecompletionData` to be clear it's the re-completion payload? Might be just me, but I read it as the actual completion data (i.e. previous state) at first, and then didn't understand the expectations!

packages/api/tests/task.test.js
```diff
@@ -3036,65 +3040,170 @@ describe('Task tests', () => {
...
+        const recompleteTaskRes = await completeRequest(fakeReqBody);
+        expect(recompleteTaskRes.status).toBe(200);
+        const recompletedTask = await knex('task').where({ task_id }).first();
+        expectTaskCompletionFields(recompletedTask, fakeRecompletionData);
+        expect(previousRevisionDate < new Date(recompletedTask.revision_date).getTime()).toBe(true);
+        expect(recompletedTask.revised_by_user_id).toBe(user_id);
+        previousRevisionDate = new Date(recompletedTask.revision_date).getTime();
+
+        const expectedTaskTypeData =
+          (await getExpectedTaskTypeData?.(taskTypeDataBeforeRecompletion)) || {};
```
kathyavini, line 3191: Similarly, could this `getExpectedTaskTypeData` be called something like `getExpectedDataAfterRecompletion`?

With `taskTypeDataBeforeRecompletion` being the only argument into the function, my first impulse was to miss the transformation aspect and feel like it's an expectation on the original data; does that make sense?

kathyavini (approved):
Continues to be a very cool test setup! 😎

There were two variable names that tripped me up within the re-complete block, but the renames are optional -- please include in one of the associated PRs only if they make sense to you!

PR #3865, opened 2025-08-20
LF-4917 Add link to ESci report in irrigation prescription view

kathyavini:
**Description**

This adds a link to the ESci report to the Irrigation tab (above the table of prescriptions).

The link requires a lower- and snake-case formatted version of the organisation and system names, which we did not have previously, so I have added fetching that and adding both to each IP of the IP list.

ESci has added the property `url_name` to their systems and organisations, I think for this exact purpose. As far as I can tell with our current data, it looks like it's just the lower- and snake-cased version of the full name. Even so -- and even though this adds a little bit of code overhead -- I have pulled this property from ESci rather than generating it, just in case they do create it or change it manually.

The link is not shown at all if a valid link cannot be generated. Routing to a bunk/undefined system is actually not a problem, but an incorrect or missing org gives this error:

Screenshot 2025-08-20 at 10 00 35 AM

so in that case, the link is skipped entirely.

## API Test Changes
Fetching the organisation details (url_name & system url_name) involved a second call to Ensemble. I couldn't get the original mock, which was of our EnsembleAPICall function, to work with the nested call I was making:

```ts
jest.mock('../src/util/ensemble.js', () => ({
  ...jest.requireActual('../src/util/ensemble.js'),
  ensembleAPICall: jest.fn(async () => ({ data: [] })),
}));
```

It already bothered me a bit that `EnsembleAPICall` is mocked in this test, while `axios` is mocked in the two other tests that call Ensemble, so I moved this one over to an `axios` mock as well, and then wrote in an explicit call to represent the second query (of Ensemble Organistions). Note that I didn't add the logic for adding organisation data to the irrigation prescription list, so the second mock just returns an empty array. Let me know if this feels like an important thing to add to the test, but at this point adding in all the org properties seemed like quite a bit of trouble for little benefit.

Jira link: https://lite-farm.atlassian.net/browse/LF-4917

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/LocationIrrigation/index.tsx
+++ b/packages/webapp/src/components/LocationIrrigation/index.tsx
@@ -18,7 +18,6 @@ import PageTitle from '../PageTitle/v2';
 import RouterTab from '../RouterTab';
 import { useTranslation } from 'react-i18next';
 import styles from './styles.module.scss';
-import { IconLink } from '../Typography';
 import { ReactComponent as ExternalLinkIcon } from '../../assets/images/icon_external_link-01.svg';
 import { getLocalizedDateString } from '../../util/moment';
 import { Variant } from '../RouterTab/Tab';
@@ -60,12 +59,6 @@ export default function PureLocationIrrigation({
     system_url_name: irrigationPrescriptions[0]?.system_url_name,
     organisation_url_name: irrigationPrescriptions[0]?.organisation_url_name,
   });
-
-  const handleExternalLink = () => {
-    if (externalLink) {
-      window.open(externalLink, '_blank');
-    }
-  };
 
   const getColumns = (): TableV2Column[] => {
     return [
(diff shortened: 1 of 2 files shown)


packages/webapp/src/util/smartIrrigation.ts
```diff
@@ -24,3 +24,18 @@ export function createSmartIrrigationDisplayName({
...
+
+export const createESciReportLink = ({
+  system_url_name,
+  organisation_url_name,
+}: {
+  system_url_name?: string;
+  organisation_url_name?: string;
+}) => {
+  if (!organisation_url_name) {
+    // missing system name is OK; link will just redirect to a list of all systems within that org
```
SayakaOno, line 36: Do we still want to show the link when we know the user will be redirected to the systems page?

packages/webapp/src/components/LocationIrrigation/index.tsx
```diff
@@ -117,6 +131,18 @@ export default function PureLocationIrrigation({
...
+      {externalLink && (
+        <div className={styles.linkContainer}>
+          <IconLink
+            className={styles.linkText}
+            icon={<ExternalLinkIcon />}
+            isIconClickable
+            onClick={handleExternalLink}
+          >
+            {t('IRRIGATION_PRESCRIPTION.FULL_IRRIGATION_REPORT')}
+          </IconLink>
```
SayakaOno: There's no good way to use `<a>` tag here, right? 😞

SayakaOno (commented):
I'll review the changes to the tests tomorrow! But it works great!

Update: Tests look good!

packages/webapp/src/util/smartIrrigation.ts
kathyavini: I would think so, yes! That UI looks like this and it's still very useable  -- just one extra click to get to the report, and otherwise just as good as the full link.

Screenshot 2025-08-22 at 9 30 45 AM

packages/webapp/src/components/LocationIrrigation/index.tsx
kathyavini: You know, I don't know why I thought it had to be an `<IconLink />` -- probably I was thinking I couldn't put an icon into an  `<a>`, which isn't true. I'll fix that, thank you!

SayakaOno (approved):
Looks good! ❤️

PR #3868, opened 2025-08-27
LF-4815 Update logic to set animal locations during task re-completion

kathyavini:
**Description**

This updates the handling of animal locations upon re-completion of movement tasks in the following ways:

### For animals that were removed
- since their location is no longer affected by this task, they are put at the location of the latest completed task, -or-
- reset to `null` location (`unknown` in the UI) if there is no such completed task, or if the location of the most recent completed task has been retired

_I think this is the more straightforward case to understand, and is also what was discussed in Jira_

### For animals that remain in the task
- if there is a newer completed task than the (updated) date of the re-completed task, animals are moved to that location. In other words, if re-completing a task moves that date to BEFORE another completed task, then the other task's location takes precedence even though completing that task originally might not have moved animals (see final test case)

_We didn't really discuss this case in Jira and I wonder if it's intuitive to do this way?_ The ticket was originally written as if we don't have a full historical record, but for movement tasks we definitely do, so it _is_ possible to compare dates and always put the animal on the most recent place. However, it's hard to know if this is the most logical (vs magical) user experience for that weird edge case of re-ordering tasks via re-completion. What do you think @SayakaOno?

Jira link: https://lite-farm.atlassian.net/browse/LF-4815

**Type of change**

**How Has This Been Tested?**

- [x] Passes test case
- [ ] Other (please explain)

I did some testing in app, although, as would be expected since it was via Postman, the RTK Query caches and thus UI didn't update in a timely fashion, by re-submitting movement tasks with new dates and related animals/batches, e.g.

PATCH `http://localhost:5001/task/complete/animal_movement_task/{{task_id}}`

```json
{
  "duration": 15,
  "happiness": 5,
  "completion_notes": "",
  "complete_date": "2025-08-06",
  "related_animal_ids": [],
  "related_batch_ids": [77, 78]
}

```

**Checklist:**

### DIFF
--- a/packages/api/tests/animalTask.test.js
+++ b/packages/api/tests/animalTask.test.js
@@ -1302,9 +1302,70 @@ describe('Animal task tests', () => {
               knex('animal_batch').select('id', 'location_id').where('id', batchA.id).first(),
             ]);
 
-            // animalA and batchA should reflect the location of the second task (locationB)
+            // animals should reflect the location of the second task (locationB)
             expect(updatedAnimalA.location_id).toBe(locationB.location_id);
             expect(updatedBatchA.location_id).toBe(locationB.location_id);
+          });
+
+          test('should null animal locations if most recent task location is retired', async () => {
+            const {
+              animals: [animalA],
+              batches: [batchA],
+            } = await createAnimalsAndBatches({ farm_id });
+
+            const [locationA, locationB] = await createLocations(farm_id);
+
+            // Complete a movement task a week ago
+            await checkAnimalMovementWithSpecificCompleteDate({
+              locationId: locationA.location_id,
+              animals: [animalA],
+              batches: [batchA],
+              completeDate: toLocal8601Extended(dateWeekAgo),
+              expectedAnimalLocations: {
+                [animalA.id]: locationA.location_id,
+              },
+              expectedBatchLocations: {
+                [batchA.id]: locationA.location_id,
+              },
+            });
+
+            // Complete a new movement task today
+            // Animals and batches should be moved to locationB
+            const { task_id } = await checkAnimalMovementWithSpecificCompleteDate({
+              locationId: locationB.location_id,
+              animals: [animalA],
+              batches: [batchA],
+              completeDate: toLocal8601Extended(dateToday),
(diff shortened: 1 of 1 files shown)


packages/api/src/controllers/taskController.js
```diff
@@ -172,17 +173,53 @@ async function updateTaskWithCompletedData(
...
             entity.location_id = locationId;
+          } else if (isRecompleting) {
+            // Case 2: Recompleting and newer tasks (now) exist - use latest task's location
+            const [latestTaskLocationId] = await TaskModel.getTaskLocationIds(
+              newerCompletedTasks[0].task_id,
+            );
+            entity.location_id = latestTaskLocationId;
+          } else {
+            // Case 3: Not recompleting and newer tasks exist - don't change location
+            continue;
```
kathyavini, line 194: This empty `else { ` was an AI readability suggest to make all logical paths clear, and I'm undecided how I feel about it! What do you think Sayaka?

packages/api/src/models/animalBatchModel.js
```diff
@@ -241,7 +241,7 @@ class AnimalBatchModel extends baseModel {
     );
   }
 
-  static async getBatchesWithNewerCompletedTasks(batchIds, taskTypeId, completedDate) {
+  static async getBatchesWithNewerCompletedTasks(batchIds, taskTypeId, completedDate, taskId) {
```
kathyavini, line 244: Previously, limiting on a `complete_date` would have been enough to exclude the current task, since it had not yet been completed.

Now that re-completion is a possibility, the re-completed task itself can also be returned from this query (depending on how the dates were set), which is why the `task_id` for the current task was added. Ordering was added for grabbing the most recent task.

packages/api/src/controllers/taskController.js
SayakaOno: I have mixed feelings about this… I usually don’t love code that does nothing, but it’s definitely helpful for reviews. Let’s keep it 😅

kathyavini:
@SayakaOno I added a test case walking through the situation you described, since Loïc gave it a warm-neutral 😂

PR #3878, opened 2025-09-12
LF-4957 Rating, Notes, and Time Spent Not Persisted During Task Re-Completion

kathyavini:
**Description**

This one wasn't quite as smooth and fun a bugfix as the other, since everything with `hookFormPersist` completely confuses me 🤣 

But in a nutshell, the issue was that in `PureCompleteStepOne`, the default form values are taken from either the (database saved) `selectedTask` -or- the `persistedFormData`, based on `need_changes`:

```js
  const defaultsToUse = formatTaskReadOnlyDefaultValues(
    persistedFormData.need_changes ? persistedFormData : selectedTask,
  );
```

This is exactly the right behaviour for the task-specific details fields, but not for the complete fields, on which `need_changes` has no relevance (we always want the `persistedFormData` to take precedence if it exists).

When unmounting `/before_complete`, `useHookFormPersist` updates the persistedFormData will the current form values:

```js
  const {
    persistedData: { uploadedFiles },
    historyCancel,
  } = useHookFormPersist(getValues, [], getFieldsToKeep(selectedTaskType));
  ```
 
Because that logic has a step that removes `null` values (`getCorrectedPayload()` in `/hookFormPersistSlice.js`), it just so happened to always take the `/complete` form values, since complete fields are all null on `selectedTask` in the original complete flow.

In the re-complete flow, those values are NOT null, won't be discarded, and will overwrite anything that has been put in `persistedFormData` subsequently by the actual `/complete` form. I'm quite sure this is the cause although I'm not 100% sure this is the best code solution 😄 I just checked for one of the non-optional fields that is always populated once visiting the second step of the form, and if it was present, made sure the `persistedFormData` values always overwrite whatever was in the originally completed task as far as completion fields. I left the `need_changes` logic untouched as it is correct for the `/before_complete` fields.

Jira link: https://lite-farm.atlassian.net/browse/LF-4957

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/Task/TaskComplete/StepOne.jsx
+++ b/packages/webapp/src/components/Task/TaskComplete/StepOne.jsx
@@ -22,6 +22,23 @@ import PureMovementTask from '../MovementTask';
 import AnimalInventory, { View } from '../../../containers/Animals/Inventory';
 import { ANIMAL_IDS } from '../TaskAnimalInventory';
 import FilePicker from '../../FilePicker';
+import {
+  ANOTHER_DATE,
+  COMPLETION_NOTES,
+  DATE_CHOICE,
+  DURATION,
+  HAPPINESS,
+  PREFER_NOT_TO_SAY,
+} from '.';
+
+const COMPLETION_FIELDS = [
+  DURATION,
+  HAPPINESS,
+  COMPLETION_NOTES,
+  PREFER_NOT_TO_SAY,
+  DATE_CHOICE,
+  ANOTHER_DATE,
+];
 
 const soilAmendmentContinueDisabled = (needsChange, isValid) => {
   if (!needsChange) {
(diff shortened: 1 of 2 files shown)


kathyavini:
> I don't understand why "duration" is preserved in the completion flow with the current code... do you know where that comes from? 😅

@SayakaOno by current code, do you mean the code currently on integration? And by completion flow you mean the initial completion, as opposed to the re-completion? (Sorry -- just want to make sure I'm thinking of the right thing with so many permutations now 😂)

kathyavini:
@SayakaOno Ah okay, just checking!

So not just the duration, but any completion field (happiness, etc.) always persisted on the initial completion because the values for those fields in `selectedTask` are all `null`. So after running that ternary check, `hookFormPersistSlice` DOES try to update `persistedFormData` to the values in `selectedTask`, but because they all `null`, those properties get filtered out! And then nothing happens -- the form set `persistedFormData` just stays the way it was!

It's more a side-effect of the values being `null` than something that was intentional, as far as I can tell! Although maybe that went into consideration when `hookFormPersistSlice` was originally written? I think the filtering out `null` part is from the oldest/original implementation.

PR #3879, opened 2025-09-16
LF-3633: The type of the Irrigation task will be written in "UPPERCASE" after the user completes the irrigation task with making changes in it

SayakaOno:
**Description**

This was layers of mystery!

#### Fix

~~Updated the format function (`getCompleteIrrigationTaskBody`) to ensure `irrigation_type_id` is preserved and sent to the backend.~~
See my [comment below](https://github.com/LiteFarmOrg/LiteFarm/pull/3879#pullrequestreview-3241907896)!

#### Root cause
We were not sending `irrigation_type_id` to the backend. As a result, completing a task **with changes** always created a new irrigation type (100% reproducible).

#### Why it looked intermittent
The query in `getAllIrrigationTaskTypesByFarmId`:
```
SELECT * FROM (
  SELECT DISTINCT ON (UPPER(irrigation_type_name)) it.*
  FROM irrigation_type AS it
  WHERE farm_id = '${farm_id}' OR farm_id IS NULL
) AS a
ORDER BY (UPPER(irrigation_type_name)) ASC;
```
does not guarantee which row is picked for duplicate names. Sometimes it returned farm-specific types (revealing the bug), other times the defaults (hiding the bug).

**Here’s what I saw:**
I had these non-default types:

Screenshot 2025-09-16 at 9 07 11 AM

But most of the types I got were default types:

Screenshot 2025-09-16 at 9 06 31 AM

#### Why it didn't break the frontend

The frontend matches options by `irrigation_type_name` rather than `irrigation_type_id` (`PureIrrigationTask` `ReactSelect`). This masked the fact that some tasks referenced types not returned by the backend.

Jira link: https://lite-farm.atlassian.net/browse/LF-3633

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/webapp/src/containers/Task/saga.js
```diff
@@ -804,6 +805,11 @@ const getCompleteIrrigationTaskBody = (task_translation_key) => (data) => {
           delete data.location_id;
         }
 
+        if (!isOtherIrrigationType) {
+          data.taskData[taskType].irrigation_type_id =
+            data.taskData[taskType].irrigation_type.irrigation_type_id;
```
kathyavini: Looks like in the case of a task re-complete, the `irrigation_type_id` (and name) nested within the `irrigation_type` object is stale -- i.e. refers to the value when the task was originally completed:

Screenshot 2025-09-16 at 4 05 49 PM

Gemini was the one who immediately called out this line (!)
```suggestion
            data.taskData[taskType].irrigation_type_id;
```

That was even before I inspected the data object. AI is getting too good 😭

kathyavini (commented):
Thank you so much for investigating and also the explanation of the intermittent behaviour!! That was really confusing (and how Denis and I ended up with bad theories like "water calculator has to be used" 😆)

I saw just one bug on re-complete where the request body was being sent with the original type id instead of the new one, noted below. The UX of the bug just looks like not being able to change the type when re-completing.

packages/webapp/src/containers/Task/saga.js
SayakaOno: Thank you so much for catching this Joyce!!!
`data.taskData[taskType].irrigation_type_id` works when the irrigation type is changed, but if it isn't changed, it's `undefined`, and a new type would be created 😭 
Here's the data when the type isn't changed:
Screenshot 2025-09-18 at 12 35 21 PM
Since this was error-prone, I refactored the whole flow. I'll explain it in another comment!

packages/webapp/src/components/Task/PureIrrigationTask/index.jsx
```diff
@@ -66,9 +66,6 @@ export default function PureIrrigationTask({
       )[0],
     };
   })[0];
-  const [irrigationTypeValue, setIrrigationTypeValue] = useState(() => {
-    if (locationDefaults?.irrigation_task_type) return locationDefaults?.irrigation_task_type;
-  });
```
SayakaOno, line 71: We now reference the value of selected ReactSelect option.

packages/webapp/src/components/Task/PureIrrigationTask/index.jsx
```diff
@@ -106,8 +103,7 @@ export default function PureIrrigationTask({
   const stateController = () => {
     return { register, getValues, watch, control, setValue, reset };
   };
-  const IRRIGATION_TYPE = 'irrigation_task.irrigation_type_name';
-  const IRRIGATION_TYPE_ID = 'irrigation_task.irrigation_type_id';
```
SayakaOno, line 110: The selected ReactSelect option used to be stored under `irrigation_task.irrigation_type_name`. It is now stored under `irrigation_task.irrigation_type`.
`IRRIGATION_TYPE_ID` must always match the selected option’s `irrigation_type_id`. Instead of duplicating it, we now reference `irrigation_task.irrigation_type.irrigation_type_id`.

packages/webapp/src/components/Task/PureIrrigationTask/index.jsx
```diff
@@ -211,22 +205,24 @@ export default function PureIrrigationTask({
...
               }}
               isDisabled={disabled}
-              value={IrrigationTypeOptions.find((options) => options.value === irrigation_type)}
-              defaultValue={getDefaultIrrigationTypeOptions}
+              value={
+                value &&
+                irrigationTypeOptions.find(
+                  (option) => option.irrigation_type_id === (value.irrigation_type_id || null),
+                )
+              }
```
SayakaOno, line 226: We now look up the option by `irrigation_type_id`.
- null value: no option is selected
- `value.irrigation_type_id === null`: the selected option is "OTHER".

`defaultValue`'s `getDefaultIrrigationTypeOptions` wasn’t being called, so it wasn’t doing anything?

packages/webapp/src/components/Task/TaskComplete/index.jsx
```diff
@@ -155,6 +155,18 @@ export default function PureTaskComplete({
...
+      const irrigationType = formData.irrigation_task.irrigation_type;
+      data.taskData.irrigation_task.irrigation_type_name =
+        irrigationType.value === 'OTHER'
+          ? data.taskData.irrigation_task.irrigation_task_type_other
+          : irrigationType.value;
+
+      if (irrigationType.value !== 'OTHER') {
+        data.taskData.irrigation_task.irrigation_type_id = irrigationType.irrigation_type_id;
+      }
+    }
```
SayakaOno: In the saga, `getObjectInnerValues` formats ReactSelect options.
(eg. `{irrigation_type: {label: "LABEL", value: "VALUE"}` => `{irrigation_type: "VALUE"}`)
To avoid unintended formatting, handle the DB formatting here instead. Irrigation type will no longer be formatted in the saga.
(Properties that should be excluded from the request body are still removed in the saga.)

SayakaOno (commented):
I looked into this more closely and found we can have up to three `irrigation_type_id`s when completing a task.
Screenshot 2025-09-18 at 12 53 55 PM

To avoid duplicating irrigation type properties:
- Updated the task slice to not duplicate `irrigation_type_translation_key` and `irrigation_type_name`
- Refactored `PureIrrigationTask` to avoid duplication as well

Since I updated the property name for irrigation type in `PureIrrigationTask`, the formatting functions were adjusted. (The goal is to convert ReactSelect option + entered new type name into a DB-ready structure.)
- We always send `irrigation_type_name` (since the `irrigation_task` table has this column -- I’d really like to drop it)
  - If the selected option is "OTHER", use `irrigation_task_type_other`
  - Otherwise, use `irrigation_type_name`
- When an existing option is selected, we also send `irrigation_type_id`

This ended up being more involved than I initially expected 😢 Let me know if you spot anything suspicious or need any further clarification. Thank you for re-reviewing! 🙏

kathyavini:
Okay wow, that flow was soooooo much messier than I ever suspected, on so many levels 😯 The Pure Component needed that refactor so badly, thank you  🙏 🙇

It came to mind because it was a recent addition, but I think `useApproveIrrigationPrescription` hook needs updating to the new data shape to continue to pre-select pivot.

Quick code search also indicates two Storybook stories (`AddIrrigationTask` and `IrrigationTaskReadOnly`) use the old properties -- maybe nice to update for clarity, too, although it doesn't matter too much I think whether the select is populated in the stories!

packages/webapp/src/containers/Task/saga.js
kathyavini: I completely underestimated the scope of the troubles!! Thank you for taking a much needed holistic view of this one 🙏

PR #3880, opened 2025-09-17
LF-4960 Add product inventory page and route

kathyavini:
**Description**

Adds the menu item, header, and container+component structure for the Product Inventory ("Inventory" in menu).

I took the approach here of copying Animal Inventory _exactly_ (literally copied the files as my starting point). I'm not entirely sure that our Animal Inventory organization is the absolute _best_ possible organization, but I think knowing exactly where to look for each piece of logic in the container and pure component will reduce mental overhead and speed up development of the current feature quite a lot.

I think keeping parallelism also leaves open the possibility of refactoring both in parallel, although Animal Inventory (which has three different use cases in app) is a _much_ more complicated component than I think Product Inventory will ever be.

As a review of how animal inventory is laid out, this means:
- the add button, filter, and searchbar are all defined on the inner pure component
- the filter logic (to be implemented) and the search logic (just went ahead added that since it's trivial) live in the container
- table columns will be defined in the container (next ticket)
- I reused `<FixedHeaderContainer />`, as the card layout with inner scrollable is exactly what was needed, and it already was being used with `null` header, and outside of animals
  - I did remove the resize observer for the case of `null` header, which should also clean up a few unnecessary event listeners that were being added in the task animal inventory views

For simplicity, I generally took the approach of leaving everything where it was -- even, for instance, the pure component's styles, which I just imported from animals rather than duplicating. Let me know your thoughts on this :) I thought the non-duplication would be an easy way to be very clear about what is shared between the inventories without extracting the shared types, components, strings, etc. into a common home. Ultimately I think that extraction could be nice, but I don't think it has to be done right now.

SORRY ABOUT THE ROUTES FILE!!

Jira link: https://lite-farm.atlassian.net/browse/LF-4960

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/webapp/src/components/Animals/FixedHeaderContainer/index.tsx
```diff
@@ -85,7 +85,7 @@ const FixedHeaderContainer = ({
   );
 
   useLayoutEffect(() => {
-    if (kind === ContainerKind.OVERFLOW) {
+    if (kind === ContainerKind.OVERFLOW || !header) {
```
kathyavini, line 88: Resize observer not needed without a header.

packages/webapp/src/containers/ProductInventory/index.tsx
```diff
@@ -0,0 +1,98 @@
...
+    makeProductsSearchableString,
+  ) as [Product[], SearchProps['searchString'], SearchProps['setSearchString']];
+
+  const searchProps: SearchProps = {
+    searchString,
+    setSearchString,
+    // These animal translations are appropriate for this view as well
+    placeHolderText: t('ANIMAL.SEARCH_INVENTORY_PLACEHOLDER'),
+    searchResultsText: t('ANIMAL.SHOWING_RESULTS_WITH_COUNT', {
+      count: searchAndFilteredInventory?.length,
```
kathyavini, line 74: In hindsight, it would have been appropriate to put these strings somewhere other than Animals, since the text has no reference to animals at all, but I didn't want to move them at this point to save the work in CrowdIn.

SayakaOno (approved):
Smart approach, looks great!

kathyavini:
> Do you remember what Loïc said about the "Clear active filters" button? I was assuming we’ll show the "Add product" button as designed, and place the clear filter button below the search input. Was that a future plan? 😅

Yeah, a future plan! He wanted update all of the screens with the corner button on desktop (tasks, animals, product inventory) at the same time, rather than having multiple patterns in app (although technically we already have two patterns because of finances 😂). But in any case, for now product inventory should match animal inventory. There will later be a ticket for the upgrade across app.

PR #3882, opened 2025-09-23
LF-4969 Migrate product tables and adjust APIs

kathyavini:
**Description**

The goal of this migration was to move the indicated properties from `product` to `product_farm` in such a way that the frontend doesn't need to be changed. This means
- frontend can continue sending the `product_farm`-destined properties in the top level of product
- API will return those products flattened back into the top level of product

For simplicity and to minimize code changes in the test files, I also flattened in the return from `productFactory()`

I realized, a bit belatedly (but maybe you always knew!) that making this compatible with the current frontend code means a given `product_id` and `farm_id` combination has to be unique in the `product_farm` table, since the frontend specifies the product with nothing more than those two properties.

For the most part, changes all lump into:
- upserting / inserting on the right table (`productController`, `product.js`, `mocks.factories.js`), OR
- checking properties on the right table (`organicCertifierSurveyController.js` <-- sorry for a bit of linting in there too, `hasFarmAccess.js`, `checkPropertyValidity.js`, `checkSoilAmendmentTaskProducts.js`

Jira link: https://lite-farm.atlassian.net/browse/LF-4969

**Type of change**

**How Has This Been Tested?**

- [x] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/productController.js
+++ b/packages/api/src/controllers/productController.js
@@ -55,7 +55,7 @@ const productController = {
             product_farm: [{ farm_id, supplier, on_permitted_substances_list }],
           });
 
-        const flattenenedResult = await ProductModel.query(trx)
+        const flattenedResult = await ProductModel.query(trx)
           .findById(inserted.product_id)
           .joinRelated('product_farm')
           .where('product_farm.farm_id', farm_id)
@@ -64,7 +64,7 @@ const productController = {
           .first();
 
         await trx.commit();
-        res.status(201).send(flattenenedResult);
+        res.status(201).send(flattenedResult);
       } catch (error) {
         await handleObjectionError(error, res, trx);
       }


packages/api/tests/product.test.js
```diff
@@ -122,7 +125,7 @@ describe('Product Tests', () => {
       );
     });
 
-    test('Should get products on my farm but not default products', async (done) => {
+    test('Should get products on my farm but not library products not yet added to inventory', async (done) => {
```
kathyavini, line 128: I noticed these tests have a concept of "default products" -- maybe those legacy library entries with null `farm_id`? -- despite those no longer having any relevance in app.

Rather than removing these `null` `farm_id` products from the tests, I left them and described them according to the model we're moving to (library products that aren't necessarily associated with the farm's inventory). Let me know if that makes sense!

packages/api/src/controllers/productController.js
```diff
@@ -46,11 +46,25 @@ const productController = {
...
+
+        const inserted = await ProductModel.query(trx)
           .context({ user_id: req?.auth?.user_id })
-          .insertGraph({ ...data, farm_id });
+          .insertGraph({
+            ...productData,
+            product_farm: [{ farm_id, supplier, on_permitted_substances_list }],
+          });
+
+        const flattenenedResult = await ProductModel.query(trx)
```
SayakaOno: This seems like a typo 😁

packages/api/src/middleware/validation/checkProductValidity.js
```diff
@@ -127,10 +127,11 @@ export function checkProductValidity() {
...
-        .where({ farm_id })
-        .andWhere({ type })
-        .andWhere({ name })
-        .whereNot({ product_id: product_id ?? null })
+        .joinRelated('product_farm')
+        .where('product_farm.farm_id', farm_id)
+        .andWhere('product.type', type)
+        .andWhere('product.name', name)
+        .whereNot('product.product_id', product_id ?? null)
         .whereNotDeleted();
```
SayakaOno, line 135: Just to double-check, are we not allowing using the same name as a removed custom product? I don’t recall our convention off the top of my head...

packages/api/src/middleware/validation/product.js
```diff
@@ -6,16 +6,54 @@ function createOrPatchProduct(taskType) {
       if (req.body[taskType]) {
         if (req.body[taskType].product_id) {
           const { product, product_id } = req.body[taskType];
-          await productModel.query().context({ user_id }).patch(product).where({ product_id });
+
+          const {
+            supplier,
+            on_permitted_substances_list,
+            // omit the body sent farm_id
+            farm_id,
```
SayakaOno, line 14: Sorry, I didn’t mention this in the ticket (I even added the "backend" tag), but would it make sense to remove `farm_id` from the API request body on the frontend?

packages/api/tests/product.test.js
SayakaOno: Yes, that makes sense, thank you!

packages/api/src/controllers/productController.js
kathyavini: Oh no! Flatten-en-ed! 😆

packages/api/src/middleware/validation/product.js
kathyavini: Hmm I'm not sure! I just took a look at the code, and although there is a form value in the `AddProduct` component which is handling adding for new products

```js
const FARM = `${type}.product.farm_id`;
```

I think for existing products, it's just the fact that `farm_id` exists on `product` itself that is sending it back in the request body. It's not added in the sagas or anything as far as I can see. I thought briefly about removing `farm_id` from the `productSlice`, but it is used in the normal way in the selectors:

```js
export const productsSelector = createSelector(
  [productSelectors.selectAll, loginSelector],
  (products, { farm_id }) => {
    return products.filter((product) => product.farm_id === farm_id);
  },
);
```

Since the API returns `product` with all the `product_farm` values flattened in, I guess it makes sense for the API to also expect those values when the frontend sends them back? 🤔 I think that's consistent with a backend-frontend contract where frontend believes `farm_id` is a property on `product` and only the backend has to know otherwise? (PS I hope it's not an anti-pattern that we're creating such a contract 😂 -- what do you think?)

packages/api/src/middleware/validation/checkProductValidity.js
kathyavini: Ah yes, that's totally right! We will allow other matching custom product names as long as the others are removed. I'll add that qualifier to this query on the delete API branch since I'm adding `removed`-related changes there 🙏

packages/api/src/middleware/validation/product.js
SayakaOno: Thank you for looking into it! This is when I want inputs from someone like Ivan...😢
I asked Claude, and it says having the same structure for GET/POST requests isn't an anti-pattern, but receiving and sending different shapes of data is very common.
~~If `productsSelector` was meant to filter out "default" products, I think it'd be okay to remove `farm_id` from the `productSlice` or even from the API response?~~
~~Let's discuss it in tech daily!~~

I noticed that we would need to prevent wrong `farm_id`s from being added, so we need this either way!

PR #3883, opened 2025-09-24
[WIP] LF-4962: Create product form and implement product creation

SayakaOno:
**Description**

TODO: Test and self-review 😵🫠

Jira link: https://lite-farm.atlassian.net/browse/LF-4962

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/webapp/src/components/Task/AddSoilAmendmentProducts/ProductCard/ProductDetails.tsx
```diff
@@ -73,65 +73,59 @@ const molecularCompoundsUnitOptions = [
...
-    [CA]: NaN,
-    [MG]: NaN,
-    [S]: NaN,
-    [CU]: NaN,
-    [MN]: NaN,
-    [B]: NaN,
-  },
-  [MOLECULAR_COMPOUNDS_UNIT]: MolecularCompoundsUnit.PPM,
-};
-
```
SayakaOno, line 115: Moved to constants file

packages/webapp/src/containers/ProductInventory/ProductForm/useSaveProduct.ts
```diff
@@ -0,0 +1,132 @@
...
+import {
+  MolecularCompound,
+  Nutrients,
+  type ProductFormFields,
+  type ProductId,
+} from '../../../components/Task/AddSoilAmendmentProducts/types';
+import { FormMode } from '../';
+
+type SoilAmendmentProductFormFields = ProductFormFields & { product_id?: ProductId; name?: string };
+
```
SayakaOno, line 39: Copied and adjusted this function:

https://github.com/LiteFarmOrg/LiteFarm/blob/72418174598f8fec85997a33ba10fa4b831f4e78/packages/webapp/src/components/Task/SoilAmendmentTask/index.tsx#L115-L162

packages/webapp/src/components/ProductInventory/index.tsx
```diff
@@ -137,21 +146,20 @@ const PureProductInventory = ({
...
-    />
-  );
-});
+    return (
+      <FloatingMenu
+        ref={ref}
+        options={[
+          {
+            label: t('INVENTORY.SOIL_AMENDMENT'),
+            onClick: () => onMenuItemClick(TASK_TYPES.SOIL_AMENDMENT),
```
kathyavini, line 159: ❤️

packages/webapp/src/components/Form/ProductDetails/index.tsx
```diff
@@ -73,65 +78,59 @@ const molecularCompoundsUnitOptions = [
...
-  expand,
-  unExpand,
-  toggleExpanded: toggleProductDetailsExpanded,
-  clearProduct,
-  setProductId,
-  onSave,
-  fertiliserTypeOptions,
-  productsVersion,
-}: ProductDetailsProps) => {
+function isNestedFormProps(
```
kathyavini: It looks like the discriminated union type you set up already works perfectly with

```
const isNestedForm = props.isNestedForm;
```

(i.e. auto-complete on prop is correctly narrowed to props that go with each type). Is there an additional benefit to having a type guard function?

packages/webapp/src/containers/ProductInventory/ProductForm/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+          <FormProvider {...formMethods}>
+            <Component mode={mode} />
+          </FormProvider>
+        )}
+        <InFormButtons
+          className={styles.inFormButtons}
+          statusText={t('common:EDITING')}
+          confirmText={t('ADD_PRODUCT.SAVE_PRODUCT')}
+          onCancel={onCancel}
+          informationalText={t('ADD_PRODUCT.BUTTON_WARNING')}
```
kathyavini: I wonder if it would be appropriate to hide the informational text on `FormMode.CREATE` since by definition no tasks could ever be associated with a new custom product 🤔

Update: Loïc already said yes we should hide it, which you heard, but I'm keeping my comment here 😅 

Image

packages/webapp/src/containers/ProductInventory/ProductForm/styles.module.scss
```diff
@@ -0,0 +1,63 @@
...
+    &:nth-of-type(3) {
+      svg {
+        width: 22px;
+      }
+    }
+  }
+}
+
+.formWrapper {
+  padding: 32px 8px 0 16px;
```
kathyavini:
```suggestion
  padding: 0 8px 0 16px;
```
I think the spacing between the form and the drawer header title is too much relative to Figma, based on [this screen here](https://www.figma.com/design/eMffC0h3JPlWtzNttxJpil/Litefarm-Updated?node-id=50076-52308) (however, in [another screen](https://www.figma.com/design/eMffC0h3JPlWtzNttxJpil/Litefarm-Updated?node-id=50843-195379) there is no drawer title at all for add product so I'm not sure which is definitive; do you know? 😂)

But it's really weird... the extra spacing disappears if I hot module reload (i.e. after editing some unrelated code); have you ever seen anything like that?

Image

There is also this fixed 35px height on the header within `<Drawer />` which it looks like I removed once and then reverted (absolutely don't remember why 😭 I think I accidentally grouped it as a change related to another feature which it is not dependent on.) It throws off the spacing here and is not within our design system's spacing, so I'm super tempted to remove it again -- what do you think? 😬

packages/webapp/src/containers/ProductInventory/ProductForm/SoilAmendmentProductForm.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { TASK_TYPES } from '../../Task/constants';
+import { FormMode } from '..';
+
+export default function SoilAmendmentProductForm({ mode }: { mode: FormMode | null }) {
+  const { t } = useTranslation();
+
+  const { data: fertiliserTypes = [] } = useGetSoilAmendmentFertiliserTypesQuery();
+  // @ts-expect-error - Selector return empty object without property
+  const { country_id } = useSelector(userFarmSelector);
+  const { interested, farm_id } = useSelector(certifierSurveySelector, shallowEqual);
```
kathyavini, line 31: Should we be using `shallowEqual` a lot more with selectors? How do you know when to reach for it? 🙏

packages/webapp/src/components/ProductInventory/ProductForm/PureSoilAmendmentProductForm.tsx
```diff
@@ -0,0 +1,74 @@
...
+}: PureSoilAmendmentProductFormProps) => {
+  const { t } = useTranslation();
+  const {
+    register,
+    reset,
+    formState: { errors },
+  } = useFormContext();
+
+  useEffect(() => {
+    reset(productDefaultValuesByType[TASK_TYPES.SOIL_AMENDMENT]);
```
kathyavini, line 45: > Currently implemented by resetting the form when the soil amendment form mounts (useEffect without dependencies). I'm open to suggestions for a better approach!

I don't really mind it, but my initial expectation would have been to look for the reset within `onCancel()`, which is already passed to both the button and the drawer header. So maybe in `<ProductForm  />` container, defining an `onCancel()` next to the `onSave()` would be a reasonable alternative?

Since the form field constants are already in that exact folder (ProductForm container), it also feels a bit more "grouped" that way.

Maybe something like this?

```ts
const onCancel = () => {
  if (productFormType) {
    formMethods.reset(productDefaultValuesByType[productFormType]);
  }
  props.onCancel();
};
```

kathyavini (commented):
This looks _fantastic_! And I'm totally impressed at how thoroughly you made the form extensible to other product types 🙇

The only thing that I wanted to double-check on was the spacing because that doesn't seem correct to me; everything else was an optional-type comment. Thank you for tackling this huge task ❤️

packages/webapp/src/components/Form/ProductDetails/index.tsx
SayakaOno: It does work without it! 😳
I think I just did something weird initially that made the type guard function seem necessary. I’ve replaced it, thank you 🙌❤️

packages/webapp/src/components/ProductInventory/ProductForm/PureSoilAmendmentProductForm.tsx
SayakaOno: Handling it in `onCancel` feels right!!!
I was confused by the close vs cancel requirement, but now that’s gone, leaving the values as-is when the form closes seems fine...?
Let me double check with the read-only view in the next PR! 🙇‍♀️

packages/webapp/src/containers/ProductInventory/ProductForm/SoilAmendmentProductForm.tsx
SayakaOno: I’ll be honest, I just copied it from somewhere 🙃 I’ll learn what it does now…

packages/webapp/src/containers/ProductInventory/ProductForm/styles.module.scss
SayakaOno: I can't remember where the 32px came from 😅, but I believe we need the title...

> the extra spacing disappears if I hot module reload (i.e. after editing some unrelated code); have you ever seen anything like that?
I’ve seen icon styling disappear too! You might notice it in the next PR as well.

I passed a longer height to Drawer in the next PR… let me check if we can remove the fixed 35px!

packages/webapp/src/containers/ProductInventory/ProductForm/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+          <FormProvider {...formMethods}>
+            <Component mode={mode} />
+          </FormProvider>
+        )}
+        <InFormButtons
+          className={styles.inFormButtons}
+          statusText={t('common:EDITING')}
+          confirmText={t('ADD_PRODUCT.SAVE_PRODUCT')}
+          onCancel={onCancel}
+          informationalText={mode === FormMode.EDIT ? t('ADD_PRODUCT.BUTTON_WARNING') : undefined}
```
kathyavini, line 119: Not for now, but I wonder if `FormMode.ADD` with library products might also need this? 🤔 Based on what we said about re-adding a removed library product. But a question for later!

kathyavini (approved):
Looks good! I wasn't thinking about the read-only view in the context of `onCancel()` -- something to revisit as the other form modes come in!

PR #3885, opened 2025-09-29
LF-4959 Create DELETE product API

kathyavini:
**Description**

This adds the DELETE endpoint for products.

The logic for removing (but not un-removing!) library products I did go ahead and add. To get this logic working all the way through actually took quite some edits that perhaps could have gone on the library PR, but they also felt like they "belonged" to the DELETE API so I have gone ahead and added here:
- `hasFarmAccess` will search a `product_farm` record by both `product_id` and `farm_id` (only for library products would multiple `product_farm` table records with the same `product_id` be possible)
- similarly, the new middleware on DELETE `checkProductUsageInTasks` will look for tasks specific to that farm (only for library products could a given `product_id` return tasks belonging to multiple farms)
- the controller logic + tests themselves handle library products. I have tagged that code with the ticket number for library products, so they can be reviewed / finalized then.

Just for ease of testing, two very minimal frontend changes were included as well -- including `.removed` on the product slice and using `setAll` in place of `upsert` in the Redux store so that deleted products can be seen to disappear from the frontend products entirely, and the `removed` boolean can be visually observed. This was just to test the API using the frontend (see testing section below), and I assumed the remaining selector / filtering on removed logic would be added when the removal functionality is added to the inventory.

Jira link: https://lite-farm.atlassian.net/browse/LF-4959

**Type of change**

**How Has This Been Tested?**

- [x] Passes test case
- [ ] Other (please explain)

Maybe it's not that useful, but I've intentionally avoided merging integration (which has the formatted table for product inventory) into this branch so that the `removed` property and the `product_id` can be visually inspected on the Inventory view (I found this a little faster than looking at the database):

Screenshot 2025-09-29 at 12 12 33 PM

Then I just fed the `product_id` into Postman to test:

```
DELETE http://localhost:5001/product/{{product_id}}
```

**Checklist:**

### DIFF
--- a/packages/api/src/middleware/validation/checkSoilAmendmentTaskProducts.js
+++ b/packages/api/src/middleware/validation/checkSoilAmendmentTaskProducts.js
@@ -16,7 +16,7 @@
 import ProductModel from '../../models/productModel.js';
 import SoilAmendmentPurposeModel from '../../models/soilAmendmentPurposeModel.js';
 
-export function checkSoilAmendmentTaskProducts() {
+export function checkSoilAmendmentTaskProducts({ newTask = false } = {}) {
   return async (req, res, next) => {
     try {
       const { soil_amendment_task_products } = req.body;
@@ -92,9 +92,10 @@ export function checkSoilAmendmentTaskProducts() {
           .where('product.product_id', product.product_id)
           .andWhere('product.type', 'soil_amendment_task')
           .andWhere('product_farm.farm_id', req.headers.farm_id)
+          .modify('flattenProductFarm')
           .first();
 
-        if (!existingProduct) {
+        if (!existingProduct || (newTask && existingProduct.removed)) {
           return res
             .status(400)
             .send(
--- a/packages/api/src/middleware/validation/checkTask.js
+++ b/packages/api/src/middleware/validation/checkTask.js
@@ -197,7 +197,7 @@ export function checkCreateTask(taskType) {
 
       const checkProducts =
         checkProductsMiddlewareMap[taskType] || checkProductsMiddlewareMap['default'];
-      checkProducts()(req, res, next);
+      checkProducts({ newTask: true })(req, res, next);
     } catch (error) {
       console.error(error);
 


packages/api/src/middleware/validation/checkProductUsageInTasks.ts
```diff
@@ -0,0 +1,49 @@
...
+      const { farm_id } = req.headers;
+
+      const tasksUsingProduct: Task[] = await TaskModel
+        // @ts-expect-error known issue on models
+        .query()
+        .where({ 'task.deleted': false })
+        .modify('onFarmByLocationTasks', farm_id)
+        .modify('usingProduct', product_id);
+
+      const plannedTasksUsingProduct = tasksUsingProduct.filter(
```
SayakaOno: I think `filter` could be replaced with `some` for efficiency :)

packages/api/src/middleware/validation/checkProductValidity.js
```diff
@@ -129,6 +129,7 @@ export function checkProductValidity() {
       const existingRecord = await ProductModel.query(trx)
         .joinRelated('product_farm')
         .where('product_farm.farm_id', farm_id)
+        .andWhere('product_farm.removed', false)
```
SayakaOno, line 132: I think `existingProduct` in `checkSoilAmendmentTaskProducts` also needs this. Could you double check? 🙏

packages/api/src/middleware/validation/checkProductUsageInTasks.ts
```diff
@@ -0,0 +1,49 @@
...
+import TaskModel from '../../models/taskModel.js';
+import { Task } from '../../models/types.js';
+
+export function checkProductUsageInTasks() {
+  return async (req: Request, res: Response, next: NextFunction) => {
+    try {
+      const { product_id } = req.params;
+      const { farm_id } = req.headers;
+
+      const tasksUsingProduct: Task[] = await TaskModel
```
SayakaOno, line 26: When supporting cleaning or pest control products in the future, I believe we’ll also need to include tasks created for wild crops, which don’t have records in `location_tasks`. Can we add a comment indicating this?

packages/api/src/middleware/validation/checkProductValidity.js
kathyavini: I don't think so -- if I am remembering the chart / UX discussion correctly, we will continue to show the `.removed` product that was used in that particular task, even after removal. Pretty sure this can only come into play on re-complete, since product removal before initial completion is prohibited. But we wouldn't want to cause a 400 error (it's a "product does not exist" error) if the task is re-completed with the removed product, as that is an allowed flow. Does that sound right?

SayakaOno: Ah, we definitely shouldn't return 400 when re-completing a task with a removed product...!
But I thought we'd need to prevent removed products from being added to new tasks. Does it make sense to return an error there for new tasks? (This will be handled in LF-4981)

Maybe something like this?
```
const existingProduct = await ProductModel.query()
  .where({
    product_id: product.product_id,
    farm_id: req.headers.farm_id,
    type: 'soil_amendment_task',
  })
  .first();

if (!existingProduct || newTask && existingProduct.removed) {
  return res
    .status(400)
    .send(
      `Soil amendment product ${product.product_id} does not exist or does not belong to the given farm`,
    );
}
```

kathyavini: Ohhh! I read the middleware too fast and thought the `checkProductsMiddlewareMap` was only used in task completion. That sounds good for limiting on creation 🙏 

(The query syntax for `existingProduct` is just an example though, right? The `farm_id` isn't on that table so it can't be queried like that anymore; just making sure it's not part of the code suggestion!)

SayakaOno: Sorry, I copied the code from the branch I was on! 😱
Thank you for making the changes!

SayakaOno (approved):

PR #3886, opened 2025-09-29
Lf 4921 adjust task re completion notifications

navdm:
**Description**

- Updates the notification list on frontend to view only latest notification per entity. Notifications are filtered by their entity ids by removing duplicates and keeping the latest. Also moved sorting by `created_at` into the Redux adapter so its done in one place and makes the filtering logic easier.
- Updates wording for task re-completion as mentioned in the ticket. 

Jira link: https://lite-farm.atlassian.net/browse/LF-4921

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/api/tests/notificationUser.test.js
+++ b/packages/api/tests/notificationUser.test.js
@@ -268,6 +268,15 @@ describe('Notification tests', () => {
       );
     });
 
+    test('Responds with 400 when notification_ids contains non-UUID values', (done) => {
+      clearAlerts({ notification_ids: [123, 'not-a-uuid'] }, {}, (err, res) => {
+        expect(err).toBe(null);
+        expect(res.status).toBe(400);
+        expect(res.body.error).toBeDefined();
+        done();
+      });
+    });
+
     test('Users can modify the status for a set of their notifications', async (done) => {
       const [notification] = await mocks.notification_userFactory({
         promisedUserFarm: [userFarm],
--- a/packages/api/src/controllers/notificationUserController.js
+++ b/packages/api/src/controllers/notificationUserController.js
@@ -13,6 +13,7 @@
  *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
  */
 
+import { validate } from 'uuid';
 import NotificationUser from '../models/notificationUserModel.js';
 
 /**
@@ -179,6 +180,9 @@ export default {
    * @async
    */
   async clearAlerts(req, res) {
+    if (req.body?.notification_ids?.some((id) => !validate(id))) {
+      return res.status(400).json({ error: 'All notification IDs must be valid UUIDs' });
+    }
     try {
       await NotificationUser.clearAlerts(
         req.auth.user_id,


packages/webapp/src/containers/notificationSlice.js
```diff
@@ -70,6 +71,23 @@ export const notificationsSelector = createSelector(
...
 
+export const latestNotificationsByEntitySelector = createSelector(
+  notificationSelectors.selectAll,
+  (notifications) => {
+    const latest = [];
+    const entityIdSet = new Set();
+
+    for (let notification of notifications) {
+      const entityId = notification.ref.entity?.id;
+      if (entityId == null || entityIdSet.has(entityId)) continue;
```
kathyavini: I don't think it would have been apparent from inspecting your local db (unless you have worked with timed notifications or irrigation prescriptions), but not all notification types have a `ref.entity` -- I think it's easiest to see the a backend function `sendDailyDueTodayTaskNotification()` or the others in that controller. These notifications we would never collapse, so I think we want to always, rather than never, push when `entityId` is undefined:

```js
// something like
  if (entityId) {
    if (entityIdSet.has(Number(entityId))) {
      continue;
    }
    entityIdSet.add(Number(entityId));
  }
  latest.push(notification);
}
```

Also, I'm a bit annoyed it got up set like that in the first place, with some task notifications sending string IDs and others number IDs (?!!), but to complete the collapsing functionality across both of those situations, I think the `Number()` is needed when both adding to and checking the Set. Otherwise looks really nice; I like the use of Set!

kathyavini (commented):
Thanks Nav! This looks great, and I love what you did with the selectors.

So, I myself didn't realize this until checking out the code on this branch, but in addition to displaying the `cardContents`, it looks like `containers/Notification/index.jsx` also calls the `clearAlerts()` saga with all the "alerted" (new) notification_ids on the page:

```js
const alertIds = notifications
  .filter((notification) => notification.alert)
  .map((notification) => notification.notification_id);
if (alertIds.length) dispatch(clearAlerts(alertIds)) //  <-- here
// Unrelatedly, I'm surprised this isn't in an useEffect()?
```

That Redux state determines the alert count in the menu here:

Image

This means one side effect of changing the selector used in `cardContents` to the filtered version, is that notifications will only get cleared for the filtered cards, and if multiple notifications referred to the same task, then the notification count for those notifications becomes "unclearable", if that makes sense? So I think that will need to be fixed in some way.

Personally, when I look at this flow now, I feel like `clearAlerts()` doesn't need to take IDs -- why not just clear all notifications when the Notification list is loaded? Another option though would be to use a second selector in `containers/Notification/index.jsx` (`latestNotificationsByEntitySelector` for the cards, and some transformation on all notifications for the alerts). I think adjusting the saga would be cleaner, but what do you think?

packages/webapp/src/containers/notificationSlice.js
```diff
@@ -78,9 +78,11 @@ export const latestNotificationsByEntitySelector = createSelector(
     const entityIdSet = new Set();
 
     for (let notification of notifications) {
-      const entityId = notification.ref.entity?.id;
-      if (entityId == null || entityIdSet.has(entityId)) continue;
-      entityIdSet.add(Number(entityId));
+      const entityId = notification.ref?.entity?.id;
+      if (entityId != null) {
```
navdm, line 82: I used the condition `if (entityId != null)` which checks for both null and undefined, rather than `if (entityId)` because with the id being a number, there might be an edge case where its 0.

navDhammu:
> I think I might backtrack and vote for just creating another selector that clearAlerts() can use

Yeah that works, but we can keep this component even more cleaner going the api route. That way `<NotificationPage />` just calls `clearAlerts()` and doesn't know anything about the alerts. Also I think it should be in a `useEffect`, like:

```javascript
  useEffect(() => {
    dispatch(clearAlerts())
  }, [])
```
Not sure why its currently not because its hitting the API multiple times with the same id's if you look at the network tab. 

Once dispatched the saga can look up the alert count from the redux store, and send the request to clear alerts only if alert count is more than 0. The backend just needs to adjust its logic for clearing all alerts. I prefer this way but I'm ok with either so please let me know which approach to go with.

packages/api/tests/notificationUser.test.js
```diff
@@ -139,6 +140,51 @@ describe('Notification tests', () => {
...
+        const otherFarmNotifications = await NotificationUser.getNotificationsForFarmUser(
+          otherFarm.farm_id,
+          user.user_id,
+        );
+
+        expect(notifications[0].alert).toBe(false);
+        expect(notifications[1].alert).toBe(false);
+        expect(otherFarmNotifications[0].alert).toBe(true);
+
+        done();
```
navdm, line 184: This done syntax with the callback in `clearAlerts` can be improved and cleaner, I just didn't want to touch it here. In the newer jest versions its not allowed so we need to update all the test files in order to update to the latest  version.

PR #3891, opened 2025-10-06
LF-4967 Implement product filter

kathyavini:
I'm not sure if the first filter is necessary as it seems redundant with the filter title; I have asked Loïc [on Figma here](https://www.figma.com/design/eMffC0h3JPlWtzNttxJpil?node-id=18173-138067#1456278353) and will update accordingly.  Drafted until confirmed.

**Description**

This adds the filter to Product Inventory, following Animal Inventory exactly.

Jira link: https://lite-farm.atlassian.net/browse/LF-4967

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/containers/Filter/ProductInventory/index.tsx
+++ b/packages/webapp/src/containers/Filter/ProductInventory/index.tsx
@@ -1,0 +1,104 @@
+/*
+ *  Copyright 2025 LiteFarm.org
+ *  This file is part of LiteFarm.
+ *
+ *  LiteFarm is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  LiteFarm is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ */
+
+import { useTranslation } from 'react-i18next';
+import FilterGroup from '../../../components/Filter/FilterGroup';
+import type { ReduxFilterEntity, ContainerOnChangeCallback, FilterState } from '../types';
+import { FilterType, type ComponentFilter } from '../../../components/Filter/types';
+import { useGetSoilAmendmentFertiliserTypesQuery } from '../../../store/api/apiSlice';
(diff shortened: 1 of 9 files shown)


packages/webapp/src/containers/filterSlice.js
```diff
@@ -261,6 +279,10 @@ export const animalsFilterSelector = createSelector(
   [filterReducerSelector],
   (filterReducer) => filterReducer.animals,
 );
+export const inventoryFilterSelector = createSelector(
+  [filterReducerSelector],
+  (filterReducer) => filterReducer.inventory ?? initialInventoryFilter,
+);
```
kathyavini, line 284: It's been a long time since we've had this kind of change to the shape of an existing Redux slice, so I don't 100% remember if we have an established way to handle it.

We could either force logout or supply a fallback as above; otherwise the selector (and setter) will error when trying to access the nonexistent `tempStateReducer.filterReducer.inventory`. I'm thinking logout is certainly more common as our pattern, but we have to then commit to bumping the app version when this goes live, and since I wasn't sure if we wanted to commit to that, I added fallback here.

packages/webapp/src/containers/Filter/ProductInventory/index.tsx
```diff
@@ -0,0 +1,104 @@
...
+        label: t(`INVENTORY.FILTER.${type.key}`),
+      })),
+    },
+  ];
+
+  return (
+    <FilterGroup
+      filters={filters.map(sortFilterOptions)}
+      filterContainerClassName={filterContainerClassName}
+      onChange={handleChange}
```
SayakaOno: Could we pass `onChange` directly?

packages/webapp/public/locales/en/filter.json
```diff
@@ -1,41 +1,50 @@
 {
   "CROP_CATALOGUE": {
-    "LOCATION": "Location",
-    "STATUS": "Status",
-    "SUPPLIERS": "Suppliers",
-    "ACTIVE": "Active",
     "ABANDONED": "Abandoned",
+    "ACTIVE": "Active",
```
kathyavini, line 4: Looks like `i18n` had not touched this file in a while, so the keys got ordered alphabetically.

packages/webapp/src/containers/Filter/ProductInventory/index.tsx
kathyavini: Thank you!! I don't think I would have ever noticed that; callback prop drilling still really trips me up 😓

PR #3992, opened 2026-01-08
LF-5110 Process survey for metadata and return scores from data

Duncan-Brain:
**Description**

This calculates the chart data.

~~@kathyavini This ticket mentions adding the redux state, but I think you already did that. I could change it to a more stable store variable not based on -- but I guess not~~

Jira link: [LF-5110](https://lite-farm.atlassian.net/browse/LF-5110)

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

[LF-5110]: https://lite-farm.atlassian.net/browse/LF-5110?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/webapp/src/containers/Insights/TapeSurvey/TapeResults.tsx
+++ b/packages/webapp/src/containers/Insights/TapeSurvey/TapeResults.tsx
@@ -28,9 +28,32 @@ import {
 import styles from './styles.module.scss';
 import { Main, Semibold } from '../../../components/Typography';
 import PageTitle from '../../../components/PageTitle';
+import TapeQuestions from './tapeQuestions.json';
 
 const CHART_COLOR = 'rgba(85, 143, 112, 1)'; // --Colors-Secondary-Secondary-green-700
 const CHART_FILL_COLOR = 'rgba(85, 143, 112, 0.2)'; // reduced opacity
+const MAX_SCORE = 4;
+
+const getChartTitleFromSurveyTitle = (surveyTitle: unknown) => {
+  if (!surveyTitle || typeof surveyTitle !== 'string') return '';
+
+  // Remove the section number from the title
+  const titleWithoutSectionNumber = surveyTitle.split(/\s+/).slice(1).join(' ');
+
+  if (titleWithoutSectionNumber === '') return '';
+
+  return (
+    titleWithoutSectionNumber.charAt(0).toUpperCase() +
+    titleWithoutSectionNumber.slice(1).toLowerCase()
+  );
+};
+
+const getAnswerKeys = (element: any): string[] => {
+  if (Array.isArray(element.elements)) {
+    return element.elements.flatMap(getAnswerKeys);
+  }
+  return element.name ? [element.name] : [];
+};
 
 ChartJS.register(RadialLinearScale, PointElement, LineElement, Filler, Tooltip, Legend);
 
@@ -71,9 +94,9 @@ function TAPEResults() {
           display: true,
         },
         suggestedMin: 0,
-        suggestedMax: 100,
+        suggestedMax: MAX_SCORE,
         ticks: {
-          stepSize: 20,
+          stepSize: 1,
         },
       },
     },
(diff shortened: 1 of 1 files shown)


packages/webapp/src/containers/Insights/TapeSurvey/TapeResults.tsx
```diff
@@ -83,7 +131,7 @@ function TAPEResults() {
       },
       tooltip: {
         callbacks: {
-          label: (context: any) => `${context.label}: ${context.parsed.r}/100`,
+          label: (context: any) => ` ${context.label}: ${context.parsed.r} %`,
```
kathyavini, line 134: I think the space before the `%` is not needed, right? No need to adjust here though; I can drop it in the follow-up state PR.

packages/webapp/src/containers/Insights/TapeSurvey/TapeResults.tsx
```diff
@@ -28,9 +28,57 @@ import {
...
+};
+
+const getAnswerKeys = (element: any): string[] => {
+  if (Array.isArray(element.elements)) {
+    return element.elements.flatMap(getAnswerKeys);
+  }
+  return element.name ? [element.name] : [];
+};
+
+const CHOSEN_SECTION_NAMES = [
```
kathyavini, line 59: I guess our solutions are brittle in similar, but also opposite ways -- your requires this property (the `name` property that goes with the section heading for each dimension) to stay consistent, while mine requires the question prefix to be fixed. There was no way really for you to know this, since I didn't write it in the ticket, but Riddhi and I only talked about the question prefix, so that's the one I have more confidence in.

kathyavini (approved):
Thanks for picking this up! It works well and looks great.

I know I didn't communicate properly beforehand which parts of the data I expected to be "trustable" in the survey JSON, but as noted below, it was actually only just the question prefixes, not the headings or anything else, that I asked Riddhi to standardize.

That's why I was saying this morning that the question JSON isn't really necessary in this component, only a dictionary holding those keys:

```ts
const DIMENSION_MAPPING = {
  Diversity: 'diversity_1',
  Synergy: 'synergy_2',
  Recycling: 'recycling_3',
// ..
};

// ...
return Object.entries(DIMENSION_MAPPING).map(([dimension, prefix]) => {
  const scores = Object.keys(data)
    .filter((key) => key.startsWith(prefix)) // all that's needed to match
    .map((key) => Number(data[key]) || 0);

  if (scores.length === 0) {
    return { dimension, score: 0, maxScore: MAX_SCORE };
  }

  const sum = scores.reduce((acc, val) => acc + val, 0);
  const averageRawScore = sum / scores.length;
  const score = (averageRawScore / RAW_MAX_SCORE) * MAX_SCORE;

  return {
    dimension,
    score,
    maxScore: MAX_SCORE,
  };
});
```

But since I didn't communicate which part Riddhi was standardizing, I could see how a parsing strategy would have made sense based on what the overall structure looks like.

Ultimately neither property will probably change at this point so I'll put this down to dev preferences :)

Unrelated at all to the logic, did you happen to like the dimension labels better at 12px @Duncan-Brain?

packages/webapp/src/containers/Insights/TapeSurvey/TapeResults.tsx
Duncan-Brain: yah I had no clue any of them were being standardized!

I think it will be easier to maintain correct spelling of 9 keys rather than 70 though. ('synergy_1`, 'synregy_1_1'...)

packages/webapp/src/containers/Insights/TapeSurvey/TapeResults.tsx
Duncan-Brain: Probably not no, but I also don't remember noticing when we looked -- maybe it was nicer looking with. That would be cool if you could handle elsewhere when you tweak all the styles to your liking.

packages/webapp/src/containers/Insights/TapeSurvey/TapeResults.tsx
kathyavini: @Duncan-Brain can you explain where you're getting the 70 from? This is entirety of the mapping structure you need to maintain with what I'm talking about; there is nothing else:

```ts
const DIMENSION_MAPPING = {
  Diversity: 'diversity_1',
  Synergy: 'synergy_2',
  Recycling: 'recycling_3',
  Efficiency: 'efficiency_4',
  Resilience: 'resilience_5',
  'Culture and food traditions': 'culture_6',
  'Co-creation and sharing of knowledge': 'knowledge_7',
  'Human and social values': 'human_8',
  'Circular economy and solidarity': 'circular_9',
  'Responsible governance': 'governance_10',
};
```

Edit: Oh do you mean on Riddhi's end and not our code?

Duncan-Brain:
Thanks Joyce yah, this seemed obvious to me so it is done without that extra knowledge ya. 

And I still think it is quite flexible -- maybe improvable by throwing an error if one of the chosen dimensions was not found.

The 12px font, I don't know Divya seemed to put it on Loic during the meeting and it was not related to my PR so much -- I figure you will be tweaking the style to your liking a few more times!

packages/webapp/src/containers/Insights/TapeSurvey/TapeResults.tsx
Duncan-Brain: yes! .. from Riddhi's side or whoever updates the survey. From our side it is the same amount to keep up.

kathyavini: Maybe, but it didn't seem like she minded 😂

Duncan-Brain: Future human error.. I dno haha. The survey probably won't be updated so the prefix way would also work fine!

PR #2201, opened 2022-07-27
Fixed map update for sensors

Cali0707:
This PR makes the map update to show newly updated sensors on both a full success and a partial success. It also resolves  a small bug described in the comments on https://lite-farm.atlassian.net/browse/LF-2645 

To test:
1. Upload a file with one ESID which is claimed on another farm and one which is free - you should get one sensor added to your map, and a valid error message.
2. Unclaim the ESID which is claimed on another farm and upload the file again - you should be able to see both sensors on the farm map now

PR #3993, opened 2026-01-08
Lf 3068 when hiding point locations on farm map and zooming out clusters of 0 are shown

Duncan-Brain:
**Description**

In addition to this old bug whose status is released. This also resolves a bug I saw this week where the upon login the point filters on map would appear not to work with clustered points. On refresh that bug would resolve.

This branch fixes the issue twice and does two things:

- When setting `markerClusterRef.current` to a `new MarkerCluster()` it first removes the older cluster from the map, and cleans up any listeners before setting the new one. Previously it assumes that `markerClusterRef.current` was yet to be defined (TLDR: Before **REDRAW**, cleanup old assets)
- It removes the unnecessary re-render of the assets when maxZoom is updated. This re-render was [originally added for old sensor architecture ](https://github.com/LiteFarmOrg/LiteFarm/pull/2201#pullrequestreview-1052872329). It is no longer necessary as we have removed old sensor architecture. I think it was because the upload sensors modal did not navigate away and back to maps like the other "add" locations. Rtk-query should cover us there when we update the add map drawer and add locations flow -- in any case this approach should not be used. (TLDR: This **REDRAW** is not needed)
  - It also cleans it up identically the same as on `LocationPicker` added for sensor bugs : https://github.com/LiteFarmOrg/LiteFarm/commit/d7a5fea5235a411ccdc8849c54bfc9b25fec5f8b

Jira link: [LF-3068](https://lite-farm.atlassian.net/browse/LF-3068)

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain):

In addition to what is outline on the Jira ticket:
1. Logout method
  a. logout and login to farm with clustered points
  b. navigate to map and filter that cluster
  c. observe number remains the same but zooming in shows items are correctly filtered
  d. refresh map, problem does not persist

2. Add point location method
  a. add a new point location to an area that will cause a maxZoom difference (ocean is easiest)
  b. filter the cluster
  c. observe the number remains the same but zooming in shows items correctly filtered
  d. refresh on map, problem does not persist 

**Checklist:**

[LF-3068]: https://lite-farm.atlassian.net/browse/LF-3068?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/webapp/src/components/LocationPicker/SingleLocationPicker/index.jsx
```diff
@@ -53,9 +53,10 @@ const LocationPicker = ({
 }) => {
   useGoogleMapsLoader(['maps', 'geometry']);
   const [isGoogleMapInitiated, setGoogleMapInitiated] = useState(false);
-  const [gMap, setGMap] = useState(null);
   const [gMaps, setGMaps] = useState(null);
-  const [gMapBounds, setGMapBounds] = useState(null);
+  // Unused: Kept in for map debugging
```
Duncan-Brain: Wasn't sure about this.

packages/webapp/src/containers/Map/index.jsx
```diff
@@ -391,13 +397,6 @@ export default function Map({ isCompactSideMenu }) {
     setShowSuccessHeader(false);
   };
 
-  useEffect(() => {
```
Duncan-Brain, line 395: Calling drawAssets twice redraws the cluster twice. MaxZoom triggers this on login (moving from undefined to defined) It also triggers when adding a new point with a different maxZoom values (ocean is easiest).

packages/webapp/src/components/Map/MarkerCluster/index.jsx
```diff
@@ -16,6 +16,11 @@ const CreateMarkerCluster = (
   selectedLocationsRef = null,
   maxZoom,
 ) => {
+  // If clusterer is being replaced, remove the old one from the map and cleanup listeners
+  if (clustererRef?.current) {
```
Duncan-Brain, line 20: If drawAssets() is ever called twice two clusters will be on the map. One would be responsive to filtering, and the other would not be.

kathyavini:
@Duncan-Brain is there any chance of being able to separate the cluster part of this PR from the assets rendered two times part (the two bullet points of your description)?

It sounds like the missing icons in offline might be a great and easy way to test the second one in isolation. And since these map files are confusing to read, it would also help my understanding of what change exactly is doing what 🙏

Duncan-Brain:
@kathyavini 

I don't think it makes sense to separate them. They each independently solve this bug. And together they prevent it for the future happening again (new design for map side menu could have the same effect as old sensor upload). We now have 4 test cases (logout, add point location, filters, offline) 

I am happy to take a look at it with you if needed but I am surprised that it is difficult as there is not many lines of code -- despite the unfamiliar map files. I think between description and the test cases there is a lot to go on but I have added a couple extra details to the description that hopefully help.

packages/webapp/src/components/LocationPicker/SingleLocationPicker/index.jsx
```diff
@@ -294,9 +292,9 @@ const LocationPicker = ({
     };
   };
 
-  const handleGoogleMapApi = (map, maps) => {
+  const handleGoogleMapApi = async ({ map, maps }) => {
     mapRef.current = map;
-    getMaxZoom?.(maps, map);
+    await getMaxZoom(maps, map);
```
Duncan-Brain, line 294: Incorrect async await syntax.

packages/webapp/src/containers/Map/useMaxZoom.js
```diff
@@ -48,7 +48,7 @@ export function useMaxZoom() {
           });
         });
       });
-      Promise.all(promises)
+      await Promise.all(promises)
```
Duncan-Brain: Incorrect async await syntax.

packages/webapp/src/components/LocationPicker/SingleLocationPicker/index.jsx
kathyavini: What kind of debugging were you thinking? Is it using React dev tools, since you have to create component state to hold the values?

Actually, looking at it now, I think a large part of why I bounced right off this PR was seeing the `_` prefix convention with the `useState()` calls; super unusual and made me think something really complicated was going on!

packages/webapp/src/containers/Map/useMaxZoom.js
kathyavini: I agree (if this is what you mean by "incorrect async await syntax") that it was probably _meant_ to be awaited, since the function was defined `async`.

But now it's still a bit syntaxically weird because you have `await` and `.then().catch()` together. It works, but probably completing the refactor would be moving to straight `try`/`catch` now?

```ts
try {
  const results = await Promise.all(promises);
  const cachedAndActivePoints = retrievedPoints.filter((element) =>
//...
} catch (error) {
  console.log('Error getting available zooms: ', error);
//...
```

kathyavini (commented):
Well, I think I did figure out what's going on in the end...! Is this right?

- `getMaxZoom()` was missing `await` on its `Promise.all`. Now awaited so the first call to draw assets / draw locations has the right zoom to start with
- this means means no longer needing the `useEffect()` that redraws when `maxZoom` becomes defined (and causes all the bugs)
- although it's not strictly required after the above changes for this bug, cleanup was also added to `CreateMarkerCluster` so it's safe to call twice (nice 👍)

> I am happy to take a look at it with you if needed but I am surprised that it is difficult as there is not many lines of code

That's a bit rude! 🤣 

I think what was messing me up was:
- how the cluster change was independent from the others (no causal relationship), so it was helpful when you explained they both independently solve the bug
- probably would have preferred if you had described the role of the `await` in the PR description as that seems the most important reason the `useEffect()` is no longer needed?
- all that unused component state and variables marked with `_` !! It sounds like from your comment you were thinking of cleaning them up? Please do 🙏

So those parts were a bit rough in review, but I think it's an excellent find and fix ❤️

packages/webapp/src/components/LocationPicker/SingleLocationPicker/index.jsx
Duncan-Brain: You created these really handy debugging tools in `/util/google-maps/debugListeners` which I used for looking into those memory leaks a bit. 

The `_` is what is used to bypass unused variables in linting! I acknowledge it is not enforced on frontend but I find it good practice -- that is all it was!

These unused variables could be passed to them or variants to those functions. They are maybe a bit more difficult to setup then other debugging constants.

Deleting!

packages/webapp/src/containers/Map/useMaxZoom.js
Duncan-Brain: Yeah I think you are right that it matches the style we normally use better. Will update!

And yes since there was no code comments about purposefully not resolving the promise, I see missing `async/await` syntax as small typo -like errors. Admittedly they can can an affect on behaviour but I just saw it as a potential bug waiting to happen. 

I think it was working fine though as it was.

packages/webapp/src/components/LocationPicker/SingleLocationPicker/index.jsx
kathyavini: I totally forgot about those functions!!! I'm glad they were useful 😂 

It looks like they take the `gMaps` object (not the `gMap`) anyway though, so not needed for those 👍

packages/webapp/src/containers/Map/useMaxZoom.js
kathyavini: > I think it was working fine though as it was

Interesting. I didn't test this particular permutation, but I assumed that Javier added `maxZoom` to the `useEffect()` dependency array at the same time as that `Promise.all()` code because things needed to re-render again with the correct `maxZoom` set.

kathyavini (approved):
Looks good!

Duncan-Brain:
I can be poor at using technical wording vs conversational wording . Thanks for bearing with me 🙏  "This branch fixes the issue twice + other context" === "these two changes independently fix the issue" in my head.

The role of await => I am not aware that it has anything to do with this fix at all! To me it is just a syntax fix like a typo -- that I try to clean up as I go along. It could have indeed caused bugs but I think it was working fine as it was. And I believe it is working the same now too.

> I am happy to take a look at it with you if needed but I am surprised that it is difficult as there is not many lines of code

> That's a bit rude! 🤣

Sorry ! I was definitely frustrated about this PR... but I was genuinely trying to _politely_ nudge a request for re-read without being passive aggressive. But I guess I failed! lol 😂 ... I still can't think of a way to ask for that nicely. 

I know you are all busy with other more important stuff! It just felt to me subjectively as if it was a skim-read pr review and my code was not seriously considered because it was not important enough. The skim reading felt _slightly_ confirmed when I read the following about the prefix 🤣 :

> I think a large part of why I bounced right off this PR was the `_` prefix...

Anyways that combined with 7 weeks of waiting, rebasing, merging, and the major changes here IMO were just ~12 lines of code (4 for clusterRef 4 for each useEffect removal) -- it all made splitting it seem like much too much work for me! 

I know you have been busy, sorry I was frustrated and failed at finding the correct polite words! Suggestions are very welcome about how I can improve.

Thank you for re-reviewing! 🙏 ❤️

kathyavini:
> Sorry ! I was definitely frustrated about this PR... but I was genuinely trying to politely nudge a request for re-read without being passive aggressive. But I guess I failed! lol 😂 ... I still can't think of a way to ask for that nicely.

The nudge was totally fine! I was just laughing at the "why are you struggling to understand so little code"

> It just felt to me subjectively as if it was a skim-read pr review and my code was not seriously considered because it was not important enough.

Ah, it was definitely not about importance. I always skim read a PR to figure out if it's something trivial or something that needs some time set aside to understand, and that's why I was saying the `_` prefix made it feel feel like the second kind of PR, when maybe it was closer to the first? Same with the splitting -- I was hoping if there was a way to make it totally trivial/tiny, we could get it into the last release; it was just time to look closely before release I didn't have. So not about importance, but just in terms of mental effort.

I think I was just throwing myself off with the pressure of the release deadline anyway, which was not that important since we have another coming up basically right away! But I just needed to be on the other side of release to sit down with any think-y code.

PR #3994, opened 2026-01-09
LF-5111 Set up Redux state to store survey data

kathyavini:
**Description**

Alright, that was a bit of an adventure! Took longer than expected.

This PR hooks the TAPE form up to Redux. Now the form can be resumed from any point, reloaded, and edited after completion. Insights tracks the state of the form and routes to either the survey or the results as applicable.

Code-wise, the PR does the following:
- create a Redux slice with selectors and actions for interacting with TAPE
  - this includes one action (`clearSurvey()`) which is not used yet in app; I personally think we might want a reset that is not logout now that there is so much saving (see below), but I'll leave that to Loïc to decide next week or for Hannah and co. to comment tomorrow
- add handlers to the general Survey component for `onValueChanged` and `onCurrentPageChanged` in addition to the original `onComplete`
  - in the case of TAPE I actually didn't end up using the `onCurrentPageChanged` -- I just go ahead and save to Redux `onValueChanged` (= `onBlur` of each input). However, I left it in, in case a different survey wanted to react to that event
  - all handlers now return the current state of the survey in terms of a) page number (so it can be resumed from the same place) and b) survey content
- add a bit of status state to the slice for use in `<Insights />` so the survey status and routing can be adjusted from that main Insight page (works pretty well, I think!)
-  adds the little button to return to the survey which also removes the "completed" state. This only matters for that Insight main page; I don't really do anything with completed or not in the results page. And actually I think the way the chart data logic was implemented (0 for skipped scores) means that if you navigate right to `/tape/results` with no data, you just see a chart with all values 0. I think that's totally fine so I removed the "No results" fallback and string

Jira link: https://lite-farm.atlassian.net/browse/LF-5111

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/SurveyComponent/index.tsx
+++ b/packages/webapp/src/components/SurveyComponent/index.tsx
@@ -13,7 +13,7 @@
  *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
  */
 
-import { useCallback, useMemo } from 'react';
+import { useCallback, useEffect, useMemo } from 'react';
 import { Model } from 'survey-core';
 import { Survey } from 'survey-react-ui';
 import { DefaultLight } from 'survey-core/themes';
@@ -85,15 +85,29 @@ export default function SurveyComponent({
     [onValueChanged],
   );
 
-  survey.onComplete.add(handleComplete);
+  useEffect(() => {
+    survey.onComplete.add(handleComplete);
 
-  if (onCurrentPageChanged) {
-    survey.onCurrentPageChanged.add(handleCurrentPageChanged);
-  }
+    if (onCurrentPageChanged) {
+      survey.onCurrentPageChanged.add(handleCurrentPageChanged);
+    }
 
-  if (onValueChanged) {
-    survey.onValueChanged.add(handleValueChanged);
-  }
+    if (onValueChanged) {
+      survey.onValueChanged.add(handleValueChanged);
+    }
+
+    return () => {
+      survey.onComplete.remove(handleComplete);
+
+      if (onCurrentPageChanged) {
+        survey.onCurrentPageChanged.remove(handleCurrentPageChanged);
+      }
+
+      if (onValueChanged) {
+        survey.onValueChanged.remove(handleValueChanged);
+      }
+    };
+  }, [survey, handleComplete, handleCurrentPageChanged, handleValueChanged]);
 
   return <Survey model={survey} />;
 }
--- a/packages/webapp/src/components/SurveyComponent/index.tsx
+++ b/packages/webapp/src/components/SurveyComponent/index.tsx
@@ -21,12 +21,17 @@ import 'survey-core/survey-core.css';
 
 interface SurveyComponentProps {
   surveyJson: any; // Survey JSON schema object
-  onComplete: (surveyData: any) => void;
+  onComplete: (currentPageNo: number, surveyData: any) => void;
   initialData?: Record<string, any>;
   initialPageNo?: number;
   onCurrentPageChanged?: (currentPageNo: number, surveyData: Record<string, any>) => void;
   onValueChanged?: (currentPageNo: number, surveyData: Record<string, any>) => void;
 }
+
+const extractSurveyState = (model: Model) => ({
+  currentPageNo: model.currentPageNo,
+  surveyData: model.data,
+});
 
 export default function SurveyComponent({
   surveyJson,
@@ -58,16 +63,15 @@ export default function SurveyComponent({
   // https://surveyjs.io/form-library/documentation/get-started-react
   const handleComplete = useCallback(
     (surveyModel: Model) => {
-      const surveyData = surveyModel.data;
-      onComplete(surveyData);
+      const { currentPageNo, surveyData } = extractSurveyState(surveyModel);
+      onComplete(currentPageNo, surveyData);
     },
     [onComplete],
   );
 
   const handleCurrentPageChanged = useCallback(
     (surveyModel: Model) => {
-      const currentPageNo = surveyModel.currentPageNo;
-      const surveyData = surveyModel.data;
+      const { currentPageNo, surveyData } = extractSurveyState(surveyModel);
       onCurrentPageChanged?.(currentPageNo, surveyData);
     },
     [onCurrentPageChanged],
@@ -75,8 +79,7 @@ export default function SurveyComponent({
 
   const handleValueChanged = useCallback(
     (surveyModel: Model) => {
-      const currentPageNo = surveyModel.currentPageNo;
-      const surveyData = surveyModel.data;
+      const { currentPageNo, surveyData } = extractSurveyState(surveyModel);
       onValueChanged?.(currentPageNo, surveyData);
     },
     [onValueChanged],
--- a/packages/webapp/src/components/SurveyComponent/index.tsx
+++ b/packages/webapp/src/components/SurveyComponent/index.tsx
@@ -13,7 +13,7 @@
  *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
  */
 
-import { useCallback } from 'react';
+import { useCallback, useMemo } from 'react';
 import { Model } from 'survey-core';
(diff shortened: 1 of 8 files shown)


packages/webapp/src/components/SurveyComponent/index.tsx
```diff
@@ -13,42 +13,101 @@
...
   surveyJson,
   onComplete,
   initialData,
+  initialPageNo = 0,
+  onCurrentPageChanged,
+  onValueChanged,
 }: SurveyComponentProps) {
-  const survey = new Model(surveyJson);
+  // Memoize to create the survey model only once, even as saved data changes and component re-renders
+  const survey = useMemo(() => {
```
kathyavini, line 45: Not an optimisation, but rather the UX gets all wonky if a new model is created on each render; learned you definitely need a stable model reference here, and now (with the `initialData` also including `savedData`), re-renders happen frequently.

Duncan-Brain: I am not seeing any wonkyness -- what did you see? Sorry I always pick on `useMemo()` it is an attempt to best use react.

If its not an optimisation from react docs:

> You should **only** rely on useMemo as a performance optimization. If your code doesn’t work without it, find the underlying problem and fix it first. Then you may add useMemo to improve performance.

Maybe `useRef()` would be a good option for stable reference? I saw that pattern used in maps with markers (geometries) , clusters, and drawing managers

packages/webapp/src/containers/Insights/TapeSurvey/tapeSurveySlice.ts
```diff
@@ -0,0 +1,68 @@
...
+    },
+    completeSurvey: (
+      state,
+      action: PayloadAction<{ currentPageNo: number; surveyData: Record<string, any> }>,
+    ) => {
+      state.currentPageNo = action.payload.currentPageNo;
+      state.surveyData = action.payload.surveyData;
+      state.isCompleted = true;
+    },
+    reopenSurvey: (state) => {
```
Duncan-Brain, line 50: This might be a personal preference but wondering if `currentPageNo` should be reset here. I might expect to go to the start of the survey if I click edit my results -- maybe not I dno.

Duncan-Brain (changes requested):
You had fun last night!

Looks great and nice job putting that reducer together so quick.

Only three small things from me, only the useMemo is important to me:

useMemo - if not used for optimisation it probably shouldnt be used, I didnt personally see a difference when removing it?

currentPageNumber on reopenSurvey click -> minor question on preferences

The breadcrumbs.. I am not sure it is fixable from `surveyjs` but they seem to change color based on a `visited` status vs a `complete` status. For future use of the library I wonder if that can be edited

packages/webapp/src/components/SurveyComponent/index.tsx
kathyavini: @Duncan-Brain Okay maybe I should have specified it's not JUST an optimisation. But to see the buggy UX that comes from re-creating the model vs maintaing the same model, use the page change handler instead of the `onChange` handle:

```ts
onCurrentPageChanged={handleDataChange}
```

Now click through some pages and see what happens when the model is re-created on page click

packages/webapp/src/containers/Insights/TapeSurvey/tapeSurveySlice.ts
kathyavini: I was trying to decide between the two last night! Initially I preferred the feeling of "editing your answers" picking up right where you left off (= the last page) but I can definitely see how reviewing your answers would more naturally mean from the start -- I'll go with start.

packages/webapp/src/components/SurveyComponent/index.tsx
kathyavini: But as for the optimisation part, also React docs on useMemo -- [How to tell if a calculation is expensive](https://react.dev/reference/react/useMemo#how-to-tell-if-a-calculation-is-expensive)

> Perform the interaction you’re measuring (for example, typing into the input). You will then see logs like filter array: 0.15ms in your console. If the overall logged time adds up to a significant amount (say, 1ms or more), it might make sense to memoize that calculation.

If you log the model creation and setup, it's >50ms each time.

Edit: oooh! Just noticed through logging I need to be passing a stable reference to the handlers as well so the listeners aren't added each time (even with `useMemo()`). Without `useMemo()` that is another thing that would happen on each form interaction, since the listeners need to be added per model.

Duncan-Brain: Thanks for letting me know how to see wonkyness I will check that out! The second part of that useMemo docs line:

>  If your code doesn’t work without it, find the underlying problem and fix it first. Then you may add useMemo to improve performance.

Just wondering if useRef() is more appropriate for the stable reference part --> it seems reasonable to have a stable reference, and by the sounds of it optimisation is a good idea here! Just want to make sure the code works as desired without it!

Edit: hmm for some reason I still cant see the wonkyness adding that line and removing useMemo -- the event listeners is good enough for me to need a stable reference but I wish I could get that going

kathyavini:
> The breadcrumbs.. I am not sure it is fixable from surveyjs but they seem to change color based on a visited status vs a complete status. For future use of the library I wonder if that can be edited

Unfortunately it looks like there aren't that many [properties available to configure](https://surveyjs.io/form-library/documentation/api-reference/survey-data-model#progressBarType) on `progressBar`. There is tracking questions instead of pages, but I don't think that's what we want either 🤔 Personally I don't like how when you return to the component it resets, although that could probably be tweaked (at the cost of storing more data in Redux) by keeping the whole [`uiState`](https://surveyjs.io/form-library/documentation/api-reference/survey-data-model#uiState) instead of just the page number. Something to play around with if we get feedback about this after this first POC.

As for the `useMemo()`, I see what you mean; if you are only using the `onValueChanged` handler it doesn't really _look_ like anything is happening when you re-recreate the model on each render. But if you use the logging suggested in the `useMemo()` docs (`console.time` + `console.timeEnd`) it is really quite a slow/expensive thing.

packages/webapp/src/components/SurveyComponent/index.tsx
kathyavini: @Duncan-Brain the scroll position doesn't reset to the top for me when I click through pages using the bottom navigation buttons -- you don't see that? (i.e. the page loads somewhere in the middle instead of the top)

Duncan-Brain: 🙏 oooh yes I did see that! Okay I was expecting like flickering or some other weirdness haha. Thanks

kathyavini: @Duncan-Brain did you still want to talk about `useMemo()` vs `useRef()` here? Is it still blocking for you? I think both can be made to work, but can we bring this to tech daily on Monday? This PR and the next have to get up to beta basically kind of now 😂

If you want to whip up the quick refactor on your end for us to look at in tech daily that would also be helpful; we could look at what's different 🙏

Duncan-Brain (approved):
Approving so that Hannah can test the Demo tonight. Everything works good enough!

I do not technically approve of the useMemo being used as it is just yet. My understanding is all functionality should work as expected without it, then be optimized once it works standalone. If we can continue that discussion that would be great.

packages/webapp/src/components/SurveyComponent/index.tsx
Duncan-Brain: Uhm yah I think its important to follow the guidelines of "If your code doesn’t work without it, find the underlying problem and fix it first." It sounds like a very strong recommendation!

Sorry if its annoying... I do take that as a strong recommendation from react. 

I think they say that because it can throw away the cached value randomly for varying reasons. They say it is "not a semantic guarantee" which could as you said could add more unnecessary listeners, make multiple instances, etc. 

We can talk monday of course since you don't agree. I don't know as much about useRef() as I would like so maybe I will learn something and do that if I have time!

PR #3995, opened 2026-01-09
LF-5109 Pre-populate survey with some LiteFarm data

kathyavini:
**Description**

A very basic proof of concept for pre-populating the initial fields of the TAPE survey before the demonstration with Hannah today. "Country" and the next two Location fields should be populated (provided Google has those address components for the given address!) So far I haven't seen any addresses that don't fill all three values, but I'm sure it's possible. Lat/Lng is populated with the values in `grid_points`

I'm not going to say that I have the utmost confidence that the code for this is ideal as it's a bit of a last-minute whip-up, I think it should serve for the demo this afternoon 🙂

Note: I didn't implement Loïc's requested info text, because the only way I could figure to do that was to add an HTML element to the question JSON; however, that means the info indicator would be present whether or not the field actually got populated, which I didn't think was ideal. This probably could be revisited next week.

Jira link: https://lite-farm.atlassian.net/browse/LF-5109

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/containers/Insights/TapeSurvey/index.tsx
+++ b/packages/webapp/src/containers/Insights/TapeSurvey/index.tsx
@@ -17,6 +17,7 @@ import { useCallback } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 import { useHistory } from 'react-router-dom';
 import { useTranslation } from 'react-i18next';
+import { useTapeSurveyPrepopulatedData } from './useTapeSurveyPrepopulatedData';
 import { saveSurveyProgress, completeSurvey, tapeSurveySelector } from './tapeSurveySlice';
 import SurveyComponent from '../../../components/SurveyComponent';
 import surveyJson from './tapeQuestions.json';
@@ -27,13 +28,7 @@ function TAPESurvey() {
   const history = useHistory();
   const dispatch = useDispatch();
 
-  // TODO LF-5109: Prepare prepopulated data
-  const prepopulatedData = {
-    // Country
-    // Lat / Lng
-    // Do you raise animals
-    // Number of unique species in crop management plans
-  };
+  const { prepopulatedData, isLoading } = useTapeSurveyPrepopulatedData();
 
   const { surveyData: savedData, currentPageNo: savedPageNo } = useSelector(tapeSurveySelector);
 
@@ -51,13 +46,15 @@ function TAPESurvey() {
   return (
     <>
       <PageTitle title={t('INSIGHTS.TAPE.TITLE')} backUrl="/Insights" />
-      <SurveyComponent
-        surveyJson={surveyJson}
-        onComplete={handleComplete}
-        onValueChanged={handleDataChange}
-        initialData={initialData}
-        initialPageNo={savedPageNo}
-      />
+      {!isLoading && ( // wait for prepopulated data to load
+        <SurveyComponent
+          surveyJson={surveyJson}
+          onComplete={handleComplete}
+          onValueChanged={handleDataChange}
+          initialData={initialData}
+          initialPageNo={savedPageNo}
+        />
+      )}
     </>
   );
 }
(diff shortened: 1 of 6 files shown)


packages/webapp/src/containers/Insights/TapeSurvey/tapeSurveySlice.ts
```diff
@@ -61,7 +61,8 @@ export default tapeSurveySlice.reducer;
 
 // Selectors
 export const tapeSurveySelector = (state: any) =>
-  state.entitiesReducer[tapeSurveySlice.name] || initialState;
+  // TODO: re-evaluate placement of tapeSurveySlice in farmStateReducer when adding backend
+  state.farmStateReducer[tapeSurveySlice.name] || initialState;
```
kathyavini, line 65: A bit of quick fix solution thus the comment to revisit.

Unlike the other reducers in `entitiesReducer`, this slice doesn't have a `farm_id`, so we can't do the normal filtering on `farm_id` that the selectors normally to do show on the farm-relevant data without purging the rest (whether that is even a good pattern being a whole different issue 🙂).

Here we could also pass a `farm_id` from the component and add the `loginSelector` logic other entities have, but the code change would be a bit more involved, and since I'm not sure how the slice will look after adding a backend (and RTK query); this seemed cleaner for now; `farmStateReducer`, _is_ cleared on farm change.

Duncan-Brain: I am not sure I would have caught on that anything was going on haha So 👍  for now haha

Should we make a specific Jira ticket to reference in the comment so we don't forget about it? "Comment out TAPE" or "Wrap up Tape for prod" ticket.

packages/webapp/src/containers/Insights/TapeSurvey/useTapeSurveyPrepopulatedData.ts
```diff
@@ -0,0 +1,77 @@
...
+export const useTapeSurveyPrepopulatedData = () => {
+  const { isLoaded } = useGoogleMapsLoader(['geocoding']);
+
+  // @ts-expect-error -- userFarmSelector issue
+  const userFarm: UserFarm = useSelector(userFarmSelector);
+
+  const [prepopulatedData, setPrepopulatedData] = useState({});
+  const [isLoading, setIsLoading] = useState(true);
+
+  useEffect(() => {
```
Duncan-Brain, line 43: I am not sure I have seen this pattern before -- maybe I have --  of editing the state of the hook using a function call in the hook. I am wondering if this hook should return the function itself instead of the data.

packages/webapp/src/containers/Insights/TapeSurvey/index.tsx
```diff
@@ -27,13 +28,7 @@ function TAPESurvey() {
...
   const dispatch = useDispatch();
 
-  // TODO LF-5109: Prepare prepopulated data
-  const prepopulatedData = {
-    // Country
-    // Lat / Lng
-    // Do you raise animals
-    // Number of unique species in crop management plans
-  };
+  const { prepopulatedData, isLoading } = useTapeSurveyPrepopulatedData();
```
Duncan-Brain, line 31: I don't think this should be called on every render when the value changes.

Duncan-Brain (dismissed):
Approving for pre-demo demo with Hannah today.

Left a few comments that I think are worth doing in the future.

packages/webapp/src/containers/Insights/TapeSurvey/index.tsx
kathyavini: I agree! And also on the comment above on the hook; I want to look at that with fresh eyes next week! 👀

Duncan-Brain (approved):
Re-approving with branch change

packages/webapp/src/containers/Insights/TapeSurvey/tapeSurveySlice.ts
kathyavini: > Should we make a specific Jira ticket to reference in the comment so we don't forget about it? "Comment out TAPE" or "Wrap up Tape for prod" ticket.

That's a good idea; I'll make a placeholder ticket in this user story and add it to release Jira when we have one. I suspect it will be comment out but we can update the ticket if Divya comes back to us with something different.

I think I have to do a little adjust on the consent logic (still waiting to hear if anything else came out of Friday's meeting) so I can update the comment reference then

PR #3997, opened 2026-01-13
LF-5111 Replace useMemo with other hooks

Duncan-Brain:
**Description**

The branch `tests/useref_for_survey` is an exploration of react hooks This PR is the result of that. 

We want to:
- avoid rebuilding the Model class when possible

useMemo is appropriate for cases where there are expensive computations. A class instantiation could be considered an expensive computation. But it is not necessary to achieve a single instance instantiation.

useRef is appropriate for cases where we want a really stable reference to something that we do not expect to need to re-render based on that reference.

The example shown in tech-daily for useRef was for the case where surveyJson is not expected to change within the component lifecycle. So changing that prop would have no effect. Something the useMemo version had that the useRef example did not. 
    
 **Notes**
- I wanted to preserve the useMemo functionality of allowing surveyJson to change
- I finally found terminology for the useRef pattern:  [“Singleton”. Use cases exist but it may not be best for this case, and there can be drawbacks](https://medium.com/@ignatovich.dm/singletons-in-react-applications-when-and-how-to-use-them-effectively-2a943691949d)
- the example shown in the useRef docs the class instance has no props. So it is more valid in that case.
- re-evaluated the need for a "strong" stable reference => our goal is to reduce not necessarily to prevent re-render
- the code still does not work properly when useMemo is removed (my main concern)
- [lazy state initialization of useState](https://react.dev/reference/react/useState#avoiding-recreating-the-initial-state) + useEffect can mimic the helpful properties we are looking for from useMemo and useRef
- weird thing: the test branch shows very low re-render time for recreating survey: makes me think we profiled the rendering wrong (useMemo recreation = ~40ms, all other methods using useEffect = ~2x ~0.01ms)
    

Jira link:

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/SurveyComponent/index.tsx
+++ b/packages/webapp/src/components/SurveyComponent/index.tsx
@@ -13,11 +13,31 @@
  *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
  */
 
-import { useCallback, useEffect, useMemo } from 'react';
+import { useCallback, useEffect, useRef, useState } from 'react';
 import { Model } from 'survey-core';
 import { Survey } from 'survey-react-ui';
 import { DefaultLight } from 'survey-core/themes';
 import 'survey-core/survey-core.css';
+
+const createSurveyAndInitialData = (
+  surveyJson: any,
+  initialData?: Record<string, any>,
+  initialPageNo?: number,
+) => {
+  const survey = new Model(surveyJson);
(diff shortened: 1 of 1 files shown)


Duncan-Brain:
> the code still does not work properly when useMemo is removed (my main concern)

Not worth investigating as slightly more complex.

PR #4003, opened 2026-01-21
LF-5117: Disable prohibited flows

SayakaOno:
**Description**

Hide or disable buttons that trigger prohibited flows when offline.
- This PR does not address disabling the creation of a new cleaning or pest control product during task creation, nor task wage changes.
- The `@navStyles` alias was introduced to make hiding elements easier.

Changes:
- Buttons/elements hidden via `@navStyles`
   - Crops: Add to your farm, Add crop / variety / crop plan / edit crop plan / delete crop plan
   - Animals: single animal view three-dot menu
   - Tasks: add custom harvest use for completion, already completed flow, manage your custom task
   - Inventory: floating button for adding product
   - Additional items not mentioned in the ticket description:
      - Add a crop (location’s Crops tab)
      - Edit / abandon / complete plan
      - Retire / edit location
- Buttons/elements hidden using `useIsOffline` hook
   - Inventory form buttons (edit, duplicate, delete)
   - Animal inventory table checkboxes
   - Map's Add Location and Export Map buttons
- Buttons disabled using `useIsOffline` hook
   - Finances' Add Transaction floating button and buttons on the carrousel
   - Animal's Add animal floating button

Note:
I went ahead and hid form buttons after confirming that buttons are being hidden as needed (e.g., edit / retire buttons for workers). If that’s not ideal, I’m open to disabling them, although disabling location form buttons would require prop drilling across many files.

Jira link: https://lite-farm.atlassian.net/browse/LF-5117

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/containers/Navigation/styles.module.scss
+++ b/packages/webapp/src/containers/Navigation/styles.module.scss
@@ -28,5 +28,9 @@
     @include xs-breakpoint {
       --global-navbar-height: calc(55px + 24px);
     }
+
+    .hideWhenOffline {
+      display: none;
+    }
   }
 }
--- a/packages/webapp/src/components/Animals/AnimalSingleViewHeader/index.tsx
+++ b/packages/webapp/src/components/Animals/AnimalSingleViewHeader/index.tsx
@@ -35,6 +35,7 @@ import {
   chooseAnimalTypeLabel,
 } from '../../../containers/Animals/Inventory/useAnimalInventory';
 import styles from './styles.module.scss';
+import navStyles from '@navStyles';
 
 type MenuItemProps = { iconName: IconName; text: string };
 
@@ -112,7 +113,9 @@ const ContainerWithButtons = ({
           <MeatballsMenu
             disabled={!!isEditing}
             options={options}
-            classes={{ button: isEditing ? styles.editingStatusButton : '' }}
+            classes={{
+              button: clsx(isEditing ? styles.editingStatusButton : '', navStyles.hideWhenOffline),
+            }}
           />
         )}
       </div>
--- a/packages/webapp/src/components/Animals/Inventory/index.tsx
+++ b/packages/webapp/src/components/Animals/Inventory/index.tsx
@@ -63,6 +63,7 @@ export type PureAnimalInventoryProps = {
   showActionFloaterButton: boolean;
   extraRowSpacing?: boolean;
   hideNoResultsBlock?: boolean;
+  disableActionFloaterButton?: boolean;
 };
 
 const PureAnimalInventory = ({
@@ -89,6 +90,7 @@ const PureAnimalInventory = ({
   showActionFloaterButton,
   extraRowSpacing,
   hideNoResultsBlock,
+  disableActionFloaterButton,
 }: PureAnimalInventoryProps) => {
   const { t } = useTranslation();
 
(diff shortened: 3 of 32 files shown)


SayakaOno:
This PR is ready for review, thank you! 🙏 

Update: There was one thing I wanted to fix, so I made one more commit. No more changes will be added 🙇

packages/webapp/src/containers/Navigation/styles.module.scss
```diff
@@ -28,5 +28,9 @@
     @include xs-breakpoint {
       --global-navbar-height: calc(55px + 24px);
     }
+
```
kathyavini, line 31: ❤️‍🔥❤️‍🔥❤️‍🔥

Never thought to use the class like this when I set it up for the hight adjust 😁 Fantastic!

kathyavini (commented):
I love `navStyles.hideWhenoffline` so much!! With the alias config it's so perfect 👌

Thank you also for catching all those areas I missed on the Jira ticket! 🙇 Having clicked around for quite a while, I think I spotted just a few more:

- **Crops**: We need to also hide the three-dot menu for the "repeat crop plan" (sorry *completely* forgot that!)
- **Finances (Main Page)**: We'll need to disable the "Report" feature. Loïc provided a disabled style [here in Figma](https://www.figma.com/design/eMffC0h3JPlWtzNttxJpil/Litefarm-Updated?node-id=62824-228985&m=dev), but I think hiding is also totally fine
- **Finances (Transactions)**: When a transaction is expanded, there is a "View & edit" option. I'm not sure what the best thing to do is here! Is it easiest to remove that link entirely? Otherwise the delete and edit options would need to be removed from the following pages. That seems like a little bit more work, but maybe more in the spirit of "readonly" finances?

While doing the offline testing, I also noticed that notifications... can't be clicked on while offline??! Do you see that behaviour? It's so weird; I wonder if there is some sort of API call dependency in that flow. I will _probably_ ticket it, but if we can get Loïc's permission, it would definitely be easier to just flat-out disable notifications as well...!

SayakaOno:
Thank you Joyce! I hid all the ones you spotted.

The notification button works fine for me! Is it not always working?
I think it'll be fine to just disable it for now.

----------
A patch tasks re-completion API test just failed:
https://github.com/LiteFarmOrg/LiteFarm/actions/runs/21376656939/job/61534227712?pr=4003#:~:text=Summary%20of%20all,js%3A3159%3A46)

It may or may not be the same test that failed before (I couldn't find it on slack). I'm not going to investigate it now given the time constraint.

kathyavini (approved):
Oh, I _love_ how easy it is to hide additional things with `navStyles.hideWhenOffline` 😁 

> The notification button works fine for me! Is it not always working?

It's not the notification list itself that isn't loading, but rather when you are viewing the list, you can't click on any notification entry to see its details. I just checked with the network tab, and indeed it is the case that every time you click on a notification tile, it's a PATCH to `/notification_user`. So when that fails, as it will when offline, the page transition also doesn't happen! Kind of wild.

(Totally fixable I'm sure -- that network request probably doesn't need to be blocking -- but disabling the list entirely is a pleasingly low-effort approach, and then we don't have to worry about re-synching up the read notification state, or the fact that new notifications aren't coming in while offline.)

SayakaOno:
Ah, okay...
I think it’s also okay for users to see the list, as long as it’s clearly indicated that the notifications aren’t clickable. If that doesn’t add much value, we could disable it. Let’s ask Loïc tomorrow, I’ve updated the tech daily document!

PR #4006, opened 2026-01-24
LF-5124 Remove the ability to set an hourly wage from the task flows

kathyavini:
**Description**

Okay, this PR went through some phases of being *quite* convoluted, but I think with the new modal, it's become much more manageable. Still a lot of files touched, but many are now just straight removal of logic -- hopefully not too bad to review 🤞 

**Summary of changes:**
- `updateUserFarmWage` and `setUserFarmWageDoNotAskAgain` options were removed from the `<HourlyWageInputs />` radio group along with all supporting logic
  - The `patchWageUrl` exists unchanged on the backend, but is now only called through `updateUserFarmSaga`
  - The `setUserFarmWageDoNotAskAgain` flow was cleaned up, and the saga itself marked as _deprecated_ rather than deleted. This is a full-stack flow that would include an irreversible migration + test changes to completely purge; the deprecation is a reminder that the functionality still exists on the backend
- The bifurcated logic between assignees who have a wage and those two don't has been removed; now all admin users see the same two radios
  - Note: the removed logic was defined in `contentForWorkerWithWage` and only seen in the task creation flow
- A new UI element reports the userFarm wage in `<TaskAssign />` if there is one 
- the saga and the API on `patchTaskWage` have been updated so that `override_hourly_wage` can be _**unset**_ instead of just set
- `override_hourly_wage` is returned back to the frontend so it can be displayed once it's set
- wage logic has been removed from the `<QuickAssignModal />`, which is accessed through `<TaskCard />` and `<TaskReadOnly />`
  - Now that logic exists only in the new `<EditTaskWageModal />`
- There is a new UI element on `<PureTaskReadOnly />` to display the task wage
- the offline adjusts related to this part of this app have been added:
  - the snackbar + optimistic update on `patchTaskWage`
  - hiding the "Assign all tasks on this date to {{name}}" checkbox when offline

Jira link: https://lite-farm.atlassian.net/browse/LF-5124

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/taskController.js
+++ b/packages/api/src/controllers/taskController.js
@@ -440,13 +440,33 @@ const taskController = {
   async patchWage(req, res) {
     try {
       const { task_id } = req.params;
-      const { wage_at_moment } = req.body;
+      const { farm_id } = req.headers;
+      const { override_hourly_wage } = req.body;
+      let { wage_at_moment } = req.body;
+
+      // When unsetting override, look up assignee's current farm wage
+      if (override_hourly_wage === false) {
+        const { assignee_user_id } = await TaskModel.query()
+          .select('assignee_user_id')
+          .findById(task_id)
+          .first();
+
+        if (assignee_user_id) {
+          const userFarm = await UserFarmModel.query()
+            .where({ user_id: assignee_user_id, farm_id })
+            .first();
+          wage_at_moment = userFarm?.wage?.amount || null;
+        } else {
+          wage_at_moment = null;
+        }
+      }
 
       const result = await TaskModel.query()
         .context(req.auth)
         .findById(task_id)
-        .patch({ wage_at_moment, override_hourly_wage: true });
-      return result ? res.sendStatus(200) : res.status(404).send('Task not found');
+        .patch({ wage_at_moment, override_hourly_wage })
+        .returning('*');
+      return result ? res.status(200).send(result) : res.status(404).send('Task not found');
     } catch (error) {
       console.log(error);
       return res.status(400).json({ error });
@@ -1114,8 +1134,9 @@ const taskController = {
   async getIrrigationTaskTypes(req, res) {
     const { farm_id } = req.params;
     try {
-      const irrigationTaskTypes =
-        await IrrigationTypesModel.getAllIrrigationTaskTypesByFarmId(farm_id);
+      const irrigationTaskTypes = await IrrigationTypesModel.getAllIrrigationTaskTypesByFarmId(
+        farm_id,
+      );
       res.status(200).json(irrigationTaskTypes);
     } catch (error) {
       return res.status(400).send(error);
--- a/packages/api/src/controllers/taskController.js
+++ b/packages/api/src/controllers/taskController.js
@@ -440,33 +440,12 @@ const taskController = {
   async patchWage(req, res) {
     try {
       const { task_id } = req.params;
-      const { farm_id } = req.headers;
-      const { override_hourly_wage } = req.body;
-      let { wage_at_moment } = req.body;
-
-      // When unsetting override, look up assignee's current farm wage
-      if (override_hourly_wage === false) {
-        const { assignee_user_id } = await TaskModel.query()
-          .select('assignee_user_id')
-          .findById(task_id)
-          .first();
-
-        if (assignee_user_id) {
-          const userFarm = await UserFarmModel.query()
-            .where({ user_id: assignee_user_id, farm_id })
-            .first();
-          wage_at_moment = userFarm?.wage?.amount || null;
-        } else {
-          wage_at_moment = null;
-        }
-      }
-
+      const { wage_at_moment, override_hourly_wage } = req.body;
       const result = await TaskModel.query()
         .context(req.auth)
         .findById(task_id)
-        .patch({ wage_at_moment, override_hourly_wage })
-        .returning('*');
-      return result ? res.status(200).send(result) : res.status(404).send('Task not found');
+        .patch({ wage_at_moment, override_hourly_wage });
+      return result ? res.sendStatus(200) : res.status(404).send('Task not found');
     } catch (error) {
       console.log(error);
       return res.status(400).json({ error });
--- a/packages/api/tests/task.test.js
+++ b/packages/api/tests/task.test.js
@@ -2041,8 +2041,9 @@ describe('Task tests', () => {
       await mocks.soil_amendment_taskFactory({ promisedTask: [{ task_id }] });
 
       const new_soil_amendment_task = fakeTaskData.soil_amendment_task(farm_id);
-      const new_soil_amendment_task_products =
-        await fakeProductData.soil_amendment_task_products(farm_id);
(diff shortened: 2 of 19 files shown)


packages/api/src/controllers/taskController.js
```diff
@@ -1114,8 +1113,9 @@ const taskController = {
   async getIrrigationTaskTypes(req, res) {
     const { farm_id } = req.params;
     try {
-      const irrigationTaskTypes =
-        await IrrigationTypesModel.getAllIrrigationTaskTypesByFarmId(farm_id);
+      const irrigationTaskTypes = await IrrigationTypesModel.getAllIrrigationTaskTypesByFarmId(
```
kathyavini, line 1116: Linting

packages/api/tests/task.test.js
```diff
@@ -2041,8 +2041,9 @@ describe('Task tests', () => {
       await mocks.soil_amendment_taskFactory({ promisedTask: [{ task_id }] });
 
       const new_soil_amendment_task = fakeTaskData.soil_amendment_task(farm_id);
-      const new_soil_amendment_task_products =
-        await fakeProductData.soil_amendment_task_products(farm_id);
+      const new_soil_amendment_task_products = await fakeProductData.soil_amendment_task_products(
```
kathyavini, line 2044: Linting

packages/webapp/public/locales/en/translation.json
```diff
@@ -2178,7 +2178,7 @@
     "ERROR": "Wage must be a valid, non-negative decimal number",
     "HOURLY_WAGE": "Hourly Wage",
     "HOURLY_WAGE_RANGE_ERROR": "Hourly Wage must be a positive number less than 999,999,999",
-    "HOURLY_WAGE_TOOLTIP": "Hourly wages can be set by selecting an individual on the People tab under 'my farm'."
+    "HOURLY_WAGE_TOOLTIP": "This task-specific wage will override the assignee's default wage. Default worker wages can be managed in the People tab."
```
kathyavini, line 2190: I rather like this tooltip, but I totally made it up 😅 Loïc was already out for the day; will need checking next week.

Also I'm not sure what to call the "People" section other than "tab" -- I took that from the original text but I don't think it's the right term.

packages/webapp/src/components/Modals/EditTaskWageModal/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+    },
+  });
+
+  const selectedHourlyWageAction = watch(HOURLY_WAGE_ACTION);
+  const shouldSetWage = selectedHourlyWageAction === hourlyWageActions.FOR_THIS_TASK;
+
+  const onSubmit = (data: EditTaskWageFormFields) => {
+    if (data[HOURLY_WAGE_ACTION] === hourlyWageActions.FOR_THIS_TASK) {
+      onSave({
+        wage_at_moment: +Number(data[HOURLY_WAGE]).toFixed(2),
```
SayakaOno: `+` seems redundant :)

packages/webapp/src/components/Task/AssignTask/useTaskAssignForm.jsx
```diff
@@ -137,20 +134,7 @@ const useTaskAssignForm = ({
...
-        wage: { amount },
-        wage_do_not_ask_again,
-      } = userData;
-
-      const hasWage = !!(amount || wage_at_moment);
-      shouldShow = !hasWage && !wage_do_not_ask_again;
-    }
-    return shouldShow;
-  }, [user.is_admin, selectedWorker, wage_at_moment, assigned]);
+  const showHourlyWageInputs = user.is_admin && selectedWorker && assigned;
```
SayakaOno, line 137: Is an assignee required to show the input? Wage can be set without assignee on the read-only view, so I'm just checking!

SayakaOno (commented):
It works great!
This review probably wouldn’t have been so challenging if the feature had been built this way from the start, but what we had before (what I wrote) made it harder 😢 So I ended up relying more on manual testing than the code review itself.

I thought we weren’t updating translation keys or strings, so translations wouldn’t be affected after a release. Or was that before we started retaining asset copies? 😅🙏

packages/webapp/src/components/Modals/EditTaskWageModal/index.tsx
kathyavini: It does seem redundant! But I think we need both 😢 Without the `+` there is a type error, because the outer `+` is for the `.toFixed()` and `Number()` is for `data[HOURLY_WAGE]`.

Is this way maybe clearer? I don't love how it reads either, but the parentheses are nice 😂 Or maybe you know of a different way?
```ts
Number(Number(data[HOURLY_WAGE]).toFixed(2))
```

Edit: wait maybe this is fixable with one of the rounding functions, let me check! (Yesssss, duh! Much better)

packages/webapp/src/components/Task/AssignTask/useTaskAssignForm.jsx
kathyavini: That's a good question. In the original form, it felt _very_ buggy to me to show the hourly inputs without an assignee (because of the references to user wage) but I guess in the new view it's pretty innocuous 🤔 You mean just like this, right?

Image

I don't personally mind the radios just not appearing at all (the combination of thinking about wage + there not being an assignee still bothers me a _bit_ personally -- at least in this view, where there isn't any context beyond assigning) but I think it could go either way; do you prefer the view in the screenshot above?

kathyavini (commented):
> So I ended up relying more on manual testing than the code review itself.

This I definitely understand 😂 The requirements of the original ticket and the way the logic had to exist in three different containers really tripped me up when I first started the refactor; the structure was very different than I had expected! I think with just the change to the QuickAssignModal though it will read much easier in the future.

> I thought we weren’t updating translation keys or strings, so translations wouldn’t be affected after a release. Or was that before we started retaining asset copies? 😅🙏

OK thank you so much; I just completely forgot! No, the asset management doesn't help with this -- translations were never previously in the assets folder. Going forward, if we maintain the translations in the cache as done on #4008, then I think the UI and translations _might_ stay in sync perpetually (at the cost of the cache size)... BUT that actually is only for the next release anyway, not this one! I restored all the translations and made a new set for this feature, thank you 🙏

packages/webapp/src/components/Modals/EditTaskWageModal/index.tsx
SayakaOno: Ohhh, I didn't know that `+` converts the result of `Number(data[HOURLY_WAGE]).toFixed(2)` to a number -- my oversight 🙏
`roundToTwo` looks good, thank you!!

packages/webapp/src/components/Task/AssignTask/useTaskAssignForm.jsx
SayakaOno: I’m not particular about it, I’m good either way 🙂 Let’s leave it as is!

SayakaOno (approved):
Looks good! 🎉

PR #4007, opened 2026-01-29
LF-5103: Frontload data that is currently loaded on demand

SayakaOno:
**Description**

- Call `useGoogleMapsLoader` in `App.jsx` to ensure maps can render offline without requiring an initial online visit to the map page
- Add null check for `e.response` in finances saga (I believe this was the cause of infinite loading on the transactions page)
- Update `fieldWorkFailure` and `fieldWorkLoading` reducers to avoid resetting `fieldWorkTypes` to `[]`, which was preventing field work types from being available offline
- Pre-fetch required data in `fetchAllSaga` for offline task creation and read-only views:
   - Field work tasks
   - Irrigation tasks
   - Soil amendment tasks
   - Animal movement tasks
   - Finance transactions (read-only)
   - Animals (read-only etc.)
- Configure RTK Query caching at the API level:
   - `keepUnusedDataFor`: 7 days
   - `refetchOnMountOrArgChange`: 60 seconds

#### Why change the defaults?

With RTK Query's default settings, cached data expires 60 seconds after the component using it unmounts. For example, if you visit the animals page, go offline, navigate away, and return after 60 seconds, the cached data will be gone and you won't see any data while offline. To support offline functionality, we need to adjust the cache configuration.

#### New Configuration

`keepUnusedDataFor`: 7 days ([docs](https://redux-toolkit.js.org/rtk-query/usage/cache-behavior#reducing-subscription-time-with-keepunuseddatafor))
- Retains cached data for 7 days (matching our token lifetime)
- Previously: Cache expired 60 seconds after leaving the component

`refetchOnMountOrArgChange`: 60 seconds ([docs](https://redux-toolkit.js.org/rtk-query/usage/cache-behavior#encouraging-re-fetching-with-refetchonmountorargchange))
- Automatically refetches if the last request was more than 60 seconds ago
- Previously: Only refetched when cache was empty

#### Impact

Example scenario: You visit the animals page, stay for an hour, navigate away, then immediately return.

- Old behaviour: No refetch (you returned within 60 seconds of leaving, so cache is still valid)
- New behaviour: Refetch happens (the last API call was over 60 seconds ago _- an hour -_ so data is considered stale)

This configuration is applied globally at the API level rather than per-endpoint. This approach:
1. Eliminates the need to configure caching for each individual endpoint required for offline functionality
2. Ensures consistent behaviour for future APIs without additional configuration

Note: 
With `refetchOnReconnect: true` ([docs](https://redux-toolkit.js.org/rtk-query/api/createApi#refetchonreconnect)) , queries should automatically refetch when going back online, but I wasn't able to test this successfully. Claude says there are known bugs with this feature in our `@reduxjs/toolkit` version, which might be the cause. (The newer version have breaking changes. I created a [ticket](https://lite-farm.atlassian.net/browse/LF-5128) for the upgrade)
The app should work fine without this feature due to the cache configuration, but I'm documenting this for future reference in case we need automatic refetching on reconnection.

Jira link: https://lite-farm.atlassian.net/browse/LF-5103

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/containers/fieldWorkSlice.js
+++ b/packages/webapp/src/containers/fieldWorkSlice.js
@@ -18,7 +18,6 @@ const fieldWorkSlice = createSlice({
     fieldWorkLoading: (state, action) => {
       Object.assign(state, {
         loading: true,
-        fieldWorkTypes: [],
       });
     },
     fieldWorkSuccess: (state, { payload }) => {
--- a/packages/webapp/src/containers/fieldWorkSlice.js
+++ b/packages/webapp/src/containers/fieldWorkSlice.js
@@ -31,7 +31,6 @@ const fieldWorkSlice = createSlice({
     },
     fieldWorkFailure: (state, action) => {
       state.loading = true;
-      state.fieldWorkTypes = [];
     },
   },
 });
--- a/packages/webapp/src/App.jsx
+++ b/packages/webapp/src/App.jsx
@@ -26,6 +26,7 @@ import { ANIMALS_URL, MAP_URL, SENSORS_URL } from './util/siteMapConstants';
 import { NavMenuControlsContext } from './contexts/appContext';
 import { useOfflineDetector } from './containers/hooks/useOfflineDetector/useOfflineDetector';
 import { useServiceWorkerListener } from './hooks/useServiceWorkerListener/useServiceWorkerListener';
+import { useGoogleMapsLoader } from './hooks/useGoogleMapsLoader';
 
 function App() {
   const location = useLocation();
@@ -36,6 +37,7 @@ function App() {
 
   useOfflineDetector();
   useServiceWorkerListener();
+  useGoogleMapsLoader();
 
   return (
     <div className={clsx(styles.container)}>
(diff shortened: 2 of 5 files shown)


packages/webapp/src/containers/fieldWorkSlice.js
```diff
@@ -31,7 +30,6 @@ const fieldWorkSlice = createSlice({
     },
     fieldWorkFailure: (state, action) => {
       state.loading = true;
-      state.fieldWorkTypes = [];
```
kathyavini, line 34: Am I missing something, or shouldn't this whole slice be named for types, like `fieldWorkTypesSlice`, `fieldWorkTypesFailure`? 😅

kathyavini (approved):
Oh WOW, it's such a delight to see such effective changes in 32 lines!! 😋😍

I love the new cache settings; much better default without the data getting any more stale 👌

> Add null check for e.response in finances saga (I believe this was the cause of infinite loading on the transactions page)

❤️

packages/webapp/src/containers/fieldWorkSlice.js
SayakaOno: I only noticed the file name... I'll update it, thank you!!

PR #4008, opened 2026-01-29
Lf 4950 translations break on page reload when offline i18n

Duncan-Brain:
**Description**

Sorry for delay pnpm build caught some type errors that pre-commit did not, rebase did not work this time either.  

Changes since a few months ago are only related to TFunction typings and reinstalling packages.

~~There might also be a bad merge too or something since api tests are failing... if you could still check to translations though.~~ Legit test failure the buildTransactions test is not pure.

Notes:
- Some formatting snuck through on on merge commit
- Uses[ i18n configuration to create a default backend](https://www.i18next.com/how-to/backend-fallback), as the namespaces are loaded it creates a fallback backend for the language of choice and English (our fallback for untranslated strings)
- Updates deprecated `xhr-backend` to `http-backend`
- Updates i18next packages (TFunction import updates)

Jira link: [LF-4950](https://lite-farm.atlassian.net/browse/LF-4950)

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain):
  - uses `pnpm run build` and `pnpm run preview` and using the offline radio button in service workers area of devtools to simulate the bug.  
  
  I also saw that two of each language file was downloaded on my branch:
  
Screenshot 2026-01-30 at 3 44 44 PM

**Checklist:**

[LF-4950]: https://lite-farm.atlassian.net/browse/LF-4950?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/webapp/src/locales/i18n.js
```diff
@@ -22,10 +26,27 @@ i18n
     react: {
       useSuspense: true,
     },
+    ns: ['crop', 'expense', 'task', 'market_directory_info'],
     backend: {
       queryStringParams: { v: APP_VERSION },
+      backends: [
+        HttpBackend,
+        resourcesToBackend((lng, ns) => {
```
Duncan-Brain, line 34: This is the core change that should create desired behaviour

packages/webapp/src/locales/i18n.js
```diff
@@ -22,10 +26,27 @@ i18n
     react: {
       useSuspense: true,
     },
+    ns: ['crop', 'expense', 'task', 'market_directory_info'],
```
Duncan-Brain, line 29: This property specifies namespaces which are loaded on init. Now that there is a fallback available, the `buildTransactions` tests have a mixed translation loaded state. I am thinking in the future it would be ideal to not load any translations on init.. or all of them.

juice928:
👋 Hi, I'm an automated AI code review bot. I ran some checks on this PR and found **1 point that might be worth attention** (could be false positives, please use your judgment):

1. <strong>Refinement of the offline i18n fallback logic and directory structure</strong>
   - [Location](https://github.com/LiteFarmOrg/LiteFarm/blob/2854ca323f80098e28eee21faa488d59f0cd1f47/packages/webapp/src/locales/i18n.js#L34-L41): `packages/webapp/src/locales/i18n.js:L34-L41`
   - Impact: Importing files from outside the `src` directory can cause build errors in many React environments, and the current conditional logic may prevent users from switching languages while offline.
   - Suggestion: Consider moving translation files into `src/locales` for better bundling and removing the `if (lng === i18n.language)` check to allow the fallback backend to load any requested language.

If you find these suggestions disruptive, you can reply "stop" , and I'll automatically skip this repository in the future.

packages/webapp/src/locales/i18n.js
SayakaOno: I agree. `ns` doesn't seem necessary...

SayakaOno (commented):
It works for me!
I removed the `String()` wrappers and initially got the `'TFunctionResult' is not assignable to type 'ReactI18NextChildren | Iterable'. Type 'object' is not assignable to type 'ReactI18NextChildren | Iterable'` error, but  restarting the TS server resolved it. 
Have you tried restarting your TS server? The `String()` wrappers might not be necessary.

kathyavini:
> I also saw that two of each language file was downloaded on my branch

@Duncan-Brain can you share further details about how you generated the screenshot you shared above, and what you were checking for there when you verified downloaded?

I see all 8 languages now in `dist/assets` and in the workbox precache -- which of course makes sense because the build is static... it couldn't have any idea of runtime user language... curious how you got just those two translation files in view?

Screenshot 2026-02-02 at 2 13 57 PM

Duncan-Brain:
@SayakaOno thanks restarting worked this time! I saw a few type errors that I restarted VSCode for but that one stayed around.

@kathyavini I go to devtool sources -- not the service worker dropdown but the other one(see pic). The service worker code was not in when I coded it 4 months ago. 

To be honest I am not sure exactly what is shown in those sources tabs lol 😂  It just seemed relevant that just two were shown so I shared... I haven't been able replicate all of them shown there but typically if I set my language to Malayalam I can see at least one set of two translation files from the first page I visit (filter.json > `/task`, animal.json> `/animals` etc) 

Screenshot 2026-02-03 at 11 26 05 AM

kathyavini:
> The service worker code was not in when I coded it 4 months ago.

Actually this part of the service worker (the precache) is unchanged relative to what we already have on production! It was just configured in a more automatic way before, but the behaviour is the same.

You can use this code snippet in the dev console to see how the cache size compares on this branch built vs integration:

```js
(async () => {
  const cacheNames = await caches.keys();
  const precacheName = cacheNames.find(name => name.startsWith('workbox-precache'));
  const cache = await caches.open(precacheName);
  const requests = await cache.keys();
  let totalSize = 0;

  for (const request of requests) {
    const response = await cache.match(request);
    if (response) {
      const blob = await response.clone().blob();
      totalSize += blob.size;
    }
  }

  // Convert to MB for easier reading
  console.log(`Total precache size: ${(totalSize / (1024 * 1024)).toFixed(2)} MB`);
})();
```

It looks like adding all the languages to the precache increases the size each user downloads from  **8.95MB** to **10.52MB**. The UX with all the translation in precache _is_ really nice -- it works very smoothly and reliably -- but that cache bump is not ideal. Especially since the worst offenders are the Indic languages and those are not complete yet; those files will grow, and everyone's download will grow with it.

I did a quick first pass attempt at the other solution described on the Jira ticket (a route-based cache on `/locales`), but since translations are a) also initially in browser memory (until reload) and b) served as-needed, I couldn't get all the `/locales` endpoints pre-hit as cleanly as the workbox pre-cache is pre-populated 🤔

Since this is a really nice UX, I think it's good for release. But I also think we should ultimately try to move off of it if we at all can in the future. As I mentioned in Jira, our starting point was that we didn't want all users to have to download all languages.

Duncan-Brain:
Thanks for the code snippet, I see the same-ish results (I see 9.0MB) !

> It looks like adding all the languages to the precache increases the size each user downloads from 8.95MB to 10.52MB.

~~I am very surprised at these results considering malayalam is  about 3KB sitting on my computer uncompressed and 3KB x 10 languages != 1.57MB. So I would be curious to know the cause of the bump and if it is all actually downloaded to the user. If that number includes the upgrade to `i18n` or something else... than to me that increase could be inevitable!~~

Edit: yup nvm was looking at wrong translation folder haha

> As I mentioned in Jira, our starting point was that we didn't want all users to have to download all languages.

That is my goal too! Thus the single if else statement attempting to store just two languages. If it is not working... then my code is not doing what I expect and this should be improved yes. I was not aware that the precache would store the entirety of the manifest so I either need to reduce the pre-cache manifest to ignore the translation files, or avoid compiling. 

I still feel very uneducated about sw stuff. But I think I finally understand more of the problem now as we go through this....

I follow that the locales are now compiled (they were not before because the path was never `import` 'ed anywhere before now).  So obviously `dist/assets` increases.. that is fine its just a compiled source available to request from the server. But the combination of that with the fact that we seem to pre-cache **all of the manifest assets** I thought was a recent addition to beta `precacheAndRoute(self.__WB_MANIFEST); ` . The POC had it I assume but should we be pre-caching it all on prod?  In any case if its the same as before, testing the branch to find the asset behavior locally was a challenge for me.

With my limited understanding it seems reasonable that we might want to be excluding any compiled assets that are not used... like locales.. or maybe even every other disabled workflow? I wonder what are some of the biggest fish in the 10MB? ~~I am really surprised 30KB turned into 1.56MB.~~

> But I also think we should ultimately try to move off of it if we at all can in the future. 

Its really hard for me to compare without a test branch to compare. I am guessing there is still value in letting i18n know whats happening but I am still trying to understand the lifecycle of the SW. Here I took an easy couple lines of code to make a low hanging fruit test branch... but I am glad we are finally testing it.

@kathyavini TLDR : I now see what is happening and I agree it does not work as intended but I don't think I will be able to resolve it myself before tomorrow. But it does work to fix the bug at least for now.

kathyavini:
> considering malayalam is ~3KB sitting on my computer uncompressed

Woah, how do you figure? I'm just inspecting my `dist/` using finder and I'm seeing the built languages range from 80KB to 178KB (for Malayalam) which is more or less consistent with that cache bump. Bump is still bigger but that could very well be library updates as you said.

Screenshot 2026-02-03 at 3 28 57 PM

(All the ones over 100KB are the Indic languages)

> In any case if its the same as before, testing the branch to find the asset behavior locally was a challenge for me.

Let's follow-up on the caching behaviour in tech daily; not quite sure what you mean by asset behaviour so let's take a look. It would be good to make sure we're on the same page. Yes, it is exactly the same behaviour as before in terms of the precache. It's just that VitePWA handled creating the service worker file under the hood for us, while now that we are using  `injectManifest` we need to [add that `precacheAndRoute` line manually](https://vite-pwa-org.netlify.app/guide/inject-manifest.html#service-worker-code). You can see the same lines by inspecting the built service worker on app -- we can look at that in tech daily too. But yeah, precaching all the built assets is exactly why the app already fundamentally worked offline, minus translations.

> I don't think I will be able to resolve it myself before tomorrow.

As I said above, I think this is fine for release! I actually want to try something with code splitting too which I think could make way more of a difference than the translations in terms of network data transfer... not for this upcoming release but for all subsequently; I'll try to open something and get it ready for tomorrow...!

Edit: ugh, on the topic of things that are worse than translations, the SurveyJS library chunk is 1.4MB by itself 🤮

kathyavini (approved):
@Duncan-Brain I think the 'ns' pre-load might be the ticket!... but not before 12pm... merging!!

Thank you!

PR #4015, opened 2026-02-04
LF-5116 Comment out TAPE for offline release

kathyavini:
**Description**

Comments out all TAPE-related code by removing the routes and the reducer.

**_Locally this cuts my precache down from 9MB to 5MB!!!_** 🤯 Yikes SurveyJS + the related code here is a monstrosity. This is something we should solve before releasing TAPE to production; absolutely can't justify doubling the cache for this one feature.

Jira link: https://lite-farm.atlassian.net/browse/LF-5116

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

Inspected routes and checked cache size with the precache-checking code:

```js
(async () => {
  const cacheNames = await caches.keys();
  const precacheName = cacheNames.find(name => name.startsWith('workbox-precache'));
  const cache = await caches.open(precacheName);
  const requests = await cache.keys();
  let totalSize = 0;

  for (const request of requests) {
    const response = await cache.match(request);
    if (response) {
      const blob = await response.clone().blob();
      totalSize += blob.size;
    }
  }

  // Convert to MB for easier reading
  console.log(`Total precache size: ${(totalSize / (1024 * 1024)).toFixed(2)} MB`);
})();
```

**Checklist:**

### DIFF
--- a/packages/webapp/src/containers/Insights/TapeSurvey/tapeSurveySlice.ts
+++ b/packages/webapp/src/containers/Insights/TapeSurvey/tapeSurveySlice.ts
@@ -61,8 +61,7 @@ export default tapeSurveySlice.reducer;
 
 // Selectors
 export const tapeSurveySelector = (state: any) =>
-  // LF-5116 Remove from Redux configuration when commenting out for next release
-  // Also re-evaluate placement in farmStateReducer when adding backend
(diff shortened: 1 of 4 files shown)


SayakaOno (approved):
Thank you!!
Just FYI, my precache dropped from 7.79MB to 6.06MB!

PR #4018, opened 2026-02-06
LF-5140: No animals are shown after creating them and going offline

SayakaOno:
**Description**

### Problem

After adding animals, the data wasn't being refetched because the component that calls `useGetAnimalsQuery` wasn't currently rendered. This caused issues when:

1. Add animals (online)
2. Go offline
3. Navigate to Animals page
4. Page shows no data because the query fails offline and the cache wasn't populated

Tag invalidation alone doesn't trigger refetches when no component is currently subscribed to the query.

### Solution

Proactively fetch all invalidated queries in `onQueryStarted` after the mutation succeeds. This populates the cache while still online, ensuring data is available when navigating to animal pages offline.

We have separate caches for queries with different params:
- `getCustomAnimalTypes()` and `getCustomAnimalTypes('?count=true')`
- `getDefaultAnimalTypes()` and `getDefaultAnimalTypes('?count=true')`

So we need to call each API twice to populate both cache entries.

### Additional Fix

In Beta, editing animal/batch → go offline → navigate to Animals page showed no KPIs. This issue is also addressed.

Reference:
- https://github.com/reduxjs/redux-toolkit/issues/1509
- https://redux-toolkit.js.org/rtk-query/api/createApi#onquerystarted

Jira link: https://lite-farm.atlassian.net/browse/LF-5140

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/store/api/apiSlice.ts
+++ b/packages/webapp/src/store/api/apiSlice.ts
@@ -168,7 +168,9 @@ export const api = createApi({
           const { data } = await queryFulfilled;
           await dispatch(api.endpoints.getAnimals.initiate());
           await dispatch(api.endpoints.getCustomAnimalTypes.initiate());
+          await dispatch(api.endpoints.getCustomAnimalTypes.initiate('?count=true'));
           await dispatch(api.endpoints.getDefaultAnimalTypes.initiate());
+          await dispatch(api.endpoints.getDefaultAnimalTypes.initiate('?count=true'));
         } catch (err) {
           // handled in component
         }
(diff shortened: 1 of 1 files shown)


Duncan-Brain (changes requested):
@SayakaOno it works for me!

Since the function is currently reused 8 times can we make it reusable?

For the future-- If we changed the UI/UX to skip the cute "done" infographic and just return to the core data (animals table) like we do for the other flows would that also work here? It has always sort of annoyed me to have to click "go back to animals" every time I make some haha

SayakaOno:
Thank you Duncan!

There are actually 4 distinct variations of the invalidate tags:

- ['Animals', 'CustomAnimalTypes', 'DefaultAnimalTypes']
- ['AnimalBatches', 'CustomAnimalTypes', 'DefaultAnimalTypes']
- ['Animals', 'DefaultAnimalTypes', 'CustomAnimalTypes', 'CustomAnimalBreeds']
- ['AnimalBatches', 'DefaultAnimalTypes', 'CustomAnimalTypes', 'CustomAnimalBreeds']

So, 4 functions x 2 times. I attempted to extract a shared function but found it hurt readability, so I left it as is. I'm happy to revisit if you have an idea in mind though!

On the UX point, I personally don't mind the "go back" click, but I know what you mean, and navigating directly to the animals page after submission would definitely work. I like this approach because it handles cases where the APIs get called from other pages in the future, and writing `onQueryStarted` for each endpoint feels like a reasonable tradeoff for that safety.

Duncan-Brain (approved):
Sorry I didn't notice that part!

OKay sounds good!

PR #4022, opened 2026-02-09
LF-5160: Fix replay queue order (Ticket title: Fix error message shown when deleting a task after toggling assignee offline)

SayakaOno:
**Description**

Replaced multiple background sync queues with a single queue to fix request replay ordering issues. Requests now replay in FIFO order regardless of type.

Changes:
- Consolidate three separate queues into one 'background-sync' queue
- Update `onSync` handler to send method instead of area
- Area resolution now happens in App's `resolveAreaFromUrl`

@kathyavini Please let me know if this approach is different from what you were expecting!

Jira link: https://lite-farm.atlassian.net/browse/LF-5160

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/hooks/useServiceWorkerListener/useServiceWorkerListener.ts
+++ b/packages/webapp/src/hooks/useServiceWorkerListener/useServiceWorkerListener.ts
@@ -39,11 +39,15 @@ type SyncConfig = {
 };
 
 /**
- * Resolve specific kinds of task patch operations from the URL
+ * Resolve specific kinds of task operations from the URL and HTTP method.
  */
-function resolveAreaFromUrl(area: string, url: string): SyncArea {
-  if (area !== 'tasks.update') {
-    return area as SyncArea;
+function resolveAreaFromUrl(method: string, url: string): SyncArea {
+  if (method === 'POST') {
+    return 'tasks.create';
+  }
+
+  if (method === 'DELETE') {
+    return 'tasks.delete';
   }
 
   if (url.includes('/task/complete/')) {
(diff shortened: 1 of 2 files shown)


packages/webapp/src/hooks/useServiceWorkerListener/useServiceWorkerListener.ts
```diff
@@ -168,16 +172,14 @@ export function useServiceWorkerListener() {
          * See: https://developer.chrome.com/docs/workbox/modules/workbox-background-sync
          */
         // We will also manually replay when the app comes back online.
-        switch (rawArea) {
-          case 'tasks.create':
+        switch (method) {
```
kathyavini, line 175: I see how it's cleaner to just use method for now when we are only queuing tasks, but when we add our next feature farm notes (also for offline syncing), I'm guessing it will be easiest to recycle `resolveAreaFromUrl()` down here too -- does that match what you're thinking?

```ts
const area = resolveAreaFromUrl(method, url);

switch (area) {
  case 'tasks.create':
    dispatch(enqueueErrorSnackbar(t('message:TASK.CREATE.SYNC.NETWORK_ERROR')));
    break;
  case 'tasks.delete':
    dispatch(enqueueErrorSnackbar(t('message:TASK.DELETE.SYNC.NETWORK_ERROR')));
    break;
```

And within `resolveAreaFromUrl()` I would think to just add a route check, does that seem fair?

```ts
if (method === 'DELETE' && url.includes('/task/')) {
```

kathyavini (approved):
@SayakaOno it's better than what I was thinking because I thought the `resolveAreaFromUrl` logic would have to go back into the service worker 😂 I didn't think to have the current handler just do the method too -- very nice!

PS: Oh I wish I had renamed this PR + ticket though, it's not exactly descriptive of the change 😜

packages/webapp/src/hooks/useServiceWorkerListener/useServiceWorkerListener.ts
SayakaOno: Ah true, I should've kept the switch as is. `resolveAreaFromUrl` was built for tasks, so I didn't think ahead about other endpoints. Your ideas look good to me!

PR #4029, opened 2026-02-09
Revert "LF-5116 Comment out TAPE for offline release"

kathyavini:
Reverts LiteFarmOrg/LiteFarm#4015

This is a temporary rollback for a call happening this afternoon. It can then be re-hidden


kathyavini:
Call is happening right now so I'll just merge 😝 We can restore to previous state tomorrow

SayakaOno (approved):

PR #4031, opened 2026-02-10
Revert "Revert "LF-5116 Comment out TAPE for offline release""

kathyavini:
Reverts LiteFarmOrg/LiteFarm#4029

Hides TAPE again!


Duncan-Brain (approved):
Should we just leave this in until last minute to avoid flipping it on again?

PR #4033, opened 2026-02-10
LF-5158 Prevent submitting unchanged 'no' task wage updates

kathyavini:
**Description**

This PR fixes two issues with the new EditTaskWageModal form:

1. The default value should be "no" when there is no task wage, not blank / empty string
2. The form should only be submittable when dirtied

To support (2) the hourly wage input is now also reset to default when "no" is selected -- this is to prevent the form from reading as dirtied (and submittable) just because a hourly wage was added, if the radio selection is then returned to "no"

To support _this_ I made the `hourlyWageActions` object an enum (and updating its naming to enum convention, i.e. `HourlyWageAction`), so I could use it in the new TS file as a type. I guess this is another use case for an enum -- when you want to use an existing JS object as a type in new files 🙂 

Jira link: https://lite-farm.atlassian.net/browse/LF-5158

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/Modals/EditTaskWageModal/index.tsx
+++ b/packages/webapp/src/components/Modals/EditTaskWageModal/index.tsx
@@ -57,11 +57,13 @@ export default function EditTaskWageModal({
     control,
     watch,
     handleSubmit,
-    formState: { errors, isValid },
+    formState: { errors, isValid, isDirty },
   } = useForm<EditTaskWageFormFields>({
     mode: 'onChange',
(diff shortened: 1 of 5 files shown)


Duncan-Brain (approved):
Works well for me! 

Is the enum a new case to use it or is it just the fastest way? When I look at it it seems like the union of strings would be our preferred here since no iteration of object?

PR #4035, opened 2026-02-11
POC - Lf 5132 maps do not work unless maps is pre visited while online db

Duncan-Brain:
**Description**

This is just a POC/WIP so showcase that if `/map` is pre-visited maps should work on refresh. For me on beta, if I refresh the map even after visinting `/map` it crashes 100% of the time -- with this branch 0%. So this focuses on the refresh and adds the Map container to Home to simulate starting a hidden map immediately.

This branch:
- imported once google map component reference for reuse (does not work as envisioned -- see memory leak)
- runs googleLoader hook once for reuse (may or may not clear that console error)
- adds map container to home screen so that map is "seen"
- conditionally loads all map related items I THINK THIS IS THE ONLY REASON THE BRANCH WORKS

Notes
- Likely all map assets that need to be looked at offline first need to be created as map objects for the cache (I think wild crops may just work fine though, which makes me question this)
- Adding Map container to Home significantly increases memory leaks -- I did not troubleshoot this first stab at it
- Console error I still see it on login -- the useEffect runs twice maybe so a simple check might work. Did not see it touring the rest of the app

Memory leak check:
Adding the map to home seems to double the leak:

Screenshot 2026-01-07 at 9 00 29 AM

Screenshot 2026-02-11 at 1 15 11 PM

Jira link:

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/App.jsx
+++ b/packages/webapp/src/App.jsx
@@ -13,7 +13,7 @@
  *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
  */
 
-import { useState } from 'react';
+import { useRef, useState } from 'react';
 import { matchPath, useLocation } from 'react-router-dom';
 import clsx from 'clsx';
 import { SnackbarProvider } from 'notistack';
@@ -23,10 +23,11 @@ import { NotistackSnackbar } from './containers/Snackbar/NotistackSnackbar';
 import styles from './styles.module.scss';
 import Routes from './routes';
 import { ANIMALS_URL, MAP_URL, SENSORS_URL } from './util/siteMapConstants';
-import { NavMenuControlsContext } from './contexts/appContext';
+import { GoogleMapInstanceContext, NavMenuControlsContext } from './contexts/appContext';
 import { useOfflineDetector } from './containers/hooks/useOfflineDetector/useOfflineDetector';
 import { useServiceWorkerListener } from './hooks/useServiceWorkerListener/useServiceWorkerListener';
 import { useGoogleMapsLoader } from './hooks/useGoogleMapsLoader';
+import GoogleMap from 'google-map-react';
 
 function App() {
   const location = useLocation();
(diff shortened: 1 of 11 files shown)


kathyavini:
@Duncan-Brain I just want to make sure I understand this comment because I'm also super interested in which of the changes is affecting the reload behaviour most (and ideally going with those that don't increase the memory leaks!)

> - conditionally loads all map related items I THINK THIS IS THE ONLY REASON THE BRANCH WORKS

You mean putting the render behind `isLoaded` in the JSX? So this pattern?

```jsx
{isLoaded && (
  <GoogleMap
    data-cy="google-map"
    style={{ flexGrow: 1 }}
    center={grid_points}
    defaultZoom={DEFAULT_ZOOM}
    yesIWantToUseGoogleMapApiInternals
 ```

What about the `<GoogleMapsInstanceContext />`? No effect from that, then?

PS: The reload behaviour looked really good to me; couldn't drop it at all! (Oh okay, it did finally drop after not using that tab for a bit)

Duncan-Brain:
@kathyavini 

The memory leak (unless their main source is solved) is unavoidable with the solution and our current setup. If you go back and forth to any map instance (even `/new_farm` with no drawn locations ) There will be a steady increase in nodes and listeners as you go back and forth. That is the page I am troubleshooting the leak on unless its truly unavoidable.

But... if you were able to successfully drop the map on just regular old page staleness... then for `/map` to be good we need Leaflet it sounds like! We might consider just making `/map` not navigable too?

So yah my hypothesis on why I was experiencing the issue 100% and you were getting it on 5th refresh was just that the library loaded at different speeds causing that error and the conditional rendering makes that behaviour predictable for us both, avoiding an unrecoverable state. 

There could be some effect from the googleMapsLoader context, not refreshing that data as much or something that just initiated the core problem.... but definitely no help yet from the GoogleMap instance.

PR #4036, opened 2026-02-11
LF-5163: Incorrect sync state and conflicting snackbars after offline changes when app is closed

SayakaOno:
**Description**

My assumption is that the network stack isn't fully ready when automatic background sync fires on mobile Chrome, causing requests to fail.
This PR disables automatic background sync by setting onSync to a no-op.

Retry won't happen automatically, so "Sorry, we are down momentarily. We will retry your task update automatically" snackbar message now feels inaccurate. We can confirm this with Loïc later.
The queue is no longer for background sync, so we should rename variables.

Also, I addressed the [comment](https://github.com/LiteFarmOrg/LiteFarm/pull/4022#discussion_r2784744595) about the `switch` for areas, and converted the `forEach` in the message event listener to `for` loop since `forEach` doesn't wait for promises ([doc](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach#description:~:text=forEach()%20expects%20a%20synchronous%20function%20%E2%80%94%20it%20does%20not%20wait%20for%20promises.%20Make%20sure%20you%20are%20aware%20of%20the%20implications%20while%20using%20promises%20(or%20async%20functions)%20as%20forEach%20callbacks.)). 

Jira link: https://lite-farm.atlassian.net/browse/LF-5163

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/hooks/useServiceWorkerListener/useServiceWorkerListener.ts
+++ b/packages/webapp/src/hooks/useServiceWorkerListener/useServiceWorkerListener.ts
@@ -167,10 +167,8 @@ export function useServiceWorkerListener() {
       } else if (type === 'SYNC_ITEM_FAILURE') {
         /*
          * This indicates a failure to reach the server (e.g. API down). It should be rare.
-         * The background-sync-api retries automatically with exponential backoff (approx. 5 min on Chrome).
-         * See: https://developer.chrome.com/docs/workbox/modules/workbox-background-sync
+         * We will manually replay when the app is re-rendered.
          */
-        // We will also manually replay when the app comes back online.
         switch (area) {
           case 'tasks.create':
             dispatch(enqueueErrorSnackbar(t('message:TASK.CREATE.SYNC.NETWORK_ERROR')));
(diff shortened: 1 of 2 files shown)


kathyavini (approved):
Okay I'm excited to try this on beta and on different devices!!🤞🤞🤞

> Retry won't happen automatically, so "Sorry, we are down momentarily. We will retry your task update automatically" snackbar message now feels inaccurate.

I think it's still quite okay because it _does_ still happen automatically when toggling offline/online, or if the component is re-mounted. It doesn't happen with exponential backoff but that was a confusing kind of replay anyway. I think I was actually thinking of the manual queue for that snackbar anyway, which is why it was vague and didn't say "in five minutes"!

>converted the forEach in the message event listener to for loop since forEach doesn't wait for promises ([doc](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach#description:~:text=forEach()%20expects%20a%20synchronous%20function%20%E2%80%94%20it%20does%20not%20wait%20for%20promises.%20Make%20sure%20you%20are%20aware%20of%20the%20implications%20while%20using%20promises%20(or%20async%20functions)%20as%20forEach%20callbacks.)).

Thank you!! ❤️

PR #4037, opened 2026-02-11
LF-5143: Error snackbar appears when offline on reload

SayakaOno:
**Description**

- Replay queue only when online to avoid triggering API calls that could show incorrect error snackbar.
- Adjust background-sync-related variable names.

Jira link: https://lite-farm.atlassian.net/browse/LF-5143

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/hooks/useServiceWorkerListener/useServiceWorkerListener.ts
+++ b/packages/webapp/src/hooks/useServiceWorkerListener/useServiceWorkerListener.ts
@@ -204,7 +204,10 @@ export function useServiceWorkerListener() {
       window.addEventListener('online', replayQueue);
 
       // Also replay queue on startup
-      replayQueue();
+      // Do not use Redux isOnline state here - it may not reflect the actual network status on initial render
+      if (window.navigator.onLine) {
+        replayQueue();
+      }
     }
 
     return () => {
--- a/packages/webapp/src/sw.js
+++ b/packages/webapp/src/sw.js
@@ -121,19 +121,19 @@ const queues = {
   [RETRY_QUEUE_NAME]: { queue: retryQueue },
 };
 
-// Push to our queue on failure
-const queueOnFailurePlugin = {
-  fetchDidFail: async ({ request }) => {
-    await retryQueue.pushRequest({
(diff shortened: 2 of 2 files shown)


Duncan-Brain:
@SayakaOno I am sorry I have not been able to reproduce locally the error to confirm the fix.

I could see it on beta, on chrome. And also not able to see it on firefox or safari.

@kathyavini would you be able to test this one since you were able to reproduce?

packages/webapp/src/hooks/useServiceWorkerListener/useServiceWorkerListener.ts
```diff
@@ -204,7 +204,10 @@ export function useServiceWorkerListener() {
       window.addEventListener('online', replayQueue);
 
       // Also replay queue on startup
-      replayQueue();
+      // Do not use Redux isOnline state here - it may not reflect the actual network status on initial render
+      if (window.navigator.onLine) {
```
kathyavini, line 208: After looking at my broken desktop Safari behaviour today, I re-read the second paragraph of [MDN on navigator.onLine](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/onLine) more carefully:

> Browsers and operating systems leverage different heuristics to determine whether the device is online... Therefore, this property is inherently unreliable, and you should not disable features based on the online status, only provide hints when the user may seem offline.

😬 Uh-oh! Well, the check does seem very reliable on mobile devices and on Chrome, our main targets! I'm curious to see if we get back any feedback from other systems on this.

kathyavini (approved):
Changes look good, and worked well on Chrome.

Couldn't test on Firefox due to "work offline" not doing what we need. And Safari... ugh. We can double-check those when it's on beta 😄

SayakaOno:
Thank you Duncan and Joyce!
> Therefore, this property is inherently unreliable, and you should not disable features based on the online status, only provide hints when the user may seem offline.

Thank you, I hadn't checked the docs 🙏 Let's find another solution if we get feedback!

PR #4038, opened 2026-02-12
LF-5144 Weather widget disappears periodically when offline

kathyavini:
**Description**

Not the highest priority, but is also an easy fix!

RTK Query does not purge existing cache on a fetch error, such as the network error that you see when offline. However, the Weather _Widget_ had been conditionally rendered on `!error`, and thus it vanished even though the RTK Query data was intact.

Since the RTK query api is now set as follows to support offline (as of #4007):

```ts
// apiSlice.ts
keepUnusedDataFor: 60 * 60 * 24 * 7, // 7 days
refetchOnMountOrArgChange: 60,
```

refetch on any component (and so widget disappearance from home) happens the first time you revisit it after 60 seconds.

### Note about cache expiry
The weather endpoint has its own 

```ts
keepUnusedDataFor: 7200, // Cache data for 2 hours (7200 seconds)
```

which I thought at first to remove, but if you are offline for a long time, I don't think you want to be seeing stale weather data... in that case I think it's preferable for the data + widget to disappear, and I left that in place.

Jira link: https://lite-farm.atlassian.net/browse/LF-5144

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

Had to `pnpm build` and `pnpm preview` to be able to get the component to remount while offline, a bit annoying 🫤 Also set the `7200` above down to `30` to double-check that the widget will disappear after that time (it does)

**Checklist:**

### DIFF
--- a/packages/webapp/src/containers/WeatherBoard/index.jsx
+++ b/packages/webapp/src/containers/WeatherBoard/index.jsx
@@ -11,11 +11,7 @@ export default function WeatherBoard() {
     units: { measurement },
   } = useSelector(userFarmSelector);
 
-  const {
-    data,
-    error,
-    isSuccess: loaded,
-  } = useGetWeatherQuery({
+  const { data } = useGetWeatherQuery({
     measurementSystem: measurement,
   });
   const language_preference = getLanguageFromLocalStorage();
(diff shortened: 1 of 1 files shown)


SayakaOno (approved):
Looks good!

PR #4040, opened 2026-02-17
Lf 5132 maps do not work unless maps is pre visited while online 2

Duncan-Brain:
**Description**

This PR conditionally renders location picker and map when the google maps loader has been successfully completed. 

It also reuses the google loader call at the top level. This avoids repeat calls triggering library fetch failure.

Renames the pre-existing context.

NOTE
- Does not solve the problem of maps **locations** breaking if not visited, this solves the refresh

Jira link: [LF-5167](https://lite-farm.atlassian.net/browse/LF-5167)

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain):
    `pnpm build + preview`
     If pre-visited the locations and map will remain until google maps cache becomes stale
     Stale state can be simulated by ignoring the files and refreshing (devtools > sources > googlemaps.api.com > right click google api and ignore folder)

**Checklist:**

[LF-5167]: https://lite-farm.atlassian.net/browse/LF-5167?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/webapp/src/hooks/useGoogleMapsLoader.ts
+++ b/packages/webapp/src/hooks/useGoogleMapsLoader.ts
@@ -18,6 +18,7 @@ import { useEffect, useState } from 'react';
 import { setOptions, importLibrary } from '@googlemaps/js-api-loader';
 
 const GOOGLE_MAPS_API_KEY = import.meta.env.VITE_GOOGLE_MAPS_API_KEY;
+export const ALL_LIBRARIES = ['maps', 'places', 'drawing', 'geometry', 'geocoding', 'marker'];
 
 let optionsInitialized = false;
 if (!optionsInitialized) {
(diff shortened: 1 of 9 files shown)


packages/webapp/src/hooks/useGoogleMapsLoader.ts
```diff
@@ -28,7 +29,7 @@ if (!optionsInitialized) {
   optionsInitialized = true;
 }
 
-export const useGoogleMapsLoader = (libraries = ['places', 'drawing', 'geometry']) => {
+export const useGoogleMapsLoader = (libraries = ALL_LIBRARIES) => {
```
SayakaOno, line 32: This hook is now intended to be called once at the app root, correct? To make that clearer, I think it'd be nice to add a comment and hardcode `ALL_LIBRARIES` instead of accepting it as a parameter.

SayakaOno (approved):
Looks good to me!

packages/webapp/src/hooks/useGoogleMapsLoader.ts
Duncan-Brain: I did the bare minimum. This PR does not replace all instances. Market directory info and Add Farm flows are not offline.

SayakaOno: Ah I though you'd replaced everything

PR #4044, opened 2026-02-18
LF-5165 Indicate offline readiness state to user and help recover it

kathyavini:
**Description**

This PR aims to communicate offline readiness to the user and give a UI prompt to recover from states that would otherwise lead to paused / stuck precache.

The readiness of the application is evaluated according to the **state of the precache** alongside the state of the service worker. If the cache is incomplete, the application is considered "not offline ready" and the user is given a warning. When they return online, they will be prompted to "refresh" the application, which is a reload to resume cache population.

Additionally, if the cache is incomplete (or missing) but the service worker is installed and active, this is considered the "stuck cache" state, because this combination is unrecoverable on any amount of reload. In this case, the "refresh" button will unregister the service worker + reload. (Note: I had originally also cleared the cache, but I don't think it's necessary; I have seen a stuck cache pick up where it left off just via the service worker unregistration). I'll describe below how I tested each case in the testing section.

### Feature components
- new Redux slice to store the offline-readiness of the application
- a hook `useOfflineReadiness.ts` that interacts with the Redux state
- new UI elements within the offline indicator to let the user know when a) the application is not yet offline ready, and to b) prompt refresh of the application when it has gotten stuck, only shown when the user is returning from state (a)
  - as discussed with Loïc on Wednesday, the user sees only one prompt, even though under the hood we sometimes reload and sometimes reset (= unregister the service worker and reload) depending on whether we are evaluating the precache as "paused" or "stuck"
- new functionality in the service worker that checks the state of the cache against the precache manifest to see if it's completed or not

### Notes
- I ended up not using the VitePWA [onOfflineReady](https://vite-pwa-org.netlify.app/guide/prompt-for-update.html#importing-virtual-modules) event, which I showed in tech daily last week -- the service worker-emitted `oncontrollerchange` event had the same timing, and this way the logic could live only in the hook, with no need to register the SW manually within `main.jsx`
- the cache is checked ~in two places: the service worker and the hook.~ Update: now only checked in the service worker! Only the service worker is capable of declaring the cache complete (because only it has the manifest)
- as usual with any frontend update, this isn't going to solve any problems that would have occurred on the release of _this_ feature / code change
  - In other words, it's very likely our bashers tomorrow and users after release next week will still be able to go offline before the new service worker has completed precache install
  - Only in the cases where the user then crashes, reloads, and otherwise ends up completely dropping the cache will the new code come into play (since the old cached frontend is then gone) :) So it will still be useful for this release, but more in recovering from crashes rather than preventing them.

### Caveats / Things I would fix if I had more time
- this will immediately change the UX of `localhost:3000` (`pnpm dev`) -- you will see the "offline not ready" and the "refresh" (which does nothing) instead of "you are offline" because localhost _can_ support service workers, but of course it does not have a service worker or a cache with our current config
  - I don't love the change, but it was a little hard to fix. In some sense it is actually the correct message, since `localhost:3000` _will_ crash if you try to use it offline, e.g. on page change or reload, but if it's bothersome maybe we can find a workaround or make a specific exception with a `.env` variable
  - if you test wih a device over the local network, that is a service worker unsupported environment and you should see the original offline message
- I think maybe the hook should only be called once? At some point in my testing, it seemed registering the listeners and `useEffects()` multiple times was causing bugginess; not sure if it's still the case as I removed the extra hook call (it had been in `<Navigation />`; now `<Navigation />` uses the Redux selector directly).
 - Not being callable multiple times seems like an anti-pattern for a custom hook. Maybe separating the setup part of the hooks (to run at top level) and the state derivation part of the hook (for the values the offline indicator component needs) would have been better...? For now it just lives in the one hook though, and just one component (the `<OfflineIndicator />`) calls it.

Jira link: https://lite-farm.atlassian.net/browse/LF-5165

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

The UI elements I have looked at on both desktop and mobile, but the cache behaviour and logic I have only been able to test on desktop (due to local network being an unsecured context that doesn't allow service workers).

Here's how I tested each type of cache situation (all on `pnpm build` + `pnpm preview`) _on Chrome only_ (Firefox and Safari don't have the kind of dev tools that would make this posible):

1. **Paused cache**: Clear your site data to install the service worker freshly. While the cache is prepopulating, go offline. You should see the offline not ready warning. When reconnecting, your precache will be paused until you reload. The "refresh" prompt should be seen and will trigger a reload
2. **Dropped cache**: Let your service worker install to a ready state. Manually drop the `workbox-precache` from within storage by right click --> "delete"
3. **Stalled cache** (this one can be a little timing-dependent): Clear your site data and reload to begin normal precache population. Right-click on the `workbox-precache` --> "delete" before the cache is completed. I _usually_ see the cache immediately recreated and populated part way, but it never finishes, and doesn't resume on reload.

Please feel free to test with any other combinations you can think of! Please try to break it!! 🙏

**Checklist:**

_Translation keys already merged in previous PR_


packages/webapp/src/registerServiceWorker.js
```diff
@@ -1,129 +0,0 @@
-/*
```
kathyavini, line 1: Sorry forgot to mention that this file is a complete holdover from a very long time ago (before we used VitePWA) and is not used at all. Finally deleted it because it was showing up in code search results.

packages/webapp/src/hooks/useOfflineReadiness/useOfflineReadiness.ts
```diff
@@ -0,0 +1,188 @@
...
+        }
+      } catch (e) {
+        // Service worker not available or failed
+      }
+    };
+
+    // Revalidate cache on controller change (new SW taking control)
+    const handleControllerChange = () => {
+      validateAndUpdateState();
+    };
```
SayakaOno: Can we use `validateAndUpdateState` directly in the listener?

packages/webapp/src/hooks/useOfflineReadiness/useOfflineReadiness.ts
```diff
@@ -0,0 +1,188 @@
...
+  const resetApplication = async () => {
+    console.log('resetApplication: unregistering SWs...');
+    if (navigator.serviceWorker) {
+      const registrations = await navigator.serviceWorker.getRegistrations();
+      for (const registration of registrations) {
+        await registration.unregister();
+      }
+    }
+    window.location.reload();
+  };
```
SayakaOno: I think it would be nice to place these outside the hook.
I feel like the less there's cache "paused" vs "stuck" logic, the better...
We could probably return eg., `restoreCache: recoveryMode ? resetApplication : reloadApp`.

packages/webapp/src/hooks/useOfflineReadiness/useOfflineReadiness.ts
```diff
@@ -0,0 +1,188 @@
...
+    } else if (controlled) {
+      // SW is active (install phase complete), but cache is incomplete.
+      // This is unrecoverable regardless of how many entries are cached
+      dispatch(setRecoveryMode(true));
+      dispatch(setWentOfflineDuringSetup(true));
+    } else if (validation.totalCached && validation.totalCached > 0) {
+      // No controller yet -- SW is still installing and actively filling the cache.
+      // Not broken, just in progress
+      dispatch(setRecoveryMode(false));
+    }
```
SayakaOno, line 92: `wentOfflineDuringSetup` and `recoveryMode` mostly depending on `validation` makes me wonder if they could be selectors not state... probably `wentOfflineDuringSetup` can't since it's set in several places?

packages/webapp/src/hooks/useOfflineReadiness/useOfflineReadiness.ts
```diff
@@ -0,0 +1,188 @@
...
+          if (totalCached > 0) {
+            // Cache is present but SW hasn't claimed the page -- treat as in-progress
+            return { isComplete: false, totalCached };
+          }
+        }
+      } catch {
+        // caches API not available or failed
+      }
+    }
+    return { isComplete: false, error: 'No service worker controller' };
```
SayakaOno: I struggled to understand how these different returns affect the final outcome 😢

SayakaOno (commented):
Sorry, I haven't fully understood how the hook works, so I'll continue reviewing it tonight.
The code is very readable, I just needed some more time to connect the dots 🙏

packages/webapp/src/hooks/useOfflineReadiness/useOfflineReadiness.ts
kathyavini: You're totally right, these branches are not needed at all anymore!! 😭

This was based upon recovery mode initially using the cache empty error... and then this code was from when I tried to handle stuck cache as well. But at the end of yesterday I moved to simply cache incomplete + controlled = recovery mode and I think this is all junk now!! I don't think anything else is needed at all -- neither the `windows.check` nor the validation errors!! 🙇

Thank you so much; it's gone!!!

packages/webapp/src/hooks/useOfflineReadiness/useOfflineReadiness.ts
kathyavini: I think `recoveryMode` can be a selector, but then `controlled` would have to be in the state too, right? Since that's the part that's not in `validation`? So maybe it _should_ be in `validation`??!

I'm not sure, but I am going to try that! It will simplify one of the many code branches that felt very messy to me 🙏

Edit: the more I look at it, the more I feel can be stripped away. I don't think it needs to even be a selector... which means that I don't think controlled needed to be in state, and the cache state could just be a single boolean... but I think at this stage I might need to just leave it be for now!

packages/webapp/src/hooks/useOfflineReadiness/useOfflineReadiness.ts
kathyavini: I really like `restoreCache: recoveryMode ? resetApplication : reloadApp`, thank you!!!

By outside the hook, I'm assuming you mean outside of the hook body (but in the same file)? I _loved_ how you separated the `offlineEventLogger` and `useOfflineActivityLogger` in your PR and really wish I had done something like that, but at the moment it's so mixed up! I tried a utils file but it didn't much help 😬

But definitely they can live outside of the hook call!

kathyavini (commented):
@SayakaOno thank you soooo, sooo much for taking the time to look carefully at this 🙇 And I'm so sorry you had to work late!! 🙇🙇🙇

Your comments were SO helpful, especially around the cache check logical branches and the Redux state -- those were accumulated from several days of trying to cover different cache states and an older concept of "recovery mode"; honestly I was SO confused by the end in terms of what was needed and not, because every time I would try to adjust an effect or remove a condition the functionality would break!

It got extremely over-complicated -- in the end I think 100% of what is needed was just the service worker giving the cache complete message, and knowing if the service worker is controlling the page. Thanks to your comments I was able to remove soooo much nonsense logic 🙇  And I think it's behaving _exactly_ as before. If you read it now, I think it should probably make a lot more sense... 🙏

> The code is very readable

It was only an illusion!! I myself had lost track over the last few days which booleans were important and I was getting afraid to refactor due to the number of times I had broken it 😓

❤️❤️❤️ THANK YOU!!! ❤️❤️❤️

Edit: Ah no, celebrated too fast. I HAVE broken one thing. This has been the general pattern. Now the reset is triggering too aggressively, including on first install 🫤 Will be looking at this.

Edit 2: No, I'm totally nuts, it's just the reload action always bumps the service worker # in the application tab, I just had missed that. Everything is working, and I think it's time for me to step away from the computer!

SayakaOno (approved):
Wow, the logic I was going to dig into is all gone! It's very intuitive now :)
It's working well; thank you for handling this challenging task! ❤️

PR #4045, opened 2026-02-19
LF-5145: Record offline logs

SayakaOno:
**Description**

This PR adds an endpoint and frontend logic to capture and send user activity logs collected while offline.

### Backend
- Add `POST /offline_event_log` endpoint
  - Auth middleware: Skips the common `checkJWT` and instead checks the token if supplied. The result is reflected in the `log_status` property, indicating how trustworthy the log is
  - Validation middleware: Validates user, farm, userFarm, and the shape of the logs
  - See the Jira [comment](https://lite-farm.atlassian.net/browse/LF-5145?focusedCommentId=23601) for log table properties
- Install `@types/ua-parser-js` to use `ua-parser-js` in TypeScript files

### Frontend
- API logs: In `useServiceWorkerListener`, store a log to a ref each time the app receives an API success message from the service worker. Send logs to the API when:
  - Logs accumulate more than 10
  - 2 seconds have passed since the first message
  - The user leaves the app
  - *(This approach avoids interrupting in-progress sync API requests)*
- User behaviour logs: `useOfflineActivityLogger` manages the full lifecycle from `session_start` to submission. User activities are stored in localStorage and sent via the logging API after 3 seconds the user comes back online

Note:
- I adopted a `visibilitychange` listener ([doc](https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilitychange_event)) to send log API requests, but the `useEffect` cleanup functions might be the ones actually taking effect. Will continue to investigate.
- Some user-agent fields are required in the log schema, so requests from non-browser clients (e.g. curl, Postman) will naturally fail validation. Therefore, I didn't add the origin property we discussed in the tech-daily on Feb 18.
- `api/package-lock.json` was regenerated as a side effect of installing `@types/ua-parser-js`. The previous lock file was generated with `legacy-peer-deps=true` in a local `.npmrc`, which caused peer dependencies of `express-oas-generator` (mongoose, mongoose-to-swagger) to be omitted. Regenerating without that setting restored them.  (If you remember, Duncan helped me with this last time!)

Additional note:
- When the browser is closed while offline, any pending API requests (queued in the visibilitychange listener or useEffect cleanup) appear to be processed the next time the app is opened. These requests won't be visible in the network tab during the original session.
- More tests are needed for edge cases where you close the browser before logging APIs are processed and open the app next time offline etc.

Jira link: https://lite-farm.atlassian.net/browse/LF-5145

**Type of change**

**How Has This Been Tested?**

Run `pnpm build` and `pnpm preview`, use the app offline, and check the database.

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/api/db/migration/20260209223930_add_offline_event_log.js
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+export const up = async function (knex) {
+  return knex.schema.createTable('offline_event_log', function (table) {
+    table.increments('id').primary();
+    table.uuid('session_id').notNullable();
+    table.string('event_name').notNullable();
+    table.string('status_code');
```
kathyavini: Should it be an integer? It's a number in the type and the model.

packages/api/src/controllers/offlineEventLogController.ts
```diff
@@ -0,0 +1,68 @@
...
+  addOfflineEventLog() {
+    return async (
+      req: LiteFarmRequest<unknown, unknown, unknown, OfflineEventLogReqBody>,
+      res: Response,
+    ) => {
+      try {
+        const { logs, went_online_at, app_version, network, session_id } = req.body;
+
+        const wentOnlineAt = went_online_at && new Date(went_online_at).toISOString();
+        const ua = parser(req.headers['user-agent']);
```
kathyavini, line 32: Not at all blocking, but I still worry we might run into space concerns repeating the header information in each record. Since only one header is sent per call (either queue flush or non-queue call from the webapp), it seems that only one record per call should have all the header information? I don't mind just `null`ing the rest, for example like:

```ts
const headerData = {
  country_id: res.locals.country_id,
  network,
  browser: ua.browser.name,
  // ...
};

const records = logs.map(({ event_name, event_at, status_code, url }, index) => {
  const isFirst = index === 0;

  return {
    session_id,
    event_name,
    // ...
    ...(isFirst ? headerData : {}),
    app_version,
    log_status: res.locals.log_status,
  };
});
```

Personally I like how that looks in the db because you can see at a glance how many events were sent with each request. But I think that's a bit subjective, and I'm also fine with waiting to see how much the db actually grows over time first!

packages/webapp/src/util/offlineEventLogger.ts
```diff
@@ -0,0 +1,101 @@
...
+}
+
+export const recordOfflineEvent = async ({ auth, farmId, logs }: OfflineEventPayload) => {
+  const { id_token, session_id, went_online_at = Date.now() } = getOfflineSession();
+
+  return fetch(offlineEventLogUrl, {
+    method: 'POST',
+    headers: {
+      'Content-Type': 'application/json',
+      Authorization: auth || id_token,
```
kathyavini, line 82: I see how it's working but you might have to talk me through this again in tech daily next week. 

This line is the reason that the queue call uses the original auth (since it's taken from payload) but the non-queue call uses the `id_token` at the time of send? So is it that the `id_token` stored at session start is never used in either flow? I think you might have mentioned that...

Testing with token expiry, it was interesting to see how how the queue call shows the 401 + `expired` log state (I love that!! ❤️ ) and the non-queue call at 3s shows `anonymous` because of the logout. I do wonder if we might like to keep the session start one so those can keep `expired` for those records too... but it's not nearly as important as capturing the queue expiry so merely a point of discussion!

kathyavini (dismissed):
Honestly I have no idea how you did this a) in that time frame, and b) so cleanly. It was _extremely_ robust in my testing and I could not spot anything that looked like a bug.

Thanks to your walkthrough yesterday, fundamentally it makes a lot of sense. I like the `log_status` addition _quite_ a lot, too! It would be fun to talk a little bit more about the auth states later, but it all definitely looks to be safe to go up to beta for me.

(There is a minor datatype thing but I don't think it's unsafe...!)

packages/api/db/migration/20260209223930_add_offline_event_log.js
SayakaOno: Oh, thank you!! I initially called it `status`, thinking that we could store "network error" as well. I'll fix it!

packages/api/src/controllers/offlineEventLogController.ts
SayakaOno: Okay, will update it!
It doesn't make sense to save other common properties, so I clean them up too.

packages/webapp/src/util/offlineEventLogger.ts
SayakaOno: I intended to use the `id_token` stored at session start for the non-queue call to send a token, so it's a bug 😢 
I don't think I get `id_token` directly... do I 😅 ?? I might be calling `startOfflineSession` as unintended... I'll take a look if I have time!

SayakaOno (commented):
Thank you so much for your review! I spent a lot of time going back and forth on this, so your comment made my week!! ❤️ I also really appreciated your feedback yesterday, it made the implementation much better than what I initially had.

I've addressed quick ones; I appreciate another eye, thank you! 🙏 
Feel free to update anything before merging!

kathyavini (approved):
Thank you so much for working on this, so late too!! 🙇  Looks great!

I'm actually genuinely so excited about this feature, it's so cool ❤️ It will be fun to inspect the logs next week and see what we find!

Have a great weekend!!!

packages/webapp/src/util/offlineEventLogger.ts
kathyavini: I'll look again too! I just now noticed that if I happen to have a second Chrome window open with a `localhost:4173` tab, interacting with one window will also send logs from the second! They will be posted afterwards and be `anonymous` log status. Interesting!

I don't think that's what I saw before -- I remember a matching session ID -- but I'll check again. Very fun stuff to inspect these logs 😁

PR #4049, opened 2026-02-20
LF-5145(2): Record offline logs

SayakaOno:
**Description**

- Save `user_id` and `role_id` in `offline_event_log` table
- Send 'reset_app' and 'reload_app' events from the client
- Prevent db operations from failing when the request contains duplicated logs (https://knexjs.org/guide/query-builder.html#onconflict:~:text=20)%22-,onConflict,-MS)
  
    Screenshot 2026-02-23 at 12 09 37 PM

  - This happens when the user opens the app in multiple tabs.
  - In such cases, we'll have multiple `session_start`. I'd rather not clean it so that we'll know the session includes events from different tabs.

Jira link: https://lite-farm.atlassian.net/browse/LF-5145

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/util/offlineEventLogger.ts
+++ b/packages/webapp/src/util/offlineEventLogger.ts
@@ -79,7 +79,7 @@ export const recordOfflineEvent = async ({ auth, farmId, logs }: OfflineEventPay
     method: 'POST',
     headers: {
       'Content-Type': 'application/json',
-      Authorization: auth || id_token,
+      ...(auth || id_token ? { Authorization: auth || `Bearer ${id_token}` } : {}),
     },
     // https://developer.mozilla.org/en-US/docs/Web/API/Request/keepalive
     keepalive: true, // Attempt to send even when the page is unloading
@@ -101,11 +101,13 @@ export const recordOfflineEvent = async ({ auth, farmId, logs }: OfflineEventPay
 };
 
 export const postEventLogs = async ({ farmId, logs }: OfflineEventPayload) => {
+  const idToken = localStorage.getItem('id_token');
+
   return fetch(offlineEventLogUrl, {
     method: 'POST',
     headers: {
       'Content-Type': 'application/json',
-      Authorization: 'Bearer ' + localStorage.getItem('id_token'),
+      ...(idToken ? { Authorization: `Bearer ${idToken}` } : {}),
     },
     // https://developer.mozilla.org/en-US/docs/Web/API/Request/keepalive
     keepalive: true, // Attempt to send even when the page is unloading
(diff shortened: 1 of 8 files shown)


kathyavini:
> In case we want to see reload/reset app event logs in the database!

Oh nice! 😍 Yes it would be fantastic to see these, thank you.

SayakaOno:
Woops! I forgot to manage `session_start`... converting to draft!

packages/webapp/src/util/offlineEventLogger.ts
```diff
@@ -99,3 +99,26 @@ export const recordOfflineEvent = async ({ auth, farmId, logs }: OfflineEventPay
     }),
   }).catch(() => {}); // Fire and forget
 };
+
+export const postEventLogs = async ({ farmId, logs }: OfflineEventPayload) => {
+  return fetch(offlineEventLogUrl, {
+    method: 'POST',
+    headers: {
+      'Content-Type': 'application/json',
+      Authorization: 'Bearer ' + localStorage.getItem('id_token'),
```
kathyavini: Okay, I was having some trouble posting logs and I realized it's because the action I use to clear the service worker also clears local storage 🙂 And I think this new function doesn't work in that case.

It looks like a result of how fetch coerces the headers into strings, but this line is creating the auth header `Bearer null`, which after the middleware

```ts
 const token = req.headers.authorization?.split(' ')[1];
```

creates the token `"null"` (the string) which will throw 'jwt malformed' and hit the catch block.

I realized when looking into this, the fact the other function works when logged out is actually maybe a bit of a fluke/accident? 😅

The header formed if there is no `id_token` (which results in undefined `auth`) is just the string 
`'null'` after

```ts
Authorization: auth || id_token
```

But since then there is is nothing after the split in the middleware (there is no space!), the token is undefined and the middleware doesn't error. I totally missed it in first pass, but looking at it now, I think the original function maybe should have been

```ts
Authorization: auth || `Bearer ${id_token ?? ''}`,
```
and this one

```ts
const token = localStorage.getItem('id_token');
//...
    Authorization: `Bearer ${token ?? ''}`,
```

does that seem right?

That seems to fix the logged out case, without breaking the logged in case, which was already good to go!

kathyavini (commented):
Looks great! Thank you for diagnosing the source of duplicate entries. Very interesting how different tabs =/= different open windows and I wonder how much we'll see that outside of testing.

I found one small bug regarding the logged out case that was keeping me from seeing the offline readiness logs; I tried the fix below and got all my reload and reset events after that🙏 Thank you for thinking to track those -- though part of me **doesn't want to know** if that painful-to-develop feature doesn't see any use at all 🤣🤣

SayakaOno:
Thank you so much for spotting those!! 🙏
I made changes not to send the `Authorization` header without a token and cleaned up one `recordOfflineEvent` call.
The issues should be resolved, but it would be great if you could double-check. Thank you!

kathyavini (approved):
> not to send the Authorization header without a token 

Ah okay, that's even better 😂 I was thinking the middleware required it, but it's because I missed the `?.` on `headers.authorization`!

(Does feel a lot cleaner and more to spec this way 😁)

PR #4051, opened 2026-02-20
POC - LF- 5100 isLoading and isFetching is not farm independent 

Duncan-Brain:
**Description**

This is the discussed companion PR for https://github.com/LiteFarmOrg/LiteFarm/pull/3988

This PR create a partition on farm_id in our rtk data store.

This took some effort to make the wrappers pass types through so before we dismiss it based on complexity please think about considering the following:
 
 - wrapper complexity to retain typing can be improved in RTK v2.0 there are much better helper types available
 - back and forth switching between farms should be faster
 - offline our rtk data is stored allowing us to switch farms offline (tasks saga is currently reset on switch farm... but once it moves to rtk it is baked in)
 
 I am now personally on the fence about which solution is best for us -- but as mentioned originally the other solution is much simpler.
 
 Notes
- I do not know why there is not an easier way to do this -- but `getState` is just not available prior to query param serializing
- It was asked about keeping the tag simple, but then adding an animal on one farm clears state for both farms which did not make sense to me. And cases like new `onQueryStarted` functions you would need the `farm_id` of the other farm to refresh that state.
 

Jira link: [LF-5100](https://lite-farm.atlassian.net/browse/LF-5100)

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain):
In the redux store you should now see queries with `farm_id` and original paramaters. Seeing `undefined` as a query arg will be library const endpoint. Anything else is likely a bug.

**Checklist:**

[LF-5100]: https://lite-farm.atlassian.net/browse/LF-5100?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/webapp/src/containers/AddFarm/saga.js
```diff
@@ -61,9 +59,6 @@ export function* postFarmSaga({ payload: { showFarmNameCharacterLimitExceededErr
     };
     yield call(axios.patch, patchStepUrl(farm_id, user_id), step, getHeader(user_id, farm_id));
 
-    // Clear old farm RTK Query data
-    yield put(invalidateTags([...FarmTags, ...FarmLibraryTags]));
```
Duncan-Brain, line 65: No longer need to invalidate tags on switch farm. Store will clear on logout only.

packages/webapp/src/containers/AddSensors/index.tsx
```diff
@@ -45,7 +45,11 @@ const AddSensor = ({ isCompactSideMenu }: AddSensorProps) => {
     const result = await addFarmAddon(values[PARTNER]);
 
     if ('error' in result) {
-      const isInvalidId = 'data' in result.error && result.error.data === 'Organisation not found';
+      const isInvalidId =
```
Duncan-Brain, line 48: error is now unknown requiring stricter narrowing

packages/webapp/src/containers/AddSensors/Partners.tsx
```diff
@@ -22,9 +22,9 @@ const Partners = ({
 }: {
   setIsEditing: React.Dispatch<React.SetStateAction<boolean>>;
 }) => {
-  const { data: esciData = [], isLoading: esciLoading } = useGetFarmAddonQuery(
-    `?addon_partner_id=${PARTNERS.ESCI.id}`,
-  );
+  const { data: esciData = [], isLoading: esciLoading } = useGetFarmAddonQuery({
+    params: `?addon_partner_id=${PARTNERS.ESCI.id}`,
```
Duncan-Brain, line 26: `params` arguments standardized for use with `useQueries` across GET endpoints.

packages/webapp/src/containers/saga.js
```diff
@@ -603,7 +602,7 @@ export function* fetchAllSaga() {
   const { has_consent, user_id, farm_id } = yield select(userFarmSelector);
   if (!has_consent) return history.push('/consent');
 
-  yield put(api.endpoints.getSensors.initiate());
+  yield put(api.endpoints.getSensors.initiate({ farm_id, payload: undefined }));
```
Duncan-Brain, line 605: Cant deny this is ugly, but the goal is to remove sagas if I recall. So using `initiate()` should be rare.

packages/webapp/src/hooks/api/useQueries.js
```diff
@@ -14,7 +14,9 @@
  */
 
 const useQueries = (queries) => {
-  const result = queries.map((query) => query.hook(query.params));
+  const result = queries.map((query) =>
+    query.hook(query.params ? { params: query.params } : undefined),
```
Duncan-Brain, line 18: Pass params only if exists.

packages/webapp/src/store/api/selectors/sensorSelectors.ts
```diff
@@ -15,27 +15,42 @@
 
 import { createSelector } from '@reduxjs/toolkit';
 import { api } from '../apiSlice';
+import { loginSelector } from '../../../containers/userFarmSlice';
 
 // For map use / to use with old selectors only (e.g. pointSelector). In components, use the query hook and selectFromWithin
 
+const selectSensorsData = createSelector(
```
Duncan-Brain, line 22: Had to separate out to get farm_id in `select()` call. This entire file will be gone in the migration of locations to rtk PR.

packages/webapp/src/store/api/types/index.ts
```diff
@@ -31,6 +31,8 @@ export interface BasicEnum {
   id: number;
   key: string;
 }
+
+export type WithFarmIdPayload<T = any> = { farm_id: string; payload: T };
```
Duncan-Brain, line 35: Originally had intersection `export type WithFarmId<T> = T & { farm_id: string };`

But that creates inference problems when T is Array-like the properties can not match up right on inference.

packages/webapp/src/store/api/util.ts
```diff
@@ -0,0 +1,329 @@
...
+  QueryStatus,
+  skipToken,
+  SubscriptionOptions,
+} from '@reduxjs/toolkit/query/react';
+import { useSelector } from 'react-redux';
+import { loginSelector } from '../../containers/userFarmSlice';
+import { WithFarmIdPayload } from './types';
+import { FarmLibraryTag, FarmTag } from './apiTags';
+
+type RefetchQueryResult<T, A> = Promise<RawQueryResult<T, A>> & {
```
Duncan-Brain, line 27: With rtk v2.0 I expect not to have to mock these types and so the wrapper should become simpler.

Hopefully inference will also be better and passing the endpoint typings everywhere will not be necessary as it is pretty bulky.

packages/webapp/package.json
```diff
@@ -43,7 +43,7 @@
     "@mui/x-date-pickers": "^8.0.0",
     "@react-oauth/google": "^0.7.0",
     "@reactour/tour": "^3.1.6",
-    "@reduxjs/toolkit": "^1.9.1",
+    "@reduxjs/toolkit": "1.9.7",
```
Duncan-Brain, line 46: This improved typings a little bit without having to upgrade to v2.0. 

Basically by reusing tome types in the previous version it could break parent types that were otherwise untouched. I never realized typescript could be so fragile.

PR #4068, opened 2026-03-02
LF-5182: TAPE Plan Phase 1 - Frontend Changes

kathyavini:
**Description**

- Update visibility
    - Hide routes/tape menu from workers or farms not in Australia
- Fetch survey from DigitalOcean
   - Handle API error
- Update survey completion function
    - Call the API to submit data instead of saving to Redux store
    - Handle API error
- Update tapeSurveySlice
    - Remove `isCompleted` state and calculate it using saved survey (if exists, isCompleted is true)
    - surveyData => surveyDataInProgress
    - tapeSurveyStatusSelector now returns `inProgress`
- Adjust built-in warning notification position
- Comment out Step 2
- Delete the Update button and logic (could be reverted: [de7b624](https://github.com/LiteFarmOrg/LiteFarm/pull/4068/commits/de7b6245a32ed222b736e69e410e9c46e4fb0ae0))

Note:
- The result page is not responsive; will take a look.
- We'll need to add the json file to DO before merging.

Jira link: https://lite-farm.atlassian.net/browse/LF-5182

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


SayakaOno (commented):
This looks pretty good to me.

TODOs:
- `getSurveyVersion` needs to be adjusted
   - eg., `COUNTRY_VERSIONS.includes(countryCode) ? countryCode.toLowerCase() : 'default'`
- `getTapeSurveyData` endpoint is missing
- Replace `tapeQuestions.json` with `useQuery` in `TAPEResults`
- Delete `tapeQuestions.json`
- ~~Refetch is not happening when choosing a farm in a different country?~~
- How do we want to handle errors?
- Test with the APIs

packages/webapp/src/containers/Insights/index.jsx
```diff
@@ -50,17 +52,24 @@ const Insights = () => {
...
   const biodiversityData = null;
   const pricesData = useSelector(pricesSelector);
+  const isAdmin = useSelector(isAdminSelector);
 
   const dispatch = useDispatch();
   const { t } = useTranslation();
 
+  const surveyVersion = getSurveyVersion(farm?.country_code);
+  const { data: tapeSurvey, isError: isTapeSurveyError } = useGetTapeSurveyQuery();
+  const isTapeSurveyCompleted = !isTapeSurveyError && !!tapeSurvey?.id;
```
SayakaOno, line 69: By following these steps:
`Go to an Australian farm -> Complete a survey -> go to another Australian farm`
You get `tapeSurvey` for the previous farm although the API is refetched and returns 404. `isTapeSurveyError` is to prevent the wrong "completed" status.

packages/webapp/src/containers/Insights/index.jsx
```diff
@@ -50,17 +52,24 @@ const Insights = () => {
...
-      data_point: 'TAPE',
-    },
+    isAdmin && surveyVersion
+      ? {
+          label: t('INSIGHTS.TAPE.TITLE'),
+          image: tape_survey,
+          route: isTapeSurveyCompleted ? 'tape/results' : 'tape',
+          data_point: 'TAPE',
+        }
+      : undefined,
```
SayakaOno: I'll update this once https://github.com/LiteFarmOrg/LiteFarm/pull/4069 is merged. ✅

packages/webapp/src/routes/index.jsx
SayakaOno, line 1: Sorry for the formatting.
I added `TapeRoutes` for Admin roles and removed routes for workers.

packages/webapp/src/routes/TapeRoutes.tsx
```diff
@@ -0,0 +1,47 @@
...
+const TapeResults = React.lazy(() => import('../containers/Insights/TapeSurvey/TapeResults'));
+
+const TapeRoutes = ({ isCompactSideMenu }: { isCompactSideMenu: boolean }) => {
+  // @ts-expect-error -- userFarmSelector issue
+  const { country_code } = useSelector(userFarmSelector);
+  const surveyVersion = getSurveyVersion(country_code);
+
+  if (!surveyVersion) {
+    return <Redirect to="/" />;
+  }
```
SayakaOno, line 32: Routes won't be rendered for non-Australian farms.

packages/webapp/src/containers/Insights/TapeSurvey/index.tsx
```diff
@@ -13,48 +13,104 @@
...
+  useEffect(() => {
+    if (isSurveyJsonError) {
+      const activeError = notifications.find(
+        ({ message }) => message === t('INSIGHTS.TAPE.LOAD_ERROR'),
+      );
+      if (!activeError) {
+        dispatch(enqueueErrorSnackbar(t('INSIGHTS.TAPE.LOAD_ERROR')));
+      }
+    }
+  }, [isSurveyJsonError]);
```
kathyavini, line 94: Interesting! If it comes up again outside of TAPE, we might want to make this part of the default behaviour for `enqueueErrorSnackbar`; it seems like a good default

packages/webapp/src/containers/Insights/TapeSurvey/TapeResults.tsx
```diff
@@ -152,42 +177,31 @@ function TAPEResults() {
...
-          </div>
           <Semibold className={styles.titleText}>{t('INSIGHTS.TAPE.RESULTS_TITLE')}</Semibold>
           {tapeData && tapeData.length > 0 && (
             <div className={styles.chartContainer}>
               <Radar data={chartData} options={options} />
             </div>
           )}
         </div>
-        <div className={styles.sectionContainer}>
+        {/* <div className={styles.sectionContainer}>
```
kathyavini: I think it would be fair to just entirely delete this Step 2 placeholder (including its CSS), as I don't think the actual new design has anything in common with this weird button list thing 😜

packages/webapp/src/containers/Insights/TapeSurvey/TapeResults.tsx
```diff
@@ -100,14 +87,52 @@ interface TAPEDimension {
...
-function TAPEResults() {
+function TAPEResults({ surveyVersion }: { surveyVersion: string }) {
   const { t } = useTranslation();
   const history = useHistory();
   const dispatch = useDispatch();
 
-  const { surveyData } = useSelector(tapeSurveySelector);
-
-  const tapeData = analyzeTAPEData(surveyData);
+  const { data: surveyJson, isError: isSurveyJsonError } = useGetTapeSurveyJsonQuery(surveyVersion);
```
kathyavini: I showed this in tech daily (I _think_... it could have also been one of those times I thought I was screen sharing but wasn't 😬) but I should have also provided a link as well, sorry! 🙇

In this results component, we shouldn't be fetching the survey questions nor building the chart data from them. I knew this before, and described why [in this comment](https://github.com/LiteFarmOrg/LiteFarm/pull/3992#pullrequestreview-3641102043) but I also wrongly said at that time it doesn't matter, since for the FAO draft both methods worked.

Riddhi built the survey question names (that is, the top-level keys of the survey response object) so that every question that should be computed into a CAET score already has a fixed prefix, which will be unchanging from version to version:

```ts
const DIMENSION_MAPPING = {
  Diversity: 'diversity_1',
  Synergy: 'synergy_2',
  Recycling: 'recycling_3',
  Efficiency: 'efficiency_4',
  Resilience: 'resilience_5',
  'Culture and food traditions': 'culture_6',
  'Co-creation and sharing of knowledge': 'knowledge_7',
  'Human and social values': 'human_8',
  'Circular economy and solidarity': 'circular_9',
  'Responsible governance': 'governance_10',
};
```

Questions that don't have the prefix can be totally ignored for CAET scores and the graph. You can see in this new Australia survey that some of the CAET sections now have free-form responses, which are in the response object but DON'T have the prefix, which is correct (they shouldn't be averaged, as they can't) 👍 

However, this method of pulling from the survey questions and iterating over the content is breaking on those free answer questions, which is why the results graph on this branch has two subscores missing:

Image

It's also just completely not necessary! If we use the response key prefixes, there is no need for `getChartTitleFromSurveyTitle`, `getAnswerKeys`, `chartSectionData`, or fetching the survey with `useGetTapeSurveyJsonQuery`. We just need to pull out the responses that match each of the given prefixes, average over them, and convert to perecent:

```ts
const analyzeTAPEData = (data: any): TAPEDimension[] => {
  if (!data) return [];

  return Object.entries(DIMENSION_MAPPING).map(([dimension, prefix]) => {
    const scores = Object.keys(data)
      .filter((key) => key.startsWith(prefix))
      .map((key) => Number(data[key]) || 0);

    if (!scores.length) {
      return { dimension, score: 0, maxScore: MAX_SCORE };
    }

    const averageRawScore = scores.reduce((sum, value) => sum + value, 0) / scores.length;

    return {
      dimension,
      score: (averageRawScore / RAW_MAX_SCORE) * MAX_SCORE, // RAW_MAX_SCORE = 4
      maxScore: MAX_SCORE,
    };
  });
};
```

This also means there is no need to worry about about the survey structure changing between versions, or the number of questions in each section changing, or free-form responses or other kinds of questions being added; we just ignore anything that is not given the standardized prefix, so it's much simpler overall.

packages/webapp/src/containers/Insights/index.jsx
```diff
@@ -52,15 +54,20 @@ const Insights = () => {
   const biodiversityData = null;
   const pricesData = useSelector(pricesSelector);
   const isOffline = useIsOffline();
+  const isAdmin = useSelector(isAdminSelector);
 
   const dispatch = useDispatch();
   const { t } = useTranslation();
 
+  const surveyVersion = getSurveyVersion(farm?.country_code);
+  const { data: tapeSurvey, isError: isTapeSurveyError } = useGetTapeSurveyQuery();
```
kathyavini: Is there some blocking we could do on `isLoading` in this view? I notice that when I first log in, before the results data has come in (even without network throttling) I keep accidentally navigating to the survey (instead of the results view), as that is the link that is live initially.

I think it would probably be even be better to just let the TAPE tile pop in after a delay, rather than give the opportunity to navigate to the wrong page. Being brought to the first page of the cleared survey risks the user thinking they have to take the whole survey again from the beginning or that we have lost their responses!

packages/webapp/src/containers/Insights/index.jsx
kathyavini: Is there some blocking we could also do on `isLoading` in this view? When I switch farms, before the results data has come in (even without network throttling) I keep accidentally navigating to the survey (instead of the results view), as that is the link that is active initially until the query completes.

I think it would be better if there was no link until the query completes (or maybe no tile? I'm not really sure!) rather than give the opportunity to navigate to the wrong page. Being brought to the first page of the cleared survey made me think that the data was lost, and I think a user might think they have to take the whole survey again from the beginning, which would be pretty bad.

kathyavini (commented):
@SayakaOno I love it! This is very clean and thank you so much for handling on those crazy scope changes in the last two days, that was a very fast adjust to all of that! 🙇🙇

The only blocking part is the inherited code for the results graph -- it shouldn't be (and in this particular case, doesn't work with) parsing the survey JSON. I am sorry; I threw that fact at you in a tech daily but I should have followed up with a link or written note!

Duncan-Brain:
Seems to be working pretty well for me! 

I noticed a few things:

- Agree with Joyce it seems like we need to handle the case for those extra inputs. I was able to achieve over 100% when I answered the survey text inputs with a number
-  When I fill out the survey and accidentally press no to any terms and conditions. I cannot recover from that.
- The farm data that is not removed when switching farms: I really think we should fix that `resetAPIState()` bug, this survey seems like a candidate where extension officers may help meaning multiple farms for a user are maybe more likely and the sharing of personal data could occur
-  [Future looking] Was our main goal for adding the json file to the DO bucket to avoid new releases for survey changes? If so I am wondering if we should think about adding `countries` and the `DIMENSION_MAPPING` Joyce mentions in her comment above to the surveyJson metadata. That way we would not have to re-release if those details change.

packages/webapp/src/containers/Insights/index.jsx
Duncan-Brain: When current data present for farm1 and error for farm2 it shows data on the wrong farm:  This seems like another showcase for why it seems important to fixi the api state when switching between farms... 

The root cause of the other farms data still being present can be fixed. @SayakaOno could you please check out : https://github.com/LiteFarmOrg/LiteFarm/pull/3988 . 

If an extension officer or anyone helps multiple farms fill out tape surveys this likelihood increases.

packages/webapp/src/containers/Insights/TapeSurvey/TapeResults.tsx
SayakaOno: Understood, thank you!
Does `// MAX_SCORE = 4` in the code snippet above mean `RAW_MAX_SCORE`?

kathyavini: Yes! Sorry I corrected it this morning; are you still seeing the previous comment?

It's updated for me 😅 

Screenshot 2026-03-11 at 10 25 44 AM

SayakaOno: I was, thank you!

packages/webapp/src/containers/Insights/TapeSurvey/index.tsx
SayakaOno: That sounds like a good idea!

packages/webapp/src/containers/Insights/index.jsx
SayakaOno: @kathyavini Thank you so much for catching it! Tape's current data will be "Loading..." while fetching the results, and I disabled onClick using CSS. It would be appreciated if you could test the new UX 🙏 

@Duncan-Brain I thought about that PR too. I've been looking for the right time for us to revisit it; let's do it tomorrow maybe after the release?

SayakaOno (commented):
Thank you so much for reviewing! 🙇  I think I've addressed everything, and the result chart should be responsive now.
Re-reviews would be appreciated 🙏 

> When I fill out the survey and accidentally press no to any terms and conditions. I cannot recover from that.

Do we need to handle this?

> If so I am wondering if we should think about adding countries and the DIMENSION_MAPPING Joyce mentions in her comment above to the surveyJson metadata.

I agree, I'm expecting some changes in the future.

kathyavini:
@SayakaOno I approve this PR!

Please merge whenever ready ❤️ You can also retag if you want to add any further responsivity adjusts first. I can't believe it -- I had just assumed the chart was non-responsive instead of trying throwing it in a responsive container 😭

The new UX of the loading state on the insights tile is perfect 👌

SayakaOno:
Thank you @kathyavini for testing!
I added a commit for the style adjustment. Thank you for re-reviewing 🙏 

I also found a bug that the survey is not available right after creating a farm in Australia. I'll open another PR for the fix.

packages/webapp/src/containers/Insights/TapeSurvey/TapeResults.tsx
```diff
@@ -140,6 +104,13 @@ function TAPEResults() {
           font: {
             size: 14,
           },
+          // Splits labels into a maximum of 2 lines (assumes English labels)
+          callback: (label: any) => {
+            const words = label.split(' ');
+            const splitIndex = words.length === 1 ? 1 : Math.floor(words.length / 2);
+
+            return [words.slice(0, splitIndex).join(' '), words.slice(splitIndex).join(' ')];
```
kathyavini, line 112: Oh it takes arrays for the labels! Thank you for figuring that out ❤️

kathyavini (commented):
Oh that looks so much better!!! 🔥🔥🔥

> I also found a bug that the survey is not available right after creating a farm in Australia. I'll open another PR for the fix.

Really?? I made a new farm when testing this yesterday. I'll check it out in the new PR 🙏

Please do approve and merge ☑️

SayakaOno (approved):

PR #4069, opened 2026-03-02
LF-5183 Workbox precache exclusion for SurveyJS

kathyavini:
**Description**

- Moves the SurveyJS library code (~1.4MB) from our `workbox-precache` to runtime caching managed by our service worker; i.e. it will be stored in cache the first time it's fetched, and never fetched at all for users who never visit TAPE
- To enable this, adds a little bit of manual chunking to our Vite (Rollup under the hood for this part) config. Comment below / inline has more details
- Hides the TAPE survey option from the Insights table when offline. Because it's no longer in the workbox precache, visiting this link _for the first time_ when offline would cause a crash
  - for this release we are offering TAPE as an online-only feature, and therefore at this stage I don't think it's necessary to do anything more complicated. Sounds like in the future we might revisit this for FAO

Jira link: https://lite-farm.atlassian.net/browse/LF-5183

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

My main testing interests were:

1) Seeing the `workbox-precache` decrease by the size of the SurveyJS code. I used the same check as we have used before to check cache size
2) **_(Most difficult to implement, but now working)_** Making sure the survey code (the `vendor-survey` chunk) was not network requested until the survey page was loaded
3) Seeing the chunk subsequently stored in cache and served from service worker after initial load

**Checklist:**


kathyavini (commented):
Well, this one is pretty wrong on first pass; probably the plan wasn't specific enough since the model has the knowledge/context to fix each of these:

1. `NetworkOnly` strategy like this is functionally equivalent to just omitting from the service worker; should omit or use `CacheFirst`. This is the one I pointed out on the plan as being wrong.
2. (This one I didn't realize until testing) Using manual chunks creates static import right in the entry HTML with `modulepreload`

Image

so it's loaded right on first page load. That's definitely not what we want if our goal is only fetching those 1.4MB on demand! `modulepreload` behaviour should be configurable though.

3. (Also a learning on my part) The 1.4MB file was the lazy-loaded route (`TapeSurvey/index.tsx`). So omitting it from the workbox-precache does not require manual chunking if we rename it to something we can glob on:

Image


TODO:
- [x] I will adjust for a lazy-load and `CacheFirst` strategy on the route
- [x] Will test again separating out the vendor code (i.e. manual chunks but disabling `modulepreload`)
  - TOOK FOREVER BUT NOW WORKING 🎉
- [x] otherwise I have tested and like the UX of runtime caching the lazy-loaded route via the rename strategy from (3)
- [x] I will add offline hiding since this is not an offline-available feature

packages/webapp/vite.config.ts
```diff
@@ -45,13 +45,39 @@ export default defineConfig({
...
+
+          if (
+            id.includes('/node_modules/survey-core/') ||
+            id.includes('/node_modules/survey-react-ui/')
+          ) {
+            return 'survey-vendor';
+          }
+
+          return undefined;
+        },
```
kathyavini, line 76: This manual chunks issue was a beast to debug and finally get working 😆

When separating out the `survey-core` and `survey-react-ui` packages using `manualChunks`, two things were happening (and they were related)

1. it looked like about ~500kB of code was being shifted from the main entrypoint chunk into `survey-vendor`
2. `survey-vendor` was being imported right at the top of the built entrypoint file and so was immediately loaded, completely negating the point of exclusion from precache

After a looooong session of trying to figure out what was going on, the answer was here at the top of the `.map` for the `survey-vendor` chunk -- a bunch of react core library packages (dependencies) were being chunked as well

Image

I'm pretty sure that this would NOT be the case if we were on Vite 5 / Rollup 4 and could set [`onlyExplicitManualChunks: true`](https://rollupjs.org/configuration-options/#output-onlyexplicitmanualchunks), which will also be the _default_ in Rollup 5. But our version of Vite uses Rollup 3.

Although this is a complete manual workaround for not having that Rollup option, I wouldn't necessarily remove it when we upgrade -- it achieves another goal of mine I had mentioned before, which is hash stability (and so cache preservation) between releases. Both `framework-vendor` and `survey-vendor` should stay unchanged as long as we don't upgrade those packages, and so users shouldn't have to re-download them between releases unnecessarily.

If you have this logging enabled for your service worker, you can test this locally by changing some app code and "re-releasing" (rebuild, preview, and reload) then searching for the survey-vendor and survey-framework chunks in this "previously cached URLs" list

Screenshot 2026-03-05 at 1 08 23 PM

packages/webapp/src/sw.js
```diff
@@ -48,6 +48,12 @@ async function validatePrecacheIntegrity() {
   }
 }
 
+// Assets omitted from precache, but cached on first fetch for fast subsequent loads
+registerRoute(
+  ({ url }) => /\/assets\/(survey-vendor)-[^/]+\.(js|css)$/.test(url.pathname),
+  new CacheFirst({ cacheName: 'dynamic-chunks' }),
+);
```
kathyavini, line 55: Designed to be extensible to any other parts of app we would like to avoid fetching upfront

kathyavini:
@SayakaOno and @Duncan-Brain I am done working on this ticket and it is ready for review 🙏 

I have removed the "Experiment" part from the title and created a normal description -- I guess at this point it can just be treated now as any other PR.

SayakaOno (approved):
Great job on figuring this out!!
Thank you for providing the details, I learned a lot from these changes! ❤️

PR #4071, opened 2026-03-02
LF-5184 (Experiment) TAPE Plan Phase 3 - Backend

kathyavini:
**Description**

As discussed today in tech daily.

We are doing an [Audit logging in place method](https://www.red-gate.com/blog/database-design-for-audit-logging/) There are downsides to this approach but we are fine with for now. 

@SayakaOno this PR may change what was originally envisioned by AI and yourself. Specifically the patch. If you would prefer a single endpoint, or a different update url to use body instead please let me know!

Jira link: https://lite-farm.atlassian.net/browse/LF-5184

**Type of change**

**How Has This Been Tested?**

- [x] Passes test case
- [ ] Other (please explain)

**Checklist:**


Duncan-Brain:
Tried to stay on theme and let the agent do most of the work. 

I tried to give commit attribution to the ai agent but I forgot to for the test boilerplate.  

Issues with the agent I saw:
- kept asking me to fetch example.com -- eventually I had to add a rule
- was writing several node commands to query and edit the db directly some it asked for permission but others I believe were done without permission -- had to add some rules

Issues with human:
- The ai caught a few bugs I missed during the endpoint creation - I think I would have caught these through testing but had I not been doing testing then they can slip through.  Essentially I am better at identifying if what is present is wrong rather than if something is absent and should be present.

Duncan-Brain:
@kathyavini I cannot add you to the reviewers unless I close PR and reopen. Depending on whose you want to review we we will have to do surrogate approvals haha

packages/api/src/controllers/tapeSurveyController.js
```diff
@@ -0,0 +1,87 @@
...
+    return async (req, res) => {
+      try {
+        const { farm_id } = req.headers;
+        // Find the latest tape survey for the farm
+        const result = await TapeSurveyModel.query()
+          .where({ farm_id })
+          .orderBy('created_at', 'desc')
+          .first();
+        if (!result) {
+          return res.status(404).json({ error: 'Tape survey not found' });
```
SayakaOno, line 49: Could we return an empty object? The reason is the same as the 404 on the empty management plan. (https://github.com/LiteFarmOrg/LiteFarm/pull/4079/changes)

packages/api/tests/tapeSurvey.test.js
```diff
@@ -0,0 +1,223 @@
+/*
+ *  Copyright (C) 2026 Free Software Foundation, Inc. <https://fsf.org/>
+ *  This file (tapeSurvey.test.js) is part of LiteFarm.
```
SayakaOno, line 3: Was this written by AI? 😂 It's the old format.

packages/api/src/controllers/tapeSurveyController.js
Duncan-Brain: I thought returning an empty array `[]` was only for endpoints that can return multiple records. This endpoint was only going to return one single record so the 404 seemed appropriate. 

But it might be good if I return all submissions now that we are doing it more inline with our other tables. @SayakaOno  I will update this endpoint to return all submissions (not submission logs) for a farm if that is okay and we only use the one? I could sort descending by date if thats helpful.

SayakaOno: I should've said null instead of an empty object.
It didn't make sense to me that "no management plans" is not an error but "no submission" is an error. But I know there's no correct answer, and I'll handle 404 on the frontend.

I don't think we want old submissions... would it make more sense if we create another endpoint `/tape_survey/latest` or receive a param `/tape_survey?latest=true`?
If there's a way to get the latest one, that would be great.

packages/api/tests/tapeSurvey.test.js
Duncan-Brain: Definitely yes, I did update the year haha

packages/api/src/controllers/tapeSurveyController.js
Duncan-Brain: Google says the most common way to do that on a collection endpoint is : 
`GET /items?sort=createdAt:desc&limit=1` or `GET /items?sort=-createdAt&limit=1`

or less frequently `GET /items/latest`

@SayakaOno do you have a preference? If I don't hear back today I will probably assume `/latest` if thats cool!

packages/api/src/controllers/tapeSurveyController.js
```diff
@@ -0,0 +1,87 @@
...
+import TapeSurveyModel from '../models/tapeSurveyModel.js';
+
+const tapeSurveyController = {
+  createTapeSurvey() {
+    return async (req, res) => {
+      try {
+        const { farm_id } = req.headers;
+        const { user_id } = req.auth;
+        const { survey_response } = req.body;
+        const { survey_version, project_id, survey_step } = survey_response;
```
SayakaOno, line 25: Could we flatten this so the request body is aligned with the table structure?
I'd like to send:
```
{ survey_response, survey_version, project_id, survey_step, ... }
```
instead of:
```
{ survey_response: { survey_version, project_id, survey_step, ... }  ... }
```

But let me confirm first...
@kathyavini `survey_response` in SurveyJS won't include survey_version, project_id, survey_step automatically, correct?

packages/api/src/controllers/tapeSurveyController.js
SayakaOno: Do you have a preference? I'm okay with either!

SayakaOno: Oh but hold on, let's go with what we have now, we don't have much time...

packages/api/src/controllers/tapeSurveyController.js
kathyavini: > survey_response in SurveyJS won't include survey_version, project_id, survey_step automatically, correct?

@SayakaOno they should! They should be top-level keys in `survey_response` (= `surveyData`), but could you double-check in your updated frontend? The last draft Riddhi provided should have the `survey_version` at least -- the others should be in the final draft.

Edit: I TAKE IT BACK! I guess calculated values are not treated like questions. Is that okay for you @SayakaOno? ~Are you able to access them some other way, or should Riddhi put the data in some other way?~  Nvm it's trivial to add a calculated value to the result, see below

kathyavini: This modification to the JSON puts them in as top-level keys:

```json
"calculatedValues": [
  {
    "name": "survey_version",
    "expression": "'TAPE_Mornington_CAET_2026_V1'",
    "includeIntoResult": true // <-- this is sufficient to see them in `survey_response`
  }
  ```

SayakaOno: Okay great, thank you Joyce!

@Duncan-Brain Please forget about what I said, this is good!

SayakaOno (approved):
It seems to be working well with the frontend! 🎉

PR #4083, opened 2026-03-12
LF-4672 (1) - Rtk upgrade/locations rebase

Duncan-Brain:
**Description**

This is phase 1 of 3 of locations migration away from sagas to rtk-query.

**Major changes:**
- creates endpoints in api for location to eventually replace all location sagas
- migrates most selectors to hooks from locationSlice
- removes selector stubs for sensors
- only render map if assets are ready -- add temporary loading screen for maps as it can take some time (Part 2 follows figma design) (sensors was broken so in order to test changes I needed this)
- removed as many references as possible to location type selectors (eg. barnSelector) without refactoring the location type containers and selectors details Part 3 will consolidate all those individual area selectors

**Review Notes:**
- Looks scarier than it is: File changes are broad because they are core selectors but lines of code changes is limited to only a few files
- Check that selectors all have a hook or that selector was unused
  - useLocation hook is core hook - update and add endpoints currently unused until part 3
  - useExternalLocations hook mirrors the locations hook very closely but does not need cleaning or flattening
  - all other hooks use the base hooks fairly simply
  - useAvailableFilterSetting is only used in `/Map` so that is why it is not top level
- Types are a lot of code but I think up to you if you want to understand it
  - essentially has to mirror our backend get construct which is annoying format
- New selector stubs made to satisfy management plan and task selectors that need that access until we migrate those parts of the app

**Other notes**
- pre-existing bugs:
  -  Flash of just farm centre before assets load on farm with external sensors
  - zoom does not work on taskLocations very well, farm center is always included so cross country locations dont behave as expected.
  - flicker on locationPicker
  - blinkiness, flashing of 'points' when selecting a menu item or filtering
- backend sends us very messy data (eg location: { type: barn, barn: {...}, greenhouse: null, area: {...}, line: null ..})
- should be able to remove `refetch` of sensors on `/Home` isFetching blocker prevents case where data does not exist -- also it didn't work for me on integration.

Jira link: [LF-4672](https://lite-farm.atlassian.net/browse/LF-4672)

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

[LF-4672]: https://lite-farm.atlassian.net/browse/LF-4672?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/webapp/src/components/Map/Footer/index.jsx
```diff
@@ -22,26 +22,7 @@ export default function PureMapFooter({
   filterSettings,
   onFilterMenuClick,
   onAddMenuClick,
-  availableFilterSettings = {
```
Duncan-Brain, line 25: This was a duplicated part of a selector migration, also locationEnum should be phased out for typescript versions.

packages/webapp/src/components/MapDrawer/index.jsx
```diff
@@ -142,6 +143,11 @@ export default function MapDrawer({
           icon: () => <Sensor style={{ transform: 'translate(-5px, 5px)' }} />,
           key: locationEnum.sensor,
         },
+        {
```
Duncan-Brain, line 146: Part of the sensors vs sensor array being filtered separately discussion with Loic.

packages/webapp/src/components/Modals/LoadingMapModal/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { Dialog } from '@mui/material';
+import { makeStyles } from '@mui/styles';
+
+const useStyles = makeStyles((_theme) => ({
+  paper: (_props) => ({
+    overflow: 'inherit',
+  }),
+}));
+
+export default function LoadingMapModal({ isOpen }: { isOpen: boolean }) {
```
Duncan-Brain, line 29: This temporary component will be deleted with part 2 where I implement Loic's loading design.

packages/webapp/src/containers/Map/index.jsx
```diff
@@ -440,140 +442,143 @@ export default function Map({ isCompactSideMenu }) {
...
-                dispatch(setSpotlightToShown('adjust_line'));
-              }}
-            />
-          )}
-        </div>
-      </>
-    )
+        </>
+      )}
+      {isLocationsFetching && <LoadingMapModal isOpen={isLocationsFetching} />}
```
Duncan-Brain, line 581: The desired behaviour for me is that this should be `isLocationsLoading` - block render if current data does not exist. And then `isLocationsFetching` is used to refresh/redraw the data if it changes.

There is a variety of rendering issues with this component that make this hard to implement right now. The old way the selector updates were instantaneous and now we wait for a round trip back from invalidating tags.

I tried a method similar to the method recently removed when redraw was needed here: https://github.com/LiteFarmOrg/LiteFarm/pull/3993/changes/9944bf88f7449d629e19221229f1e1cc646e56f1 As noted in that PR we will need something like that again when we start adding locations in the new add location redesign. But this is a good compromise until that time. 

There is a too many re-rendering problem with this component and I think if we can solve that in the future this component will look much simpler. But lets also wait until we update our drawing library.

packages/webapp/src/containers/barnSlice.js
```diff
@@ -86,10 +86,3 @@ export const barnsSelector = createSelector(
 
 export const barnSelector = (location_id) =>
   createSelector(barnEntitiesSelector, (entities) => entities[location_id]);
-
-export const barnStatusSelector = createSelector(
```
Duncan-Brain, line 90: These are unused

packages/webapp/src/store/api/types/index.ts
```diff
@@ -413,3 +407,475 @@ export interface SupportTicketData {
...
+}
+
+export enum TotalAreaUnit {
+  M2 = 'm2',
+  HA = 'ha',
+  FT2 = 'ft2',
+  AC = 'ac',
+}
+
+export enum PerimiterUnit {
```
SayakaOno: AI says `PerimiterUnit` should be `PerimeterUnit`.

packages/webapp/src/hooks/location/types.ts
```diff
@@ -0,0 +1,208 @@
...
+// Shared
+export enum GroupByOptions {
+  TYPE = 'type',
+  FIGURE = 'figure',
+  FIGURE_AND_TYPE = 'figure_and_type',
+}
+
+type MapLocationType = InternalMapLocationType | ExternalMapLocationType;
+
+export type UseLocationsPropsWithFilterBy<AvailableLocationTypes extends MapLocationType> = {
```
SayakaOno: I think `AvailableLocationTypes` should be `AvailableLocationType` (singular).

packages/webapp/src/store/api/locationApi.ts
```diff
@@ -0,0 +1,85 @@
...
+          return { error: result.error };
+        }
+        return { data: result.data as InternalMapLocation[] };
+      },
+      providesTags: ['Locations'],
+    }),
+    checkDeleteLocation: build.mutation<void, WithLocationId>({
+      query: ({ location_id }) => ({
+        url: `${checkDeleteLocationUrl}/${location_id}`,
+        method: 'GET', // Mutations are not the normal pattern for GET, but the result should never be cached
```
SayakaOno, line 41: Did you find this solution somewhere?
I'm curious because `keepUnusedDataFor: 0` on a `build.query` or `refetchOnMountOrArgChange: true` at the hook call site seem more like common patterns.

packages/webapp/src/hooks/location/useLocations.ts
```diff
@@ -0,0 +1,228 @@
...
+    : location.figure;
+
+  cleanedLocation.figure = cleanedFigure;
+
+  return cleanedLocation;
+};
+
+export const flatten = (location: any): FlattenedInternalMapLocation => {
+  const clone: InternalMapLocation = { ...location };
+  let flattened = {} as any;
```
SayakaOno: I think this should be `const`?

packages/webapp/src/hooks/location/useLocations.ts
```diff
@@ -0,0 +1,228 @@
...
+function useLocations(
+  props?: undefined | UseInternalLocationProps,
+): UseLocationsReturn<FlattenedInternalMapLocation[] | undefined>;
+
+function useLocations(
+  { filterBy, groupBy, deleted }: UseInternalLocationProps = { deleted: false },
+): any {
+  const { data: locations, isLoading, isFetching } = useGetLocationsQuery();
+
+  if (isLoading) {
```
SayakaOno: Could we change this to `if (isLoading || !locations?.length)` and remove `if (locations && locations.length)`?

packages/webapp/src/containers/Map/useMapAssetRenderer.js
```diff
@@ -114,11 +109,27 @@ const useMapAssetRenderer = ({ isClickable, showingConfirmButtons, drawingState
...
+  });
+  const {
+    locations: externalLocations,
+    isLoading: isLoadingExternalLocations,
+    isFetching: isFetchingExternalLocations,
+  } = useExternalLocations({ groupBy: 'figure_and_type' });
+  const isLocationsLoading = [isLoadingInternalLocations, isLoadingExternalLocations].some(Boolean);
+  const isLocationsFetching = [isFetchingInternalLocations, isFetchingExternalLocations].some(
+    Boolean,
+  );
```
SayakaOno: Could we simplify this using an `a || b` pattern?

packages/webapp/src/containers/Map/useAvailableFilterSettings.ts
```diff
@@ -0,0 +1,48 @@
...
+
+import { MAP_LOCATION_TYPE_BY_FIGURE } from './constants';
+import useLocations from '../../hooks/location/useLocations';
+import useExternalLocations from '../../hooks/location/useExternalLocations';
+import { GroupByOptions } from '../../hooks/location/types';
+import { FigureType } from '../../store/api/types';
+
+const DEFAULT_FILTER_SETTINGS = MAP_LOCATION_TYPE_BY_FIGURE;
+
+// From map filter setting slice -- map utils insteasd of hooks?
```
SayakaOno: Could this comment be removed now? 😄

packages/webapp/src/containers/Map/useAvailableFilterSettings.ts
```diff
@@ -0,0 +1,48 @@
...
+const useAvailableFilterSettings = () => {
+  const { locations: internalLocations, isLoading: isLoadingInternalLocations } = useLocations({
+    groupBy: GroupByOptions.FIGURE_AND_TYPE,
+  });
+  const { locations: externalLocations, isLoading: isLoadingExternalLocations } =
+    useExternalLocations({ groupBy: GroupByOptions.FIGURE_AND_TYPE });
+  const isLocationsLoading = [isLoadingInternalLocations, isLoadingExternalLocations].some(Boolean);
```
SayakaOno: Could we change this to `isLoadingInternalLocations || isLoadingExternalLocations`?

packages/webapp/src/containers/Map/constants.js
```diff
@@ -165,3 +168,29 @@ const adjustAngle = (currentAngle) => {
   }
   return currentAngle;
 };
+
+// Shared location types
+export const MAP_LOCATION_TYPE_BY_FIGURE = {
+  [FigureType.AREA]: [
```
SayakaOno, line 174: AI caught that `farm_site_boundary` is missing.

packages/webapp/src/store/api/locationApi.ts
```diff
@@ -0,0 +1,85 @@
...
+
+import { api } from './apiSlice';
+import { checkDeleteLocationUrl, getLocationsByFarmIdUrl, locationURL } from '../../apiConfig';
+import { InternalMapLocation, InternalMapLocationType, WithLocationId } from './types';
+import { RootState } from '../store';
+
+export const locationApi = api.injectEndpoints({
+  endpoints: (build) => ({
+    getLocations: build.query<InternalMapLocation[], void>({
+      queryFn: async (_arg, { getState }, _extraOptions, baseQuery) => {
```
SayakaOno, line 24: I saw your [comment](https://github.com/LiteFarmOrg/LiteFarm/pull/4083/changes#r2926300716) that `isLocationsLoading` is ideal for showing the loading modal. Why not pass `farm_id` as an argument to the query? Was it too complicated?

packages/webapp/src/containers/Map/index.jsx
```diff
@@ -440,140 +442,143 @@ export default function Map({ isCompactSideMenu }) {
...
-                  confirmLine={handleLineConfirm}
-                  updateLineWidth={setLineWidth}
-                  system={system}
-                  lineData={overlayData}
-                  typeOfLine={drawingState.type}
-                  onLineParameterChange={() => {
-                    dispatch(setIsRedrawing(true));
-                  }}
+    <>
+      {!isLocationsFetching && isLoaded && (
```
SayakaOno, line 432: Can't we render the map while locations are being fetched?

packages/webapp/src/hooks/location/types.ts
```diff
@@ -0,0 +1,208 @@
...
+  deleted?: boolean;
+};
+
+export type UseLocationsPropsWithGroupBy = {
+  filterBy?: never;
+  groupBy?: GroupByOptions;
+  deleted?: boolean;
+};
+
+export type UseLocationsReturn<T> = {
```
SayakaOno, line 189: AI says `isFetching` is missing.

packages/webapp/src/hooks/location/useExternalLocations.ts
```diff
@@ -0,0 +1,158 @@
...
+      return { locations: filteredLocations, isLoading, isFetching };
+    }
+
+    if (filterBy && allFigureTypes.includes(filterBy)) {
+      const filteredLocations = locations.filter(({ figure_type }) => figure_type === filterBy);
+      return { locations: filteredLocations, isLoading, isFetching };
+    }
+
+    if (groupBy === GroupByOptions.TYPE) {
+      const groupedLocations = Object.groupBy(locations, ({ type }) => type);
```
SayakaOno: AI pointed that we use polyfill for `groupBy()` in other places.
eg., `packages/webapp/src/containers/Finances/useTransactions.js`

packages/webapp/src/hooks/location/useLocations.ts
```diff
@@ -0,0 +1,228 @@
...
+
+    if (filterBy && allFigureTypes.includes(filterBy)) {
+      const filteredLocations = flattenedLocations.filter(
+        ({ figure_type }) => figure_type === filterBy,
+      );
+      return { locations: filteredLocations, isLoading, isFetching };
+    }
+
+    if (groupBy === GroupByOptions.TYPE) {
+      const groupedLocations = Object.groupBy(flattenedLocations, ({ type }) => type);
```
SayakaOno: Polyfill?

packages/webapp/src/hooks/location/useLocations.ts
```diff
@@ -0,0 +1,228 @@
...
+    ? Object.fromEntries(
+        Object.entries(location.figure).filter(([key]) => {
+          // Keep shared figure props (type, location_id, figure_id)
+          // OR the active geometry key
+          return !allFigureTypes.includes(key) || key === figureType;
+        }),
+      )
+    : location.figure;
+
+  cleanedLocation.figure = cleanedFigure;
```
SayakaOno, line 73: AI says:

> clean() mutates RTK Query cached data in one edge case
> packages/webapp/src/hooks/location/useLocations.ts:51-69 — when locationType is falsy, cleanedLocation is assigned the original location object reference. If figureType is non-null, the assignment cleanedLocation.figure = cleanedFigure then mutates the RTK Query cache entry. RTK Query's immer-frozen state will throw in dev mode (strict freeze) and silently corrupt the cache in prod. The fix is to spread: const cleanedLocation = locationType ? Object.fromEntries(...) : { ...location }.

According to the types, `locationType` doesn't seem to become falsy. Could you confirm?

packages/webapp/src/containers/Map/useMapAssetRenderer.js
```diff
@@ -114,11 +109,27 @@ const useMapAssetRenderer = ({ isClickable, showingConfirmButtons, drawingState
...
-  const lineAssets = useSelector(lineSelector);
-  const internalPoints = useSelector(pointSelector);
-  const externalPoints = useSelector(externalPointSelector);
-  const pointAssets = { ...internalPoints, ...externalPoints };
+  const {
+    locations: internalLocations,
+    isLoading: isLoadingInternalLocations,
+    isFetching: isFetchingInternalLocations,
+  } = useLocations({
+    groupBy: 'figure_and_type',
```
SayakaOno: AI says we should use `GroupByOptions.FIGURE_AND_TYPE` here too. It seems to work...

packages/webapp/src/hooks/location/useLocations.ts
```diff
@@ -0,0 +1,228 @@
...
+  props: UseLocationsPropsWithGroupBy & { groupBy: GroupByOptions.FIGURE_AND_TYPE },
+): UseLocationsReturn<
+  LocationsGroupedByFigureAndType<InternalMapLocationType, FlattenedInternalMapLocation> | undefined
+>;
+
+function useLocations(
+  props?: undefined | UseInternalLocationProps,
+): UseLocationsReturn<FlattenedInternalMapLocation[] | undefined>;
+
+function useLocations(
```
SayakaOno, line 167: AI says:

> packages/webapp/src/hooks/location/useLocations.ts:177-178 — every call to useLocations re-runs activeLocations.map(clean).map(flatten) plus all the groupBy logic. Since RTK Query returns a stable reference when the cache is unchanged, wrapping this in useMemo with locations as the dependency would prevent redundant work across re-renders. This is especially relevant because useLocations is called multiple times in the Map tree with different groupBy/filterBy options.

I wonder if we should apply `clean` and `flatten` in [`transformResponse`](https://redux-toolkit.js.org/rtk-query/usage/customizing-queries#customizing-query-responses-with-transformresponse) so we could avoid using `useMemo`. Any thoughts?

SayakaOno (commented):
Hi Duncan! How are you doing? I'm so sorry it took a while to review this.
Great work, I think it's well thought out and will be a helpful reference for migrating management plans and tasks etc.

In addition to inline comments, some files have license with the year 2025, so it would be great if you could update them to 2026.

Let me know if you have any questions, thank you!

Duncan-Brain:
Hey @SayakaOno ! Doing well, very busy on the farm been waiting for a few rainy days to check in!

I am going to try to address your comments this week/tonight if possible.

For the license year 2025 is likely correct, as I started working on this late last year.

If I have trouble with conflicts I may need assistance to move this through, I have tried to watch what new was going in but haven't been able to watch everything. The new maps upgrade is what I am will be most unsure about if something goes in there.

packages/webapp/src/store/api/locationApi.ts
Duncan-Brain: This seemed liked the best solution I forget exactly where it came from but I think I decided it was best. Essentially the query is the more common pattern you are right but this solution removes any mental overhead about thinking about cache lifetime and refetching -- the mutation is always correct and refetched never cached. I think this was in the spirit of the original backend check.

One specific case where this approach wins is concurrency. I try to delete a location but it fails because there is a task. On another tab or another user deletes that task. I try again on the other tab it should not fail but it does because it is cached.

Ai explained it to me like so:

"With a mutation, there's no reasoning required: every call is unconditionally a fresh network request, no cache involved at any stage. That's the real benefit — not that the lazy query is broken in your scenario, but that the mutation makes the guarantee explicit by design rather than by convention. For an endpoint whose entire purpose is "always get fresh state before a destructive action," that's the stronger contract.

The semantic oddness of useCheckDeleteLocationMutation is a real cost, but I'd consider it worth it here — or you could wrap it with a better name at the call site. The pattern is valid."

packages/webapp/src/hooks/location/useLocations.ts
Duncan-Brain: I thought only constants were const and lets were used for utility variables (example: `for (let x = 0; x < 5; x++)`).  Maybe its a style thing?

packages/webapp/src/containers/Map/useMapAssetRenderer.js
Duncan-Brain: sure!

packages/webapp/src/hooks/location/useLocations.ts
Duncan-Brain: okay!

packages/webapp/src/containers/Map/useAvailableFilterSettings.ts
```diff
@@ -21,14 +21,13 @@ import { FigureType } from '../../store/api/types';
...
 
-// From map filter setting slice -- map utils insteasd of hooks?
 const useAvailableFilterSettings = () => {
   const { locations: internalLocations, isLoading: isLoadingInternalLocations } = useLocations({
     groupBy: GroupByOptions.FIGURE_AND_TYPE,
   });
   const { locations: externalLocations, isLoading: isLoadingExternalLocations } =
     useExternalLocations({ groupBy: GroupByOptions.FIGURE_AND_TYPE });
-  const isLocationsLoading = [isLoadingInternalLocations, isLoadingExternalLocations].some(Boolean);
+  const isLocationsLoading = isLoadingInternalLocations || isLoadingExternalLocations;
```
Duncan-Brain, line 30: So weird thing happened to me here.. I committed the comment change above, then before the pre-commit was finished I continued updating the file -- and this seems to have gotten added.. its fine but I was not aware of that haha

packages/webapp/src/containers/Map/constants.js
Duncan-Brain: Woops! Thanks AI

packages/webapp/src/store/api/locationApi.ts
Duncan-Brain: I am not sure I understand how `farm_id` would help. The comment is not related to the bug where another farms data shows -- I saw this problem occurred recently on an time consuming endpoint. That is still present here until we fix the `resetApiState()`

What I was saying in the comment is that there is render logic on map does not seem correct to me (multiple re-renders, infinite renders when trying some solutions, etc) and it causes problems. I didn't want to overcomplicate this PR by attempting to solve those render issues. So regardless of if I solved the farm data there would still be issues until that is solved.

packages/webapp/src/containers/Map/index.jsx
Duncan-Brain: I did not have good results with that approach. But I agree it is ideal. Related to the comment you mentioned above.

packages/webapp/src/hooks/location/types.ts
Duncan-Brain: Im slightly surprised it wasn't caught by tsc.. fixing!

packages/webapp/src/hooks/location/useExternalLocations.ts
Duncan-Brain: interesting, ts doesnt like that solution too much so I put in some cheat ts-expect-errors for now

packages/webapp/src/containers/Map/useMapAssetRenderer.js
Duncan-Brain: sure that should work

packages/webapp/src/hooks/location/useLocations.ts
Duncan-Brain: This is definitely one of my frontend gaps, I think the right approach will be to make a structure clone earlier in the hook to prevent any chance of cache mutation. We use these data all over and there no reason to allow the possibility I think. The solution committed is what AI suggests to address that concern completely.

If we run into performance issues, we can lump this into a memoization.

packages/webapp/src/hooks/location/useLocations.ts
Duncan-Brain: My thought is that the root issue is the backend data retrieval, it is sending the frontend useless data. I would prefer to fix that in the future rather than having a beefy frontend api. 

That said my AI said to me: "Given no perf issues yet, I'd leave it as-is and add `useMemo` only if you see the hook showing up in profiler traces." -- If we are optimizing maps components probably the most gains would be to spend time on the google maps instance. I am guessing most farms have much much less than <<1000 locations. Which is my mental cheat amount for when to think about it.

Also `deleted` prop filters the activeLocations before it goes into those functions making it a display specific prop that would have to be communicated to the api to see that benefit. 

So `transformResponse` could indeed be used to optimize if needed!

Duncan-Brain:
@SayakaOno it should be updated! No issue with conflicts.

packages/webapp/src/store/api/locationApi.ts
SayakaOno: No, I agree that the cached data should never be returned for this. I was just wondering if using mutation was the right approach.

> When the trigger function returned from a LazyQuery is called, it always initiates a new request to the server even if there is cached data. (https://redux-toolkit.js.org/rtk-query/api/created-api/hooks#uselazyquery)

This seems good to me (Thanks AI!). I’d prefer to avoid the semantic oddness.

packages/webapp/src/hooks/location/useLocations.ts
SayakaOno: I think by convention, const is used for variables that aren't reassigned.
This [ESLint page](https://eslint.org/docs/latest/rules/prefer-const) explains why. (I’d like to enable the rule so pre-commit can fix it automatically)

packages/webapp/src/hooks/location/useLocations.ts
```diff
@@ -168,7 +167,10 @@ function useLocations(
 function useLocations(
   { filterBy, groupBy, deleted }: UseInternalLocationProps = { deleted: false },
 ): any {
-  const { data: locations, isLoading, isFetching } = useGetLocationsQuery();
+  const { data: rawLocations, isLoading, isFetching } = useGetLocationsQuery();
+
+  // Deep clone to prevent mutating original data from cache
+  const locations = rawLocations?.map((rawLocation) => structuredClone(rawLocation));
```
SayakaOno: It seems like the result is the same as `structuredClone(rawLocations)`?

packages/webapp/src/store/api/locationApi.ts
SayakaOno: Sorry that I misunderstood. (What I meant was that if you take `farm_id` as an argument, each farm gets its own cache data, and when switching to a new farm, `isLocationsLoading` becomes `true`, which is irrelevant here sorry)

So simple question, why did you choose to not pass `farm_id` as an argument?

Thank you for not overcomplicating the PR!!

SayakaOno (commented):
I'm glad you're doing well!
Thank you for responding so quickly, especially with how busy you’ve been!
A few more follow-up questions 🙏

packages/webapp/src/hooks/location/useLocations.ts
Duncan-Brain: thanks! lol

packages/webapp/src/store/api/locationApi.ts
Duncan-Brain: @SayakaOno Can I try to make the case to keep it (for now) one more time? lol If not I can change it as you wish.

1) **This middleware check is 99% of what is done on the delete mutation.** So really it is a "dry-run" of the mutation -- and should be deleted altogether. The backend endpoint would likely be best served as returning a specific response if the delete fails instead of running a pre-flight -- but thats a UI question and more extra TODO work not relevant here. 

2) **The semantic oddness is less important to me than being foolproof,** the result will never be cached. If LazyQuery is used with the incorrect option, or plain Query is used **the cache could be served**. No need for future devs to think about cache lifetime at all as is (I don't have a firm grasp on the core value of LazyQueries yet myself since they are rare).

3) Oddness as an visual indicator for weird or "anti-patterns" - if it blends in we forget to fix it!

Maybe a renaming of it would help with the semantics `checkDeleteLocationMutation() >>> checkDeleteLocationBeforeDeleteMutation()` ? lol

packages/webapp/src/hooks/location/useLocations.ts
Duncan-Brain: fair enough! The eslint rule would be great -- I will mess this up again for sure.

packages/webapp/src/store/api/locationApi.ts
Duncan-Brain: It is not a convention to pass it anywhere yet. To my knowledge we have not decided anything regarding farm data partitions prs I have shared elsewhere. So right now I am treating it as others have described "edge case" and I don't worry about switching farms at all like all our other endpoints.

I think our location api is pretty fast that is is edge casey here.

packages/webapp/src/store/api/locationApi.ts
SayakaOno: I knew you would challenge that! lol

1&3. Understood!
2. I believe LazyQuery is also safe, since there’s no option to serve cached data on trigger. But I recognize the risk that someone might use useQuery, which can return cached data.

I would personally still prefer lazyQuery, but I can compromise this time 👍

SayakaOno (commented):
We're releasing early next week, so I'll approve and merge this right after.
Thank you!

packages/webapp/src/store/api/locationApi.ts
Duncan-Brain: Lol ! Thanks.. my AI mentioned there was a lazyQuery option called `preferCacheValue` that defaults to false so thats what I heard for that one. Agree it would probably be fine too!

Duncan-Brain:
Awesome thanks!

SayakaOno (approved):

PR #4087, opened 2026-03-13
LF-4672 (2) - Map Loader for rtk upgrade for locations

Duncan-Brain:
**** THIS PR BASE IS ON THE rtk upgrade pr"

**Description**

This updates the temporary loading screen in the other [PR](https://github.com/LiteFarmOrg/LiteFarm/pull/4083).

Figma link in Jira ticket.

Jira link: [LF-4672](https://lite-farm.atlassian.net/browse/LF-4672)

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

[LF-4672]: https://lite-farm.atlassian.net/browse/LF-4672?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/webapp/src/components/Loading/LoadingV2/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+    };
+  }, [isOpen, showDelay, hideDelay]);
+
+  const customStyles = isCompactSideMenu
+    ? {
+        width: 'calc(100vw + var(--global-compact-side-menu-width))',
+      }
+    : {
+        width: 'calc(100vw + var(--global-side-menu-width))',
+      };
```
SayakaOno, line 81: Could we change these to `paddingLeft: 'var(--global-compact-side-menu-width)'` and `paddingLeft: 'var(--global-side-menu-width)'`?
I think it would be best to handle it within the viewport.

packages/webapp/src/components/Loading/LoadingV2/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+  return (
+    <Backdrop
+      open={isVisible}
+      sx={{
+        ...customStyles,
+        top: 'var(--global-navbar-height)',
+        zIndex: (theme: any) => theme.zIndex.appBar + 2,
```
SayakaOno: Which part is `theme.zIndex.appBar`?

packages/webapp/src/components/Loading/LoadingV2/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+        backgroundColor: 'rgba(246, 251, 250, 0.80)',
+        pointerEvents: 'none', // ← remove if you want to block clicks
```
SayakaOno: Let's remove it!

packages/webapp/src/components/Loading/LoadingV2/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+  dataName,
+}: LoadingBackdropProps) {
+  const [isVisible, setIsVisible] = useState(false);
+  const timerRef = useRef<NodeJS.Timeout | null>(null);
+
+  useEffect(() => {
+    if (timerRef.current) {
+      clearTimeout(timerRef.current);
+      timerRef.current = null;
+    }
```
SayakaOno: This part is never reached because the cleanup always runs first and clears the timer before the effect re-runs. It can be removed :)

SayakaOno (commented):
I love the animation, thank you! Just a few minor comments :)

packages/webapp/src/components/Loading/LoadingV2/index.tsx
Duncan-Brain: Ai definitely did this and I recognized it so I didn't question it. My understanding is it is a material ui const (1100). While we don't use the appBar here... Material UI has ordered their constants according to what makes most sense from some ui perspective. 

So that said... when I make the Loading bar always visible I expected the give feedback drawer + menus to be higher... but it is not.. even though I checked the css in inspector and drawer z-index is 1110 and the loader is 1102. So it is weird. Any ideas? 😬

packages/webapp/src/components/Loading/LoadingV2/index.tsx
Duncan-Brain: I think one is on unmount the cleanup, and one is if a second timer gets scheduled before unmount.

packages/webapp/src/components/Loading/LoadingV2/index.tsx
Duncan-Brain: It seems the stacking in the top menu AppBar takes precedent over sub-elements z-index like drawer.. which I still don't understand.


To achieve the desired effect of TopBar menu drawers going over this loading screen, I did -2 instead of +2.

packages/webapp/src/components/Loading/LoadingV2/index.tsx
SayakaOno: > if a second timer gets scheduled before unmount

I believe this is covered by the cleanup function. Could you try testing it?

packages/webapp/src/components/Loading/LoadingV2/index.tsx
SayakaOno: I was initially wondering why a z-index was needed since the backdrop didn't cover the header, but now I see you were accounting for the feedback drawer and menus.

I confirmed that `AppBar` is used for the header (z-index 1100)! As long as the header has a higher z-index or the backdrop doesn't have one, drawers/menus will appear on top, but without a z-index, the three buttons at the bottom are displayed on top of the backdrop.

So I interpret `theme.zIndex.appBar - 2` as the highest value that still covers even high z-index content without overlapping the top menu/drawers!

It looks good 😁

packages/webapp/src/components/Loading/LoadingV2/index.tsx
Duncan-Brain: Ah yes the dangers of trusting ai lol.. I think it probably wrote this and then when I asked it your question it gave a reasonable sounding answer saying that it was correct lol. Sorry about that I will improve!

SayakaOno (approved):

PR #4105, opened 2026-04-01
LF-5217 Drawer container and Home integration

kathyavini:
**Description**

The main content of this PR is just the three new files within `/containers/FarmNotes/` and the Delete Confirmation Modal, but the style adjusts on components ended up being quite a lot too :) I'll note those inline anywhere they don't seem self-explanatory.

The containers should be pretty simple though. @SayakaOno I tried separating out the floating action button and putting just that in `<Home />`, but since its read indicator dot requires the farm note list, this required either repeating the GET query in both containers, or using render props. Letting the FAB live with the rest of the feature was therefore a little simpler. I guess the mental modal would be dropping `<FarmNotes />` into a view means dropping both the button and drawer together? I think that's okay... 😅

Other notes I will leave inline!

Jira link: https://lite-farm.atlassian.net/browse/LF-5217

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/webapp/public/locales/en/translation.json
```diff
@@ -1008,6 +1008,10 @@
     "ATTACH_PHOTO": "Attach a photo",
     "CANCEL": "Cancel",
     "COLLAPSE": "Collapse note",
+    "DELETE_CONFIRMATION": {
```
kathyavini, line 1011: I just made up these strings; I think Loïc will give us this modal on Tuesday!

Update: these strings got the OK from Loïc

packages/webapp/src/components/Drawer/style.module.scss
```diff
@@ -125,6 +125,8 @@
 }
 
 .title {
+  display: flex;
+  align-items: center;
```
kathyavini, line 129: To fix the close icon and text not being aligned. Same adjust added to modal's title and close button (below)

packages/webapp/src/components/FarmNotes/FarmNoteItem/styles.module.scss
```diff
@@ -168,9 +179,10 @@
 }
 
 .thumbnail {
-  width: 120px;
+  width: auto;
   height: auto;
-  object-fit: cover;
+  max-width: 175px;
+  max-height: 175px;
```
kathyavini, line 185: The variable portrait / landscape orientation I told Loïc would be impossible was actually as simple as setting both height and width to `auto` and constraining with the same max in both dimensions 😂 Oops!

packages/webapp/src/components/ImageUploadCapture/styles.module.scss
```diff
@@ -88,20 +114,17 @@
 }
 
 .preview {
-  width: 108px;
-  height: 154px;
-  // height: auto; check with Loïc for preference
-  object-fit: cover;
+  min-width: 108px;
+  max-width: 154px;
+  max-height: 154px;
```
kathyavini, line 119: The addition of the min-width here (compared to the `.thumbnail` class above) allows flex shrink down to the originally-defined portrait width, to reduce the amount the action buttons are pushed out of their container on a small mobile screen:

Image

Image

(Even at 108px the buttons are pushed out a bit in English -- probably even worse in other languages -- but I'm not sure we can do too much better than that. Reducing the image further makes the preview too small to be useful)

packages/webapp/src/containers/FarmNotes/index.tsx
```diff
@@ -0,0 +1,149 @@
...
+          )
+        }
+        desktopVariant={DesktopDrawerVariants.SIDE_DRAWER}
+        fullHeight
+        addBackdrop={false}
+        classes={{ desktopSideDrawerContainer: styles.sideDrawerContainer }}
+      >
+        {isFormOpen ? (
+          <FarmNoteFormContainer
+            key={isEditState ? formState.note.id : 'add'}
```
kathyavini: I particularly like the way Claude handled the remount here _upstream_ of the component defining `defaultValues` -- very clean form state handling imo. I'd like to use this again!

packages/webapp/src/components/Modals/ModalComponent/v3/styles.module.scss
```diff
@@ -33,7 +33,7 @@
   display: flex;
   width: 100%;
   justify-content: space-between;
-  align-items: flex-start;
+  align-items: center;
```
kathyavini, line 36: Fixing the same issue as the drawer title -- close button and text slightly mis-aligned

Before:
Image

After:

Image

SayakaOno: I think I avoided this because the title can wrap to two lines and the close button gets some space above it. But looking at it now, it doesn't bother me... what do you think?

Image

packages/webapp/src/containers/FarmNotes/index.tsx
```diff
@@ -0,0 +1,149 @@
...
+      dispatch(enqueueSuccessSnackbar(t('message:FARM_NOTE.SUCCESS.DELETE')));
+    } catch (error) {
+      console.error(error);
+      dispatch(enqueueErrorSnackbar(t('message:FARM_NOTE.ERROR.DELETE')));
+    }
+    setNoteToDelete(null);
+  };
+
+  const isFormOpen = formState !== null;
+  const isEditState = formState !== null && formState !== 'add';
```
SayakaOno: Could we use `formState.mode === 'edit'`?

packages/webapp/src/containers/FarmNotes/index.tsx
```diff
@@ -0,0 +1,149 @@
...
+          />
+        ) : (
+          <FarmNoteList
+            notes={farmNotes || []}
+            userDisplayNameMap={userDisplayNameMap}
+            currentUserId={userFarm?.user_id ?? ''}
+            onAddNote={() => setFormState('add')}
+            onEditNote={(note) => setFormState({ mode: 'edit', note })}
+            onDeleteNote={(note) => setNoteToDelete(note)}
+            onImageClick={(src) => setLightboxSrc(src)}
```
SayakaOno: I think arguments can be omitted here:
```
onDeleteNote={setNoteToDelete}
```

packages/webapp/src/containers/FarmNotes/index.tsx
SayakaOno: When do we benefit from this? It seems to work without the key, so I'm just wondering if I’m missing a case!

SayakaOno (commented):
It's working well and the code is very readable!
If I'm not mistaken, we need to use `useMediaWithAuthentication` when rendering images since they are private. Could you confirm? (I hope I set it up correctly...!)

I also found that putting just the button in `<Home />` would make things more complicated. Thank you for checking! 🙏

packages/webapp/src/components/Modals/ModalComponent/v3/styles.module.scss
kathyavini: Ah, that makes sense!

That image looks correct to me, but I think I have a consistent subjective preference for centre-alignment 😂

FWIW, v2 modal uses `align-items: center` too, so at the very least this would match the previous modals (I don't know how relevant that is, though; I doubt many titles wrap to two lines!)

packages/webapp/src/containers/FarmNotes/index.tsx
kathyavini: Thank you for spotting this ❤️  This one I have added into CLAUDE.md explicitly now, as it generates this pattern a lot and I usually miss it 😉

packages/webapp/src/containers/FarmNotes/index.tsx
kathyavini: Oh 😳 I loved this `id` based `key` so much, but in this feature there is no way at all to change state or content without unmounting the form anyway so nothing at all is needed 🤣 Elegant _and_ useless!

packages/webapp/src/containers/FarmNotes/index.tsx
kathyavini: Not without Type Error, but that's just because the shape of `FormState` is inconsistent between `'edit'` and `'add'`... which suddenly I don't like very much! So let me fix both of those 🙏

packages/webapp/src/components/ImageUploadCapture/index.tsx
```diff
@@ -46,25 +46,27 @@ export default function ImageUploadCapture({
   const deviceType = getDeviceType();
   const showTakePhoto = deviceType !== 'desktop';
 
-  const [previewUrl, setPreviewUrl] = useState(defaultUrl);
+  const [localUrl, setLocalUrl] = useState<string | null>(null);
   const [showFileSizeExceedsModal, setShowFileSizeExceedsModal] = useState(false);
   const dropContainerRef = useRef<HTMLDivElement>(null);
 
+  const previewUrl = localUrl !== null ? localUrl : defaultUrl;
```
kathyavini: This is handling how `defaultUrl` now is initially undefined until the `fetch()`  in `useMediaWithAuthentication` completes. The alternative (which also works) would have been a useEffect to re-synch `previewUrl` to the `defaultUrl` prop, but I think that particular use of useEffect (to synch props to component state) is a React code smell?

packages/webapp/src/containers/hooks/useMediaWithAuthentication.ts
```diff
@@ -0,0 +1,134 @@
...
+
+    fetchMediaUrls();
+
+    return () => {
+      subscribed = false;
+      if (blobUrl) {
+        URL.revokeObjectURL(blobUrl);
+      }
+    };
+  }, [fileUrlsKey]);
```
kathyavini, line 127: I wish that GitHub had registered this as a rename in order to show the diff, but this is a new addition; the hook did not previously re-fetch when `fileUrls` changed; apparently this situation was not previously relevant. In FarmNotes, having a `[]` in this dependency array meant the preview image became stale after it was updated via Edit. Alongside this, the cleanup on `blobUrl` was added.

Other hook changes:
- `extensionName` was removed from the props; it was never used in the hook body
- `responseType: 'arraybuffer',` was removed from the fetch config since this is an axios config property, [not a fetch() option](https://developer.mozilla.org/en-US/docs/Web/API/RequestInit) (would previously have been ignored)
- minor stylistic cleanup like moving from `.then()` to `await` within the `Promise.all()` and removing this pattern: `subscribed && setMediaUrl(fileUrl);`

packages/webapp/src/components/FarmNotes/FarmNoteItem/index.tsx
```diff
@@ -50,11 +52,15 @@ export default function FarmNoteItem({
   onImageClick,
 }: FarmNoteItemProps) {
   const { t } = useTranslation(['translation', 'common']);
+  const { mediaUrl: authenticatedImageUrl } = useMediaWithAuthentication({
+    fileUrls: note?.image_url ? [note.image_url] : [],
```
kathyavini, line 56: We could delay fetching the image until the note is expanded, e.g. here:

```ts
fileUrls: isExpanded && note?.image_url ? [note.image_url] : []
```

but I didn't like the idea of the `authenticatedImageUrl` being purged when the note was collapsed 🤔 I think there would probably be a workaround (some holding in state) but I thought maybe for now it's okay to just fetch all the images upfront when the Farm Note List is loaded, and maintain them regardless of toggle state? What do you think?

kathyavini (commented):
> If I'm not mistaken, we need to use useMediaWithAuthentication when rendering images since they are private. Could you confirm? (I hope I set it up correctly...!)

Thank you so much @SayakaOno!!! Minio had been giving me false confidence since of course there is no private/public distinction with that bucket!

Both `useMediaWithAuthentication` and `ImageUploadCapture` ended up needing a bit of update to work correctly in this feature -- hopefully not too much of a scope creep 🤞 and if you could kindly re-review 🙏

packages/webapp/src/components/ImageUploadCapture/index.tsx
```diff
@@ -46,24 +46,27 @@ export default function ImageUploadCapture({
   const deviceType = getDeviceType();
   const showTakePhoto = deviceType !== 'desktop';
 
-  const [previewUrl, setPreviewUrl] = useState(defaultUrl);
+  const [localUrl, setLocalUrl] = useState<string | null>(null);
   const [showFileSizeExceedsModal, setShowFileSizeExceedsModal] = useState(false);
+  const dropContainerRef = useRef<HTMLDivElement>(null);
+
+  const previewUrl = localUrl ?? defaultUrl;
```
kathyavini, line 53: This handles how `defaultUrl` now is initially undefined until the fetch() in useMediaWithAuthentication completes. The alternative (which also works) would be a useEffect to re-synch previewUrl to the defaultUrl prop, but I think that particular use of useEffect (to synch props to component state) is a React code smell?

packages/webapp/src/components/FarmNotes/FarmNoteItem/index.tsx
SayakaOno: Fetch all the images upfront sounds better to me!
But I completely forgot about thinking about offline...😢 `authenticatedImageUrl` will be broken after refresh when offline, so I think we need another solution replacing the hook. 
Let me confirm if images need to be available offline!

kathyavini: Oh no! I also forgot about offline. I think they almost certainly do need to be available offline so they can be viewed in the field...

Edit: could runtime caching be added to the digital ocean private droplet GET?

SayakaOno: I thought it should be available too!
I think runtime caching can be added, and it seems simpler than what I initially envisioned. I think we could just go for it 😂

Your changes look good, so should we merge this and open a new PR?

kathyavini: > so should we merge this and open a new PR?

Let's do that! Or does it belong on the offline ticket with the other service worker changes? I was thinking it needed the rest of the offline ticket to be tested anyway, but that's not true... this only needs the GET so it could stand alone.

Did you want to open a draft on it -- did you already get something working?

SayakaOno: Sounds good!
Claude provided some code that I believe will work, but I haven't tested it yet. I have two PRs to open before getting to it, so feel free to beat me!

kathyavini: Ugh it took too much fiddling with dev / minio but I finally got something into cache 😂  Was struggling for a while

Screenshot 2026-04-08 at 1 06 09 PM

SayakaOno (approved):

PR #4108, opened 2026-04-08
LF- 5213: Farm notes offline support

SayakaOno:
**Description**

- Update API error handling in the components
   - Prevent error snackbars from showing for network errors. (network errors are handled in the API slices)
- Add optimistic UI for farm notes and the farm notes unread dot
- Update `getFarmNotes` endpoint
   - When offline, `queryFn` returns the cached data to keep the query in "fulfilled" status; otherwise optimistic updates won't appear on the UI once the query moves to "rejected".
- Sync farm notes
  - Update `useServiceWorkerListener` to handle farm note routes, consolidating snackbar messages in one place.
  - Add farm note routes to `RETRY_ROUTES` in `sw.js`.

Jira link: https://lite-farm.atlassian.net/browse/LF-5213

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/public/locales/en/message.json
+++ b/packages/webapp/public/locales/en/message.json
@@ -90,8 +90,6 @@
   },
   "FARM_NOTE": {
     "CREATE": {
-      "FAILED": "Failed to create note",
-      "SUCCESS": "Successfully created note",
       "SYNC": {
         "NETWORK_ERROR": "Sorry, we are down momentarily. We will retry your note creation automatically",
         "ONLINE": "Note created. Will appear when you're back online",
@@ -99,8 +97,6 @@
       }
     },
     "DELETE": {
-      "FAILED": "Failed to delete note",
-      "SUCCESS": "Successfully deleted note",
       "SYNC": {
         "NETWORK_ERROR": "Sorry, we are down momentarily. We will retry your note deletion automatically",
         "ONLINE": "Note deleted. Will save when you're back online",
@@ -108,8 +104,6 @@
       }
     },
     "EDIT": {
-      "FAILED": "Failed to update note",
-      "SUCCESS": "Successfully updated note",
       "SYNC": {
         "NETWORK_ERROR": "Sorry, we are down momentarily. We will retry your note update automatically",
         "ONLINE": "Note updated. Will save when you're back online",
--- a/packages/webapp/public/locales/en/message.json
+++ b/packages/webapp/public/locales/en/message.json
@@ -125,6 +125,11 @@
       "ADD": "Successfully saved note",
       "DELETE": "Successfully deleted note",
       "EDIT": "Successfully updated note"
+    },
+    "SYNC": {
+      "FAILED": "Task changes failed to save",
+      "NOT_FOUND": "Note failed to save: note no longer exists",
+      "UNAUTHORIZED": "Note failed to save: insufficient permissions"
     }
   },
   "HELP_REQUEST": {
--- a/packages/webapp/public/locales/en/message.json
+++ b/packages/webapp/public/locales/en/message.json
@@ -89,6 +89,33 @@
     "SUCCESS_DISCONNECT_ADDON": "Successfully disconnected addon"
   },
   "FARM_NOTE": {
+    "CREATE": {
+      "FAILED": "Failed to create note",
+      "SUCCESS": "Successfully created note",
+      "SYNC": {
+        "NETWORK_ERROR": "Sorry, we are down momentarily. We will retry your note creation automatically",
+        "ONLINE": "Note created. Will appear when you're back online",
+        "SUCCESS": "Note created offline has been saved successfully"
+      }
+    },
+    "DELETE": {
+      "FAILED": "Failed to delete note",
+      "SUCCESS": "Successfully deleted note",
+      "SYNC": {
+        "NETWORK_ERROR": "Sorry, we are down momentarily. We will retry your note deletion automatically",
+        "ONLINE": "Note deleted. Will save when you're back online",
+        "SUCCESS": "Note deleted offline has been saved successfully"
+      }
+    },
+    "EDIT": {
+      "FAILED": "Failed to update note",
+      "SUCCESS": "Successfully updated note",
+      "SYNC": {
+        "NETWORK_ERROR": "Sorry, we are down momentarily. We will retry your note update automatically",
+        "ONLINE": "Note updated. Will save when you're back online",
+        "SUCCESS": "Note updated offline has been saved successfully"
+      }
+    },
     "ERROR": {
       "ADD": "Failed to save note",
       "DELETE": "Failed to delete note",
--- a/packages/webapp/public/locales/en/message.json
+++ b/packages/webapp/public/locales/en/message.json
@@ -89,27 +89,6 @@
     "SUCCESS_DISCONNECT_ADDON": "Successfully disconnected addon"
   },
   "FARM_NOTE": {
-    "CREATE": {
-      "SYNC": {
-        "NETWORK_ERROR": "Sorry, we are down momentarily. We will retry your note creation automatically",
-        "ONLINE": "Note created. Will appear when you're back online",
-        "SUCCESS": "Note created offline has been saved successfully"
-      }
-    },
-    "DELETE": {
-      "SYNC": {
-        "NETWORK_ERROR": "Sorry, we are down momentarily. We will retry your note deletion automatically",
-        "ONLINE": "Note deleted. Will save when you're back online",
-        "SUCCESS": "Note deleted offline has been saved successfully"
-      }
-    },
-    "EDIT": {
-      "SYNC": {
-        "NETWORK_ERROR": "Sorry, we are down momentarily. We will retry your note update automatically",
-        "ONLINE": "Note updated. Will save when you're back online",
-        "SUCCESS": "Note updated offline has been saved successfully"
-      }
-    },
     "ERROR": {
       "ADD": "Failed to save note",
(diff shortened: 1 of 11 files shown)


packages/webapp/src/containers/FarmNotes/FarmNoteForm/index.tsx
```diff
@@ -22,18 +22,14 @@ import PureFarmNoteForm, {
 } from '../../../components/FarmNotes/FarmNoteForm';
 import type { FarmNote } from '../../../store/api/types';
 import { enqueueErrorSnackbar, enqueueSuccessSnackbar } from '../../Snackbar/snackbarSlice';
+import { isNetworkError } from '../../../util/apiUtils';
 
 interface FarmNoteFormContainerProps {
   note?: FarmNote;
-  onSuccess: () => void;
-  onCancel: () => void;
```
SayakaOno, line 29: Renamed `onSuccess` and `onCancel` to a single `onClose` callback to make it clearer why it's called after a network error.

packages/webapp/src/hooks/useServiceWorkerListener/useServiceWorkerListener.ts
SayakaOno, line 1: Extracted snackbar messages and the function into a separate file. Let me know if I broke your original intent!

packages/webapp/public/locales/en/message.json
```diff
@@ -98,6 +98,26 @@
...
+        "NETWORK_ERROR": "Sorry, we are down momentarily. We will retry your note deletion automatically",
+        "ONLINE": "Note deleted. Will save when you're back online",
+        "SUCCESS": "Note deleted offline has been saved successfully"
+      },
+      "EDIT": {
+        "NETWORK_ERROR": "Sorry, we are down momentarily. We will retry your note update automatically",
+        "ONLINE": "Note updated. Will save when you're back online",
+        "SUCCESS": "Note updated offline has been saved successfully"
+      },
+      "FAILED": "Task changes failed to save",
```
kathyavini: "Note changes failed to save?"

packages/webapp/src/hooks/useServiceWorkerListener/useServiceWorkerListener.ts
kathyavini: Thank you for making it so much cleaner + more extensible ❤️  It had become truly behemoth since that original file!

packages/webapp/src/hooks/useServiceWorkerListener/utils.ts
```diff
@@ -0,0 +1,138 @@
...
+    successMessage: i18n.t('message:TASK.DELETE.SYNC.SUCCESS'),
+    errors: {
+      404: i18n.t('message:TASK.SYNC.NOT_FOUND'),
+    },
+    retryMessage: i18n.t('message:TASK.DELETE.SYNC.NETWORK_ERROR'),
+  },
+  'farm_notes.create': {
+    successMessage: i18n.t('message:FARM_NOTE.SYNC.ADD.SUCCESS'),
+    errors: {},
+    retryMessage: i18n.t('message:FARM_NOTE.SYNC.ADD.NETWORK_ERROR'),
```
kathyavini, line 115: Were you at all bothered by the task pattern being `<FEATURE>.<ACTION>.SYNC` and this one being `<FEATURE>.SYNC.<ACTION>`? 😅

I guess it's probably easier to group the SYNC messages together top-level so that should be our pattern going forward, right?

packages/webapp/src/store/api/farmNoteApi.ts
```diff
@@ -59,6 +76,38 @@ export const farmNoteApi = api.injectEndpoints({
...
+        const patchResult = dispatch(
+          farmNoteApi.util.updateQueryData('getFarmNotes', undefined, (draft) => {
+            draft.unshift(optimisticNote);
+          }),
+        );
+
+        try {
+          await queryFulfilled;
+        } catch (error: any) {
+          if (isNetworkError(error)) {
```
kathyavini, line 103: Edit: how did this comment lose it's formatting??! 😅 Lost its link too

In tasks, we  only show the to-be-synced snackbar when offline, not in the case of the API being down (which is also a network error). Here the to-be-synced when online would be shown in both cases. I think the [original concern from tasks](https://github.com/LiteFarmOrg/LiteFarm/pull/4005#discussion_r2718856905) (two snackbars) doesn't exist anymore now that the queue is always replayed manually, though? 🤔

Should we remove the `if (isOffline)` check in the task sagas now so these two features behave equivalently?

In an ideal world, I think in both places we would use the NETWORK_ERROR string if the user is online, and the ONLINE string if the user is offline. Do you think that's worth the refactor in tasks? Or is it too much of edge case? I don't love where tasks ended up... when the API is down you get:

1. optimistic update
2. no snackbar explaining it
3. and then (on replay of the queue) the series of "task ____ when offline" snackbars! 😬

Here you get the wrong (offline) snackbars, but it's still less confusing!

packages/webapp/src/store/api/farmNoteApi.ts
```diff
@@ -59,6 +76,38 @@ export const farmNoteApi = api.injectEndpoints({
...
+            dispatch(enqueueSuccessSnackbar(i18n.t('message:FARM_NOTE.SYNC.ADD.ONLINE')));
+          } else {
+            // Server error: rollback the optimistic updateconsole.error(error);
```
kathyavini: Is the `console.error(error)` meant to be active / on its own line?

packages/webapp/src/util/apiUtils.ts
```diff
@@ -0,0 +1,21 @@
...
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ */
+
+import { FetchBaseQueryError } from '@reduxjs/toolkit/query';
+
+export type QueryResult<T> = { data: T } | { error: FetchBaseQueryError };
+
+// In RTK Query, network errors result in undefined status or specific error structure
```
kathyavini: In which case were you able to create an error with `undefined` status? I saw both the offline case and the API down case matching the type of `FetchBaseQueryError` (the one imported in this file) which always had a status of 'FETCH_ERROR`!

Additionally, although this works in the component context (with `.unwrap`), when it's called from `onQueryStarted`, it is giving a false positive to every API error, because status is nested within error like this for a regular HTTP error:

Image

So regular API errors like 404 were also giving the sync snackbar (from `onQueryStarted`) and then the regular error snackbar from the component:
 
https://github.com/user-attachments/assets/512b196e-6462-4e4a-92c4-920519e7d296

I think the logic can be updated like this:

```ts
export const isNetworkError = (error: any) => {
  const status = error?.error?.status ?? error?.status;
  return status === 'FETCH_ERROR';
};
```
but again I wasn't able to identify an `undefined` status state (which this would not work for)...!

kathyavini (commented):
Thank you so much for taking on this development task, and for figuring out all the nuances of working with RTK Query for this!! Just a few hours of `pnpm build` + `pnpm preview` testing reminded me OFFLINE IS SO HARD! Thank you for establishing the patterns we can use going forward 🙏

The code looked excellent, and the only bug I could spot was in the `isNetworkError` function, which was giving false positive to all API errors.

packages/webapp/src/hooks/useServiceWorkerListener/utils.ts
SayakaOno: I think so; it made more sense to me.

The task pattern uses `<FEATURE>` as the top-level key, but in farm note, it's already nested under `SUCCESS`/`ERROR`, so adding `<FEATURE>` felt redundant (here's how [farm note looks with the task pattern](https://github.com/LiteFarmOrg/LiteFarm/pull/4108/changes/bfb200a60d3bae57a4851f70020cf9cf19c8738d#diff-16593f68e540ab548f72f1d1c237df53add16481bdf8965c5921e470448004b2)).
If I'm missing something, please let me know!

packages/webapp/src/store/api/farmNoteApi.ts
SayakaOno: 🫨 The component logs it, so it shouldn't be here. Thank you for catching 🙏🙏🙏

packages/webapp/src/store/api/farmNoteApi.ts
SayakaOno: Thank you for bringing this up, I didn't remember the discussion 🙏 

> In an ideal world, I think in both places we would use the NETWORK_ERROR string if the user is online, and the ONLINE string if the user is offline. 

Let me quickly update it!

packages/webapp/src/util/apiUtils.ts
SayakaOno: I'm so sorry, that was an AI hallucination I didn’t catch 😭
Thank you so much!!!

SayakaOno (commented):
Thank you so much for reviewing!
I addressed your comments and added `storeActivity('/', 'open_farm_notes')` to record the event in `offline_event_log` table.
I'd appreciate another review 🙏

packages/webapp/src/store/api/farmNoteApi.ts
kathyavini: Ah, the context aware-snackbars ARE so pleasing!! I love them.

kathyavini (approved):
Looks fantastic!! Thank you!

PR #4110, opened 2026-04-09
LF-5231: Fix unread indicator missing for notes fetched after seen at is set

SayakaOno:
**Description**

- Update the farm note API URL from `/farm_note` to `/farm_notes`.
- Rename `last_seen_at` in `farm_notes_read` table to `read_through`.
- Save latest other user note's `updated_at` as `read_through`.
- Avoid sending `markFarmNotesRead` request when notes have been read.

Jira link: https://lite-farm.atlassian.net/browse/LF-5231

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/api/db/migration/20260409180010_rename_farm_notes_read_last_read_at.js
+++ b/packages/api/db/migration/20260409180010_rename_farm_notes_read_last_read_at.js
@@ -13,28 +13,22 @@
  *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
  */
 
-import Model from './baseFormatModel.js';
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+export const up = async function (knex) {
+  return knex.schema.alterTable('farm_notes_read', (table) => {
+    table.renameColumn('last_read_at', 'read_through');
+  });
+};
 
-class FarmNotesReadModel extends Model {
-  static get tableName() {
-    return 'farm_notes_read';
-  }
-
-  static get idColumn() {
-    return ['user_id', 'farm_id'];
-  }
-
-  static get jsonSchema() {
-    return {
-      type: 'object',
-      required: ['user_id', 'farm_id', 'read_through'],
-      properties: {
-        user_id: { type: 'string' },
-        farm_id: { type: 'string' },
-        read_through: { type: 'string', format: 'date-time' },
-      },
-    };
-  }
-}
-
-export default FarmNotesReadModel;
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+export const down = async function (knex) {
+  return knex.schema.table('farm_notes_read', (table) => {
+    table.renameColumn('read_through', 'last_read_at');
+  });
+};
--- a/packages/api/db/migration/20260409180010_rename_farm_notes_read_last_read_at.js
+++ b/packages/api/db/migration/20260409180010_rename_farm_notes_read_last_read_at.js
@@ -19,7 +19,7 @@
  */
 export const up = async function (knex) {
   return knex.schema.alterTable('farm_notes_read', (table) => {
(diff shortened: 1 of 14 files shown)


packages/api/db/migration/20260409180010_rename_farm_notes_read_last_read_at.js
```diff
@@ -0,0 +1,34 @@
...
+ *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
+ */
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+export const up = async function (knex) {
+  return knex.schema.alterTable('farm_notes_read', (table) => {
+    table.renameColumn('last_read_at', 'read_through');
```
kathyavini: `read_through` is a bit ambiguous -- here it means "read (the notes) through to this particularly timestamp", but the meaning that actually came to mind for me first is "has read through the entire list". Maybe that's just me?

I think I would prefer `read_up_to` or `last_seen_note_time`? But if `read_through` makes more sense to you, then it's also okay.

packages/webapp/src/sw.js
```diff
@@ -132,16 +132,16 @@ const RETRY_ROUTES = [
     method: 'DELETE',
   },
   {
-    matcher: ({ url }) => url.pathname.includes('/farm_note'),
+    matcher: ({ url }) => url.pathname.includes('/farm_notes'),
     method: 'POST',
   },
   {
     // This matcher will include farm_note/{uuid} paths
```
kathyavini: Re: the comment, should this now be `/farm_notes/{uuid}` ?

packages/webapp/src/containers/FarmNotes/index.tsx
```diff
@@ -54,9 +54,17 @@ export default function FarmNotes() {
...
+  const readThrough = farmNotesRead?.read_through;
+  const latestOtherUserNote = farmNotes?.find((note) => note.user_id !== userFarm?.user_id);
+  const hasUnread =
+    !!latestOtherUserNote &&
+    (!readThrough || new Date(latestOtherUserNote.updated_at) > new Date(readThrough));
+
   const handleOpenDrawer = () => {
     setIsDrawerOpen(true);
-    markFarmNotesRead();
+    if (hasUnread) {
```
kathyavini, line 71: ❤️

packages/webapp/src/sw.js
```diff
@@ -132,16 +132,16 @@ const RETRY_ROUTES = [
...
-    matcher: ({ url }) => url.pathname.includes('/farm_note'),
+    matcher: ({ url }) => url.pathname.includes('/farm_notes'),
```
kathyavini: Oh, this was equally true for the original code, but I didn't catch it!

This will already match `/farm_notes_read` so the lower/dedicated matcher is never used. Perhaps narrow this one with the trailing slash (`/farm_notes/`) or expand the comment to be explicit?

kathyavini (dismissed):
Fantastic! The new logic makes total sense, and not calling the `farm_notes_read` endpoint every time the drawer is opened is 🔥

packages/webapp/src/sw.js
SayakaOno: Oh, thank you! I'll update the matcher!

packages/api/db/migration/20260409180010_rename_farm_notes_read_last_read_at.js
SayakaOno: Thank you for the suggestions! I wasn't sure about this although I went with one of Claude's suggestions. I'll update it to `read_up_to`!

SayakaOno (commented):
Thank you so much for paying attention to the details! 🙏❤️❤️ 
I'd appreciate it if you could check the updates!

kathyavini (approved):
Looks great!

PR #4113, opened 2026-04-10
LF-5237 Implement immediate validation for unsupported file formats on upload in Farm Note creation

kathyavini:
**Description**

Despite the ticket title, this fix is applied to all image uploads throughout app in this PR

There were two existing patterns:
1. No check for / warning about filetype. This was the pattern in `<ImagePicker />`'s delayed upload flow (i.e. Farm Image, Help Request image), and inherited by `<ImageUploadCapture />` (Farm Notes), which was based on this flow
2. A check for image MIME-type, with an immediate "Failed to upload attachments" snackbar if the file wasn't an image. This was in the original `<ImagePickerWrapper />` (crop images) and in the immediate upload flow of `<ImagePicker />` using `useImagePicker` (e.g. animals, market directory). This was fundamentally sound but the snackbar was unhelpful as it didn't explain the cause of the upload error

This PR:
- Creates a single validator function to consolidate the MIME-type checks into a single shared function. It's slightly more permissive than the original checks as it allows specific file extensions if a MIME-type is not present
- Creates a more helpful snackbar that directs towards uploading the right kind of file
- Uses the validator + the snackbar across the various image upload situations in app (should now cover all of them)

Jira link: https://lite-farm.atlassian.net/browse/LF-5237

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/public/locales/en/translation.json
+++ b/packages/webapp/public/locales/en/translation.json
@@ -1109,7 +1109,7 @@
     "OVER_255_CHARS": "Text cannot exceed 255 characters"
   },
   "HELP": {
-    "ATTACHMENT_LABEL": "Upload screenshot or file",
+    "ATTACHMENT_LABEL": "Upload screenshot",
     "EMAIL": "Email",
     "FEEDBACK_INVITATION": "Get help or give us feedback",
     "MESSAGE_LABEL": "Message",
--- a/packages/webapp/public/locales/en/translation.json
+++ b/packages/webapp/public/locales/en/translation.json
@@ -2228,6 +2228,7 @@
     "CLICK_TO_UPLOAD": "Click to upload",
     "DRAG_DROP": "or drag and drop",
     "REMOVE_IMAGE": "Remove Image",
+    "UNSUPPORTED_FILE_TYPE": "Please upload a PNG, JPG, GIF, or WebP image",
     "UPLOAD_IMAGE": "Upload Image"
   },
   "WAGE": {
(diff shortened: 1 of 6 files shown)


packages/webapp/public/locales/en/translation.json
```diff
@@ -1109,7 +1109,7 @@
     "OVER_255_CHARS": "Text cannot exceed 255 characters"
   },
   "HELP": {
-    "ATTACHMENT_LABEL": "Upload screenshot or file",
+    "ATTACHMENT_LABEL": "Upload screenshot",
```
kathyavini, line 1112: The Feedback Form DOES accept a file currently, and it will never go through imaginary so there is really no issue with uploading a file via the `<ImagePicker />` component (it will just get attached to the support email); however, 1) the image preview is broken in the case of a file:

Image

and 2) it doesn't really make sense to use the `<ImagePicker />` component for file upload!

I suspect what happened here was that Loïc requested the new upload _component_ when we revamped Feedback Form, and we had ignored or just overlooked that this new component is specific to images so probably shouldn't have been used (at least, without changing the label string as belatedly done here!)

According to Denis, file upload is not really used in support requests anyway. I think we even discussed this before as a team and had talked about restricting to images? In any case, that is what Denis suggested today when we discussed this ticket, so I have gone with that and updated the label accordingly.

SayakaOno (approved):
Looks good to me!

PR #4118, opened 2026-04-13
LF-5232 Fix farm notes modal not closing when clicking my profile icon or feedback button

kathyavini:
**Description**

This PR centralizes the visibility state for all the right side-drawers and the TopMenu Profile Menu into a single global value `activeDrawer`, so that a most a single drawer (or menu) can be opened at a time.

Each individual's drawer (or menu)'s open state is then derived from the global `activeDrawer`:

```ts
const isOpen = activeDrawer === drawerName;
```

The existing `AppUIContext` (which previously tracked `isFeedbackFormOpen`, `setIsFeedBackFormOpen`), was already perfect for holding this state, and a common exported hook (`useDrawerState`) can now replace localized `[isOpen, setIsOpen]` style state in each component.

### Extending the pattern to future drawers

Now to create a new drawer that participates in this mutually exclusive application state, register the name in the `appContext.ts` type:

```ts
type DrawerName = 'feedbackSurvey' | 'farmNotes' | /*... */ | 'newDrawer'
```

and interact with it like this in the component, destructuring only the needed values

```ts
const { isOpen, openDrawer, closeDrawer, toggleDrawer } = useDrawerState('newDrawer');
```

Jira link: https://lite-farm.atlassian.net/browse/LF-5232

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/contexts/appContext.ts
+++ b/packages/webapp/src/contexts/appContext.ts
@@ -35,3 +35,26 @@ export const useAppUIContext = () => {
     return AppUIData;
   }
 };
+
+export const useDrawerState = (drawerName: DrawerName) => {
+  const { activeDrawer, setActiveDrawer } = useAppUIContext();
+
+  const isOpen = activeDrawer === drawerName;
+
+  const openDrawer = () => setActiveDrawer(drawerName);
+
+  /**
+   * Safely closes this specific drawer.
+   * Can safely be used even as the onClose triggered by ClickAway listeners
+   * because it verifies it is still the active drawer before applying `null`.
+   */
+  const closeDrawer = () => {
+    setActiveDrawer((prev) => (prev === drawerName ? null : prev));
+  };
+
+  const toggleDrawer = () => {
+    setActiveDrawer((prev) => (prev === drawerName ? null : drawerName));
+  };
+
+  return { isOpen, openDrawer, closeDrawer, toggleDrawer };
+};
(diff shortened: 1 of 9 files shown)


packages/webapp/src/contexts/appContext.ts
```diff
@@ -35,3 +35,26 @@ export const useAppUIContext = () => {
...
+
+  const openDrawer = () => setActiveDrawer(drawerName);
+
+  /**
+   * Safely closes this specific drawer.
+   * Can safely be used even as the onClose triggered by ClickAway listeners
+   * because it verifies it is still the active drawer before applying `null`.
+   */
+  const closeDrawer = () => {
+    setActiveDrawer((prev) => (prev === drawerName ? null : prev));
```
kathyavini, line 52: The `useDrawerState` hook was actually created specifically to make it easier for components to use this functional setter in `closeDrawer()`. The [original implementation ](https://github.com/LiteFarmOrg/LiteFarm/pull/4118/changes/155c05609291051c706d006380dc1f0cdad62bce) which just had

```ts
setActiveDrawer(null)
```
on each close, produced a bug specific to the TopMenu Profile Menu, where clicking on a new drawer initiator (e.g. FeedbackForm) would trigger `setActiveDrawer('feedbackForm')` AND THEN also trigger the click-away listener on TopMenu, triggering its `onClose()` and thus `setActiveDrawer(null)`. The result was that the feedback form would not open until you clicked again.

Right now only the TopMenu is set up with a click-away listener such that this style of `onClose()` is necessary, but I figured putting it here would make it safer to consume this pattern in the future, in case other participants did have click away listeners set up eventually.

And then as a bonus this hook API rather cleaned up the component logic as well!

SayakaOno (approved):
Nice refactor! 🙌

PR #4120, opened 2026-04-14
LF-5219: Incorrect redirect to create farm page after changing password

SayakaOno:
**Description**

###  Fix
Prevent unnecessary Route re-renders when form submissions update Redux state

###  Problem
When users submit the password reset form (in `/password_reset`), the Routes component would unexpectedly re-render and redirect them to `/welcome`, where they should see a modal and then be redirected to `/farm_selection`.

###  Root Cause

8bb5402 (Implementing offline navigation logging, [v3.10.0](https://github.com/LiteFarmOrg/LiteFarm/releases/tag/v3.10.0) release on Feb 25, 2026)

The `useOfflineActivityLogger` hook was using `useSelector(userFarmSelector)`, which subscribes to all changes in the `userFarm` Redux state. When the form submission updated `user_id`:

https://github.com/LiteFarmOrg/LiteFarm/blob/5f2510f2192cdf05b690fdc08633745a4b251b12/packages/webapp/src/containers/PasswordResetAccount/index.jsx#L19-L22

https://github.com/LiteFarmOrg/LiteFarm/blob/5f2510f2192cdf05b690fdc08633745a4b251b12/packages/webapp/src/containers/PasswordResetAccount/saga.js#L28-L49

https://github.com/LiteFarmOrg/LiteFarm/blob/5f2510f2192cdf05b690fdc08633745a4b251b12/packages/webapp/src/containers/userFarmSlice.ts#L104-L106

it would:

1. Trigger `<App />` to re-render (parent component)
2. Cause `<Routes />` to re-render as a child
3. Re-evaluate routing logic unnecessarily and render `<OnboardingFlow />` which doesn't have `/password_reset`

https://github.com/LiteFarmOrg/LiteFarm/blob/5f2510f2192cdf05b690fdc08633745a4b251b12/packages/webapp/src/routes/index.jsx#L343-L344

4. Redirect the user unexpectedly

https://github.com/LiteFarmOrg/LiteFarm/blob/5f2510f2192cdf05b690fdc08633745a4b251b12/packages/webapp/src/routes/Onboarding.jsx#L198-L200

###  Solution
Made the offline activity logger's Redux subscription more granular by changing selector:
Use `currentFarmIdSelector` instead of `userFarmSelector`
 - Only subscribes to `farm_id` changes, not all userFarm state
 - Prevents re-renders from unrelated form submissions

I found this bug while testing the React Router upgrade and wanted to ship a fix as soon as possible.

Jira link: https://lite-farm.atlassian.net/browse/LF-5219

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/containers/userFarmSlice.ts
+++ b/packages/webapp/src/containers/userFarmSlice.ts
@@ -321,3 +321,7 @@ export const userDisplayNameMapSelector = createSelector(userFarmsByFarmSelector
     return userMap;
   }, {});
 });
+
+export const currentFarmIdSelector = createSelector(userFarmSelector, (userFarm) =>
+  'farm_id' in userFarm ? (userFarm?.farm_id as string) : undefined,
+);
--- a/packages/webapp/src/hooks/useOfflineActivityLogger.ts
+++ b/packages/webapp/src/hooks/useOfflineActivityLogger.ts
@@ -23,15 +23,15 @@ import {
   storeActivity,
   updateOfflineSessionOnline,
 } from '../util/offlineEventLogger';
-import { userFarmSelector } from '../containers/userFarmSlice';
+import { currentFarmIdSelector } from '../containers/userFarmSlice';
 import { useIsOffline } from '../containers/hooks/useOfflineDetector/useIsOffline';
 
 const useOfflineActivityLogger = () => {
   const isOffline = useIsOffline();
   const location = useLocation();
 
-  const farm: { farm_id?: string } = useSelector(userFarmSelector);
-  const farmIdRef = useRef(farm?.farm_id);
+  const farmId = useSelector(currentFarmIdSelector);
+  const farmIdRef = useRef(farmId);
 
   const flushLogs = async () => {
     const activities = getActivities();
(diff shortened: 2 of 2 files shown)


kathyavini (approved):
Ooooh, I would never have come across that on my own or traced the route cause...!

> and render `<OnboardingFlow />` which doesn't have /password_reset

I'm not great on the logic of the routes files, but you don't think this is also worth fixing on its own, separate from the re-rendering? Am I understanding correctly the re-render exposed a little bit of faulty routing logic in that branch?

  And also just checking my understanding that this would be something to watch for as we keep adding custom hooks to App.jsx (as has become a small pattern since the offline release!)... I don't think I was thinking at all about `<Routes />` re-rendering when writing selectors in those hooks 🫤

SayakaOno:
@kathyavini You’re right, the same thing happens when you go offline while the modal is shown.
The routing conditions seem quite complex, so I’ve been hesitant to update them 😅 I’ll create a bug ticket to make them more robust!
Ticket: https://lite-farm.atlassian.net/browse/LF-5256

PR #4124, opened 2026-04-16
LF-5242 Fix issue where Cross button automatically highlights the feedback icon

kathyavini:
**Description**

On iOS the tap's pseudo-click was was triggering a sticky `:hover` state on the element behind the close button. [MDN writes](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:hover) about mobile `:hover`:

> Note: The :hover pseudo-class is problematic on touchscreens. Depending on the browser, the :hover pseudo-class might never match, match only for a moment after touching an element, or continue to match even after the user has stopped touching and until the user touches another element.

Gating the `:hover` styles behind a

```scss
@media (hover: hover) {
```

has fixed this bug. @SayakaOno I see you have already known to use this media query in places! But I have not reached for it before. It probably would be a good pattern moving forward for _all_ our buttons. That said, Loïc gives us pretty subtle `:active` states that might not make any sense without the hover to contrast with. I have added this as a topic to next week's dev-design because I would like to get this pattern straight!

Jira link: https://lite-farm.atlassian.net/browse/LF-5242

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

Tested on iPadOS Safari and double-checked for regressions on Android phone.

**Checklist:**

### DIFF
--- a/packages/webapp/src/containers/FeedbackSurvey/styles.module.scss
+++ b/packages/webapp/src/containers/FeedbackSurvey/styles.module.scss
@@ -26,7 +26,15 @@
   background-color: var(--Colors-Accent---singles-Blue-light);
   @include svgColorFill(var(--Colors-Accent---singles-Blue-full));
 
-  &:hover {
+  @media (hover: hover) {
+    &:hover {
+      width: 48px;
+      background-color: var(--Colors-Accent---singles-Blue-full);
+      @include svgColorFill(var(--White));
+    }
+  }
+
+  &:active {
     width: 48px;
     background-color: var(--Colors-Accent---singles-Blue-full);
     @include svgColorFill(var(--White));


packages/webapp/src/containers/FeedbackSurvey/styles.module.scss
```diff
@@ -26,7 +26,15 @@
...
-  &:hover {
+  @media (hover: hover) {
+    &:hover {
+      width: 48px;
+      background-color: var(--Colors-Accent---singles-Blue-full);
+      @include svgColorFill(var(--White));
+    }
+  }
+
+  &:active {
```
kathyavini, line 37: This particular component [happens to have identical hover and active states](https://www.figma.com/design/eMffC0h3JPlWtzNttxJpil/Litefarm-Updated?node-id=6550-51368&m=dev), although I think that's generally more rare in the codebase?

SayakaOno (approved):
Wow sorry, I still can't quite remember what problems I fixed with `@media (hover: hover)`!
Using it for all buttons sounds like a great idea!

PR #4129, opened 2026-04-20
LF-5247: Add hover state and visual boundaries to delete note button

SayakaOno:
**Description**

- Add `secondary-red` and `link-btn` as options for the `<Button />`'s `color` prop.
- Add [stories](http://localhost:6006/?path=/story/components-button--link-btn).
- Replace farm note's delete `<TextButton />` with `<Button />`.

Jira link: https://lite-farm.atlassian.net/browse/LF-5247

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/FarmNotes/FarmNoteItem/styles.module.scss
+++ b/packages/webapp/src/components/FarmNotes/FarmNoteItem/styles.module.scss
@@ -229,17 +229,9 @@
 
   button {
     min-width: 0;
-    display: flex;
     flex: 1 1 0;
-    align-items: center;
-    justify-content: center;
-    gap: 4px;
-    font-size: 14px;
-    font-weight: 600;
-    color: var(--Colors-Neutral-Neutral-500);
 
     span {
-      @include fontFamily();
       @include truncateText();
     }
   }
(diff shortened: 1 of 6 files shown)


packages/webapp/src/components/FarmNotes/FarmNoteItem/styles.module.scss
```diff
@@ -252,9 +252,3 @@
     }
   }
 }
-
-/* ──── Focus outline styles ──── */
```
kathyavini, line 256: In this stylesheet, I think almost all the styles under the `.actions button` and its `span` could also be removed now? (the colour, fontFamily etc.?)

I would have thought the whole `button` section could be removed, actually, but when I tested it just now, the `flex: 1 0 0` appeared necessary 🤔 I guess that makes sense; base `<Button />` must be agnostic about flex!

kathyavini (commented):
Looks great!

I swear I have never seen `secondary-red` before 😂 I wonder if it is a new addition, or maybe I just missed it because it is in one of the Figma button component charts, but not the other!

packages/webapp/src/components/FarmNotes/FarmNoteItem/styles.module.scss
SayakaOno: Thank you Joyce!
I removed all styles except for `flex: 1 0 0` and the ones needed for text truncate. Could you please double check? 🙏

kathyavini (approved):
❤️

PR #4130, opened 2026-04-20
LF-5101 Migrate off of Google Maps drawing manager to Terra Draw

kathyavini:
**Description**

A functional ~but buggy~ draft of the migration to Terra Draw, generated from the [hosted plan ](https://lite-farm.atlassian.net/wiki/x/AoB1jg) with some refactor subsequently. Not fully sure about `coordinatePrecision: 9`... this came from the [Google Maps Terra Draw Demo code](https://developers.google.com/maps/documentation/javascript/examples/map-drawing-terradraw), but I think it's the default.

I want to keep browsing [the package GitHub issues](https://github.com/JamesLMilner/terra-draw/issues) in general in case someone did report this premature polygon closing (Edit: it wasn't really closing so much as a dropped `pointerup` event that halted drawing until a second click).

Update: I think the bug is fixed, but required a `pnpm patch` to `terra-draw`; explanation is in the readme.md in the patchfile. I didn't find an exact bug match, although I also observed this one: [[Bug] First click after finishing polygon not registering (Google Maps adapter)](https://redirect.github.com/JamesLMilner/terra-draw/issues/720).

Jira link: https://lite-farm.atlassian.net/browse/LF-5101

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/webapp/src/containers/Map/createShapeCapture.ts
```diff
@@ -0,0 +1,108 @@
...
+    const config = getTerraDrawModeConfig(locationType);
+    if (config.mode === 'polygon') {
+      draw.updateModeOptions<typeof TerraDrawPolygonMode>('polygon', { styles: config.styles });
+    } else if (config.mode === 'linestring') {
+      draw.updateModeOptions<typeof TerraDrawLineStringMode>('linestring', {
+        styles: config.styles,
+      });
+    } else {
+      draw.updateModeOptions<typeof TerraDrawPointMode>('point', { styles: config.styles });
+    }
```
SayakaOno, line 96: Could we simplify these?
`draw.updateModeOptions(config.mode, { styles: config.styles })`

packages/webapp/src/containers/Map/terraDrawStyles.ts
```diff
@@ -0,0 +1,95 @@
...
+};
+
+type PointConfig = {
+  mode: 'point';
+  styles: {
+    pointColor: HexColor;
+    pointWidth: number;
+    pointOutlineColor: HexColor;
+    pointOutlineWidth: number;
+  };
```
SayakaOno: I don't think we need `styles` for points. Can we drop it? (we would need to make some adjustment in `createShapeCapture.ts`)

packages/webapp/src/util/google-maps/terraFeatureToOverlay.ts
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  if (geometry.type === 'Point') {
+    const position = toLatLng(geometry.coordinates);
+    const overlay = new maps.Marker({
+      position,
+      map,
+      icon: (icons as Record<string, string | undefined>)[locationType],
+      draggable: true,
+      crossOnDrag: false,
```
SayakaOno: I think this should be replaced with `...drawingOptions?.markerOptions`.

SayakaOno (commented):
I don't see any bugs, and I like how it works! ❤️ 
Just a few minor cleanup suggestions :)

packages/webapp/src/containers/Map/createShapeCapture.ts
kathyavini: 🤔 Hmmm it's a lot more concise!

But I think the point of the branching is type safety on the specific config objects, which seems like it would be worth preserving, particularly if we want to use something beyond `style` in the future:

Screenshot 2026-04-23 at 3 10 31 PM

---

Edit/update: reading your next comment though, I think the point logical branch is useless, so that will clean it up a little, too!

packages/webapp/src/util/google-maps/terraFeatureToOverlay.ts
kathyavini: Thanks for catching that ❤️ 

Why did AI lose track of the pattern 2/3 of the way through..??! Sometimes so inexplicable!

kathyavini (commented):
Thank you @SayakaOno 🙏🙏 Those were good catches! After cleanup the code is actually pretty... reasonable, isn't it? Maybe even an improvement from the original drawing manager?!

I don't feel totally done / satisfied with the dropped clicks bug yet (I don't think our code is blameless 😂) but probably best to investigate alongside QA...

SayakaOno (approved):

PR #4133, opened 2026-04-21
LF-5254: Fix deleted notes reappearing after going online

SayakaOno:
**Description**

When farm notes are synced offline, `getFarmNotes` is triggered for each operation (update, delete, etc.). If a previous `getFarmNotes` request is still in-flight, RTK Query deduplicates — the second call subscribes to the in-flight request instead of firing a new fetch. Since the in-flight request was initiated before the latest sync completed, it can return stale data.

Depending on timing, one of the following is observed (~5% reproduction rate locally, higher on Beta):

- `getFarmNotes` is called once and returns correct data (no race)
- `getFarmNotes` is called once and returns stale data that includes the deleted note ← bug
- `getFarmNotes` is called twice and returns correct data on the second call

**Solution**

Abort any in-flight getFarmNotes request before firing a new one, ensuring the response always reflects the latest server state.

Jira link: https://lite-farm.atlassian.net/browse/LF-5254

**Type of change**

**How Has This Been Tested?**

Since the race condition is timing-dependent (~5% locally), the abort isn't easily observable without slowing things down. To verify, apply a commit that delays `getFarmNotes` by 2 seconds:

`git cherry-pick --no-commit 1a5b655f4` (https://github.com/LiteFarmOrg/LiteFarm/commit/1a5b655f4f7096827c72c555e398cf396bcef660)

With the delay applied, the network tab should show one canceled request followed by one successful one.

- [ ] Passes test case
- [x] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/farmNoteController.ts
+++ b/packages/api/src/controllers/farmNoteController.ts
@@ -40,6 +40,7 @@ const farmNoteController = {
           .where('updated_at', '>', new Date(Date.now() - 30 * 24 * 60 * 60 * 1000)) // last 30 days
           .orderBy('updated_at', 'desc');
 
+        await new Promise((resolve) => setTimeout(resolve, 2000));
         return res.status(200).json(notes);
       } catch (error: unknown) {
         console.error(error);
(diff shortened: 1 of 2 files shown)


kathyavini (approved):
Sorry, took me way too long to test this since my corrupt Redux state (oddly, on both desktop and mobile...) was causing all sorts of unrelated errors. I did at last get to see the cancelled request -- thank you for the helpful testing commit!

Thank you for the GitHub link too 🙇 Another intermittent bug that makes total sense now 🙏

PR #4136, opened 2026-04-22
LF-5248: Improve User Experience for Image Upload in Farm Notes by Early File Size Notification and UI Enhancements

SayakaOno:
**Description**

Compress camera-captured images exceeding 5MB before upload.

Jira link: https://lite-farm.atlassian.net/browse/LF-5248

**Type of change**

**How Has This Been Tested?**

1. Apply a temporary commit (8a302c2e4618c8779c4154e9e4a455f75625c1c6) that enables the compression for all images and logs the compression result to the console:
    ```
     git cherry-pick 8a302c --no-commit
     ```
    Then select or drag and drop an image over 5MB. The console should show the before/after size in MB.
2. To test on mobile, lower the threshold to 2MB to make it easier to trigger:
     In `packages/webapp/src/components/ImageUploadCapture/index.tsx`, replace `5e6` with `2097152` (2MB), then run `vite preview --host` and take a photo with the camera.

- [ ] Passes test case
- [x] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/api/src/middleware/image/uploadFarmNoteImage.ts
+++ b/packages/api/src/middleware/image/uploadFarmNoteImage.ts
@@ -69,7 +69,7 @@ export default function uploadFarmNoteImage() {
       console.error(error);
 
       const err = error as HttpError;
-      const status = err.status || err.code || 500;
+      const status = err.status ?? (typeof err.code === 'number' ? err.code : 500);
       return res.status(status).json({ error: err.message || err });
     }
   };
--- a/packages/webapp/src/components/ImageUploadCapture/index.tsx
+++ b/packages/webapp/src/components/ImageUploadCapture/index.tsx
@@ -30,6 +30,35 @@ import getDeviceType from '../../util/getDeviceType';
 import { isImageFile } from '../../util/validation';
 import styles from './styles.module.scss';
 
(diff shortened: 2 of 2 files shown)


kathyavini (commented):
I love it! The compression looks great to my eyes... couldn't even notice degradation with `Image compressed from 12.19 MB to 2.23 MB` (thank you for the testing commit!)

It's not at all from this change, but in testing it (with my local imaginary down) I fully crashed my API, have you gotten this one before?

Image

It was the call to imaginary post returning `err.code === 'ECONNREFUSED'` and trying to use that as a status code here:

```js
const status = err.status || err.code || 500; // <-- no status; string err.code
return res.status(status).json({ error: err.message || err });
```

We use that error pattern everywhere, and I think it's normally safe, but imaginary is on a separate droplet so not inconceivable that it could `ECONNREFUSED` or `ENOTFOUND` -- maybe worth protecting from that crash? What do think? It would just be making sure not to send a code that isn't numeric

```js
const status = err.status ?? (typeof err.code === 'number' ? err.code : 500);
```

SayakaOno:
I've added your suggestion. Good call, thank you!

kathyavini (approved):
❤️

PR #4139, opened 2026-04-23
LF-5265 Old farm's note data persists after switching farms to a newly invited farm

kathyavini:
**Description**

This PR makes sure that `clearOldFarmStateSaga` (with the included `invalidatesTags`) is called in two flows where it was either missing entirely, or not called due to early return:

1. On switching to a new-to-you (userFarm status 'Invited') farm
2. On routing to consent on an existing userFarm due to need to re-consent (version bump, role change)

The code in the original Jira ticket / Slack thread is taken from pathway (2), but it so happens that the behaviour we were looking at was (1), which made debugging a bit needlessly hard!

It's pretty clear in `containers/ChooseFarm/index.jsx` though, I just failed to look at it 😝

```js
const onProceed = () => {
  const farm = userFarmEntities[selectedFarmId][user_id];
  if (farm.status === 'Active') {
    // ...
    dispatch(selectFarmAndFetchAll({ farm_id: selectedFarmId })); // <-- the saga we looked at
  } else {
    dispatch(patchUserFarmStatusWithIDToken(farm)); // <-- the invited case we were testing
  }
};
```

it didn't help that there are a lot of pathways to `/consent` and these flows are a bit convoluted...

Jira link: https://lite-farm.atlassian.net/browse/LF-5265

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

#### Testing Invited flow
- Manually set farms to Invited status via database (`status` 'Invited', `has_consent` and `step_three` FALSE, `step_three_end` NULL)

#### Testing consented flow within `selectFarmAndFetchAll()`
- Toggled a role between worker & admin using a different device

#### (Bonus) Testing Redux behaviour of `clearOldFarmStateSaga`
(Not really necessary, but was the most interesting finding from my debug!)
- Place debugger in farm switch saga(s). (Pausing via debugger is necessary within selectFarmAndFetchAll only; the consent screen actually pauses in the right place when looking at patchUserFarmStatusWithIDToken
- clear Redux state (pause and resume) and initiate farm switch. Step through with debugger if applicable
- result: you can see each piece of invalidated farm tag state being REMOVED from the api Redux tree! (Or, equivalently, you can see that at the end of tag invalidation + removeQueryResult, only library queries are left!)

Screenshot 2026-04-23 at 1 19 06 PM

**Checklist:**

### DIFF
--- a/packages/webapp/src/containers/saga.js
+++ b/packages/webapp/src/containers/saga.js
@@ -690,6 +690,7 @@ export function* selectFarmAndFetchAllSaga({ payload: farm }) {
   try {
     yield call(selectFarmSaga, { payload: farm });
     const userFarm = yield select(userFarmSelector);
+    yield call(clearOldFarmStateSaga);
     if (!userFarm.has_consent) {
       // has_consent is derived in the userFarmSelector from DB has_consent && consent_version === CONSENT_VERSION
       // Reachable when CONSENT_VERSION was bumped and the user hasn't re-accepted, or when an admin has changed the user's role (userFarmController.updateRole resets has_consent but leaves status='Active')
@@ -697,7 +698,6 @@ export function* selectFarmAndFetchAllSaga({ payload: farm }) {
       return history.push('/consent');
     }
     history.push({ pathname: '/' });
-    yield call(clearOldFarmStateSaga);
     yield call(fetchAllSaga);
   } catch (e) {
     console.error('failed to fetch farm info', e);
--- a/packages/webapp/src/containers/saga.js
+++ b/packages/webapp/src/containers/saga.js
@@ -690,7 +690,12 @@ export function* selectFarmAndFetchAllSaga({ payload: farm }) {
   try {
     yield call(selectFarmSaga, { payload: farm });
     const userFarm = yield select(userFarmSelector);
-    if (!userFarm.has_consent) return history.push('/consent');
+    if (!userFarm.has_consent) {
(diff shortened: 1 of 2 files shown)


packages/webapp/src/containers/saga.js
```diff
@@ -690,9 +690,14 @@ export function* selectFarmAndFetchAllSaga({ payload: farm }) {
   try {
     yield call(selectFarmSaga, { payload: farm });
     const userFarm = yield select(userFarmSelector);
-    if (!userFarm.has_consent) return history.push('/consent');
-    history.push({ pathname: '/' });
     yield call(clearOldFarmStateSaga);
+    if (!userFarm.has_consent) {
+      // has_consent is derived in the userFarmSelector from DB has_consent && consent_version === CONSENT_VERSION
+      // Reachable when CONSENT_VERSION was bumped and the user hasn't re-accepted, or when an admin has changed the user's role (userFarmController.updateRole resets has_consent but leaves status='Active')
+      // Status 'Invited' farms do not use selectFarmAndFetchAllSaga, but instead use patchUserFarmStatusWithIdTokenUrl
```
kathyavini, line 697: Is it too much comment??! 😅 This is my reaction to how long I was confused by this NOT being the saga involved in selecting and switching to an invited farm! But maybe I overdid it with the comment length 😬

SayakaOno: I think it's helpful!

SayakaOno (approved):
Looks good, thank you!

PR #4152, opened 2026-04-29
LF-5248 (2) Adjust ImageUploadCapture image compression logic to handle more real world cases

kathyavini:
**Description**

This PR adjusts the compressorjs logic within `<ImageUploadCapture />` to account for the following real-world cases that were discovered through QA and further testing with files from various devices:

1. The compression loop was taking an unreasonable amount of time on a mid-range older Android device (see videos in Jira ticket description + discussion thread) with its default settings of 64MP / 25-30MB per image
2. 5MB was a very unrealistic target for an image produced from a typical device! Even on my phone which is set to an un-adjustable 12.5MP, it is possible to take photos (in the normal photo app) that are > 6MB and cannot subsequently be uploaded. On the devices I checked with adjustable resolution settings, JPEGs were ranging from 25MB to 40MB to 60MB+ (Samsung Galaxy S24 Ultra)
3. Some filetypes available from the normal image file picker were hitting the `Image compression failed` catch block. This was DNG (RAW) on Android, and HEIC on MacOS Chrome. Note that Safari can compress both, and Imaginary can always handle both (i.e. they are perfectly fine filetypes to send to backend)

To handle the above:
- the `compressorjs` logic was adjusted to mostly deal in filesize instead of compression (it does still do an 0.8 compression pass, but only one)
- `checkOrientation` was set to false due to [this warning from the readme](https://github.com/fengyuanchen/compressorjs#checkorientation) -- I never saw a memory crash, but it seemed safer to follow the guideline!
> If the size of the target image is too large (e.g., greater than 10 MB), you should disable this option to avoid an out-of-memory crash.
- all images are now compressed, including file picker uploaded ones, so that the 5MB limit can be increased to something that covers the default for more phones
  - I've tentatively set this to 50MB and would like to discuss; note that the 40MB image I tested with compressed down to less than 1MB but this took 7 seconds with throttling. Denis's test image took ~4 sec with throttling and compressed to 1.6MB. I have not seen compression output larger than 2MB with the currently committed settings
- Hitting the "Image compression failed block" now triggers the wrong filetype snackbar; I never saw compressorjs fail for any other reason in my testing

Jira link: https://lite-farm.atlassian.net/browse/LF-5248

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

There is a logging patchfile included in the Jira ticket that can be run on integration's code.

This patchfile can be applied to this branch to log the compression time and output size. What I realize now after watching Denis' video is that is that we should probably be doing more of is logging with CPU throttling, not just network throttling!
[UPDATED-diagnostic-logging-image-compression.patch](https://github.com/user-attachments/files/27186289/UPDATED-diagnostic-logging-image-compression.patch)

If interested, I can share a testing kit to Slack with some different files and sizes I used (HEIC, DNG, large JPEGs) to test the compression behaviour, but if you have your own from different devices, even better!

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/ImageUploadCapture/index.tsx
+++ b/packages/webapp/src/components/ImageUploadCapture/index.tsx
@@ -30,19 +30,14 @@ import getDeviceType from '../../util/getDeviceType';
 import { isImageFile } from '../../util/validation';
 import styles from './styles.module.scss';
 
+const COMPRESSION_THRESHOLD = 5e6; // 5MB
+const MAX_ACCEPTED_FILE_SIZE = 50e6; // 50MB for orignal file, matching high MP images from most phones
+
 const compressImage = async (file: File): Promise<Blob> => {
-  if (file.size <= 5e6) {
-    return file;
-  }
-
-  return compressImageWithQuality(file, 0.8); // best-effort compression
-};
-
-const compressImageWithQuality = async (file: File, quality: number): Promise<Blob> => {
   const Compressor = await import('compressorjs').then((compressor) => compressor.default);
   return new Promise((resolve, reject) => {
     new Compressor(file, {
-      quality,
+      quality: 0.8,
       maxWidth: 2560,
       maxHeight: 2560,
       checkOrientation: false,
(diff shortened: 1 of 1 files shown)


SayakaOno (approved):
The changes look good! Thank you for the write-up!

PR #4169, opened 2026-05-11
LF-5276: Save and return which crops or animals each expense is linked to models endpoints tests

SayakaOno:
**Description**
 
- Add `farmExpenseAnimalModel` and `farmExpenseCropVarietyModel`
- Update `farmExpense` relations
- Add `checkFarmExpense` middleware for validation
   - Create crop variety model method `cropVarietiesBelongToFarm`
   - Extract common utility functions from `checkSale`
- Update `farmExpenseController` to handle allocation modifications
- Add tests
- (Irrelevant to the ticket) Update GET response to return `[]` instead of `404` when there are no expenses
   - Adjust saga

PATCH sample requests:
- Delete existing animal expense
   ```
   { "farm_expense_animal": [] }
   ```
   or
   ```
   { "farm_expense_animal": null }
   ```
   or add `farm_expense_crop_variety`
   ```
   { "farm_expense_crop_variety": [...] }
   ```

Jira link: https://lite-farm.atlassian.net/browse/LF-5276

**Type of change**

**How Has This Been Tested?**

- [x] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/api/src/controllers/farmExpenseController.js
kathyavini: Not from this PR, but could this be moved inside the `try/catch`?

I crashed my backend when testing in Postman sending a non-existent `expense_type_id`; didn't expect that to bring the whole API down 😝

Additionally (for future safety) should `baseController.isDeleted` also return `record?.deleted`? Caller is not really set up to handle that case though (not deleted but not found)...

packages/api/src/middleware/validation/checkFarmExpense.ts
```diff
@@ -0,0 +1,202 @@
...
+  };
+}
+
+const hasInvalidAllocation = async (expense: ExpenseBody, expenseId: string | undefined) => {
+  const { value: newValue, farm_expense_animal, farm_expense_crop_variety } = expense;
+  const isNewExpense = expenseId === undefined;
+  const isAddingAnimalExpense = !!farm_expense_animal?.length;
+  const isAddingCropVarietyExpense = !!farm_expense_crop_variety?.length;
+
+  if ((isNewExpense || !newValue) && !isAddingAnimalExpense && !isAddingCropVarietyExpense) {
```
kathyavini: Since values can be `0`, I think it will have to be `newValue === undefined` rather than `!newValue` to avoid early return on checking the allocations for a value that's been updated to `0`.

packages/api/src/middleware/validation/checkFarmExpense.ts
```diff
@@ -0,0 +1,202 @@
...
+  getUniqueAnimalAndBatchIds,
+  hasInvalidAnimalIds,
+  hasInvalidBatchIds,
+} from '../../util/finance.js';
+import { HttpError, LiteFarmRequest } from '../../types.js';
+import { BaseProperties } from '../../models/types.js';
+
+interface AnimalExpenseItem {
+  id?: number;
+  farm_expense_id: number;
```
kathyavini: `farm_expense_id` is a uuid, but it looks like it's `number` in three of these types (and `string` in the 4th!)

packages/api/src/middleware/validation/checkFarmExpense.ts
```diff
@@ -0,0 +1,202 @@
...
+  animal_id?: number | null;
+  animal_batch_id?: number | null;
+  allocated_value: number;
+}
+
+interface CropVarietyExpenseItem {
+  farm_expense_id: number;
+  crop_variety_id: string | null;
```
kathyavini: Can the `crop_variety_id` ever be `null`?

packages/api/src/controllers/farmExpenseController.js
kathyavini: Not this PR, but this caused a server crash in my testing! This needs a `return` before the `res.status` or sending an object will crash with a

```
Error [ERR_HTTP_HEADERS_SENT]: Cannot set headers after they are sent to the client
```

I think the controller definitely needs fixing there anyway, but I wonder why Claude went with a more permissive pattern in the middleware that allows non-arrays, do you know? 🤔

```ts
// checkFarmExpense.ts
const expenses = Array.isArray(body) ? body : [body];
```

packages/api/src/controllers/farmExpenseController.js
```diff
@@ -108,8 +109,37 @@ const farmExpenseController = {
           return res.status(400).send('failed to patch data');
         }
 
+        const isAddingCropVarietyExpense = !!farm_expense_crop_variety?.length;
+        const isAddingAnimalExpense = !!farm_expense_animal?.length;
+
+        if (farm_expense_crop_variety !== undefined || isAddingAnimalExpense) {
```
kathyavini, line 115: I didn't understand the conditionals at first glance, but this is very clever! ❤️

packages/api/src/middleware/validation/checkSale.js
```diff
@@ -114,7 +117,7 @@ export function checkSaleBody(operation = 'add') {
         let batchIds = [];
 
         try {
-          ({ animalIds, batchIds } = getUniqueAnimalAndBatchIdsFromAnimalSale(animal_sale));
+          ({ animalIds, batchIds } = getUniqueAnimalAndBatchIds(animal_sale));
```
kathyavini: It's just for the error message, but looks like it needs the new 2nd parameter `entityName` here too

packages/api/src/middleware/validation/checkFarmExpense.ts
```diff
@@ -0,0 +1,202 @@
...
+          return res
+            .status(400)
+            .send('an expense cannot have both animal and crop variety allocations');
+        }
+
+        if (operation === 'add' && (expense.value === undefined || expense.value === null)) {
+          return res.status(400).send('value is required when creating a farm expense');
+        }
+
+        if (farm_expense_animal?.length) {
```
kathyavini, line 95: I think an array check will be needed here as well. `insertGraph` is happy to accept a single object as well as array, but since `length` is falsy the check won't run. So sending an object gets around the farm ownership check:

```js
"farm_expense_animal": {
  "animal_id": 45, // another farm's animal ID
  "allocated_value": 150
}
```

packages/api/tests/expense.test.js
```diff
@@ -212,6 +241,284 @@ describe('Expense Tests', () => {
...
+        });
+        const res = await postExpenseRequest(expense, {
+          user_id: owner.user_id,
+          farm_id: farm.farm_id,
+        });
+        expect(res.status).toBe(201);
+      });
+
+      test('Returns 400 if value is missing', async () => {
+        const { value: _value, ...expenseWithoutValue } = mocks.fakeExpense({
```
kathyavini, line 479: Did you get an unused variable linting error here? For me it's fine even without the rename to `_value` 🤔

kathyavini (commented):
Really nice! The middleware is very well organized and the controller logic looks great. Thank you for handling and thinking of so many required checks -- again I'm surprised by how much can go wrong when multiple farm entities are involved 😅

In testing in Postman I (accidentally 😬) found two points in the controllers that cause a full API crash; neither is from the new code and hopefully those are the only ones!

packages/api/src/controllers/farmExpenseController.js
SayakaOno: The request body is an array in `POST`, but an object in `PATCH`.
I initially wondered that too 😂

packages/api/src/middleware/validation/checkFarmExpense.ts
SayakaOno: No 😅 Thank you! 🙏

packages/api/src/middleware/validation/checkFarmExpense.ts
SayakaOno: Ohhh thank you!

packages/api/tests/expense.test.js
SayakaOno: AI did it, so I didn't check, but I'm not getting an error either. I guess I would if it were TypeScript though 🤔

SayakaOno (commented):
Thank you so much for catching all those errors!! I think I’ve addressed them all.
Could you please re-review? Thank you 🙏

kathyavini (approved):
Looks good!

PR #4170, opened 2026-05-14
LF-5274 Add animal batch selector with quantity and sale value to the revenue entry form

kathyavini:
**Description**

Completes the frontend for Animal-associated Revenues / Sales:

1. Adds the `AnimalSaleInputs` container and `AnimalSaleItem` components mirroring their crop versions
2. Updates the `CropSaleTable` shown within the transaction expandable  to a more general `EntitySaleTable` (very simple change; only needed adjust was a label)
3. Adds animal logical branches to `useTransactions` and the formatting functions `mapSalesToRevenueItems` (Transactions, Actual Revenue), and `mapRevenueFormDataToApiCallFormat`
4. Removes the filter on and completes the frontend (title + description) for the new Animal Sales default revenue type

This PR also makes a few visual adjusts that were present on Figma or discussed in dev-design:
- Notes field is now a Text Area with dynamic placeholder based on entity type
- Footer of `EntitySaleTable` has been updated as discussed
- The multi-selects for crops and animals have updated labels and placeholders relative to the original form

Additionally, this PR resolves the naming/folder structure issues discussed in #4150 as follows (after quite some rounds of iteration as you can see in the commit history 😜)
- `AnimalSaleInputs` and `CropSaleInputs` have been gathered under a new `Finances/EntitySaleInputs/` folder
- The default component export from that Folder `<EntitySaleInputs>` fetches and shows either the Crop or Animal inputs based on the enitity type (as AI had originally proposed)
  - However rather than being passed as an unconditional prop, `<EntitySaleInputs>` is simply imported directly into the Revenue Form
- Consequently, the form HAS been renamed from `<GeneralRevenue>` to `<RevenueForm>`
  - I think it's a good change overall, but I have to admit I still catch myself typing `GeneralRevenue` all the time out of habit! Hopefully only an issue right now since I've been working in these files for the last two weeks 🤞
  - Note: the `entitySaleDefaultValues` (only relevant for `RevenueDetails` / existing records) are still passed as a prop. This keeps them available for the initial `useForm` call, which doesn't update on re-render
- If you recall from the first pass refactor on the other branch, `CropSaleInputs` used to wrap an entity-agnostic container called `<EntitySaleInputs>` (which held the multi-select and its relationship to the following sale item rows); this structure and naming felt incorrect, because as a name, `CropSaleInputs` feels more like a child/type/peer of `EntitySaleInputs` rather than a parent of it. Now the inner component has been made strictly a proper component, is called `EntitySalePicker`, and resides within `components/Form/RevenueForm`
  - I apologize if you have created any sort of mental model based on the `EntitySaleInputs` naming from the previous PR, but I _think_ this will be better moving forward!
  
**Finally, there is a small adjust for the Storybook crash:**

```
ReferenceError: Cannot access 'loginSelector' before initialization at alertSlice.js:45:35
```

Claude (Opus -- Sonnet was clueless!) traced this crash to a circular/race import of the Redux store, specifically `userFarmSlice` importing from `util/` which itself imports from the (somewhat sketchy) `getFromReduxStore/` with a top-level store import. I think selecting from `userFarm` at two levels within the current flow made reloading on the `GeneralRevenue` (oops, I mean `RevenueForm` 😅) Story a particularly easy way to create this crash, although I think some other `useFarmSlice`-calling components in Storybook could potentially have triggered it as well. In any case, it requires loading the store via component import which can only happen in Storybook; in app the store is provided top-level and I can't imagine a situation where the circular import could matter... but I'm very curious to know if I'm wrong about that! 😁 @SayakaOno you said you saw it in App once?

It does seem to be fully resolved now and I haven't seen it since the import fix. (Fix was the trivial change of moving the utility that `userFarmSlice` needs outside of the default `utils/index.js` file).

Jira link: https://lite-farm.atlassian.net/browse/LF-5274

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/webapp/src/components/Finances/Transaction/ExpandedContent/EntitySaleTable.jsx
```diff
@@ -20,10 +20,10 @@ import history from '../../../../history';
 import styles from './styles.module.scss';
 import { createRevenueDetailsUrl } from '../../../../util/siteMapConstants';
 
-const getColumns = (t, mobileView, totalAmount, quantityTotal, currencySymbol) => [
+const getColumns = (t, titleLabel, mobileView, totalAmount, currencySymbol) => [
```
kathyavini, line 23: In addition to making the table general to both crops and animals by passing the column title, this PR removes the quantity sum and changes "DAILY TOTAL" --> "TOTAL" as discussed in dev-design.

Image

Note: only the mobile view of this table is ever used in app (i.e. the footer is always defined by `<FooterCell>`, and never by the `Footer` property of `getColumns`); therefore the changes to Footer here in `getColumns` are just for completeness; they don't have any visual effect in app.

packages/webapp/src/components/Forms/GeneralRevenue/styles.module.scss
```diff
@@ -1,129 +0,0 @@
-/*
```
kathyavini, line 1: It was not deleted, but rather renamed to `RevenueForm/styles.module.scss`, but I deleted so many unused CSS classes apparently the diffing algorithm could not figure out they were the same file!

packages/webapp/src/components/Forms/RevenueForm/styles.module.scss
```diff
@@ -0,0 +1,48 @@
+/*
```
kathyavini, line 1: Not a new file, but rather a rename of `GeneralRevenue/styles.module.scss`. See note above!

Most of the original SCSS was not used anywhere.

packages/webapp/src/containers/Finances/constants.js
```diff
@@ -33,11 +33,6 @@ export const UPDATE_SALE = 'UPDATE_SALE';
 export const DELETE_EXPENSE = 'DELETE_EXPENSE';
 export const SET_IS_FETCHING_DATA = 'SET_IS_FETCHING_DATA';
 
-export const REVENUE_FORM_TYPES = {
```
kathyavini, line 36: Wasn't actually used anywhere in app. Same for the `getRevenueFormType()` that goes along with this constant (which makes sense since there is only the one form 😉)

packages/webapp/src/components/CardWithStatus/TaskCard/TaskCard.jsx
```diff
@@ -33,7 +33,7 @@ export const taskStatusTranslateKey = {
 
 import { languageCodes } from '../../../hooks/useLanguageOptions';
 import { getIntlDate } from '../../../util/date-migrate-TS';
-import { getFirstNameWithLastInitial } from '../../../util';
+import { getFirstNameWithLastInitial } from '../../../util/getFirstNameWithLastInitial';
```
kathyavini, line 36: Change related to the Storybook crash, as described in the PR description.

packages/webapp/src/components/Forms/RevenueForm/index.jsx
```diff
@@ -87,6 +87,15 @@ const GeneralRevenue = ({
   const selectedRevenueType = revenueTypes?.find(
     (rt) => rt.revenue_type_id === selectedTypeOption?.value,
   );
+  const selectedEntityType = selectedRevenueType?.entity_type;
+  const isEntitySale = selectedEntityType === 'crop' || selectedEntityType === 'animal';
```
SayakaOno: The util functions could be used 😄

packages/webapp/src/components/Forms/RevenueForm/EntitySalePicker.tsx
```diff
@@ -94,27 +92,29 @@ export default function EntitySaleInputs({
...
-          disabledInput,
-        }),
-      )}
-    </>
+      <div className={styles.saleItemList}>
+        {selectedOptions.map((option) =>
+          children({
+            option,
+            system,
+            currency,
```
SayakaOno: `system` and `currency` don't seem to be necessary since `AnimalSaleInputs` and `CropSaleInputs` can pass them directly.

packages/webapp/src/containers/Finances/EntitySaleInputs/AnimalSaleInputs.tsx
```diff
@@ -0,0 +1,137 @@
...
+      system={system}
+      currency={currency}
+    >
+      {({ option, system, currency, disabledInput }) => (
+        <AnimalSaleItem
+          key={option.value}
+          animalName={String(option.label)}
+          entityId={String(option.value)}
+          system={system}
+          currency={currency}
```
SayakaOno, line 127: Related to the comment in `EntitySalePicker`, I don't think `EntitySalePicker` has to pass `system` and `currency` props to children since they can be passed directly.

packages/webapp/src/containers/Finances/EntitySaleInputs/AnimalSaleInputs.tsx
```diff
@@ -0,0 +1,137 @@
...
+      label: chooseIdentification(a),
+      value: generateInventoryId(AnimalOrBatchKeys.ANIMAL, a),
+    }));
+    const batchOptions = (animalBatches ?? []).map((b: AnimalBatch) => ({
+      label: chooseIdentification(b),
+      value: generateInventoryId(AnimalOrBatchKeys.BATCH, b),
+    }));
+    return [...animalOptions, ...batchOptions].sort((a, b) =>
+      String(a.label).localeCompare(String(b.label)),
+    );
```
SayakaOno, line 103: `CheckboxMultiSelect` has sorting logic, so this might not be necessary.

SayakaOno (commented):
Looks great! I appreciated `CustomFormChildren` being removed - I think it made things cleaner 😄 

Just curious, `AnimalSaleDetail` story doesn't render animal options or inputs, right?

> ReferenceError: Cannot access 'loginSelector' before initialization at alertSlice.js:45:35

I think I've seen this kind of error when I accidentally created an incorrect selector with `createSelector`, so it wasn't a real issue 🙇‍♀️

packages/webapp/src/containers/Finances/EntitySaleInputs/AnimalSaleInputs.tsx
kathyavini: The select's built-in sorting logic applies to the open dropdown, but not the pills. This sort alphabetizes the pills and also the `AnimalSaleItems` for an existing sale.

I'm not even 100% sure that alphabetical sorting (as opposed to the order in which the user entered the items) is what is desired, though! I copied that from crops, but I'd like to check that with Loic tomorrow in dev-design. The expanded content tables are in insertion order so I think maybe the pills and the children sale items should match insertion order too?

packages/webapp/src/components/Forms/RevenueForm/index.jsx
kathyavini: Kind of funnily, I started that refactor but AI (I think Gemini?) said that direct string compare was more readable??

kathyavini (commented):
Thank you for your comments @SayakaOno; I think they are each addressed now!

Thank you also for bringing up the alphabetical ordering of the selected entity items. Alphabetical re-sort in the from is an inherited pattern from Crop Sale, but I don't think we've discussed with Loïc if it's desirable. Personally, I'm starting to lean against it! I've added it to the agenda for tomorrow's dev-design 🙏

Edit: Also just asked Claude though who said we should go the opposite way and sort the EntitySaleTable entries alphabetically as well :)

> Just curious, AnimalSaleDetail story doesn't render animal options or inputs, right?

That's correct -- animal and batches come from the Redux store and the Storybook mock state has not been populated with them. For this I just maintained the pattern from Crop Sale (which also doesn't render crops, since the mock store lacks management plans). I left this situation as it is because AI advised that adding the data to the mock store (particularly for animals with its RTK Query data) would be "quite difficult"!

packages/webapp/src/containers/Finances/EntitySaleInputs/AnimalSaleInputs.tsx
SayakaOno: Ah right, I just realized how I used `options` for expenses is different from here...
I'll follow whatever is decided tomorrow - thank you!

packages/webapp/src/components/Forms/RevenueForm/index.jsx
SayakaOno: Really!? Did I replace them with a function just for myself? 😂
I saw you added the `isAnimalSale` function, so I thought, why not use it? 🤣

SayakaOno (dismissed):

packages/webapp/src/components/Forms/RevenueForm/index.jsx
kathyavini: No I liked them too; I don't know why I let the AI boss me around like that 😆

SayakaOno (approved):
Looks good!

PR #4176, opened 2026-05-20
LF-5284 Finance: Newly created transaction not visible under "Year To Date" filter until filter is changed

kathyavini:
**Description**

The initial calculation of dates that runs on first mount / first render of `useDateRange` was being skipped if `startDate` and `endDate` were provided, which means the date ranges could become indefinitely stale (i.e. would not be re-computed without logout or switching the date dropdown option).

For finances (unlike sensor readings) these values are persisted in Redux, meaning this initial mount `useEffect` action was never re-running.

Since relative dates ("Last week" etc.) need to be recomputed relative to the current date, this part of the conditional was removed. Therefore the dates will be calculated a bit more often with this code change, but this is definitely preferable to them going stale!

Jira link: https://lite-farm.atlassian.net/browse/LF-5284

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

I didn't really test, honestly. To test organically you would need to have not switched your date range selector over a date boundary, which I had not done. Maybe there is another way to mimic this situation but I was not that clever about it 😅

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/DateRangeSelector/useDateRange.ts
+++ b/packages/webapp/src/components/DateRangeSelector/useDateRange.ts
@@ -20,7 +20,9 @@ import { MONDAY, SUNDAY } from '../../util/dateRange';
 
 /**
  * Returns startDate and endDate in state (local or Redux).
- * If they are not defined initially, set them using the DateRange utility class.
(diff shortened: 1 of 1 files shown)


SayakaOno (approved):
Looks good to me!

PR #4179, opened 2026-05-21
LF-5277: Add crop/animal association section to expense add and edit forms

SayakaOno:
**Description**

Add `ExpenseEntitySection` to the expense forms. I had to try several form patterns to make validation work correctly. The final approach simplifies the form structure using an `allocations` field array shared between crop and animal types. Data is converted from DB format to form values before display and back before submission.

``` 
{ …, allocations: [], entityType: ‘crop’ | ‘animal’ | null }
```
 When `entityType` is switched or cleared, `allocations` is reset via `useFieldArray.remove`.

- Add `ExpenseEntitySection` component (`EntityAssociationToggle` + allocation inputs)
- Integrate `ExpenseEntitySection` into:
    -  `AddExpense` (`ExpenseItemInputs`)
        - Wrap the component with `FormProvider` instead of passing form props through `ExpenseItemsForType` ➡️  `ExpenseItemInputs`
        - [Story](http://localhost:6006/?path=/story/page-finance-addexpense--primary)
    - `PureExpenseDetail`
        - Wrap the component with `FormProvider`
        * Format expense data from DB and set it as default form values
        - [Story](http://localhost:6006/?path=/docs/page-finance-pureexpensedetail--docs)
- Fix `NumberInput` logic
- Update `CheckboxMultiSelect` to accept `noOptionsMessage`
    - [Story](http://localhost:6006/?path=/story/components-form-reactselect-checkboxmultiselect--with-custom-no-options-message)
- Update `AnimalSaleInputs`
    - Add “No options” message
    - Add “(removed)” label
- Update `CropSaleInputs`
    - Add “No options” message
    - Use crop varieties instead of managementPlans (Remove the no longer used selector)
    - Extract label generation logic into `formatCropVarietyLabel`
- Update `AddExpense` and `ExpenseDetail` containers
    - Pass `cropVarietyOptions` and `animalOptions` to pure components
    - Update `handleSubmit` to format allocations
- Add `cropVarietyOptionsSelector` and `animalOptionsSelector`
    - `animalApi` file is expected to contain animal APIs in the future

Jira link: https://lite-farm.atlassian.net/browse/LF-5277

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/Finances/ExpenseEntitySection/index.tsx
+++ b/packages/webapp/src/components/Finances/ExpenseEntitySection/index.tsx
@@ -124,7 +124,7 @@ function ExpenseEntitySection({
       return existingAllocation ?? { id, allocated_value: null };
     });
     setValue(activeFieldPath, newAllocations);
-    trigger(activeFieldPath);
+    // trigger(activeFieldPath);
   };
 
   return (
@@ -153,37 +153,42 @@ function ExpenseEntitySection({
               isDisabled={disabled}
             />
           </div>
-          {!disabled && selectedOptions.length > 0 && (
-            <div className={styles.allocationSummary}>
-              <span>{t('EXPENSE.ENTITY_SECTION.TOTAL_TO_ALLOCATE')}</span>
-              <Badge remaining={remaining} currency={currency} />
-            </div>
+
+          {selectedOptions.length > 0 && (
+            <>
+              {!disabled && (
+                <div className={styles.allocationSummary}>
+                  <span>{t('EXPENSE.ENTITY_SECTION.TOTAL_TO_ALLOCATE')}</span>
+                  <Badge remaining={remaining} currency={currency} />
+                </div>
+              )}
+              <Controller
+                key={activeFieldPath}
+                name={activeFieldPath}
+                control={control}
+                rules={{
+                  validate: (value) => {
+                    const sum = sumAllocated(value);
+                    return sum <= totalValue || t('EXPENSE.ENTITY_SECTION.EXCEEDED_ERROR');
+                  },
+                }}
+                render={({ fieldState }) => (
+                  <>
+                    {selectedOptions.map((option, index) => (
+                      <EntityAllocationInput
+                        key={option.value}
+                        inputName={`${activeFieldPath}.${index}.allocated_value`}
+                        option={option}
+                        currency={currency}
+                        disabled={disabled}
+                      />
+                    ))}
+                    {fieldState.error?.message && <Error>{fieldState.error?.message}</Error>}
+                  </>
+                )}
+              />
+            </>
           )}
-          <Controller
-            key={activeFieldPath}
-            name={activeFieldPath}
-            control={control}
-            rules={{
-              validate: (value) => {
-                const sum = sumAllocated(value);
-                return sum <= totalValue || t('EXPENSE.ENTITY_SECTION.EXCEEDED_ERROR');
-              },
-            }}
-            render={({ fieldState }) => (
-              <>
-                {selectedOptions.map((option, index) => (
-                  <EntityAllocationInput
-                    key={option.value}
-                    inputName={`${activeFieldPath}.${index}.allocated_value`}
-                    option={option}
-                    currency={currency}
-                    disabled={disabled}
-                  />
-                ))}
-                {fieldState.error?.message && <Error>{fieldState.error?.message}</Error>}
-              </>
-            )}
-          />
         </>
       )}
     </div>
--- a/packages/webapp/src/components/Finances/ExpenseEntitySection/index.tsx
+++ b/packages/webapp/src/components/Finances/ExpenseEntitySection/index.tsx
@@ -13,17 +13,7 @@
  *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
  */
 
-import {
-  Control,
-  Controller,
-  FieldErrors,
-  UseFormRegister,
-  UseFormSetValue,
-  UseFormTrigger,
-  UseFormUnregister,
-  UseFormWatch,
-  useFormState,
-} from 'react-hook-form';
+import { Controller, useFormContext } from 'react-hook-form';
 import { MultiValue } from 'react-select';
 import { useTranslation } from 'react-i18next';
 import clsx from 'clsx';
@@ -39,12 +29,6 @@ import { EXPENSE_ANIMAL, EXPENSE_CROP_VARIETY, VALUE } from '../AddExpense/const
 import styles from './styles.module.scss';
 
 interface ExpenseEntitySectionProps {
-  control: Control;
-  register: UseFormRegister<any>;
-  unregister: UseFormUnregister<any>;
-  watch: UseFormWatch<any>;
-  setValue: UseFormSetValue<any>;
-  trigger: UseFormTrigger<any>;
   fieldNamePrefix: string;
   cropVarietyOptions: SelectOption[];
   animalOptions: SelectOption[];
@@ -54,10 +38,7 @@ interface ExpenseEntitySectionProps {
 interface EntityAllocationInputProps {
   inputName: string;
   option: SelectOption;
-  register: UseFormRegister<any>;
-  formErrors: FieldErrors;
   currency: string;
-  onChange?: () => void;
   disabled?: boolean;
 }
 
@@ -68,13 +49,14 @@ const sumAllocated = (allocations?: { allocated_value: number | undefined }[]) =
 function EntityAllocationInput({
   inputName,
   option,
-  register,
-  formErrors,
   currency,
-  onChange,
   disabled,
 }: EntityAllocationInputProps) {
   const { t } = useTranslation();
+  const {
+    register,
+    formState: { errors },
+  } = useFormContext();
 
   return (
     <div key={option.value} className={styles.entityBlock}>
@@ -87,9 +69,8 @@ function EntityAllocationInput({
           required: true,
           valueAsNumber: true,
           min: { value: 0, message: '' },
-          onChange,
         })}
-        errors={getInputErrors(formErrors, inputName)}
+        errors={getInputErrors(errors, inputName)}
         disabled={disabled}
       />
     </div>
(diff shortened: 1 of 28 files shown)


packages/webapp/src/components/Form/NumberInput/index.tsx
```diff
@@ -60,8 +60,9 @@ export default function NumberInput<T extends FieldValues>({
   ...props
 }: NumberInputProps<T>) {
   const { field, fieldState, formState } = useController({ name, control, rules, defaultValue });
+  const rawInitialValue = field.value ?? get(formState.defaultValues, name) ?? defaultValue;
   const { inputProps, numericValue, increment, decrement, clear } = useNumberInput({
-    initialValue: field.value || get(formState.defaultValues, name) || defaultValue,
+    initialValue: rawInitialValue === '' ? NaN : rawInitialValue,
```
SayakaOno, line 65: `field.value` can be `''` (empty string) when RHF requires an empty initial state, but `useNumberInput` treats `''` as 0 via `isNaN` (`isNaN("") === false`). Converting to `NaN` here ensures the input displays empty instead. Using `??` instead of `||` prevents `0` from being incorrectly ignored.

packages/webapp/src/components/Finances/ExpenseEntitySection/index.tsx
```diff
@@ -0,0 +1,212 @@
...
+  disabled,
+}: ExpenseEntitySectionProps) {
+  const { t } = useTranslation();
+  const currency = useCurrencySymbol();
+  const { watch, getValues, setValue } = useFormContext();
+
+  const prefixField = (fieldName: string) =>
+    fieldNamePrefix ? `${fieldNamePrefix}.${fieldName}` : fieldName;
+
+  const allocationsFieldPath = prefixField(ALLOCATIONS);
```
kathyavini, line 92: Ahhh this is such a nice handling of only sometimes being within a field array! 😍

And the validation behaviour with the exceeded allocations is also soo good 👌

packages/webapp/src/components/Form/ReactSelect/CheckboxMultiSelect/index.tsx
```diff
@@ -200,6 +202,7 @@ export const CheckboxMultiSelect = ({
     <div ref={ref}>
       <Select
         options={sortedOptions}
+        noOptionsMessage={noOptionsMessage || (() => t('common:NO_OPTIONS'))}
```
kathyavini, line 205: ❤️

packages/webapp/src/containers/Finances/EntitySaleInputs/CropSaleInputs.tsx
```diff
@@ -74,33 +76,29 @@ export default function CropSaleInputs({ sale, disabledInput }: CropSaleInputsPr
   const { t } = useTranslation();
   const system = useSelector(measurementSelector);
   const currency = useCurrencySymbol();
-  const managementPlans = useSelector((state) =>
-    selectManagementPlansForSale(state, sale?.crop_variety_sale),
-  );
+  const farmCropVarieties = useSelector(cropVarietiesSelector);
 
   // Management plans determine which crop varieties are sale-eligible,
```
kathyavini: Oh a good example of a comment quickly becoming dead / junk, sorry. In fact it was a bad comment to start with because it was only in reference to moving off the old pattern 🙇 

Could you please remove it? 🙏

packages/webapp/src/containers/Finances/EntitySaleInputs/CropSaleInputs.tsx
```diff
@@ -74,33 +76,29 @@ export default function CropSaleInputs({ sale, disabledInput }: CropSaleInputsPr
...
     const tileDataById: Record<number, CropVarietySaleTileData> = {};
     const optionList: SelectOption[] = [];
-    for (const mp of managementPlans ?? []) {
-      if (!(mp.crop_variety_id in tileDataById)) {
-        tileDataById[mp.crop_variety_id] = {
-          crop_variety_name: mp.crop_variety_name,
-          crop_translation_key: mp.crop_translation_key,
-          crop_variety_photo_url: mp.crop_variety_photo_url,
+    for (const cv of farmCropVarieties ?? []) {
+      if (!(cv.crop_variety_id in tileDataById)) {
```
kathyavini: This `if` isn't necessary anymore now that `cropVarietiesSelector` is being used since there will only be one of each crop variety (unlike before with management plans).

Just curious -- what you you think about using the `cropVarietyOptionsSelector` here too though, and just building the tile data separately?

```js
const options = useSelector(cropVarietyOptionsSelector);
const farmCropVarieties = useSelector(cropVarietiesSelector);

const cropVarietyTileDataById: Record<number, CropVarietySaleTileData> = useMemo(
  () =>
    Object.fromEntries(
      (farmCropVarieties ?? []).map((cv) => [
        cv.crop_variety_id,
        {
          crop_variety_name: cv.crop_variety_name,
          crop_translation_key: cv.crop_translation_key,
          crop_variety_photo_url: cv.crop_variety_photo_url,
        },
      ]),
    ),
  [farmCropVarieties],
);
```

I like the idea of revenue and expense using the same selector! The only thing is that the `cropVarietyOptionsSelector` is not currently sorted (while the `optionList` here was), but I wonder if sorting in the selector would be fine / good anyway, because it would then match the sorted `animalsOptionsSelector`?

packages/webapp/src/store/api/animalApi.ts
```diff
@@ -0,0 +1,39 @@
...
+    ...animals.map((animal) => ({
+      value: generateInventoryId(AnimalOrBatchKeys.ANIMAL, animal),
+      label: generateSelectOptionLabel(animal),
+    })),
+    ...animalBatches
+      .map((batch) => ({
+        value: generateInventoryId(AnimalOrBatchKeys.BATCH, batch),
+        label: generateSelectOptionLabel(batch),
+      }))
+      .sort((a, b) => String(a.label).localeCompare(String(b.label))),
```
kathyavini: I'm guessing that the whole array was meant to be sorted? And not just the batches?

I think the difference is functionally silent in Expenses (due to `sortAllocations`) but you can just barely notice it in Sales, if you set up your animals right 😉

packages/webapp/src/store/selectors/animals.ts
```diff
@@ -0,0 +1,39 @@
+/*
```
kathyavini, line 1: If we follow the current codebase convention, `store/api/<entity>Api.ts` has so far been used for API endpoints -- I believe this one looks like it should be in `store/selectors/`? Like the `location.ts` there?

Edit: oh sorry I just belatedly re-read this part of your description:

> animalApi file is expected to contain animal APIs in the future

Does that mean that `selectors/location.ts` should be folded into the `api/locationApi.ts` as well? I don't know if I have a particularly strong preference but it would be nice to decide if selectors belong with the  endpoints or not, going forward.

packages/webapp/src/stories/Pages/Finance/PureExpenseDetail.stories.jsx
```diff
@@ -110,3 +113,27 @@ EditView.args = {
   buttonText: 'Update',
   ...args,
 };
+
+export const ReadOnlyWithCrops = Template.bind({});
```
kathyavini, line 117: Seeing the working Storybook makes me wish I had refactored further and just made `CropSaleInputs` and `AnimalSaleInputs` pure components 😝 It's so nice!

kathyavini (commented):
This is lovely on so many levels!! The new form addition looks and works perfectly, and the form <--> API logic is so clean and readable 🔥 And I love the new options selectors, thank you!

This is not for this PR, but did you talk to Loïc about what's expected in the Transaction List Expanded Content for these allocated expenses? I don't think I ever saw a design, but I don't love that there is no sign of allocation in the `<GeneralTransactionTable>`; that felt a bit jarring to add the breakdowns and then not see them again until re-opening the form:

Image

Image

I don't know how he would prioritize that compared to e.g. the Weight/Unit quantities, but it seems like it would be a nice addition if we have some extra time left before release...!

packages/webapp/src/containers/Finances/EntitySaleInputs/CropSaleInputs.tsx
SayakaOno: Oh sorry! Thank you for catching!!

packages/webapp/src/containers/Finances/EntitySaleInputs/CropSaleInputs.tsx
SayakaOno: I added sorting in `cropVarietyOptionsSelector` and updated the useMemo!

packages/webapp/src/store/selectors/animals.ts
SayakaOno: Did you look at Duncan's PR!?? I completely forgot about the new folder 😅 
I've moved this file there!

packages/webapp/src/store/api/animalApi.ts
SayakaOno: What a mistake 😥 Thank you so much for catching!

SayakaOno (commented):
Thank you so much for reviewing Joyce!!
I didn't ask Loïc about the allocations in the transaction list, but we can ask him on Wednesday!
Thank you for re-reviewing in advance 🙏

packages/webapp/src/store/selectors/animals.ts
kathyavini: Haha no I didn't 😂  I was actually looking for/remembering the old `sensorSelectors.ts` which I had once wrote a long time ago and put in `store/api/selectors/`. Actually I *just* now noticed that in addition to removing that file, the locations RTK Query upgrade moved that folder up a level into `store/selectors` 🤷

kathyavini (approved):
❤️

packages/webapp/src/store/selectors/animals.ts
SayakaOno: Ah, I see! I'm sorry I didn't notice the folder path had changed!!

PR #4186, opened 2026-06-05
LF-5318 Unsupported language user is redirected to welcome page after login despite having farms

kathyavini:
**Description**

I worked waaaay too long trying to really nail this bug down, so my brain is a bit fried now and I'm afraid in the end this explanation will STILL not be as clear as I had hoped! There are two diagnostic log commits included on the branch which I will describe below.

This is actually quite a few things across multiple files interacting, and race (and network) conditions involved. I'll try to organize it in a readable way!

## When was the bug introduced?
Starting from the commit that would have introduced this bug as a detectable issue, that would be the [change to i18n.ChainedBackend](https://github.com/LiteFarmOrg/LiteFarm/pull/4008), which I confirmed via manual testing before and after (note that a `pnpm install` is needed when switching between these two commits):

Commit | i18n backend | Result
-- | -- | --
a16fb6ad41e58e183ca8b737d46b78c3c444d936 (before LF-4950) | i18next-xhr-backend | Choose Farm
a2dffa088b110d506ffe1e7182d55ff777f9618e (after LF-4950) | ChainedBackend + HttpBackend | Welcome Screen

There is nothing wrong with that change; it is just altered the timing of how `i18n.changeLanguage` resolves, most likely from something that is fundamentally synchronous (with the old fallback) to something that resolves just a split second later. The timing change interacted with other parts of the app as described below.

This means that the issue has been on production since the [offline release](https://github.com/LiteFarmOrg/LiteFarm/releases/tag/v3.10.0). Unfortunately a) any user with a non-LiteFarm language in their `language_preferences` OR any user with a browser language to LF language mismatch would be affected, and b) due to the nature of the bug (blocking normal login) it's probably horrficially under-reported. We can check this against db logs next week, but it's likely that this would have caused many users to just bounce off the platform, rather than filing a support request (which you can't do unless you re-create a farm each time you try to log in!)

## When / why is i18n.changeLanguage called

There are at least three parallel calls, but I'll focus on the SSO flow one:

```js
// loginWithGoogleSaga
localStorage.setItem('litefarm_lang', user.language_preference); // <-- can be LF language or not
if (i18n.language !== getLanguageFromLocalStorage()) {
  i18n.changeLanguage(getLanguageFromLocalStorage());
}
```

You can group this call to `i18n.changeLanguage` into two categories, one of which I've simply removed in this PR

1. (REMOVED) user's language_preference is an unsupported language --> i18n.changeLanguage is called to change 'en' to 'en'

The comparison ` if (i18n.language !== getLanguageFromLocalStorage())` was introduced in https://github.com/LiteFarmOrg/LiteFarm/pull/3470 as a solution to keep the browser language from overriding your LiteFarm language preference. Unfortunately when we were testing this, I believe we completely forgot about the case of Google SSO language not existing as a LF language 😬😬😬. That wouldn't have yet produced the bug being addressed in this PR, but it was the source of the local storage language being overwritten, and the end of native language dates when logging in via SSO, as discussed in #4178. If your `user.language_preference` is not a LF language (~5% of users on prod), then `i18n.changeLanguage` will make an unnecessary `'en'` --> `'en'` call, which also writes to local storage. Note there is no network fetch associated with this because there are no such files that can be fetched.

2. (REMAINS) user's language_preference IS a supported language, but local storage language (set by navigator.language after logout / on fresh session) is different --> i18n.changeLanguage called, and fetches the new language upon login

This second pathway is why the easiest way to reproduce this bug is simply log in via SSO on an `'en'` brower, change language, log out, log back in. A minor point about this is that you can actually prevent the bug (at least in my testing) in this case, by throttling your network down to Slow 4g.  I believe it sorts out the race condition below.

## Cause of the redirect

Diagnostic logs are provided in the branch; the before state can be observed with

```
git checkout 9b1a777e79d4f4fc34835813cef7825d57fdd03d
```

This was a bit of a pain to diagnose because multiple AI models consistently hypothesized multiple wrong mechanisms, most particularly a failed `useTranslation()` call in `ChooseFarm` causing `App.jsx` teardown thanks to Suspense (not at all accurate or relevant, but the reason the diaganostic logs are included in ChooseFarm and App.jsx to rule this out conclusively). There was also some confusion and inconsistency on whether the fix requires `i18n.changeLanguage` being `yield`-ed (awaited), and whether the router mismatch (below) is problematic or not. I'm pretty sure all of these are red herrings and can be ignored.

What _is_ confirmed is that the the buggy push is from within the fallback section (`<RequireCondition />`) of `Onboarding.jsx`. This is what the tracking of the two possible `history.push('/welcome')` calls in the flow (marked in the logs as #A and #B, since there is one in ChooseFarm.jsx as well) refer to. The root issue appears to be just a minor missing conditional there.

Here are the logs for the original bug, showing the push to `/welcome` due to `hasUserFarms: 0` despite `farmsLoaded` still being false:

```
App RENDER { routerLocation: '/' }
OnboardingFlow render { routerLocation: '/', windowLocation: '/farm_selection', MISMATCH: true,  hasUserFarms: 0, farmsLoaded: false }
PUSH #B (RequireCondition) -> /farm_selection { windowPath: '/farm_selection', hasUserFarms: 0, farmsLoaded: false }
App RENDER { routerLocation: '/welcome' }
OnboardingFlow render { routerLocation: '/welcome', windowLocation: '/welcome', MISMATCH: false, hasUserFarms: 0, farmsLoaded: false }
App RENDER { routerLocation: '/welcome' }
OnboardingFlow render { routerLocation: '/welcome', ... }      <-- stuck here
```

## The two-part fix

The logs and the fix can be viewed together by checking out the second temporary commit:

```
git checkout 85a476e7fb3c316e4dd373840cf6e7be6f503529
```

The fix for the non-LF language SSO account is as described above; `i18n.changeLangauge` is simply never be called so there isn't that async re-render trigger that is causing the issues (same as for current users -- includng all of us testers, unfortunately -- who only use browser + LF in one language).

For farms that do need to fetch / change languages, in the end I simply adjusted `Onboarding.jsx` to make sure it didn't route to `/welcome` before the farms were loaded. I went back and forth on this part a lot because ultimately I couldn't decide if mis-matched router state (the situation prevented by `yield`-ing on `i18n.changeLanguage`) was really a problem. With the routing logic tightened up, I didn't notice any ill effects from it, even though the mismatch was still produced in the console logs, and there was a small performance improvement (not a lot, but noticable at throttled network speeds) in NOT yielding / awaiting. At the same time, merely `yield`-ing does also, independently, resolve the bug.

In other words, the solution ultimately committed to this this PR doesn't really try to tackle the race condition which leads to unpredictable routing order, but  makes it benign whichever way it resolves. Anyway, I think your knowledge of routing is much better than mine @SayakaOno so I think I have to just have to leave it here and defer to you on this one!!

Jira link: https://lite-farm.atlassian.net/browse/LF-5318

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

Used logs as included in the testing commits. Tested both legitamate (`en` --> `pa` on log) and pointless (`en` --> `en` due to `ja` in user table) change language calls on SSO. Also verified with invite to existing email, but not yet fresh email.

**Checklist:**

### DIFF
--- a/packages/webapp/src/containers/Callback/saga.js
+++ b/packages/webapp/src/containers/Callback/saga.js
@@ -65,6 +65,11 @@ export function* patchUserFarmStatusSaga({ payload }) {
     const { user: userFarm, id_token } = result.data;
     localStorage.setItem('id_token', id_token);
     localStorage.setItem('litefarm_lang', userFarm.language_preference);
+    const requestedLang = getLanguageFromLocalStorage();
+    const targetLang = i18n.options.supportedLngs?.includes(requestedLang) ? requestedLang : 'en';
+    if (i18n.language !== targetLang) {
+      i18n.changeLanguage(targetLang);
+    }
     i18n.changeLanguage(userFarm.language_preference);
     purgeState();
     yield put(acceptInvitationSuccess(userFarm));
@@ -75,7 +80,11 @@ export function* patchUserFarmStatusSaga({ payload }) {
       // and message === 'user does not exist
       console.log(e);
       localStorage.setItem('litefarm_lang', language);
-      i18n.changeLanguage(language);
+      const requestedLang = getLanguageFromLocalStorage();
+      const targetLang = i18n.options.supportedLngs?.includes(requestedLang) ? requestedLang : 'en';
+      if (i18n.language !== targetLang) {
+        i18n.changeLanguage(targetLang);
+      }
       history.push('/accept_invitation/sign_up', invite_token);
     } else if (e?.response?.status === 401) {
       const { email: currentEmail } = yield select(userFarmSelector);
--- a/packages/webapp/src/containers/Callback/saga.js
+++ b/packages/webapp/src/containers/Callback/saga.js
@@ -70,7 +70,6 @@ export function* patchUserFarmStatusSaga({ payload }) {
     if (i18n.language !== targetLang) {
       i18n.changeLanguage(targetLang);
     }
-    i18n.changeLanguage(userFarm.language_preference);
     purgeState();
     yield put(acceptInvitationSuccess(userFarm));
     yield put(startInvitationFlow(userFarm.farm_id));
--- a/packages/webapp/src/App.jsx
+++ b/packages/webapp/src/App.jsx
@@ -13,7 +13,7 @@
  *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
  */
 
-import { useState } from 'react';
+import { useEffect, useState } from 'react';
 import { matchPath, useLocation } from 'react-router-dom';
 import clsx from 'clsx';
 import { SnackbarProvider } from 'notistack';
@@ -31,6 +31,10 @@ import useOfflineActivityLogger from './hooks/useOfflineActivityLogger';
 
 function App() {
   const location = useLocation();
+  // TEMP welcome-trace: render-body log. Fires on every render of App with the current
+  // router location. Pairs with the mount/unmount log below to prove Suspense boundary A
+  // (main.jsx: <Suspense fallback={null}> wrapping <App/>) never tears App down.
+  console.log('[welcome-trace] App RENDER', { routerLocation: location.pathname }); // TEMP welcome-trace
   const [isCompactSideMenu, setIsCompactSideMenu] = useState(false);
   const [activeDrawer, setActiveDrawer] = useState(null);
   const FULL_WIDTH_ROUTES = [MAP_URL, ANIMALS_URL, SENSORS_URL];
@@ -40,6 +44,17 @@ function App() {
   useServiceWorkerListener();
   useOfflineActivityLogger();
   const { isLoaded } = useGoogleMapsLoader();
+
+  // TEMP welcome-trace: mount/unmount probe. If Suspense boundary A (above <App/> in
+  // main.jsx) ever caught a suspension from the side menu / Navigation, App would UNMOUNT
+  // and later REMOUNT. Seeing exactly one MOUNTED and no UNMOUNTED through the SSO ->
+  // /welcome repro proves boundary A never fired -> App's Suspense is not in the chain.
+  useEffect(() => {
+    console.log('[welcome-trace] App MOUNTED (boundary A did NOT tear App down)'); // TEMP welcome-trace
+    return () => {
+      console.log('[welcome-trace] App UNMOUNTED (boundary A caught -> App torn down)'); // TEMP welcome-trace
+    };
+  }, []);
 
   return (
     <div className={clsx(styles.container)}>
--- a/packages/webapp/src/App.jsx
+++ b/packages/webapp/src/App.jsx
@@ -13,7 +13,7 @@
  *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
  */
 
-import { useEffect, useState } from 'react';
+import { useState } from 'react';
 import { matchPath, useLocation } from 'react-router-dom';
 import clsx from 'clsx';
 import { SnackbarProvider } from 'notistack';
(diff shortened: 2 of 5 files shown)


packages/webapp/src/containers/Callback/saga.js
```diff
@@ -65,6 +65,11 @@ export function* patchUserFarmStatusSaga({ payload }) {
     const { user: userFarm, id_token } = result.data;
     localStorage.setItem('id_token', id_token);
     localStorage.setItem('litefarm_lang', userFarm.language_preference);
+    const requestedLang = getLanguageFromLocalStorage();
+    const targetLang = i18n.options.supportedLngs?.includes(requestedLang) ? requestedLang : 'en';
+    if (i18n.language !== targetLang) {
+      i18n.changeLanguage(targetLang);
+    }
```
kathyavini, line 72: @SayakaOno okay I did some further testing of this saga!

The `language` here can be an unsupported language; you just need to invite to the account that has that language preference. Logging here when accepting such an invite:

Image

~It's set in `userController.addInvitedUser`:~

```js
const { language_preference } = isUserAlreadyCreated ?? { language_preference: language };
```

EDIT: Sorry I take that back! It's set by reading the `user.language_preference` right above... I wasn't reading the controller code correctly 😓 Wait, why is it named as a boolean but it's a db record 😭

---

(Unrelated to the above and still true) But looking at it now... the following `i18n.changeLanguage()` in the saga should have been deleted too, right??? (The line immediately following this change.)

packages/webapp/src/containers/Callback/saga.js
```diff
@@ -75,7 +80,11 @@ export function* patchUserFarmStatusSaga({ payload }) {
       // and message === 'user does not exist
       console.log(e);
       localStorage.setItem('litefarm_lang', language);
-      i18n.changeLanguage(language);
+      const requestedLang = getLanguageFromLocalStorage();
+      const targetLang = i18n.options.supportedLngs?.includes(requestedLang) ? requestedLang : 'en';
+      if (i18n.language !== targetLang) {
+        i18n.changeLanguage(targetLang);
+      }
```
kathyavini, line 86: As for this code block though, unfortunately I just traced it and was able to set a `'null'` language in my user table 😞

Image

I think this line is not correct

```
localStorage.setItem('litefarm_lang', language);
```

combined with the fact that if you don't interact with the language dropdown when sending the invite, you don't create a value here. I think one or the other has to be changed.

packages/webapp/src/containers/Callback/saga.js
SayakaOno: Thank you for looking into it!
`isUserAlreadyCreated` being an object is tricky 😅 

And yes, I didn't notice it, but I agree that the line below should be deleted!

packages/webapp/src/containers/Callback/saga.js
SayakaOno: I was able to reproduce it too! But there are no such records in prod/beta? 🤔 
Anyway, thank you for catching!!

kathyavini:
@SayakaOno sorry I didn't see your comments right away for some reason!

It looks like the invite flow is dropping `language` in two separate places. It seems quite wrong right now, but given as it doesn't seem to have any impact on prod, I think maybe that fix can be left out of the patch candidate change? What do you think?

As for the two Onboarding conditional branches, I am able to enter the second in the (admittedly contrived) case of being on the `/farm_selection` page and then clicking on an invite link in a different tab. That seems... correct, right? It could stay for that case?

SayakaOno (approved):
@kathyavini

> I think maybe that fix can be left out of the patch 

I agree! Maybe we can ship a fix in the upcoming release? `'null'` doesn't seem ideal...


> As for the two Onboarding conditional branches, I am able to enter the second in the (admittedly contrived) case of being on the /farm_selection page and then clicking on an invite link in a different tab. That seems... correct, right? It could stay for that case?

Yes, that's correct! It's hard to tell which scenarios end up in which route, so I was wondering if it could be cleaned up... but it seems to require some proper refactoring!


This is good to merge, right? If so, please go ahead and merge! 🙏

PR #4190, opened 2026-06-09
Backport 3.11.1

SayakaOno:
Backport 3.11.1


packages/api/src/controllers/userFarmController.js
kathyavini, line 416: The version numbers make sense but why is there a diff here? 🤔

SayakaOno: I had no idea until I asked AI, but it should be because of cherry-picking...!

kathyavini (approved):

PR #4194, opened 2026-06-11
LF-5333 Repair Ensemble sensor list fetching June 2026

kathyavini:
**Description**

Changes required to restore fetching of sensors and sensor arrays to LiteFarm from valid organisations:

1. The failure to authenticate messages we have been getting from Ensemble were due not to Ensemble, but to OUR database!

We store the tokens to avoid unnecessary calls to the authentication API, but the recent addition of a new field to the tokens pushed the character length just over 300, and we stopped being able to save them. The fact that this error can only be created in app (and not via Postman) made me think the authentication endpoint had been down momentarily when in fact the authentication endpoint was always fine; here our DB flow was at fault.

(This uncovered a minor bug, pre-existing, where the refresh token was not saved correctly. I'll explain that one in context)

2. In addition to the structure of the access token, Ensemble updated the `profiles/` endpoint relative to last year, and the shape of the profiles return. Now the previous return is specified in `results`:

```ts
{
  count: 2,
  next: null,
  previous: null,
  results: [
    {
      id: 29,
      description: "South",
      zone: null,
      // ...
    }
    {
     //...
    }
  ],
}
```

(AI informs that this is a **Django REST Framework paginated envelope: { count, next, previous, results }**) Other than that, it seems to contain the same data.

_I did not investigate the irrigation prescription or sensor data endpoints yet  -- they don't seem to be working but Ensemble said there was no change there, so I will follow-up with them._

Jira link: https://lite-farm.atlassian.net/browse/LF-5333

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

I'll message in Slack with the org ID returning sensors.

**Checklist:**

### DIFF
--- a/packages/api/src/util/ensemble.js
+++ b/packages/api/src/util/ensemble.js
@@ -513,7 +513,7 @@ async function refreshTokens() {
     await AddonPartnerModel.patchAccessAndRefreshTokens(
       ENSEMBLE_BRAND,
       response.data?.access,
-      response.data?.access,
+      response.data?.refresh,
     );
     return response.data;
   } catch (error) {
@@ -541,7 +541,7 @@ async function authenticateToGetTokens() {
     await AddonPartnerModel.patchAccessAndRefreshTokens(
       ENSEMBLE_BRAND,
       response.data?.access,
-      response.data?.access,
+      response.data?.refresh,
     );
     return response.data;
   } catch (error) {
--- a/packages/api/src/util/ensemble.js
+++ b/packages/api/src/util/ensemble.js
@@ -77,10 +77,10 @@ const getEnsembleSensors = async (farm_id) => {
 
   const farm = await FarmModel.query().findById(farm_id);
   const farmCenterCoordinates = farm.grid_points;
-
   const sensorArrays = [];
   for (const system of orgSystems) {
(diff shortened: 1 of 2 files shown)


packages/api/src/util/ensemble.js
```diff
@@ -513,7 +513,7 @@ async function refreshTokens() {
     await AddonPartnerModel.patchAccessAndRefreshTokens(
       ENSEMBLE_BRAND,
       response.data?.access,
-      response.data?.access,
+      response.data?.refresh,
```
kathyavini, line 516: The result of this random error (pre-existing) is that we store access tokens in both the access and refresh token fields of our database.

When hitting this endpoint `/token/refresh/` with what we think is a "refresh" token, we therefore get a 401 with

```json
{
    "detail": "Token has wrong type",
    "code": "token_not_valid"
}
```

But because 401 `isAuthError(error)` it just calls `authenticateToGetTokens()`, which works.

So overall this bug didn't have much effect beyond calling `authenticateToGetTokens()` more often than necessary (refresh tokens have 7 day expiry).

SayakaOno (approved):
Looks great! Thank you!!

PR #2577, opened 2023-04-12
Lf 2964 augment varietal read only view

MwayaB:
**Description**

Updated the header on the crop (read-only) hybrid varietal 

Jira link: https://lite-farm.atlassian.net/browse/LF-2964

**Type of change**

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

**How Has This Been Tested?**

https://lite-farm.atlassian.net/wiki/spaces/LITEFARM/pages/1287684104/LF-2964+Test+plan

- [x] Passes test case
- [x] UI components visually reviewed on desktop view
- [x] UI components visually reviewed on mobile view
- [ ] Other (please explain)

**Checklist:**

- [ ] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] The precommit and linting ran successfully
- [ ] I have added or updated language tags for text that's part of the UI
- [ ] I have added "MISSING" for all new language tags to languages I don't speak
- [ ] I have added [the GNU General Public License](https://lite-farm.atlassian.net/l/cp/BT0Dd7WW) to all new files

PR #4196, opened 2026-06-15
LF-5297 Incorrect labels displayed when creating and viewing crops and crop varieties

kathyavini:
**Description**

Adjusts the following:

- Form header updated from 'Add a crop' to 'Add a crop variety' when accessing that form page directly from "Add a new variety" link. The header remains 'Add crop' when this page appears as the second page in the 'Add crop' flow and the (already existing) `isNewCrop` value is passed to disambiguate these two cases.

Add a crop variety header

- Header that incorrectly read `<crop name> variety` is corrected to `{{ crop_name }} varieties`. This was actually an error in the ENGLISH STRING (the key is correctly pluralized 'VARIETIES' but 'variety' was used). This error was propagated to all languages and had the following effects:
  - extremely confusing UX: you click on a crop tile, and are taken to the list of all varieties for that crop, but the header suggests you have somehow landed on a single variety!
  - without context, the Punjabi + Malayalam translators picked a word that means "diversity" (a synonym for 'variety') rather than the word that means varieties of the crop. I used AI to query the right term for those and I will fix manually on CrowdIn, since I don't think PA and ML are actively translated anymore (ditto German)

- The so-called `<CropHeader>` that is only shown at the top of pages referring to crop varieties now shows the Crop Variety name in the header since it was _never shown_. It appears to have been lost, via Product decision, in the redesign [implemented here](https://github.com/LiteFarmOrg/LiteFarm/pull/2577). But this makes no sense. The photo is the crop _variety_ and so is the page content. So I am now showing the Crop Variety in that header with Crop as fallback (since Crop Varieties can be empty strings). I consider this a more minor issue than the two above.
- A few dev cleanups: component imports in the routes file adjusted to match the component names (incorrectly named as Add/Edit Crop for containers that are actually Add/Edit Crop Variety) and cleaning up some longstanding console warnings on the Crop Catalogue page (will indicate in-line)

Jira link: https://lite-farm.atlassian.net/browse/LF-5297

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/CropTile/index.jsx
+++ b/packages/webapp/src/components/CropTile/index.jsx
@@ -101,6 +101,7 @@ PureCropTile.propTypes = {
     active: PropTypes.number,
     planned: PropTypes.number,
     completed: PropTypes.number,
+    noPlans: PropTypes.number,
   }),
   needsPlan: PropTypes.bool,
   title: PropTypes.string,
--- a/packages/webapp/src/containers/CropCatalogue/index.jsx
+++ b/packages/webapp/src/containers/CropCatalogue/index.jsx
@@ -15,10 +15,8 @@ import { useHistory } from 'react-router-dom';
 import { getCropsAndManagementPlans } from '../saga';
 import CropCatalogueFilterPage from '../Filter/CropCatalogue';
 import {
-  cropCatalogueFilterDateSelector,
   cropCatalogueFilterSelector,
   isFilterCurrentlyActiveSelector,
-  setCropCatalogueFilterDate,
   resetCropCatalogueFilter,
 } from '../filterSlice';
 import { isAdminSelector } from '../userFarmSlice';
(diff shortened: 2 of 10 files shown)


packages/webapp/src/containers/CropCatalogue/index.jsx
```diff
@@ -130,8 +125,6 @@ export default function CropCatalogue() {
             <CropStatusInfoBox
               status={{ active, abandoned, completed, planned, noPlans }}
               style={{ marginBottom: '16px' }}
-              date={date}
```
kathyavini, line 133: To fix this console warning:

Image

`<CropStatusInfoBox />` doesn't do anything with date or setDate. The commented out code below (in `CropVarieties/index.jsx`) suggests the props were meant to be removed at that time... but something got a little confused in that refactor 🙂

packages/webapp/src/components/CropTile/index.jsx
```diff
@@ -101,6 +101,7 @@ PureCropTile.propTypes = {
     active: PropTypes.number,
     planned: PropTypes.number,
     completed: PropTypes.number,
+    noPlans: PropTypes.number,
```
kathyavini, line 104: Resolves console warning

Image

kathyavini:
Merging without review this week as we only have one dev

PR #4206, opened 2026-06-19
LF-5345 Fix irrigation prescription date input styles

kathyavini:
**Description**

Repairs the styles of the date input component ([Storybook](http://localhost:6006/?path=/docs/components-form-dateinput--docs)) used in the Irrigation Prescription view.

Current integration:
Screenshot 2026-06-19 at 9 14 33 AM

Original and on this branch:
Screenshot 2026-06-19 at 9 13 35 AM

The visual regression was due to a component version bump. The component was built on **8.0.0**, but since the lockfile regeneration in https://github.com/LiteFarmOrg/LiteFarm/pull/3984/changes/e6c5c28e725bd116edf4df737ce1a48962897f07 it has been **8.17.0**.

I don't think that minor version bump should have created a breaking change except the component was styled using a more v7-esque method of targetting the input field (`slotProps.textField.InputProps`) which was functional in 8.0.0 but not in 8.17.0. And there is also a newer styling method (will note inline) we cannot access because our Material remains on version 5.

Jira link: https://lite-farm.atlassian.net/browse/LF-5345

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/Inputs/DateTime/index.tsx
+++ b/packages/webapp/src/components/Inputs/DateTime/index.tsx
@@ -13,11 +13,10 @@
  *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
  */
 
-import { useState } from 'react';
 import moment from 'moment';
-import { InputAdornment } from '@mui/material';
 import { DatePicker as MuiXDatePicker } from '@mui/x-date-pickers';
 import { TimePicker as MuiXTimePicker } from '@mui/x-date-pickers/TimePicker';
+import type { FieldOwnerState } from '@mui/x-date-pickers/models';
 import clsx from 'clsx';
 import styles from './styles.module.scss';
 import InputBaseLabel, { InputBaseLabelProps } from '../../Form/InputBase/InputBaseLabel';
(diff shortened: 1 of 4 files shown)


packages/webapp/src/components/Inputs/DateTime/index.tsx
```diff
@@ -31,30 +30,13 @@ interface DateTimePickerProps extends InputBaseLabelProps {
...
 }
 
-const createSlotProps = (
-  Icon: React.ElementType,
-  handleOpen: React.Dispatch<React.SetStateAction<boolean>>,
-  disabled: boolean,
-) => ({
-  textField: {
-    fullWidth: true,
-    InputProps: {
```
kathyavini, line 41: It's these `InputProps` which are now dropped / overridden so all the passed classes were lost.

packages/webapp/src/components/Inputs/DateTime/index.tsx
```diff
@@ -63,8 +45,6 @@ export const DateInput = ({
   label,
   ...labelProps
 }: DateTimePickerProps) => {
-  const [isDatePickerOpen, setIsDatePickerOpen] = useState(false);
```
kathyavini, line 66: The manual usage of `startAdornment` + component state was already unnecessary in version 8.0.0 (`openPickerButtonPosition: 'start'` should have already been available), but I  _suspect_ I used the version 7 method simply because version 8 was newly released and the AI models of a year ago had no awareness of it?

The native trigger (now used) behaves the same, and has slightly better accessibility as it is a true `<button>`.

packages/webapp/src/components/Inputs/DateTime/styles.module.scss
```diff
@@ -13,43 +13,39 @@
...
+// Styles for the MUI X DatePicker / TimePicker field, taken from the <Input /> component.
 
-.input {
-  height: 48px;
-  width: inherit;
-}
-
-.inputDisabled.inputDisabled {
-  background-color: var(--inputDisabled);
+.field {
```
kathyavini, line 18: If we weren't on Material version 5, we could use a `slotProps.input` for these styles, but it is version gated (library code checks for Material >= 6), so `:global()` has to be used instead

SayakaOno (approved):
It looks like the commit that updated the version was: https://github.com/LiteFarmOrg/LiteFarm/commit/17b9dc6003a124a04ccc3f714d18d394262ed115#diff-2be8e5e6f2f51881a11b8ae995fcddd361fe1f0e14c7cc564bd114d2febd2fff (you prevented that huge changes, Joyce! 😁)

I checked the fix on Storybook, and it looked great!

PR #4208, opened 2026-06-19
LF-5344 400 Error After Using Browser Back Navigation During Farm Map Area Creation

kathyavini:
**Description**

Although superficially similar to the bug fixed in #4207, this flow is a bit simpler, because it's only concerned with the form data part of the hookFormPersistSlice, and not with the history stack. In fixing the bug I've decided to leave it this way -- I haven't added the hook call that will clear the history stack (whether it's a multi-step form is a bit debatable) so you can browser back right into the form still. Only if the figure data is missing will the solution in this PR route you back to `/map`.

The advantage is a very minimal change to this flow, the tradeoff being that where you land is different based on whether you navigate to /tasks or /crops, which reset that hookFormPersistSlice, or to somewhere else (/animals, /inventory) that doesn't.

Loïc's position is that it's better to allow users to resume the form from where they left of. The app's general pattern is absolutely the opposite (can't re-enter forms). I guess this change is leaving this flow agnostic -- you can continue your form if the data is still there; otherwise restart. Maybe not ideal, but a way to keep the scope and change minimal for release.

Jira link: https://lite-farm.atlassian.net/browse/LF-5344

(Also now fixes https://lite-farm.atlassian.net/browse/LF-5348 as well)

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

Followed the flow in the Jira video:

Draw location > Confirm > Route to elsewhere in app (see above for the differing behaviour based on destination) > Browser back.

**Checklist:**

### DIFF
--- a/packages/webapp/src/components/LocationDetailLayout/PureLocationFormWrapper.jsx
+++ b/packages/webapp/src/components/LocationDetailLayout/PureLocationFormWrapper.jsx
@@ -13,6 +13,7 @@
  *  GNU General Public License for more details, see <https://www.gnu.org/licenses/>.
  */
 
+import { useEffect } from 'react';
 import { PersistedFormWrapper } from './PersistedFormWrapper';
 import { getFormDataWithoutNulls } from '../../containers/hooks/useHookFormPersist/utils';
 import { PureLocationDetailLayout } from './PureLocationDetailLayout';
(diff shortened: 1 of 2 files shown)


packages/webapp/src/components/LocationDetailLayout/PureLocationFormWrapper.jsx
```diff
@@ -81,6 +82,21 @@ export default function PureLocationFormWrapper({
...
+
+  useEffect(() => {
+    if (!hasGeometry) {
+      history.replace({ pathname: '/map' });
+    }
+  }, [hasGeometry, history]);
+
+  if (!hasGeometry) {
+    return null;
+  }
```
kathyavini, line 98: Note: `persistedFormData` is not actually read from the useHookFormPersist slice in the Edit flow -- it's populated by the deletion-filtered return from useGetLocationsQuery():

```ts
// EditLocationDetailForm
const { locations: locationData } = useLocationsById(location_id);

// ...
<PureLocationFormWrapper persistedFormData={locationData} ... />
```

In the Add flow it genuinely is from the slice. So although this fix in the pure component will fix both, resetting of the slice only is relevant in the Add flow.

SayakaOno (approved):
The only thing I don't love about this approach is duplicated `/map` paths, but this should be an edge case, and I like the minimal changes!

PR #4213, opened 2026-06-23
LF-5351 "Sit back we are fetching your map data modal" is showing every two minutes

kathyavini:
**Description**

Title is a _bit_ imprecise. Technically it's showing every time you revisit the map after 60 seconds has elapsed, since that's our RTK `refetchOnMountOrArgChange`.

As I wrote in the Jira ticket, using `isFetching` might have been a workaround for the RTK Query calls re-fetching instead of purging on farm switch? But in any case a completely blocking modal is not the right use case for isFetching.

In practice, we only see this when fetching sensors, where unfortunately it produced an almost unusable UI since the a multi-second modal was re-shown every 60 seconds. Using isLoading guarantees it is only shown on first fetch, when no cached data is available.

Jira link: https://lite-farm.atlassian.net/browse/LF-5351

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/containers/Map/index.jsx
+++ b/packages/webapp/src/containers/Map/index.jsx
@@ -189,7 +189,7 @@ export default function Map({ isCompactSideMenu }) {
       fullscreenControl: false,
     };
   };
-  const { drawAssets, assetGeometriesRef, markerClusterRef, isLocationsFetching } =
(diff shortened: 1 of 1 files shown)


SayakaOno (approved):
Much better UX, thank you!!

PR #4214, opened 2026-06-23
LF-5340 Blank Page Shown After Uploading Valid Sensors and No Success Confirmation Displayed

kathyavini:
**Description**

I'm actually not sure when this stopped working, because I swear I remember seeing the loading screen when it was first added 🙂 Probably the return from the API was different originally.

In a funnily parallel (but opposite) case to the change in #4213, here the isLoading gate is not appropriate to block on, because for any farm that was loaded _disconnected_ to Ensemble (i.e. a farm that is still to go through the Add Sensors flow), the getSensors() API request has already returned with

```json
{
    "sensors": [],
    "sensor_arrays": []
}
```

Since there _is_ data, isLoading won't be true, and we need to check on isFetching -- but not EVERY isFetching, just when the arrays are still empty!

Jira link: https://lite-farm.atlassian.net/browse/LF-5340

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/containers/SensorList/useGroupedSensors.ts
+++ b/packages/webapp/src/containers/SensorList/useGroupedSensors.ts
@@ -159,19 +159,14 @@ const useGroupedSensors = (): {
   sensorSummary: SensorSummary;
   groupedSensors: GroupedSensors[];
 } => {
-  const {
-    data: sensors,
-    isLoading: isLoadingSensors,
-    isFetching: isFetchingSensors,
-  } = useGetSensorsQuery();
+  const { data: sensors, isFetching: isFetchingSensors } = useGetSensorsQuery();
   const system = useSelector(measurementSelector);
   const { locations: farmAreas, isLoading: isLoadingFarmAreas } = useLocations({
     filterBy: FigureType.AREA,
   });
 
   const hasNoSensorData = !sensors?.sensors.length && !sensors?.sensor_arrays.length;
-  const isLoading =
-    isLoadingSensors || isLoadingFarmAreas || (isFetchingSensors && hasNoSensorData);
+  const isLoading = isLoadingFarmAreas || (isFetchingSensors && hasNoSensorData);
   const dataNotReady = !sensors || !farmAreas;
 
   const sensorsData =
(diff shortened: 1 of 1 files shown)


packages/webapp/src/containers/SensorList/useGroupedSensors.ts
```diff
@@ -105,14 +105,17 @@ const formatSensorToGroup = (
...
+        acc[name] = 0;
+      }
+      acc[name] += 1;
 
-    return acc;
-  }, {} as Record<Sensor['name'], number>);
+      return acc;
+    },
+    {} as Record<Sensor['name'], number>,
+  );
```
kathyavini, line 118: Above is pre-commit lint. Only change is below 👇

packages/webapp/src/containers/SensorList/useGroupedSensors.ts
```diff
@@ -156,13 +159,14 @@ const useGroupedSensors = (): {
...
-  const { data: sensors, isLoading: isLoadingSensors } = useGetSensorsQuery();
+  const { data: sensors, isFetching: isFetchingSensors } = useGetSensorsQuery();
   const system = useSelector(measurementSelector);
   const { locations: farmAreas, isLoading: isLoadingFarmAreas } = useLocations({
     filterBy: FigureType.AREA,
   });
 
-  const isLoading = isLoadingSensors || isLoadingFarmAreas;
+  const hasNoSensorData = !sensors?.sensors.length && !sensors?.sensor_arrays.length;
+  const isLoading = isLoadingFarmAreas || (isFetchingSensors && hasNoSensorData);
```
SayakaOno, line 169: It's been a while since I last worked on sensors, so I can't remember... when do `isFetching: true` and `hasNoSensorData: false` happen at the same time??

kathyavini: Sorry it took me a minute to parse the double negative 😅 

It should be the normal state reloading the sensor list once the sensors are fetched and the 60 sec of `refetchOnMountOrArgChange` has elapsed.

SayakaOno: Got it, thank you! 🙏

SayakaOno (approved):

PR #4215, opened 2026-06-24
LF-5350 Support new types of sensors or filter them out of the app

kathyavini:
**Description**

This one is a _bit_ large (and AI-generated) for my tastes given we are less than 24 hours to release... _but_ I think it's okay! It impacts only the sensor code, and is fixing what was literally non-functional before. I've been clicking around on it locally for quite some hours as well now 😁

### Background
Ensemble is returning a lot of new sensors to us this year. In fact, four of them (IR Temperature sensor, DripDrain Sensor, Humidity Sensor, and Soil Water Content sensor) have JUST gone online today, _after_ I started coding this ticket!

Our setup was actually _almost_ able to receive and show the readings for all of these (the sensor readings part of app was written to be so flexible ❤️🙏 )  so rather than filtering out the new sensors, I've extended the allow lists to include them.

Farming Smarter has confirmed with us that this new list looks exhaustive, and the priority sensors for this project are only Soil Water Content, ET, and the SDI-12 Weather Station. Based on that, I've made Soil Water Content + ET chartable. No other changes were made to the chart data, but the allowed reading types for KPI were expanded.

### Summary of Changes

1. A few small adjusts to prevent against crashes, _were_ we to get back unsupported sensor types from Ensemble:

```ts
const readingTypes = SENSOR_READING_TYPES[sensor.name] ?? [];   // ← fixing the crash from Jira
// ...
return SENSOR_READING_TYPES[sensor.name] ? getGeneralSensorDefaultKPIProps(sensor.name, t) : [];
// ...
sensors: mappedSensors[sensorArray.id] ?? []
```

As one of the last adjust on this PR I have added _filtering on the sensors API return!_ So none of these safeguard branches should be actually reachable now. However, the guards remain just in case we return to non-filtering in the future, e.g. to pick up new types as Ensemble adds them.

2. Adding the new sensors types visible as of Jun 23, 2026 and and their _KPI_ reading types (`SENSOR_READING_TYPES`). As mentioned above, just the paramters of the Soil Water Content and ET sensors were additionally added to the line chart `CHART_SUPPORTED_PARAMS`.

3. The KPIs are now populated with their latest reading, even if that was not within the last 2 hour window. Previously the search for recent data was last 15 min --> last 2 hours. Now it will be last 15 min--> last 2 hours --> latest timestamp period. The "update readings" button runs through the same sequence, so that that KPIs stay populated on update, even if there is still no new data.

Jira link: https://lite-farm.atlassian.net/browse/LF-5350

Note: PR is tagged with New Translations but they don't have to get TRANSLATED for this project. English-only is fine!

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/webapp/src/containers/SensorReadings/v2/LatestReadings/utils.tsx
```diff
@@ -60,8 +61,10 @@ const generateTMeasurement = (
 ): TMeasurement => {
   return {
     measurement: t(`SENSOR.READING.${readingType.toUpperCase()}`),
-    value: value && system ? convertEsciReadingValue(value, readingType, system) : '-',
```
kathyavini, line 63: The change here is so that values of 0, which I am seeing with a working ET sensor, report properly as 0 instead of as a dash (missing).

I don't recall if Loïc wanted a '-' for zero values based on some of the sensors we started with (like Weather Station?) but I don't think it's appropriate for e.g. evapotranspiration.

Image

packages/webapp/src/containers/SensorReadings/v2/constants.ts
```diff
@@ -26,15 +26,46 @@ import type {
...
+  'Tipping Bucket Rain Gauge': ['cumulative_rainfall'],
+  'ET sensor': ['evapotranspiration', 'latent_energy_flux', 'heat_flux'],
+  'Soil Water Content Sensor': ['moisture', 'temperature', 'electrical_conductivity'],
+  'Humidity Sensor': [
+    'relative_humidity',
+    'temperature',
+    'vapor_pressure_deficit',
+    'wet_bulb_temperature',
+  ],
+  // DripDrain reading types are not yet modelled; renders no KPI for now.
```
kathyavini, line 58: The Ensemble readings return for this one is a little differently formatted. But it's not a priority sensor for this project so we'll skip worrying about its readings for now.

kathyavini:
Merging early so QA and code review can happen simultaneously 🙏

PR #4218, opened 2026-06-25
LF-5354 (VERSION #2) Farm map does not update after area creation or retirement until page reload

kathyavini:
**Description**

The main part of this solution was an implementation Claude Opus 4.8 came up with based on the original version in #4217.

The core of the change I'm satisfied with and I don't think it's too much to ship: it just moves the call to `drawAssets()`, and the following related code into a useEffect that is gated on `isLoadingExternalLocations` and a ref to make sure drawing happens only once. This part actually makes pretty intuitive sense.

The part that was a little harder was the modal behaviour 😂

My aim was to keep the restored UX from the second commit 5316d5b5f4cc01f3d56267186825d4e3237344ec of #4217, where the map renders without delay again. That commit had been cherry-picked onto this branch and worked great for non-sensor farms. The problem was entirely for farms with sensors, in that
- the drawing of the assets DOES have to be gated on the sensors being loaded, because unlike in the redraw solution, they will _not_ pop in later
- this would be an extremely poor UX without a loading screen. Actually even for the pop-in case, I think a loading screen becomes necessary when the sensor endpoint is taking 10-20 seconds as it does now

The problem was that since every single farm calls `getSensors()` (even if only to confirm that they have none), a modal that was gated on only `isLoadingExternalLocations` was flashing for every farm again, basically negating the UX of the commit I had cherry-picked!

The ultimate solution is a bit weird, but I actually think it is functional. The loading modal **only shows if the sensors request finishes after the locations request by more than 400 ms**. In other words, the modal is gated like this:

```ts
isOpen={!isLoaded || (!isFetchingInternalLocations && isLoadingExternalLocations)}
```

and it already had a feature built in `showDelay` such that the modal is only shown if the open condition is true for a given period of time. I think the tables and joins involved in the locations controller are probably inherently slower than the simple farm_addon check needed to return the empty response of a farm without sensors. Actually even with the delay set to 200ms, I never again saw the loading modal for a non-sensor farm again! But just as an extra precaution and to cover more network conditions, I bumped it to 400ms.

I hope I'm not missing anything, but in my testing, that seemed to achieve the best of both words -- the immediately loading map UX of `isOpen={!isLoaded}` for sensor-free farms WHILE still showing the modal when needed (sensor partner farms). And so no need to switch back to modal on every fetch 🤞

Jira link: https://lite-farm.atlassian.net/browse/LF-5354

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

Just a lot of testing on switching to farms with and without sensors!

**Checklist:**

### DIFF
--- a/packages/webapp/src/containers/Map/index.jsx
+++ b/packages/webapp/src/containers/Map/index.jsx
@@ -189,12 +189,11 @@ export default function Map({ isCompactSideMenu }) {
       fullscreenControl: false,
     };
   };
-  const { drawAssets, assetGeometriesRef, markerClusterRef, isLocationsLoading } =
-    useMapAssetRenderer({
-      isClickable: !drawingState.type,
-      drawingState: drawingState,
-      showingConfirmButtons: showingConfirmButtons,
-    });
+  const { drawAssets, assetGeometriesRef, markerClusterRef } = useMapAssetRenderer({
+    isClickable: !drawingState.type,
+    drawingState: drawingState,
+    showingConfirmButtons: showingConfirmButtons,
+  });
 
   // Cleanup listeners on map instance objects
   useEffect(() => {
@@ -424,7 +423,7 @@ export default function Map({ isCompactSideMenu }) {
 
   return (
     <>
-      {!isLocationsLoading && isLoaded && (
+      {isLoaded && (
         <>
           {!drawingState.type && !showSuccessHeader && <PureMapHeader farmName={farm_name} />}
           {showSuccessHeader && (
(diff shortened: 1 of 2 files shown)


kathyavini:
Merging early so QA and code review can happen simultaneously 🙏

SayakaOno (commented):
I love that the changes are simple and easy to understand, thank you!

The only edge cases I see are when navigating to the map while adding sensors or disconnecting ESci, but I don't think they're blockers.

PR #4234, opened 2026-06-30
LF-5366 Survey draft persists when creating a new farm via switch farm

kathyavini:
**Description**

### Background

The more I dig around in our Redux the more interesting it is 😂 

It turns out that we had YET ANOTHER "clear data on farm switch" mechanism that I had completely missed, because it had only been applied to notifications (the list, not the alerts/new notification dot), and _only when the farm switch modal was shown_. (Note that the farm switch modal is not shown when you Add Farm, which is the cause of the bug in both the two linked Jira tickets).

The mechanism was a bit fascinating. There was a dedicated logical branch in the rootReducer:

```js
// reducer.js
const rootReducer = (state, action) => {
  if (state && action.type === ActionTypes.SWITCH_FARMS) {
    // selectively only reset farmStateReducer state when switching farms
    const { farmStateReducer, ...otherReducers } = state;
    state = otherReducers;
  }
 ```
 
And that action type constant `SWITCH_FARMS` was the only entry in this `actionTypes.js` file (that I had also missed)!
 
 ```js
 export const ActionTypes = {
  SWITCH_FARMS: 'chooseFarmFlowReducer/startSwitchFarmModal',
};
```

That type held by the constant is the type dispatched by (as it says) `startSwitchFarmModal` in `chooseFarmFlowSlice.js`. What's super interesting about this was:
- it's just piggy-backing! Even though it's called "SWITCH_FARM" and it's used to drop the farmStateReducer state, what the action ACTUALLY does is simply set the "show modal" state to true for the given farm:

```js
// chooseFarmFlowSlice.js
startSwitchFarmModal: (state, { payload: farm_id }) => {
  updateFarmState(state, { farm_id, showSwitchFarmModal: true });
},
```
  - that means if, for any reason we refactored the UI to stop showing this modal and removed the action, we would incidentally have also lost the Redux state reset
- as mentioned above, the modal isn't shown on every case of moving from an old farm to a new farm. Add Farm in particular will never call `startSwitchFarmModal`

### Chosen Solution

To address these issues, this PR:
- adds a dedicated action specifically for what we want to do in (resetting the FarmStateReducer) in place of the piggy-back on showing the farm switch modal
- calls this in `clearOldFarmState()` since that is now reliably shared between _all_ the pathways that involve from moving from one farm to another without logging out (switch, Add Farm, and accept invite)

### Implications for future refactor
An implication of this not pursued further in _this_ PR, is that we actually could have -- and probably should have -- thrown all the `tempStateReducer` Redux state that we wanted dropped on farm switch here into `farmStateReducer`. I'm thinking in particular the filter state and the date ranges (finances, profitability).

What both `tempStateReducer` and `farmStateReducer` share is that they aren't keyed to `farm_id` via `loginSelector` the way that the entries in `entitiesReducer` are. It's not within scope here, but it should be safe and effective to rearrange them a little bit so that all the ones that should be dropped on switch live only in `farmStateReducer`. If we follow that convention moving forward, we will have three predictable ways to drop Redux state on farm switch (=`loginSelector`, `invalidatesTags`, and `resetFarmStateReducer`) and we should be clear of this category of bugs moving forward 🤞

Jira link: 
https://lite-farm.atlassian.net/browse/LF-5366 (surveyDraftSlice) and 
https://lite-farm.atlassian.net/browse/LF-5368 (notificationSlice)

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/containers/saga.js
+++ b/packages/webapp/src/containers/saga.js
@@ -101,7 +101,7 @@ import {
   userFarmsByFarmSelector,
 } from './userFarmSlice';
 import { logUserInfoSuccess, userLogReducerSelector } from './userLogSlice';
-import { resetFarmState } from '../store/actionTypes';
+import { resetFarmStateReducer } from '../store/actionTypes';
 import { api, invalidateTags } from '../store/api/apiSlice';
 import { locationApi } from '../store/api/locationApi';
 import { FarmLibraryTags, FarmTags } from '../store/api/apiTags';
@@ -556,7 +556,7 @@ export function* fetchAllSaga() {
 }
 
 export function* clearOldFarmStateSaga() {
-  yield put(resetFarmState());
+  yield put(resetFarmStateReducer());
   yield put(resetTasks());
   yield put(resetDateRange());
   releaseFarmScopedQuerySubscriptions();
--- a/packages/webapp/src/containers/saga.js
+++ b/packages/webapp/src/containers/saga.js
@@ -101,6 +101,7 @@ import {
   userFarmsByFarmSelector,
 } from './userFarmSlice';
 import { logUserInfoSuccess, userLogReducerSelector } from './userLogSlice';
+import { resetFarmState } from '../store/actionTypes';
 import { api, invalidateTags } from '../store/api/apiSlice';
 import { locationApi } from '../store/api/locationApi';
 import { FarmLibraryTags, FarmTags } from '../store/api/apiTags';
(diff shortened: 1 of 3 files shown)


SayakaOno (approved):
I hadn't known about this part of the app - thank you for digging into it!
The solution looks great, and rearranging `tempStateReducer` and `farmStateReducer` sounds good!

PR #4253, opened 2026-07-07
LF-5383: Certification export email is not sent when using the khmer language

SayakaOno:
**Description**

Certification export fails with `ENOENT` when the farm language is Khmer. The Khmer translation of "Crop Production Aids" — `សម្ភារៈ/សារធាតុជួយការផលិតដំណាំ` — legitimately contains a `/`. Record I generation interpolates this translated string directly into the output file path, so the filesystem treats the `/` as a directory separator and tries to write into a subdirectory that doesn't exist. The export job dies and the email is never sent.

* New `fileUtils.js` — `sanitizeFileName()` replaces path separators and Windows-reserved characters `(/ \ : * ? " < > |)` with spaces and collapses whitespace. The Windows characters matter because users download the export zip and may extract it on Windows, where such file names fail or get mangled.
* `record_i_generation.js` — the failing case; the full `RECORD I - {title}` filename is now sanitized. The Khmer file becomes `កំណត់ត្រា អាយ - សម្ភារៈ សារធាតុជួយការផលិតដំណាំ.xlsx`.
* `survey_record` — same bug waiting to happen with surveyName, which comes from `SurveyStack` data and could contain anything.
* `record_a_generation.js` — defensive, so a future Crowdin sync can't reintroduce the crash via EXPORT_DOCUMENT_NAME.

Jira link: https://lite-farm.atlassian.net/browse/LF-5383

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)
         Tested manually.

**Checklist:**

### DIFF
--- a/packages/api/src/jobs/certification/fileUtils.js
+++ b/packages/api/src/jobs/certification/fileUtils.js
@@ -1,0 +1,27 @@
+/*
+ *  Copyright 2026 LiteFarm.org
+ *  This file is part of LiteFarm.
+ *
+ *  LiteFarm is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  LiteFarm is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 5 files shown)


SayakaOno:
Merged without review since we only have one dev for a while.

PR #4281, opened 2026-07-28
LF-5379 Remove legacy certification API shim and /organic_certifier_survey routes

kathyavini:
**Description**

Claude's **WHAT THIS PR DOES - IN SIX FACTS**

- The field-name shim moves from the models to the legacy controller. `certifierModel` and `certificationSystemTypeModel` now emit real column names; only `certificationController.js` produces the old names.
- `/certifications` gains two lookup endpoints — `supported_certifiers` and `supported_system_types` — returning real column names, farm-scoped by the `farm_id` header.
- The persisted redux flag `certifierSurveySelector().interested` is replaced everywhere by a new hook, `useHasCertifications()`, reading the RTK Query certifications cache.
- The whole `OrganicCertifierSurvey` tree is deleted — 3 redux slices, 1 saga, 3 hooks, 4 Pure components, 3 containers, 5 stories/tests. No dangling imports remain (verified).
- The survey-export PDF renderer is corrected to the real column names. Its queue has been unregistered since 2023-03-01, so this fixes dormant code — the live export ships the survey as an .xlsx instead.
- The legacy `/organic_certifier_survey` routes stay, deliberately: cached service-worker frontends still call them. Deletion is tracked as LF-5410.

This refactor naturally fixes the bug we were seeing on integration this morning where the "interested" state was not tracking the updated list of certifications; Claude explains

> previously the slice was only written to from `getCertificationSurveys()`, dispatched from one place — `containers/saga.js:511`, i.e. login and farm select. Nothing refetched it afterwards, so adding a certification left interested stale for the whole session. Here `addCertification` carries `invalidatesTags: ['Certifications']`, so the hook re-reads immediately. 

Jira link: https://lite-farm.atlassian.net/browse/LF-5379

**Type of change**

**How Has This Been Tested?**

- [x] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/api/tests/certifications.test.ts
```diff
@@ -253,6 +290,94 @@ describe('Certifications CRUD tests', () => {
     });
   });
 
+  describe('GET /certifications/supported_system_types', () => {
+    test('Returns system types under their real column names', async () => {
```
SayakaOno: Can we drop "real column names"? It'll only make sense while the shims exist...

packages/api/tests/certifications.test.ts
```diff
@@ -253,6 +290,94 @@ describe('Certifications CRUD tests', () => {
...
+      const pgs = res.body.find(
+        (systemType: SystemType) => systemType.translation_key === PGS_TRANSLATION_KEY,
+      );
+      expect(thirdPartyOrganic.id).toBe(thirdPartySystemTypeId);
+      expect(thirdPartyOrganic.name).toBe('Third-party Organic');
+      expect(pgs.id).toBe(pgsSystemTypeId);
+      expect(pgs.name).toBe('Participatory Guarantee System');
+      // The legacy /organic_certifier_survey route still serves these shimmed names
+      expect(thirdPartyOrganic).not.toHaveProperty('certification_id');
+      expect(thirdPartyOrganic).not.toHaveProperty('certification_translation_key');
```
SayakaOno: Do you expect these assertions to be removed along with the shims?
I've become a bit sensitive to comments/tests AI writes that might go stale in the future 😂

packages/webapp/src/containers/saga.js
```diff
@@ -507,10 +503,9 @@ export function* fetchAllSaga() {
   yield call(openFarmScopedQuery, api.endpoints.getSensors.initiate());
 
   const isAdmin = yield select(isAdminSelector);
+  // Only admin roles hold get:certification
   const adminTasks = [
-    put(getCertificationSurveys()),
-    put(getAllSupportedCertifications()),
-    put(getAllSupportedCertifiers()),
+    call(openFarmScopedQuery, certificationsApi.endpoints.getCertifications.initiate()),
```
SayakaOno, line 508: Could you confirm that this is only necessary to evaluate `hasCertifications` when offline?
I'd like to confirm whether I should replace this with a call to get "hasCertifications" for all roles.

SayakaOno (commented):
This feels great, thank you!!
Just a few minor comments!

packages/api/tests/certifications.test.ts
kathyavini: Yes, good call! 🙏

packages/webapp/src/containers/saga.js
kathyavini: > I'd like to confirm whether I should replace this with a call to get "hasCertifications" for all roles.

I think that would be my expection/preference!

In addition to offline (but I think that's by far the main one) there is a spotlight using hasCertifications (the DocumentsSpotlight.jsx from this PR) that it would be nice to have the correct value for before loading the component because flipping that one in the dependency array seemed a bit sketchy. And maybe marginally better UX for the compliance questions to not flip after the request completes (especially the one we looked at today that substitutes a further set of questions in place of a "submit"). But realistically yes mostly offline!

SayakaOno: Ah okay, thank you!!
And thank you for the clean up!

PR #4284, opened 2026-07-29
LF-5408 Add issue date to DFC payload

kathyavini:
**Description**

After confirmation (see Jira ticket) that a date on Certification is consistent with the DFC Ontology and just missing from the connector, this adds the date manually to the certification part of the DFC payload.

'YYYY-MM-DD' has been confirmed the correct format... which then we had to generate from our db format which is a datetime. Oops, I think I would have requested a datetime from OFN + DFC if I had tested the flow first 😝

Jira link: https://lite-farm.atlassian.net/browse/LF-5408

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

The DFC Adapter test was initially updated, but since that just passes through the string unchanged, the more important test was hitting the /enterprises endpoint locally to observe the date format:

```
http://localhost:5001/dfc/enterprises/
```

and adding to the API test, which actually uses the database value.

**Checklist:**


packages/api/src/server.ts
```diff
@@ -214,6 +214,7 @@ app.set('json replacer', (key: string, value: string) => {
     'effective_date',
     'germination_date',
     'harvest_date',
+    'issue_date',
```
kathyavini, line 217: This change is not related to the DFC payload at all, but for the value returned to the frontend, so that it matches `valid_until` (already included).

This fixes a bug where the issue date is seen to shift backwards by one day compared to what you input, on a UTC+ farm. I did not notice until now because I so rarely make UTC+ farms!

SayakaOno (approved):
Looks good to me!

PR #4306, opened 2026-08-07
LF-5418 Send the user to the dashboard after password new account and google sign in

kathyavini:
*I'm not too sure about RTK query vs Saga for this one.* There is nothing to hold in the store here, and in fact the ticket should be kept out of the store, which is done in the RTK Query version with an explicit reset. Do you have a preference for the two versions needed based on that final commit https://github.com/LiteFarmOrg/LiteFarm/pull/4306/changes/6411dc73ca82c6a3e5dd21810a7c9309f116961f @SayakaOno? 🙏 

**Additional open questions (Answered Aug 10)**
- should we be using Redux at all for the dashboard ticket state (`dashboardTicketSlice.ts`)? The same thing can be achieved with a module variable, in fewer lines -- but would be a new pattern
  - we have gone with new pattern 🙂 
- what should the UX be for new account creation, one of the paths covered here? We would be sending them back to Dashboard with no farms (and no user account according to the dashboard backup copy) Does the data team have a message for this case as well?
  - dashboard will adjust messaging for us
- not in spec but seems needed to me: a different screen for the login page in this flow? Perhaps like this?
  -  implemented and just needs a run by Loïc tomorrow
Screenshot 2026-08-07 at 8 24 34 PM

**Description**

The Analytics Dashboard (`data.litefarm.org`) will let people sign in with their existing LiteFarm account. LiteFarm acts as the identity provider: it issues a ticket — a 30-second, single-use JWT — which the Dashboard exchanges server-to-server for the user's identity and farm memberships.

This PR is the webapp half: someone with no Dashboard session arrives at LiteFarm carrying `?return_to=<allowlisted address>`, and once signed in the browser is sent to that address with `?ticket=…` attached.

### Four arrival paths, one shared generator

`handOffToDashboardIfRequested` in `containers/dashboardTicketSaga.ts` is the only place that calls the endpoint. Three sign-in sagas gate their existing navigation on it:

```js
// containers/CustomSignUp/saga.js — customLoginWithPasswordSaga
yield put(loginSuccess({ user_id }));
const handedOff = yield call(handOffToDashboardIfRequested);
if (!handedOff) {
  history.push('/farm_selection');
}
```

The same two lines wrap `customCreateUserSaga` and the Google saga's `isSignUp` branch.

The fourth path is arriving with a LiteFarm session already live. Access tokens last seven days, so this should be considered a common, rather than an edge, case. `useDashboardHandoff` starts the hand-off on mount, and `Routes` renders a `Spinner` in place of the route tree while it is in flight.

### Why the return address is read during render

`CustomSignUp` is a child of `Routes`, and on arrival at `/` it replaces the location with one that carries no search string:

```jsx
// containers/CustomSignUp/index.jsx
history.replace(
  { pathname: '/' }, // <-- search string dropped
  { user: { email }, component: CUSTOM_SIGN_UP },
);
```

So the hook captures in a `useState` initialiser, which runs during the first render, before any effect has had a chance to clear it.

### For reviewers

- `farm_id` is not sent. The exchange returns `farm_id: null` and the Dashboard runs its own farm choice; arriving with a farm already named ships in a follow-up PR.
- An address that is not allowlisted gets a `400`, an error snackbar, and the ordinary destination rather than a dead end.
- Invite acceptance and password reset completion are untouched. Neither is an entry point for this direction.

Jira link: https://lite-farm.atlassian.net/browse/LF-5418

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

Will share a testing document, but the fundamental way is to set local API `.env` as follows

```
DASHBOARD_ALLOWED_RETURN_TO=https://example.com/auth/finish
```

And then paste the following into browser:

```
http://localhost:3000/?return_to=https%3A%2F%2Fexample.com%2Fauth%2Ffinish
```

in different states of "logged-in ness".

**Checklist:**


packages/webapp/public/locales/en/message.json
```diff
@@ -148,6 +148,7 @@
   },
   "LOGIN": {
     "ERROR": {
+      "DASHBOARD_TICKET": "Could not open the Data Dashboard. Please try again.",
```
kathyavini: Incidentally, I'm not sure I like this message for a snackbar triggered by our ticket-creating endpoint.

I guess from the user flow perspective the ticket creation and redirect is one flow, but I still find this misleading!

SayakaOno:
Sorry this took longer than I expected, I was looking at it from a few different angles! The changes look great, and it was more complicated than I expected too 😇 

1. RTK Query vs Saga (vs plain fetch)
My preference is a plain fetch... as you mentioned, the result won't be used anywhere, and I'd value the simplicity over staying consistent with the `store/api/*` pattern. That said, a plain fetch would be a new pattern, so I'm happy to go with RTK Query! (I prefer RTK Query to Saga)

2. Selectors vs module variable
I prefer the module variable. Selectors feel like more than we need for this scale of state.

3. Separately — how `return_to` gets to the sagas (written by Claude, but I think it's more understandable than my explanation...)
This is independent of the above: even with the module variable, `return_to` is read from a global inside the sagas and inside `handOffToDashboardIfRequested`, rather than being passed explicitly. The version I'd find more intuitive is a dedicated route (e.g. `/dashboard-signin`) that wraps `CustomSignUp`, capturing `return_to` the same way `useDashboardHandoff` already does (a `useState` initializer, so it's captured before `CustomSignUp`'s own `history.replace` can drop the query string), and handing it down as a prop instead of a global read:
    ```jsx
    <Route exact path="/dashboard-signin">
      <CustomSignUp returnTo={returnTo} isDashboardSignIn />
    </Route>
    ```
     `return_to` would then flow explicitly into the three sagas' payloads and into `handOffToDashboardIfRequested` as an argument, instead of each of them reaching into `dashboardReturnTo.ts`. This is a genuine preference, but a matter of taste rather than a correctness issue — I'm happy to go with the current solution as-is.

      I think this approach would avoid both the selectors and the global variable!

kathyavini:
That all sounds good! I also thought super briefly about a plain fetch but it seemed so strange to call our own API in a plain fetch.

I'm excited to see the route-passed version! And sorry, I read it wrong at first glance -- I thought the URL was keeping the query string but it's not, it's just a different dedicated route. So we would also need a tiny update from the data team to point to `/dashboard-signin` instead, is that right?

packages/webapp/public/locales/en/message.json
kathyavini: I am going to just change a a vague "connect to" instead of "open"!

SayakaOno (approved):
Looks and works great!

PR #4307, opened 2026-08-10
LF-4982: Improve nginx.conf

SayakaOno:
**Description**

- Add `application/javascript`, `application/json`, and `image/svg+xml` to `gzip_types`, and change `gzip_proxied` to `any` — previously proxied requests (anyone behind a carrier/corporate proxy sending a `Via` header) were served uncompressed.
- Remove the `proxy_set_header 'Access-Control-Allow-Methods'` / `'Access-Control-Allow-Origin'` lines on the `api.beta.litefarm.org` block — these don't do anything as request headers (they would need to be **response** headers), and CORS is already handled by the backend's `cors()` middleware.
- Forward `X-Forwarded-For`, `X-Forwarded-Host`, and `X-Forwarded-Proto` from nginx to the backend so it sees the real client IP/host/protocol instead of nginx's own.
- Enable Express `trust proxy` in `server.ts` so `req.ip` resolves the real client IP from the `X-Forwarded-For` chain.
- Simplify `loginController.js` and `userLogController.js` to use `req.ip` instead of manually parsing `x-forwarded-for` with a `req.connection.remoteAddress` fallback.

NOTE:
Storing users' IP addresses and removing incorrect request headers are unrelated to the epic this ticket belongs to, and I'd be happy to exclude them if necessary!

Jira link: https://lite-farm.atlassian.net/browse/LF-4982

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

Built docker container locally and ran a script.

Before:
integration

After:
LF-4982

`gzip_min_length`is set to 10240 (10 KB), so smaller files like `common.json` won't be compressed.

**Checklist:**

### DIFF
--- a/packages/webapp/nginx.conf
+++ b/packages/webapp/nginx.conf
@@ -37,7 +37,15 @@ http {
     gzip_min_length  10240;
     gzip_comp_level  6;
     gzip_proxied expired no-cache no-store private auth;
-    gzip_types text/plain text/css text/xml text/javascript application/x-javascript application/xml;
+    gzip_types
+        text/plain
+        text/css
+        text/xml
+        text/javascript
+        application/x-javascript
+        application/xml
+        application/javascript
+        application/json;
 
     # include /etc/nginx/conf.d/*.conf;
 
--- a/packages/webapp/nginx.conf
+++ b/packages/webapp/nginx.conf
@@ -98,11 +98,12 @@ http {
 
         location / {
           limit_req zone=one burst=5 nodelay;
-
-          proxy_set_header 'Access-Control-Allow-Methods' 'GET, POST, OPTIONS, PUT, DELETE';
+ 
           proxy_pass_request_headers      on;
-          proxy_set_header 'Access-Control-Allow-Origin' 'https://beta.litefarm.org';
           proxy_pass http://backend:5000/;
+          proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+          proxy_set_header X-Forwarded-Host $host;
+          proxy_set_header X-Forwarded-Proto $scheme;
         }
         listen 443 ssl; # managed by Certbot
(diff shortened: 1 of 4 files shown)


packages/webapp/nginx.conf
```diff
@@ -36,8 +36,17 @@ http {
     gzip_vary        on;
     gzip_min_length  10240;
```
kathyavini: > `gzip_min_length` is set to 10240 (10 KB), so smaller files like common.json won't be compressed.

My Claude actually highlighted this as an error in our file -- it thought 10 kB is actually quite too high a default and **`1024`** (1 kB) would be better. I think I agree... the nginx default is actually 20 bytes (!!!) and a cursory Google suggests 256 or 1024 as a good number.

Claude calculated our files betwen 1kB and 10kB as follows:

> 413 JavaScript chunks between 1 KB and 10 KB — 957,749 bytes raw, 352,507 at gzip -6 — and every locale file except crop.json. That is about 695 KB per install

kathyavini (commented):
Ahhhhhh I am so excited for this PR!! ❤️‍🔥 This might be the most exciting PR of the year in my opinion, and I love that we're just sneaking it in to this little release 😁 This deserves a rat emoji! 🐀 

The only thing I would like to add/modify would be compression down to 1 kB, as described below. It would still be well above the range of 50-100 bytes where CPU pressure would become a concern.

> Storing users' IP addresses and removing incorrect request headers are unrelated to the epic this ticket belongs to, and I'd be happy to exclude them if necessary!

No I love it, thank you for fixing everything!

SayakaOno:
Thank you Joyce! I've updated `gzip_min_length` to 1KB - could you double-check? 🙏

kathyavini (approved):
🥳

PR #4308, opened 2026-08-11
LF-5443 Signing in while a second tab sits on the sign on screen locks the user on the welcome screen with no way back

kathyavini:
**Description**

As mentioned in tech daily today, I've been landing on this particular `/welcome` screen, prompted by having an `id_token` but no Redux store / `persist:root`, a lot while testing the dashboard SSO login.

I'm not completely convinced that two tabs are _always_ involved, but I haven't figured out another mechanism and it's not an outrageous assumption (I might easily have a LiteFarm tab side-by-side with the instance called up by the dashboard redirect). Certainly two tabs is the easiest and most reliable reproduction, as described in the Jira ticket and reproduced in "testing" below.

In any case, the code solution on this branch (requiring both an `id_token` and Redux `user_id` to qualify for `isAuthenticated`), definitely addresses the state we saw today on Beta, regardless of how that state was produced.

### Tracing the bug
Our original `isAuthenticated` read only the `id_token` value from local storage:

```js
// util/jwt.js
export const isAuthenticated = () => !!localStorage.getItem('id_token'); // <-- never reads Redux
```

Routes renders its signed-in branch on that alone. No `farm_id` is set, so it renders `OnboardingFlow` and routes as follows:

```jsx
// routes/Onboarding.jsx
const hasUserFarms = useSelector(userFarmLengthSelector); // 0 — userFarmsByUserSelector returns [] with no user_id, so "no identity" counts as "no farms"
const { loaded: farmsLoaded } = useSelector(userFarmStatusSelector); // false — set true once a userFarm fetch finishes, succeed or fail
// ...
if ((!farm_id || !step_one) && !hasUserFarms) {
  const target = farmsLoaded ? '/welcome' : '/farm_selection';
  return <Redirect to={target} />;
}
```

`farmsLoaded` is false, so the first redirect is to `/farm_selection` — the correct target, because `ChooseFarm` normally fetches the farm list and fills the store.

```jsx
// containers/ChooseFarm/index.jsx
useEffect(() => {
  dispatch(getUserFarms()); // <-- runs on on mount
  dispatch(getSpotlightFlags());
}, []);
```

`getUserFarmsSaga` selects the absent `user_id` and requests `GET /user_farm/user/undefined`. That request satisfies `checkJwt` but can't (of course) locate any userFarms, so the controller answers with `200 []`. `getUserFarmsSuccess([])` sets `loaded: true`, which `ChooseFarm` reads as an answer about farms:

```jsx
// containers/ChooseFarm/index.jsx
const farms = useSelector(userFarmsByUserSelector); // [] for the same reason as above
const { loaded } = useSelector(userFarmStatusSelector); // now true
useEffect(() => {
  loaded && farms.length === 0 && history.push('/welcome');
}, [farms, loaded]);
```

### Fix 1 — require both `id_token` and `user_id` to be present for authenticated routes

This is a new hook combining both the `id_token` and Redux:

```ts
// hooks/useAuthenticatedSession.ts
export default function useAuthenticatedSession(): boolean {
  const hasIdentity = useSelector((state: RootState) => !!userFarmReducerSelector(state).user_id);

  return hasIdentity && isAuthenticated();
}
```

now used in place of the old `isAuthenticated` that controls route rendering:

```jsx
// routes/index.jsx — Routes
const isSignedIn = useAuthenticatedSession();
const isHandingOffToDashboard = useDashboardHandoff(isSignedIn);
// ...
if (isSignedIn) {
  // ... OnboardingFlow, and every signed-in Switch
} else {
  // ... CustomSignUp at '/', and a Redirect to '/' for anything else
}
```

(As a side note there is one other use of `isAuthenticated` in app that was untouched -- `useIsFarmSelected`. That one didn't need any adjust because it already combines the `id_token` with selecting from a Redux slice.)

### Fix 2a — preventing `GET /user_farm/user/undefined`

The adjust in the saga just exits `getUserFarmsSaga` if there is no `user_id` to form the request with

```js
// containers/ChooseFarm/saga.js — getUserFarmsSaga
const { user_id } = yield select(loginSelector);
if (!user_id) {
  return; // <-- early return before adjusting the loading state or calling the API
}
const header = getHeader(user_id);
yield put(onLoadingUserFarmsStart());
const result = yield call(axios.get, userFarmUrl + '/user/' + user_id, header);
```

### Fix 2b — the route now errors on `GET /user_farm/user/undefined`

Previously this would have returned `200 []`; now it would return 403 (if it were called!)

```js
// packages/api/src/routes/userFarmRoute.js
- router.get('/user/:user_id', userFarmController.getUserFarmByUserID());
+ router.get('/user/:user_id', isSelf, userFarmController.getUserFarmByUserID());
```

This change in isolation does trigger `logout()` via `handle403Saga` and therefore works on its own, but the failed request will also set `loaded: true` which would push the user to `/welcome` briefly before logout; therefore pairing with the saga change (never sending the malformed request at all) is preferable.

Jira link: https://lite-farm.atlassian.net/browse/LF-5443

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

As described in Jira, the two-tab method is the quickest and most reliable to differentiate this branch and integration:

Tab 1: open LiteFarm, stop at the sign-on screen, do not sign in.
Tab 2, same window: open LiteFarm and sign in normally. Do not choose a farm (note: this is also where the dashboard redirect would naturally leave you)
Tab 1: type any valid email and press Continue.
Tab 1 redirects itself to Welcome

**Checklist:**

### DIFF
--- a/packages/api/src/routes/userFarmRoute.js
+++ b/packages/api/src/routes/userFarmRoute.js
@@ -25,8 +25,7 @@ import checkInvitationTokenContent from '../middleware/acl/checkInviteTokenConte
 import checkUserFarmStatus from '../middleware/acl/checkUserFarmStatus.js';
 
 // Get all userFarms for a specified user
-// no permission limits
-router.get('/user/:user_id', userFarmController.getUserFarmByUserID());
+router.get('/user/:user_id', isSelf, userFarmController.getUserFarmByUserID());
 
 // Get info on all users (userFarm) at a farm
 router.get(
--- a/packages/api/tests/userFarm.test.js
+++ b/packages/api/tests/userFarm.test.js
@@ -40,8 +40,13 @@ import userModel from '../src/models/userModel.js';
 describe('User Farm Tests', () => {
   // let middleware;
 
-  function getUserFarmsOfUserRequest({ user_id }, callback) {
-    chai.request(server).get(`/user_farm/user/${user_id}`).end(callback);
+  // The mocked checkJwt above reads the authenticated user from the user_id header, so that header
+  // is what stands in for the caller's token
+  function getUserFarmsOfUserRequest({ user_id, requesting_user_id = user_id }) {
+    return chai
+      .request(server)
+      .get(`/user_farm/user/${user_id}`)
+      .set('user_id', requesting_user_id);
   }
 
   // note: the object that is sent should be adjusted to not include consent_version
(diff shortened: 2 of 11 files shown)


SayakaOno (approved):
Looks good, and I appreciated the perfect reproduction steps!
It seems like my app was in a weird state several times, and it took some time to test, but I believe this is a good fix!
(I didn't fully understand how the tests work 😅)

PR #4314, opened 2026-08-17
LF-5462 Take the new FAO TAPE in LiteFarm

kathyavini:
**Description**

This updates LiteFarm to handle the new **FAO Tape Step 0 & 1**.

The changes are:
1. Display the CAET scores directly, rather than via calculation, as they are now calculated within the survey and embedded in the survey responses. The keys of the survey-calculated CAET scores have been added to DIMENSIONS as `scoreField`
2. Dimension labels are now translation keys so they can be shown in other languages (needed to be done; not particularly a requirement for the new version)
3. Allows the user to finish their legacy FAO TAPE if they happen to have an in-progress draft in their browser. Since drafts are frontend-only on current production, as of now a draft cannot be retained for more than a week and it's not likely many users have one. But this anticipates the general rule we will build for all surveys, where drafts can always be completed on the version they were started with (*)(**)
4. Updates the logic that pre-populates the location, now also filling the correct world region for the new survey

1, 3, and 4 all branch on whether the survey follows the new schema introduced here -OR- the original logic, since the Australia TAPE from the Mornington research project is unchanged.

(*) As discussed Mon Aug 17 with the data team, the desired behaviour for draft invalidation needs to be confirmed with FAO. But either way, drafts must be either discarded (implemented) or reconciled with old keys removed, or both sets of keys are passed to the results backend in the survey response.
(**) Update Aug 18: the future-looking behaviour requested was actually neither draft invalidation or reconciliation, _but letting the user finish their in-progress draft_! That's a bit of a complication because we had initially decided against CDN versioning, and this will require going back on that. So draft versioning will become its own ticket. This ticket implements letting users finish the original FAO TAPE if they had started one only.

Jira link: https://lite-farm.atlassian.net/browse/LF-5462

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

The easiest (and I also think release-accurate) way to see the draft pinning is to checkout `integration`, begin the old survey, and then check out this branch.

**Checklist:**


packages/webapp/src/containers/Insights/Survey/index.tsx
```diff
@@ -48,36 +49,53 @@ function Survey({ isCompactSideMenu }: SurveyProps) {
   // @ts-expect-error - userFarmSelector is not typed with TypeScript yet
   const { farm_id, country_code } = useSelector(userFarmSelector);
 
-  const surveyVersion = getSurveyVersion(surveyId, country_code);
+  const cdnFileName = getCdnFileName(surveyId, country_code);
```
kathyavini: Renamed these variables for accuracy since it's the other one (the new one passed to `getSurveyCdnPath`) that actually tracks the `survey_version` value in the survey JSON.

I did leave `versionsByCountry` untouched in the config object (i.e. didn't update to `cdnFileNamesByCountry`) but let me know if that is confusing and maybe should be updated too.

kathyavini: Just FYI, the commit 7ffd8786737625039eb2357e2908d61c051cb00d which removed draft invalidation changed a lot of lines and made this comment outdated; however the rename still stands as shown in the diff above.

packages/webapp/src/containers/Insights/Survey/surveyConfig.ts
```diff
@@ -55,27 +55,90 @@ export const SURVEY_INFO: Record<string, SurveyInfo> = {
...
+    surveyId === 'tape' &&
+    hasDraft &&
+    usesGlobalSurvey &&
+    (draftSurveyVersion === undefined || draftSurveyVersion === LEGACY_FAO_SURVEY_VERSION);
+
+  if (isLegacyFaoDraft) {
+    return 'fao';
+  }
+
+  return latest;
```
kathyavini: In tech daily I showed a version of this that already went ahead and did the upcoming proposed draft pinning:

```ts
// Pinned draft fallback to be completed in LF-5473
if (hasDraft && draftSurveyVersion !== undefined && latest) {
  return `${latest}/${draftSurveyVersion}`;
}
```

That was premature... that logic has to be added with the code that handles a fallback to latest when that draft file doesn't exist (currently scoped for [LF-5473](https://lite-farm.atlassian.net/browse/LF-5473)), or it fails immediately on the non-FAO surveys as soon as their survey version is written to the `surveyDraftSlice`, e.g. looking for `idems/cathi-gao/{current-cathi-gao-version}.json`

SayakaOno (approved):
I think both survey types are well handled with minimal changes. Great job!

PR #4315, opened 2026-08-18
LF-5448: Take tape in other languages

SayakaOno:
**Description**

Adds localized-language support for the TAPE survey's base version, with a fallback to the default English file when no translation exists for the user's preferred language.

**The new flow:**
1. When the farm's country code is `AU`, the behaviour is unchanged.
2. When user's preferred language isn't English, fetch `fao_{language}.json`.
3. If the file isn't returned, fetch the English version: `fao.json`.

Adding a new language later needs no code changes — just upload `fao_<lang>.json` to the DO bucket.

**Changes:**
- Update `getSurveyVersion` to return `version` and `fallbackVersion`.
- Update `surveyApi` to fetch the English version if the file for the user's preferred language isn't returned.
- Add a Spinner to the component.
- Fix the `min-height` of `insightContainer` (which was `100vh`, ignoring the navbar). This was needed to centre the spinner, and also prevents insight pages (e.g. /insights) from becoming unnecessarily scrollable.

Jira link: https://lite-farm.atlassian.net/browse/LF-5448

**Type of change**

**How Has This Been Tested?**

Added `fao_es.json` (the English version with the title translated to Spanish) to the DigitalOcean bucket, tested the behaviour with the language preferences `es` and `fr`, and observed the Network tab. For supported languages, the API call happens once and succeeds. For unsupported languages, the first call (`fao_{language}.json`) fails, followed by a second call for `fao.json`.

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/containers/Insights/Survey/index.tsx
+++ b/packages/webapp/src/containers/Insights/Survey/index.tsx
@@ -132,7 +132,7 @@ function Survey({ isCompactSideMenu }: SurveyProps) {
       <PageTitle title={surveyTitle} backUrl="/Insights" />
       <div className={clsx(styles.surveyContainer, isCompactSideMenu && styles.compactSideMenu)}>
         {/* wait for prepopulated data and survey JSON to load */}
-        {isLoading && !surveyJson && (
+        {isLoading && (
           <div className={styles.spinner}>
             <Spinner />
           </div>
--- a/packages/webapp/src/containers/Insights/Survey/index.tsx
+++ b/packages/webapp/src/containers/Insights/Survey/index.tsx
@@ -48,7 +48,7 @@ function Survey({ isCompactSideMenu }: SurveyProps) {
   // @ts-expect-error - userFarmSelector is not typed with TypeScript yet
   const { farm_id, country_code } = useSelector(userFarmSelector);
 
-  const surveyVersion = getSurveyVersion(surveyId, country_code);
+  const { version: surveyVersion } = getSurveyVersion(surveyId, country_code) || {};
   const cdnDirectory = SURVEY_INFO[surveyId]?.cdnDirectory;
 
   const { prepopulatedData, isLoading: isPrepopulatedDataLoading } =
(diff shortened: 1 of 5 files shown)


kathyavini (commented):
@SayakaOno what I think the data team will now set up on Digital Ocean is not `fao_es.json` but rather `fao_es/step0_step1.json` (i.e. a folder per language), with all the interior files having the same filename.

Sorry this is what I meant in [Slack here](https://litefarm.slack.com/archives/C0BDN4P4H9A/p1786986929542899?thread_ts=1786654725.351669&cid=C0BDN4P4H9A) but I should have written something in the Jira ticket too.

Is that a lot more complicated to implement? From our end it was just to prevent having 45 files (9 surveys x 4 languages) in the base folder, but I think Norinne said it was simpler from their end too to do separate folders and not suffixed files. That said, they are outputing their translations via script, so if the suffix is necessary we can let them know and I don't think it should be too hard to pivot.

SayakaOno:
@kathyavini No worries, I misunderstood and thought this one and the new folder structure wouldn't interfere with each other. I'll adjust the code!

kathyavini (commented):
Sorry for the file path mixup 🙇

Otherwise this all looks really good to me! I like the 403/404 method for fallback a lot. Mmmmm it's so delicious when we don't have to release code to add things 😁

packages/webapp/src/containers/Insights/Survey/index.tsx
```diff
@@ -125,6 +132,11 @@ function Survey({ isCompactSideMenu }: SurveyProps) {
       <PageTitle title={surveyTitle} backUrl="/Insights" />
       <div className={clsx(styles.surveyContainer, isCompactSideMenu && styles.compactSideMenu)}>
         {/* wait for prepopulated data and survey JSON to load */}
+        {isLoading && !surveyJson && (
```
kathyavini: Oh -- one minor thing, and it doesn't much bother me, but the `&& !surveyJson` here means that the spinner doesn't load on a second visit (with the survey in cache), even though the `isLoading` will still be true from the call for the pre-populated data.

Quick demo of what I mean is
1. load survey
2. set to slow network like 3G
3. menu navigate out and back
4. (Result) neither spinner nor survey

It's really the maps geocode and not the survey that is loading, but since the survey is blocked on getting that data, I guess it feels equivalent and should probably have the same spinner? (But that's also exactly how it was before the spinner was added, so... a very minor thing imo.)

SayakaOno:
As discussed this morning, the code was adjusted to fetch `fao_{language}/step0_step1_{language}.json` for a localized survey.

kathyavini (approved):
It works really well!

Thank you for being flexible with the changing requirements. Hopefully no other surprises from here through release!

PR #4319, opened 2026-08-24
LF-5476 Survey validation error and navigation buttons remain in English when LiteFarm is set to French, Portuguese, or Spanish

kathyavini:
**Description**

To get localized UI (inbuilt validation and buttons, etc), we have to import the dictionary files for the SurveyJS UI in other languages, see [Survey Localization | SurveyJS](https://surveyjs.io/form-library/examples/survey-localization/documentation). Each survey is already configured with a `locale: es` (etc.) property at the top so that does the switching for us, once the dictionaries are available.

Adding three language dictionaries without other configuration changes added 5.4 kB gz to `survey-vendor`, as measured with `bundle-compare`, my new favourite tool. It's not very much, but it means adding a language will re-hash survey vendor.

Therefore I have created a new manual chunk to live alongside survey-vendor, called `survey-locales`. This holds just the language dictionaries, so that adding new languages in the future (e.g. German or Hindi if those are requested; note that pa/ml/km are not available) will be a quick invalidation of that small chunk only.

Since we know we will be adding Italian this month, it is also imported now.

Jira link: https://lite-farm.atlassian.net/browse/LF-5476

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

Used `bundle-compare` to observe changes in chunk sizes with language import, with and without the separate locales chunk.

Built and previewed the webapp to verify cached chunks:

Screenshot 2026-08-28 at 9 54 52 AM

**Checklist:**


SayakaOno (approved):
Looks good to me!

PR #4320, opened 2026-08-25
LF-5432 Re-export three oversized raster in svg illustrations at display size

kathyavini:
**Description**

These three SVGs were raster images wrapped inside SVGs and the larger two were 432 KB (😭) and 142 KB respectively.

Loïc kindly re-vectorized them; they were simplified a little in that process (you can compare the before and after below) but I would say that small loss of detail is _well_ worth it... the new files are now 28 KB and 13 KB respectively instead.

bundle-compare estimates the savings as 14.5% of our precache

Screenshot 2026-08-25 at 1 18 37 PM

As a small bonus, they are no longer in-lined into the JS (i.e. no longer imported with `?react`) so they can keep a stable hash between releases. This doesn't change the precache bundle size, but it will be a re-release savings.

Jira link: https://lite-farm.atlassian.net/browse/LF-5432

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

I used `bundle-snapshot` and `bundle-compare` between integration and this branch.

Two of the 3 components can be checked out in Storybook: [Farm Switch Modal](http://localhost:6006/?path=/docs/components-modals-farmswitchmodal--docs) and [Add Farms Summary](http://localhost:6006/?path=/story/components-addanimalssummarycard--default)

I don't think the Unknown Record page has a Storybook? So the emoji I inspected with

```
http://localhost:3000/tasks/<a nonsense number>/read_only
```

**Checklist:**


packages/webapp/src/components/Table/TableV2.jsx
```diff
@@ -15,8 +15,7 @@
 import { useState, useMemo } from 'react';
 import PropTypes from 'prop-types';
 import { useTranslation } from 'react-i18next';
-import { Checkbox } from '@mui/material';
-import Box from '@mui/material/Box';
+import { Box, Checkbox } from '@mui/material';
```
kathyavini, line 18: Never really figured out why, but Storybook started crashing for me with the error `createTheme_default is not a function` after the last few PRs.

Claude says the optimizer wrote a bad chunk for the MUI Box component that doesn't call `createTheme()` and this is the fix, but couldn't trace it to a commit or any of the recent code changes. Fix does work for me though to restore Storybook.

SayakaOno (approved):
🎉

PR #4322, opened 2026-08-27
LF-5471: Persist survey drafts to the server

SayakaOno:
**Description**

Persists survey progress to the server as the farmer fills out the survey, and resumes from the server instead of trusting whatever's sitting in local Redux.

**What changed initially (see the comments for subsequent changes)**

**Backend:**

* `GET /survey_drafts/:survey_key/:survey_step?` — fetch the farm's current draft (or `null` if there isn't one).
* `PUT /survey_drafts/:survey_key/:survey_step?` — create-or-replace the draft at that key (idempotent upsert).
* Submitting a survey (`POST /survey_responses`) now soft-deletes the live draft in the same transaction and reuses its `submission_id`, so the finished response and the draft that produced it are traceably linked.
* A draft write is rejected (`409`) once a response already exists for that `survey_key`/`survey_step`.

**Frontend:**

* Redux store shape: drafts are now keyed by `survey_step` in addition to `survey_id` `(bySurveyId[surveyId][step]` instead of `bySurveyId[surveyId]`), so a multi-step survey can have an independent in-progress draft per step.
* `useInitialDraft` — on mount, fetches the server's draft and reconciles it against whatever's in local Redux (comparing `updated_at` timestamps), resolving *once* which one wins. It always forces a fresh server check on mount rather than trusting a stale cache, since this decision only gets made once per mount.
* `useSurveyDraftSync` — on mount, reconciles local vs. server draft (adopts the server's, discards a stale local one, or registers a never-before-saved local one), and thereafter saves progress to the server as the farmer works (on page change and on unmount).
* The Insights page tile for a survey (`SurveyInsightTile`) now shows "in progress" based on this same server-reconciled state instead of local-only Redux.

Jira link: https://lite-farm.atlassian.net/browse/LF-5471

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/surveyResponseController.ts
+++ b/packages/api/src/controllers/surveyResponseController.ts
@@ -23,6 +23,7 @@ import SurveyDraftModel from '../models/surveyDraftModel.js';
 interface SurveyResponseData {
   survey_version: string;
   project_id: string;
+  survey_step?: string;
   [key: string]: unknown;
 }
 
@@ -60,7 +61,7 @@ const surveyResponseController = {
           await trx.rollback();
           return res.status(400).json({ error: 'survey_key is required' });
         }
-        const { survey_version, project_id } = survey_response;
+        const { survey_version, project_id, survey_step } = survey_response;
 
         // A live draft, if one exists, hands its submission_id over to the response so a draft
         // write that arrives after this point can recognize its own lineage is now complete.
@@ -79,6 +80,7 @@ const surveyResponseController = {
           survey_response,
           survey_version,
           project_id,
+          survey_step,
         });
 
         if (liveDraft) {
--- a/packages/api/src/controllers/surveyResponseController.ts
+++ b/packages/api/src/controllers/surveyResponseController.ts
@@ -109,7 +109,7 @@ const surveyResponseController = {
           .where({ farm_id, survey_key })
           .orderBy('created_at', 'desc')
           .first();
-        return res.status(200).send(result ?? null);
+        return res.status(200).json(result ?? null);
       } catch (error) {
         console.error(error);
         return res.status(500).json({ error });
--- a/packages/api/src/controllers/surveyResponseController.ts
+++ b/packages/api/src/controllers/surveyResponseController.ts
@@ -23,7 +23,6 @@ import SurveyDraftModel from '../models/surveyDraftModel.js';
 interface SurveyResponseData {
   survey_version: string;
   project_id: string;
-  survey_step?: string;
   [key: string]: unknown;
 }
 
@@ -61,14 +60,14 @@ const surveyResponseController = {
           await trx.rollback();
           return res.status(400).json({ error: 'survey_key is required' });
         }
-        const { survey_version, project_id, survey_step } = survey_response;
+        const { survey_version, project_id } = survey_response;
 
         // A live draft, if one exists, hands its submission_id over to the response so a draft
         // write that arrives after this point can recognize its own lineage is now complete.
         /* @ts-expect-error known issue with models */
         const liveDraft = await SurveyDraftModel.query(trx)
           .whereNotDeleted()
-          .findOne({ farm_id, survey_key, survey_step: survey_step ?? '' });
+          .findOne({ farm_id, survey_key });
 
         const insertQuery = SurveyResponseModel.query(trx).context({
           user_id,
@@ -80,7 +79,6 @@ const surveyResponseController = {
           survey_response,
           survey_version,
           project_id,
-          survey_step,
         });
 
         if (liveDraft) {
--- a/packages/api/src/controllers/surveyResponseController.ts
+++ b/packages/api/src/controllers/surveyResponseController.ts
@@ -14,8 +14,11 @@
  */
 
 import { Response } from 'express';
+import { transaction, Model } from 'objection';
 import { LiteFarmRequest } from '../types.js';
+import { handleObjectionError } from '../util/errorCodes.js';
 import SurveyResponseModel from '../models/surveyResponseModel.js';
+import SurveyDraftModel from '../models/surveyDraftModel.js';
 
 interface SurveyResponseData {
   survey_version: string;
@@ -49,19 +52,29 @@ const surveyResponseController = {
       req: LiteFarmRequest<unknown, unknown, unknown, CreateSurveyResponseReqBody>,
       res: Response,
     ) => {
+      const trx = await transaction.start(Model.knex());
       try {
         const { farm_id } = req.headers;
         const user_id = req.auth?.user_id;
         const { survey_key, survey_response } = req.body;
         if (!survey_key) {
+          await trx.rollback();
           return res.status(400).json({ error: 'survey_key is required' });
         }
         const { survey_version, project_id, survey_step } = survey_response;
 
-        const insertQuery = SurveyResponseModel.query().context({
+        // A live draft, if one exists, hands its submission_id over to the response so a draft
+        // write that arrives after this point can recognize its own lineage is now complete.
+        /* @ts-expect-error known issue with models */
+        const liveDraft = await SurveyDraftModel.query(trx)
+          .whereNotDeleted()
+          .findOne({ farm_id, survey_key, survey_step: survey_step ?? '' });
+
+        const insertQuery = SurveyResponseModel.query(trx).context({
           user_id,
         }) as unknown as InsertableQuery;
         await insertQuery.insert({
+          ...(liveDraft ? { submission_id: liveDraft.submission_id } : {}),
           farm_id,
           survey_key,
           survey_response,
@@ -70,10 +83,15 @@ const surveyResponseController = {
           survey_step,
         });
 
+        if (liveDraft) {
+          /* @ts-expect-error known issue with models */
+          await SurveyDraftModel.query(trx).context({ user_id }).deleteById(liveDraft.id);
+        }
+
+        await trx.commit();
         return res.status(201).send();
       } catch (error) {
-        console.error(error);
-        return res.status(500).json({ error });
+        return await handleObjectionError(error as Error, res, trx);
       }
     };
   },
--- a/packages/api/src/routes/surveyDraftRoute.ts
+++ b/packages/api/src/routes/surveyDraftRoute.ts
@@ -23,11 +23,9 @@ import surveyDraftController from '../controllers/surveyDraftController.js';
 
 const router = express.Router();
 
-// One live draft per farm_id + survey_key + survey_step (survey_step is optional,
-// and it defaults to '' on the survey_draft row).
+// One live draft per farm_id + survey_key
 router.get(
-  // TODO: LF-5475 Update the Express 4-specific `?` optional-param syntax to `{/:survey_step}`
-  '/:survey_key/:survey_step?',
+  '/:survey_key',
   checkScope(['get:survey_draft']),
   checkSurveyDraftKey(),
   surveyDraftController.getDraft(),
@@ -36,8 +34,7 @@ router.get(
 // Idempotent create-or-replace at this exact key — PUT, not POST, since the resource's
 // address is fully known up front rather than server-assigned.
 router.put(
-  // TODO: LF-5475 Update the Express 4-specific `?` optional-param syntax to `{/:survey_step}`
-  '/:survey_key/:survey_step?',
+  '/:survey_key',
   checkScope(['edit:survey_draft']),
   checkSurveyDraftKey(),
   checkDraftNotCompleted(),
--- a/packages/api/src/routes/surveyDraftRoute.ts
+++ b/packages/api/src/routes/surveyDraftRoute.ts
@@ -1,0 +1,43 @@
+/*
+ *  Copyright 2026 LiteFarm.org
+ *  This file is part of LiteFarm.
+ *
+ *  LiteFarm is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  LiteFarm is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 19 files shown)


packages/api/src/controllers/surveyResponseController.ts
```diff
@@ -70,10 +83,15 @@ const surveyResponseController = {
...
+          /* @ts-expect-error known issue with models */
+          await SurveyDraftModel.query(trx).context({ user_id }).deleteById(liveDraft.id);
+        }
+
+        await trx.commit();
         return res.status(201).send();
       } catch (error) {
-        console.error(error);
-        return res.status(500).json({ error });
+        return await handleObjectionError(error as Error, res, trx);
```
SayakaOno: Claude wrote this, but I'm not sure about the proper error handling... I'll confirm that later!

packages/api/src/routes/surveyDraftRoute.ts
```diff
@@ -0,0 +1,47 @@
...
+  checkDraftNotCompleted,
+} from '../middleware/validation/checkSurveyDraft.js';
+import surveyDraftController from '../controllers/surveyDraftController.js';
+
+const router = express.Router();
+
+// One live draft per farm_id + survey_key + survey_step (survey_step is optional,
+// and it defaults to '' on the survey_draft row).
+router.get(
+  // TODO: LF-5475 Update the Express 4-specific `?` optional-param syntax to `{/:survey_step}`
```
SayakaOno: https://stackoverflow.com/a/41748728

packages/api/src/routes/surveyDraftRoute.ts
```diff
@@ -0,0 +1,47 @@
...
+  '/:survey_key/:survey_step?',
+  checkScope(['get:survey_draft']),
+  checkSurveyDraftKey(),
+  surveyDraftController.getDraft(),
+);
+
+// Idempotent create-or-replace at this exact key — PUT, not POST, since the resource's
+// address is fully known up front rather than server-assigned.
```
SayakaOno, line 35: https://stackoverflow.com/questions/18470588/in-rest-is-post-or-put-best-suited-for-upsert-operation

packages/api/src/controllers/surveyResponseController.ts
```diff
@@ -91,7 +109,7 @@ const surveyResponseController = {
           .where({ farm_id, survey_key })
           .orderBy('created_at', 'desc')
           .first();
-        return res.status(200).send(result ?? null);
+        return res.status(200).json(result ?? null);
```
SayakaOno, line 112: Return `null` instead of `''`(empty string)

packages/webapp/src/containers/Insights/Survey/useInitialDraft.tsx
```diff
@@ -0,0 +1,105 @@
...
+        submissionId: string | undefined;
+        surveyVersion: string | undefined;
+        surveyData: Record<string, any>;
+        currentPageNo: number;
+        updatedAt: number | undefined;
+        needsLocalSync: boolean;
+      };
+    };
+
+function useInitialDraft(surveyId: string, surveyStep?: string) {
```
SayakaOno: The hook doesn't flag when there's already a completed survey. (It should work when the update/retake feature is implemented -- I accidentally covered the update/retake cases.)

packages/webapp/src/store/api/surveyApi.ts
```diff
@@ -80,12 +105,32 @@ export const surveyApi = api.injectEndpoints({
...
+      providesTags: (_result, _error, { surveyKey, surveyStep }) => [
+        { type: 'SurveyDraft', id: formatSurveyKeyStep(surveyKey, surveyStep) },
+      ],
+    }),
+    upsertSurveyDraft: build.mutation<SurveyDraftRecord, UpsertSurveyDraftReqBody>({
+      query: ({ surveyKey, surveyStep, ...body }) => ({
+        url: getSurveyDraftUrl(surveyKey, surveyStep),
+        method: 'PUT',
+        body,
+      }),
```
SayakaOno, line 113: `invalidatesTags` would trigger a redundant refetch after every autosave PUT (the same component is always subscribed), so it was removed. Freshness on mount is guaranteed separately via `refetchOnMountOrArgChange: true`, which forces a fresh `GET` every time regardless of cache age.

packages/webapp/src/containers/Insights/Survey/useSurveyDraftSync.tsx
```diff
@@ -0,0 +1,143 @@
...
+        }).unwrap();
+
+        submissionIdRef.current = created.submission_id;
+
+        dispatch(
+          setDraftSubmissionId({ surveyId, surveyStep, submissionId: created.submission_id }),
+        );
+      } catch (error) {
+        if (isFetchBaseQueryError(error) && error.status === 409) {
+          // TODO: Handle 409 (survey already completed)
```
SayakaOno: TODO: Confirm the behaviour. It's such an edge case that it may be fine to ignore, but we could display an error snackbar or redirect the user to the results page.

kathyavini (commented):
I was a little late to wrap up everything else so I haven't finished reading yet, but as far as testing, this works SO WELL!!! 😍 I've done a lot of cross-device testing too and it persists and saves and clear the drafts exactly as it's supposed to.

And ahhh I'm so sorry, I missed a very important line in your Jira comment which I think we'll want to circle back on next week, about `survey_step`, though.

It's not actually true that the `survey_key` for the step 2 modules will be `tape`! I'm sorry that's the part that I missed, and that's not recorded in any shared document now except for a completely hidden [AI generated Jira ticket](https://lite-farm.atlassian.net/browse/LF-5464) 😞

I think we can actually just key on `survey_key` and `farm_id` therefore, and I'm sorry because I can see it was rather quite complex to add the additional layer 😢 But I could be missing something on this so let's sync up on Monday morning!

SayakaOno:
I've removed `survey_step` to treat `survey_key` as the identifier.

@kathyavini
I've finished the code changes and will be testing this afternoon. Thank you! 🙏

packages/webapp/src/containers/Insights/Survey/useInitialDraft.tsx
```diff
@@ -0,0 +1,120 @@
...
+  // back to it instead of reading as "no draft".
+  const { data: cachedServerDraft } = useGetSurveyDraftQuery(
+    { surveyKey: surveyId },
+    { skip: !surveyId },
+  );
+
+  const [resolved, setResolved] = useState<InitialDraftResult>(loadingResult);
+
+  // Resolves once on mount. Uses the lazy trigger, not the normal query hook, since a normal
+  // subscription can briefly show a stale cached value (e.g., isFetching: false).
```
SayakaOno, line 44: I noticed this while visiting `/insights` a second time: `SurveyInsightTile`'s survey status was resolving against stale cached data before the forced refetch with `{ refetchOnMountOrArgChange: true }` had even gone out.

SayakaOno:
I've made two changes after un-drafting the PR (I'm sorry if that was annoying):
- added `survey_step` back to response controller, which I shouldn't have removed
- addressed the TODO comment (instead of handling the errors, we'll report them to Sentry)

The auto-save indicator is still missing, but I'm not sure when the design will be finalized.

packages/webapp/src/containers/Insights/Survey/useInitialDraft.tsx
```diff
@@ -0,0 +1,120 @@
...
+  useEffect(() => {
+    if (!surveyId) {
+      return;
+    }
+    let cancelled = false;
+    (async () => {
+      const serverDraft = await fetchDraft({ surveyKey: surveyId })
+        .unwrap()
+        .catch(() => cachedServerDraft);
```
kathyavini: _I **think** that with the change below, `cachedServerDraft` and the second query would not be necessary anymore, but can you double-check that it covers the situation that warranted having both the lazy and regular query?_

-----

This was originally just an AI catch for data overwrite in a particular sequence with a failed network request:

1. Load a fresh survey so the surveyDraft is fetched (correctly) as `null`
2. Fill in a few answers and navigate away to initiate the PUT (RTK Query draft remains `null`, since as noted in the slice the update doesn't invalidate)
3. (Block the GET in Network tools to simulate a dropped request) & return to survey

--> `if (isLocalDraftStale && !serverDraft)` block fires (since serverDraft is still `null`), so PUT is sent with `surveyData: {}` overwriting the previous server-saved response. Local slice draft also cleared.

I'm always a bit skeptical of bugs that require a _particular_ network request to fail, but in spotty network that seemed... possible, and the protection against it looked pretty minimal and reasonable:

```ts
const { data, isSuccess } = await fetchDraft({ surveyKey: surveyId });
// cancelled part the same
const serverDraft = isSuccess ? data : cachedServerDraft;
// isLocalDraftStale part same
// Only send the empty `{}` PUT at all on a failed fetch
if (isLocalDraftStale && !serverDraft && isSuccess) {
```

After looking at that, though, I am also now thinking that with that guard, it's actually not necessary to keep the `cachedServerDraft`? Because even with simply

```ts
const serverDraft = isSuccess ? data : undefined;
```

a server fetch failure would then fall back to the local draft, and I can't think of a situation where `cachedServerDraft` would have gotten ahead of the local redux slice. But the flow is complex so please let me know if I'm missing something!

packages/webapp/src/containers/Insights/Survey/useSurveyDraftSync.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+  const submissionIdRef = useRef(initialDraft.submissionId);
+
+  // Upserts the draft to the server, then syncs the returned submission_id into Redux.
+  const persistDraft = useCallback(
+    async (
+      payload: { survey_data: Record<string, any>; current_page_no?: number },
+      { shouldReportErrors = false }: { shouldReportErrors?: boolean } = {},
+    ) => {
+      if (!surveyVersion) {
```
kathyavini: I don't recall if I remember seeing this on Friday, but right now navigating to a new survey and then away (without answering any questions) sends a PUT with `survey_version` and `survey_id` but no `survey_data`, which fails with Model Validation Error:

```jsonc
{
  "error": {
    "name": "ValidationError",
   // ...
    "error": {
      "message": "survey_data: must have required property 'survey_data'"
    }
  }
}
```

I think it would be safe to also early return in this case, when `!payload.survey_data`

packages/webapp/src/containers/Insights/Survey/useSurveyDraftSync.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+  useEffect(() => {
+    return () => {
+      persistDraft({
+        survey_data: latestDraftRef.current.surveyData,
+        current_page_no: latestDraftRef.current.currentPageNo,
+      });
+    };
+    // persistDraft must be a dependency, or the cleanup stays frozen on the mount-time closure
+    // where surveyVersion is still undefined, and no-ops forever.
```
kathyavini, line 94: Claude's `/code-review` skill flagged this for me last week, saying that the PUT on unmount wouldn't fire, but I couldn't reproduce it!! It always fired for me!

Is there a trick to triggering the no-op?

packages/webapp/src/containers/Insights/Survey/useSurveyDraftSync.tsx
```diff
@@ -0,0 +1,154 @@
...
+            updatedAt: initialDraft.updatedAt,
+          }),
+        );
+        return;
+      }
+
+      // Local's submission_id points to a draft that's already been completed, and no new server
+      // draft replaced it — discard the stale local content rather than keep building on it.
+      dispatch(clearSurvey({ surveyId }));
+      persistDraft({ survey_data: {}, current_page_no: 0 });
```
kathyavini: Will this `persistDraft` call always 409, then, according to the comments? Should it either be logged (like the other one) or skipped?

To me this 409 is actually a bit more alarming than the one on page change -- if the survey is reloaded after it's completed elsewhere, (which isn't... the most outrageous thing to do, if you were trying to get the updated status on your first device) this would present a cleared out survey and no redirect to `/complete` unless you return to Insights.

I would _almost_ want to add a manual redirect for that case, but it seems that Loïc isn't fussed about these edge cases, so I, too, will let it go 😅 It's crazy how many different worrisome states you can create with just two devices, though!

kathyavini (commented):
Wow, I have to say, despite how well-organized both of those beautiful hooks were, the potential states involved in this one were almost too complex for my mind to fathom! 🤯 Really lovely job getting it all handled.

I used to see a dropped draft on reload with a fresh survey, but I see you have fixed that, and everything I had been thinking about on my first pass is already covered (and then some)! 🙌

packages/webapp/src/containers/Insights/Survey/useSurveyDraftSync.tsx
SayakaOno: I thought I'd fixed it! Thank you for catching that!!

packages/webapp/src/containers/Insights/Survey/useSurveyDraftSync.tsx
SayakaOno: I didn't have `persistDraft` as dependency initially. Without it, it'll be no-op most of the time!

packages/webapp/src/containers/Insights/Survey/useSurveyDraftSync.tsx
SayakaOno: Ahhh, this was ready for retake, and I missed revisiting it!
I've added a temporary redirect; I'm with you 😅

packages/webapp/src/containers/Insights/Survey/useInitialDraft.tsx
SayakaOno: That's an awful bug! Thank you for catching it!!!

SayakaOno (commented):
Thank you for your comments!
Could you double-check that I've addressed them correctly? 🙏

kathyavini (approved):
Looks great! ❤️

PR #4324, opened 2026-08-28
LF-5478 Add pure components for TAPE step 2 modules

kathyavini:
**Description**

This adds the TAPE Step 2 Components and their Storybook stories. Nothing is yet imported into app.
- Speedometer gauge [[Storybook](http://localhost:6006/?path=/docs/components-surveymodulecard-gauge--docs)]
- Module card with all states shown in Figma [[Storybook](http://localhost:6006/?path=/docs/components-surveymodulecard--docs)]
- Legend explaining the colour bands [[Storybook](http://localhost:6006/?path=/story/components-surveybandlegend--default)]
- Full Section (grid of modules + legend) [[Storybook](http://localhost:6006/?path=/docs/components-surveymodulesection--docs)]

I'm not sure if that last component should serve as the Pure Component for this feature when the next ticket is coded, or if it can just be deleted and the components placed directly into the container. For now it allows observing the full view in Storybook including the responsive behaviour of the grid + legend together.

Jira link: https://lite-farm.atlassian.net/browse/LF-5478

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

Only viewed/tested in Storybook.

**Checklist:**


packages/webapp/public/locales/en/translation.json
```diff
@@ -1254,6 +1254,30 @@
...
+        "MEDIUM": "Medium",
+        "VERY_HIGH": "Very High",
+        "VERY_LOW": "Very Low"
+      },
+      "BAND_RANGE": {
+        "INCLUSIVE": "{{min}} – {{max}}%",
+        "GREATER_THAN_MIN": "> {{min}}% – {{max}}%"
+      },
+      "CARD": {
+        "COMPLETED_ON": "Completed · {{date}}",
```
SayakaOno, line 1269: I wonder if the dot is intentional here. "Last updated" and "started" don't have one 🤔

SayakaOno (approved):
Looks great!!

One thing that caught my eye was the alignment of the title and the new version label. I felt like the title deserved a bit more space...?
Image

I hadn't noticed until I played around with the new version label, but the card width changes depending on the content. I expected it to be fixed, so I figured I'd bring it up!

https://github.com/user-attachments/assets/19f4160e-8c0f-4e99-9b26-815289d61716

packages/webapp/public/locales/en/translation.json
kathyavini: Could ask in dev-design, but to me it actually looks correct because the other two read more like sentences, and "Completed" is the only one that stands alone as a state? Although I think maybe I'm also responding to the dot feeling like a full stop, so in that sense it goes with only Completed?

(But I'll be honest, I'm just partial to the look of the dot separator so I don't want to ask and get it removed 😂)

kathyavini:
Thank you @SayakaOno!! Those are all good points and I think they will need to go back to Loïc.

> One thing that caught my eye was the alignment of the title and the new version label. I felt like the title deserved a bit more space...?

Yes, I played around with a few different options and landed on that one, but I don't know why I did, it looks ridiculous! 😆 I've asked Loïc for a proper long-title + new version design [on Figma here](https://www.figma.com/design/eMffC0h3JPlWtzNttxJpil/Litefarm-Updated?node-id=81175-1774&m=dev#1914359604).

> but the card width changes depending on the content

Yes... I think this was my response to seeing the buttons wrap when I provided a fixed width and wanting to prevent that! The responsive behaviour I *totally* made up though (including using grid here instead of the usual flexbox), and ideally the whole responsive design needs to be brought to dev-design at some point to see what Loïc thinks.

I think in this case it would be super helpful to get all languages in first too, to see how those behave. Claude flagged that even with the current settings and the cards accommodating more space, there would be some languages that would _still_ overflow these cards, but I think it will be easiest to put off adjusting until we actually have all our strings in. Luckily all the titles are already in the JSONs, so at least that part could be done almost immediately with just a little Claude-led parse of the FAO files!

kathyavini:
Thank you @SayakaOno!! Those are all good points and I think they will need to go back to Loïc.

> One thing that caught my eye was the alignment of the title and the new version label. I felt like the title deserved a bit more space...?

Yes, I played around with a few different options and landed on that one, but I don't know why I did, it looks ridiculous! 😆 I've asked Loïc for a proper long-title + new version design [on Figma here](https://www.figma.com/design/eMffC0h3JPlWtzNttxJpil/Litefarm-Updated?node-id=81175-1774&m=dev#1914359604).

> but the card width changes depending on the content

Ah yes, this is responding to the changed button text. This was my response to seeing one of the buttons wrap when I provided a fixed width, and wanting to prevent that not, while not yet knowing how long the buttons would be when translated.

Adjusting the SCSS like this

```scss
// SurveyModuleSection/styles.module.scss
-  grid-template-columns: repeat(4, minmax(min-content, var(--card-max-width)));
+  grid-template-columns: repeat(4, minmax(0, var(--card-max-width)));
```

Would remove that size shift, but at the cost of a long button overflowing:

Screenshot 2026-09-04 at 3 11 47 PM

I think what would be ideal is getting in all the languages first, and then seeing if we can move to that default and maybe adjust the max-width accordingly to accommodate all languages?

Once all the strings are in, I wouldn't mind getting the responsive behaviour as a whole checked out, too... I kind of made it up (including the use of grid instead of flexbox) and I think it needs Loïc's approval at one point!

I'll merge to clean up + open the container ticket, but I expect a cleanup ticket for the styles will be needed once we hear back.

PR #4338, opened 2026-09-04
LF-5127 Take TAPE Step 2 surveys and see scores within LiteFarm

kathyavini:
**Description**

This PR adds the TAPE Step 2 Modules to the Tape Results Page. I confirmed with Riddhi on Friday that she is blocked on building the data dashboard's visualizations until she can take these surveys and see what the results look like.

It would therefore be good to coordinate with Riddhi and merge to beta a little early, and then let adjustments happen alongside her own testing (even after I get back next week), as long as the data in `survey_response` is likely not to change -- and it shouldn't, as it's not touched by any of the logic in this PR. I've put all the files on Digital Ocean so beta should work as soon as the code is merged.

### In this PR:

- (Only backend change) Adds two batching endpoints, used by the module/results view, for the survey responses and the survey drafts. This allows populating the modules with data from a single query instead of up to 18 (9 x 2).
  - The batched survey response endpoint returns the full survey data, as keys within that object are needed to show the scores
  - The batched draft endpoint only returns the properties needed to show the status of an in-progress survey, and omits `survey_data`
- Adds the functions given to us by the data team for use in the step 2 biodiversity survey
- Extracts the original TAPE Results Radar Chart calculations and display into their own component + utility file
- Gates the display of (and direct URL navigation to) the modules on having completed the FAO TAPE Step 1. Two modules are additionally gated on specific responses within FAO TAPE Step 1
  - To see all the modules, answer "What types of holding is this --> Household" and then put 1 or more in "Young female household members" as this will unlock both Dietary Diversity -- whose instructions read "The dietary diversity module should be answered by any woman (young or adult) in the household." -- and Youth
- Adds the Step 2 modules to surveyConfig, including the JSON-derived `pages` and `estimatedMinutes` as needed by the modules
  - Minutes are based on the question count, but are almost certainly poor estimates; those will need updating
- (Unconfigured but happens to work) The update button *kind of* works as a retake right now. The button will let you re-take a module (all the persisted drafts will error / 409 though) and create a fresh `survey_response` in the DB with a new `submission_id`. That might be useful to Riddhi as a way to generate a number of responses quickly, or she might not need to use it at all. If you could kindly let her know that is how it works, though, until real update is implemented 🙏

### What I did not get to / what is not included:
- the new version tag behaviour and display goes with the version-pinning ticket and not this one, as specified in the entrypoint in `claude-docs/`. All docs should all be up-to-date.
- I don't love the names and URLs for the two batched endpoints (drafts and responses), particularly the `/latest` but I couldn't think of a pleasing alternative in time
- need to check with Loïc if the page-based progress indicator is sufficient. It means one page surveys will show either not started, 50%, or complete. I considered a question-based progress indicator but this seemed simpler
- I'm not stoked that we'll have to keep the numbers of pages on the config object up-to-date with the surveys, but I think it's preferable to actually pulling the full survey JSONs into the results view only to grab the page count. Another alternative would be to store the number of pages with the draft (like a new column on `survey_drafts`) which I did not implement only because I didn't want to migrate at this time. But maybe that method would have been better?
- the last feature to be added and which I did not particularly well-review is the URL gating / redirect in `Insights/Survey/index.tsx` based on which modules are available; I have tested it and it works, but I was surprised at the amount of code logic considering that we already have a functional redirect on, e.g., cathi-gao survey for a non-Niger farm
- not sure if the reconciliation logic for `serverDraft` + `hasLocalDraft` ("in-progress" state) in `getSurveyState` is ideal

I hope that's about everything! Have a great week!

Jira link: https://lite-farm.atlassian.net/browse/LF-5127

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [ ] Other (please explain)

**Checklist:**

### DIFF
--- a/packages/webapp/src/containers/Insights/Survey/TapeResults.tsx
+++ b/packages/webapp/src/containers/Insights/Survey/TapeResults.tsx
@@ -17,50 +17,14 @@ import { useEffect } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 import { useTranslation } from 'react-i18next';
 import { useHistory } from 'react-router-dom';
-import { Radar } from 'react-chartjs-2';
-import {
-  Chart as ChartJS,
-  RadialLinearScale,
-  PointElement,
-  LineElement,
-  Filler,
-  Tooltip,
-} from 'chart.js';
 import styles from './styles.module.scss';
 import insightStyles from '../styles.module.scss';
 import { Semibold } from '../../../components/Typography';
 import PageTitle from '../../../components/PageTitle';
-import { roundToOne } from '../../../util/rounding';
+import TapeRadarChart from './TapeRadarChart';
+import { getTAPEDimensionScores } from './caetScores';
 import { useGetLatestSurveyResponseQuery } from '../../../store/api/surveyApi';
 import { enqueueErrorSnackbar, snackbarSelector } from '../../Snackbar/snackbarSlice';
-
-const CHART_COLOR = 'rgba(85, 143, 112, 1)'; // --Colors-Secondary-Secondary-green-700
-const CHART_FILL_COLOR = 'rgba(85, 143, 112, 0.2)'; // reduced opacity
-const MAX_SCORE = 100;
-const RAW_MAX_SCORE = 4;
-
-ChartJS.register(RadialLinearScale, PointElement, LineElement, Filler, Tooltip);
-
-const DIMENSIONS = [
-  { id: 'diversity', prefix: 'diversity_1', scoreField: 'div_score' },
-  { id: 'synergy', prefix: 'synergy_2', scoreField: 'synergy_score' },
-  { id: 'recycling', prefix: 'recycling_3', scoreField: 'recycling_score' },
-  { id: 'efficiency', prefix: 'efficiency_4', scoreField: 'efficiency_score' },
-  { id: 'resilience', prefix: 'resilience_5', scoreField: 'resilience_score' },
-  { id: 'cultureAndFood', prefix: 'culture_6', scoreField: 'cultfood_score' },
-  { id: 'cocreationAndKnowledge', prefix: 'knowledge_7', scoreField: 'cocrea_score' },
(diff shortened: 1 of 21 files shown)


packages/webapp/src/containers/Insights/Survey/TapeResults.tsx
```diff
@@ -92,125 +58,28 @@ function TAPEResults({ surveyId = 'tape' }: { surveyId?: string }) {
...
-        callbacks: {
-          label: (context: any) => ` ${context.label}: ${context.parsed.r}%`,
-        },
-      },
-    },
-  };
+  const openModule = useCallback(
+    (moduleSurveyId: string) => history.push(`/insights/survey/${moduleSurveyId}`),
+    [history],
+  );
```
SayakaOno, line 67: `useCallback` doesn't seem necessary here!?

Update: I've removed it in my PR!

SayakaOno (approved):
Looks great! I love `claude-docs` - it's so useful for understanding the thinking behind decisions!

Here are a few non-blocking points:
1. Endpoint response shape
    It's a little unusual to see a map as an API response, so I wondered whether an array response + formatting in `transformResponse` was considered. I'm not sure which approach we should go with, but I figured I'd mention it.
    I'm personally positive about `/latest`!
2. Progress indicator
    What about storing the actual progress percentage while the user is taking the survey? According to Claude, SurveyJS already computes this for us, so we'd get an accurate, question-based progress indicator.
3. Completed modules becoming unavailable
    A completed module can silently disappear from the results grid if the parent response later changes (e.g. household composition — `hh_women`/`hh_fyoung` dropping to 0 after a Step 1 retake) such that `isAvailable` now returns `false`.
    From the user's point view, the card would disappear with no explanation. From the researcher's point of view, the response would still exist, but for a reason that's no longer valid. It doesn't seem like a big deal, but I think it's worth confirming how we want to handle this. (Note to myself)
4. Reconciliation logic
    The reconciliation logic for `serverDraft` + `hasLocalDraft` will need to be updated when we implement the retake feature, so let's see how that evolves...

PR #4342, opened 2026-09-09
LF-5192: No Option to Redo Survey After Completion

SayakaOno:
**Description**

Enables retaking a survey after it's already been completed.

* **API**:
   * Update `checkDraftNotCompleted` to return 409s only when the request's `submission_id` matches an already-completed response's `submission_id`, not merely because *any* response exists for that farm+key.
   * Add `submission_id` to `GET /survey_drafts` response.
* **Retake flow**:
   * Remove redirection to the results page to allow retaking a survey after it has already been completed.
   * Uncomment `persistDraft({ survey_data: {} })` in `useSurveyDraftSync` to register a fresh draft when starting a retake.
   * `getSurveyInitialData` now prefills `initialData` from the previous response's `survey_response` when there's no live draft and no new version.
* **Status consistency**:
   * `SurveyInsightTile`'s `route` now checks `isCompleted && !inProgress` instead of just `isCompleted`, so an in-progress state navigates to the form.
   * `getSurveyState` now checks the in-progress-draft branch before the completed-response branch.
   * Extract the staleness check in `useInitialDraft` into a shared `isLocalDraftStale` function and also call it from `getSurveyState`'s `hasLocalDraft` computation, so both places share one definition of "stale".
* **Fixes for unexpected PUT /survey_draft API call**:
   * Prevent `persistDraft` from being called when leaving for the results page
   * Return in `handleCurrentPageChanged` when the survey has already been completed (the function is called when clicking "Complete" on the preview page)

Notes:
* There's currently no way to access the results page once Step 0/1 is in progress again.
* The retake button's placement on the results page is a deliberate rough placeholder. It will be implemented as part of [LF-5491](https://lite-farm.atlassian.net/browse/LF-5491).
* A real version-bump comparison (`hasNewSurveyVersion`) will be implemented in [LF-5473](https://lite-farm.atlassian.net/browse/LF-5473) — every retake prefills and every completed card/button reads "Update" until that ships. 

Jira link: https://lite-farm.atlassian.net/browse/LF-5192

**Type of change**

**How Has This Been Tested?**

- [x] Passes test case
- [ ] Other (please explain)

**Checklist:**

[LF-5491]: https://lite-farm.atlassian.net/browse/LF-5491?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/webapp/src/containers/Insights/Survey/TapeResults.tsx
```diff
@@ -76,6 +76,9 @@ function TAPEResults({ surveyId = 'tape' }: { surveyId?: string }) {
   const caetScores = survey_response ? getTAPEDimensionScores(survey_response) : [];
   const modules = useSurveyModules(surveyId, survey_response);
 
+  // modules.length > 0 means both "retake is available" and "there's a module section to show"
+  const hasModules = modules.length > 0;
```
SayakaOno, line 80: Now I'm a little confused 😂  Do we want people who have taken the `fao.json` survey to take a new version? If so, this doesn't work right...?

kathyavini: Yeah 😞 I think we need the ability to move / update from the old FAO (which now I'm really regretting putting on prod!) to the new one, so that the Step 2 modules eventually become available.

But I had forgotten that old FAO matches Mornington TAPE in being the old format, but would need the new version tag and the "Retake."

Would it be easier or harder to scrap version pinning for the old FAO TAPE and just discard any drafts? 😬 In other words just force moving to the updated FAO in this one case?

SayakaOno: Oh, I might have been a little dramatic! 😅
Would `modules.length > 0 || hasNewSurveyVersion() ? <Button /> : null` work?
Once users complete their old FAO TAPE drafts, they'll see the Retake button.

packages/webapp/src/containers/Insights/Survey/useSurveyDraftSync.tsx
```diff
@@ -37,14 +37,15 @@ function useSurveyDraftSync({
   const [upsertSurveyDraft] = useUpsertSurveyDraftMutation();
 
   const submissionIdRef = useRef(initialDraft.submissionId);
```
kathyavini, line 39: This value is initially `undefined` and doesn't get synched when the first response concludes. That means on the first PUT on any device, even one that has resolved from server draft, there is no `submission_id`. I can't decide if that is a problem though!

The potentially problematic behaviour looks like this:
- start a draft on device A, and load it up on device B
- server draft is adopted on device B, but `submissionIdRef` remains undefined
- complete survey on device A, and then trigger a PUT on device B with page change
- device B sends without `submission_id` in the payload, so the middleware check is skipped and a draft is created with a new `submission_id` even though the shared draft between them has just been completed

From the discussions during the persisted server draft ticket, I think that was meant to trigger a 409 and report to Sentry. But whether this feels buggy or not depends on the device you continue with... on device B, your subsequent changes are saved and that seems fine. On device A, it means you will go back from completed to in-progress state. Is this wrong?

The 409 can be restored by synching the ref in a useEffect and also the Redux state with `setDraftSubmissionId` but I can't decide if that's necessary!

kathyavini (dismissed):
This looks really good, and I like how it isn't too much code! ❤️ 

There was one part I went down a bit of a testing rabbit hole with, but as with most concurrency bugs I worry that this is over-thinking! 😜 Can I describe and defer to you? I'll put it in-line below.

It's mergeable to me, though!

packages/webapp/src/containers/Insights/Survey/TapeResults.tsx
kathyavini: Ummmmm I think so... ~if we can include the old FAO (but exclude Mornington) within the CDN version manifest so that `hasNewSurveyVersion()` is true for it...! But I don't think the plan covers it. In fact, I think that issue was specifically listed as "Out of scope"~

Or will it already work as-is, without further adjust?? Because the compare will be against the survey version of the new FAO TAPE?? That would be great! 🙌

SayakaOno: Let me confirm! I need to refresh my memory on country/new versions, so I'll handle it in the version ticket!

packages/webapp/src/containers/Insights/Survey/useSurveyDraftSync.tsx
SayakaOno: Thank you so much for catching this! This kind of situation was exactly what I was trying to avoid!!!

I've made two changes:
- Update `submissionIdRef` when `initialDraft` is ready
- Save `submissionId` in Redux

I'd really appreciate it if you could test it again 🙏

SayakaOno (commented):
It wasn't working as I wanted it to, so I've fixed them. There were two bugs!
Thank you! ❤️

kathyavini (approved):
Yay, I see the 409! 🎉

I'm really glad we'll be reporting to Sentry because I have truly no idea how edge case these concurrency issues are 😅 It may be that they never actually happen, despite how stressful they are to reason about and test 😜

PR #4346, opened 2026-09-15
LF-5356 Side Menu Becomes Non-Scrollable After Minimizing Twice and Expanding Finances Menu

den4ik1203:
**Description**

This change makes the navigation side menu scrollable after minimizing.

Jira link: https://lite-farm.atlassian.net/browse/LF-5356

**Type of change**

**How Has This Been Tested?**

- [X] Passes test case
- [ ] Other (please explain)

**Checklist:**


kathyavini (approved):
Really neat bug find!! I've never noticed before these parts of the menu behaviour. The code fix is also good.

Just two notes about the mechanism behind the issue, for our future reference:

1) The CSS class containing this override of `overflow: auto` is not applied on initial load, just after expansion, which explains the unusual timing of the issue:

```jsx
className={clsx(
  // ...
  !isCompact && hasBeenExpanded && styles.expandedSideMenuContent,
)}
```

2) Both the `overflow: visible` here on `.expandedSideMenuContent` and the parallel `overflow: hidden` on `.compactSideMenuContent` where added in commit https://github.com/LiteFarmOrg/LiteFarm/pull/3511/changes/df53c6b9821ff5ebf164e097ea0fa704674202bc for Malayalam. Were these part of the 'hover to show' behaviour? With the expanded menu width for `ml` I have never seen that behaviour; perhaps it wasn't implemented in the end.

The `overflow: hidden` does make the collapsed sidebar unscrollable at reduced/tablet heights, and I suspect that could be removed as well, but would need to be tested in a separate ticket.

@den4ik1203 since you are speeding through the PRs, I think you will have time to ask Codex to please inspect the `git blame` for the line of code you are changing. In this case, for instance, it suggests that verification needs to be done in Malayalam in particular, since the PR removes a line of code that was added in that context 🙏

PR #4348, opened 2026-09-16
LF-5493 People page should show everyone by default

den4ik1203:
**Description**

This Change changes the default number of people to 50 and removes the 5, 10, and 20 options.

Jira link: https://lite-farm.atlassian.net/browse/LF-5493

Screenshots:

<table>
   <tr>
    <th align="center">Before</th>
    <th align="center">After</th>
  </tr>
  <tr>
    <td align="center">
      Before - People page
    </td>
    <td align="center">
      After - People page
    </td>
  </tr>
</table>

**Type of change**

**How Has This Been Tested?**

- [X] Passes test case
- [ ] Other (please explain)

**Checklist:**


packages/webapp/src/components/Table/TableV1.jsx
```diff
@@ -51,6 +51,7 @@ function TableV1({
       showPagination={showPagination && data?.length > defaultPageSize}
       pageSizeOptions={pageSizeOptions}
       defaultPageSize={pageSize}
+      pageSize={pageSizeOptions?.length === 1 ? pageSize : undefined}
```
SayakaOno, line 54: This logic doesn't really make sense to me (and it works for me without it). What does it do?

packages/webapp/src/components/Profile/People/index.jsx
```diff
@@ -128,8 +128,8 @@ export default function PurePeople({ users, history, isAdmin }) {
         columns={summaryColumns}
         data={getFilteredUsers()}
         showPagination={true}
-        pageSizeOptions={[5, 10, 20, 50]}
-        defaultPageSize={5}
+        pageSizeOptions={[50]}
```
SayakaOno, line 131: Divya wrote in the ticket:
> The filter options of 5, 20 results should be removed.

Do we want to remove `10` as well?

packages/webapp/src/components/Table/TableV1.jsx
den4ik1203: Without this line , once I was adding more than 5 people to the farm it was not shown Immediately, only after refresh, that is why I asked Codex to look into it as well , and he proposed add this

packages/webapp/src/components/Profile/People/index.jsx
den4ik1203: I will ask Divya, because I  understood it that way

packages/webapp/src/components/Table/TableV1.jsx
den4ik1203: Adding this line of code   `pageSize={pageSizeOptions?.length === 1 ? pageSize : undefined}` let the table update its page size when the calculated
  value changes—for example, after data loads—when there’s only one option. defaultPageSize alone only sets the initial  value.
  
  Without this change while testing I wasn't able to see the person I added without refreshing the page.

PR #4351, opened 2026-09-18
LF-5495 Create and commit the CDN backup & version-manifest generating script

kathyavini:
**Description**

Despite my good initial intentions through the first several commits, by the time the CLI-run entrypoint and report-generator (`index.ts`) was added, this devolved into vibe-coded content and I would not say it's worth careful review -- I'll just explain how it relates to the rest of `packages/api`, and then describe how it works in general (optional reading).

#### Repo placement
- this was placed in `packages/api` to share the API's `.env` and Digital Ocean (`DO_ENDPOINT`) config, although in the end I also created two dedicated `.env` variables for this rather than over-writing the dev variables for bucket write access, which point to Minio
- it is outside of `src`, which I think is correct (it is unreachable from `server.ts` code), but is copied into the backend docker image, which is the same situation as `tests/`
  - being unreachable from `server.ts` is why I think it is safe to merge without too-careful review, but limited bucket access (below) is also a factor
- I used the existing `scripts/` folder and added this folder to tsc (mirroring `/tests`) and to the linting

####  Bucket access
- This was tested with an access key I generated on DO that is scoped specifically to the dev bucket only; for safety I'll probably create per-environment keys and switch between them manually, but the script will also work (and be a bit faster to use) with a key that grants access to all buckets

## Usage

Please refer to the [README.md](https://github.com/LiteFarmOrg/LiteFarm/blob/0dce8e0bc2ac3cbec35a409afd9c2f348286090b/packages/api/scripts/surveyCdnRelease/README.md) file for usage, including the `--dry-run` flag and in particular the example commands at the end.

## Concepts & terminology

Here are a few terms the scripts refer to and what they mean

- **Pointer** (`latestObjectKey`): `tape_surveys/fao/step01.json`  
    The spaces key for the latest/current file downloaded by the webapp. It is mutable and always points to the latest survey
- **Archive** (`archivedObjectKey`): `tape_surveys/fao/step01/2026_v1.json`  
   The key for an immutable snapshot tagged with the survey's embedded `survey_version`
- **Version Manifest**: `tape_surveys/versions.json`
    The JSON map pairing every survey path to its current active version (e.g. `"fao/step01"`: `"2026_v1"`). The planfile might refer to this as a 'version list' as well
- **Rescued Archive** (`preservedArchivedObjectKey`):  
    If the remote bucket contains a live pointer with a version (e.g. `2025_v2`) that was never archived (i.e. `tape_surveys/fao/step01/2025_v2.json` doesn't exist), the tool archives the old pointer before overwriting it.

### Flow in brief
1. CLI invocation (`index.ts`) parses the arguments and checks credentials, initializes the `S3Client` (`spacesClient.ts`)
2. Reads the files (`surveySources.ts`) - either from `tape_surveys` on the hosted bucket (`--from-bucket`) or recursively from the provided folder
3. Extracts the `survey_version` from each file (`surveySources.ts`)
4. Constructs the latest pointer and archive keys (`surveySources.ts`). Only processes keys within `tape_surveys` matching the `fao/` and `/fao_*/` format. (_**_Note:_** this is from the plan specification; if we wanted to update survey pinning to be more general beyond FAO TAPE, the `MANAGED_DIRECTORY_PATTERN` and `TAPE_SURVEYS_DIRECTORY` here would need updating as well)_
5. Diff + plan (`publishToCdn.ts`): Check the live pointer and rescue the old version (save it to an archive path) if none exists for it; evaluate whether the archive is `new` (doesn't exist yet, safe to write), `unchanged` (already identical to the incoming file), or `differs` (archive file exists but is not byte-identical to the incoming file, which flags a conflict and prompts `--refresh` to overwrite the archive; live pointer will be overwritten either way)

    If `--dry-run` is not set, then for each file, in this order:
7. Write the rescued archive if applicable
8. Write the versioned archive
9. Updates the live pointer

    Then, once per environment after all files are written:
10. Updates the version manifest (`versionManifest.ts`)
11. Finally, reports to terminal with a summary

## Current bucket status

I have run this on the development DO bucket (after archiving your manually-configured `versions.json` as `versions-ORIG.json` @SayakaOno, in case you needed it!) with

```bash
publish-surveys --from-bucket --env development
```
_Note: I'm using an alias as described in the README.md_

This archived and created a `versions.json` entry for every FAO TAPE JSON on dev.

I then updated **just the productivity step 2 module** with a new version (with an updated `survey_version`) that Riddhi had put on beta today.

Command (from the survey-containing folder)

```bash
publish-surveys --prefix fao --env development step2-productivity-biodiversity.json
```

So already completed drafts of that module shoulder trigger the `New version` logic (when it's ready for testing).

Jira link: https://lite-farm.atlassian.net/browse/LF-5495

**Type of change**

**How Has This Been Tested?**

- [ ] Passes test case
- [x] Other (please explain)

Ran against the development (`litefarm`) Digital Ocean Bucket as described above and inspected the bucket output

**Checklist:**


packages/api/scripts/surveyCdnRelease/publishToCdn.ts
```diff
@@ -0,0 +1,200 @@
...
+
+    if (shouldWriteArchivedObject(state, options.refresh)) {
+      await putObject(target, file.archivedObjectKey, file.body);
+      archivedObjectKeysWritten.push(file.archivedObjectKey);
+    }
+
+    if (!isPointerUnchanged) {
+      await putObject(target, file.latestObjectKey, file.body);
+      latestObjectKeysWritten.push(file.latestObjectKey);
+    }
```
SayakaOno: When you republish a file under an existing version with different content, you get `! Version conflict: archive already exists with different contents (use --refresh to replace)`, but the pointer has already been updated.

The code matches the `claude-docs` plan, but I think the plan missed this case:
A farmer who starts fresh initially gets the updated pointer file, but if they leave the survey and come back later with a draft, they can end up getting the untouched archive instead.

Could we avoid updating the pointer in this case and tell the operator to rerun with `--refresh` or bump the version?

packages/api/scripts/surveyCdnRelease/surveySources.ts
```diff
@@ -0,0 +1,136 @@
...
+
+    const expression = parsed?.calculatedValues?.find(
+      (calculatedValue) => calculatedValue.name === 'survey_version',
+    )?.expression;
+
+    if (typeof expression !== 'string') {
+      return undefined;
+    }
+
+    return expression.replace(/^['"](.*)['"]$/, '$1');
```
SayakaOno: Claude is pointing out a potential issue where the `survey_version` string gets parsed differently between the web app and the script, when a survey file's version is in an unexpected quoting format.

Could we make the parsing consistent between them?

| Input `expression` | Webapp today (`/^'(.*)'$/`) | Script today (`/^['"](.*)['"]$/`) | Proposed fix (`/^(['"])(.*)\1$/`) |
| --- | --- | --- | --- |
| `'TAPE_V1'` (single, well-formed) | `TAPE_V1` ✓ | `TAPE_V1` ✓ | `TAPE_V1` ✓ |
| `"TAPE_V1"` (double, well-formed) | `"TAPE_V1"` ✗ (unchanged, quotes kept) | `TAPE_V1` ✓ | `TAPE_V1` ✓ |
| `"TAPE_V1'` (mismatched quotes) | `"TAPE_V1'` (unchanged, accidentally safe) | `TAPE_V1` ✗ (wrongly accepted) | `"TAPE_V1'` (unchanged, correctly rejected) |
| `TAPE_V1` (no quotes at all) | `TAPE_V1` (unchanged) | `TAPE_V1` (unchanged) | `TAPE_V1` (unchanged) |

packages/api/scripts/surveyCdnRelease/index.ts
```diff
@@ -0,0 +1,244 @@
...
+        TAPE_SURVEYS_DIRECTORY,
+        values.prefix,
+      );
+
+  for (const target of targets) {
+    const sourced = localFiles ?? (await readSurveyFilesFromBucket(target, TAPE_SURVEYS_DIRECTORY));
+    const files = sourced.filter((file) =>
+      isManagedSurveyPointerKey(file.latestObjectKey, TAPE_SURVEYS_DIRECTORY),
+    );
+    const skipped = sourced.length - files.length;
```
SayakaOno, line 194: (Claude)
When running with `--from-bucket`, the files are already filtered before being assigned to `sourced`, so `sourced.filter` won't change anything and the `skipped` count will always be 0.

packages/api/scripts/surveyCdnRelease/index.ts
```diff
@@ -0,0 +1,244 @@
...
+function getSurveyStem(objectKey: string): string {
+  return objectKey.slice(TAPE_SURVEYS_DIRECTORY.length + 1).replace(/\.json$/, '');
+}
+
+function formatVersionTransition(report: SurveyFileReport): string {
+  if (report.state === 'differs') {
+    return `${report.version} (modified)`;
+  }
+
+  if (!report.currentPointerVersion || report.currentPointerVersion === report.version) {
```
SayakaOno, line 106: I don't believe the version will ever be empty, but Claude brought this up, so I'll mention it... 

`report.currentPointerVersion` being `undefined` (no previous pointer) or `''` (a previous pointer with a broken/empty version) is treated the same here — both suppress the "before" text and show only the new version.

SayakaOno (commented):
This looks great and thorough, and I can see how this will help mitigate human errors, thank you!!! 🙏 I also appreciate the design where archived files are created when adding a new version - it'll work well with my code.

I was initially going to approve after reading the README, but Claude offered to review it as well...😂 It found two good points to fix, and I've briefly commented on a few minor points too!

packages/api/scripts/surveyCdnRelease/surveySources.ts
kathyavini: Ahhh! I committed the fix for this on Friday but I forgot to push it! 🤦‍♀️

packages/api/scripts/surveyCdnRelease/publishToCdn.ts
kathyavini: > A farmer who starts fresh initially gets the updated pointer file, but if they leave the survey and come back later with a draft, they can end up getting the untouched archive instead.

Ack, actually I had assumed in the case above the farmer would be routed to the pointer, but I think that must have been in a previous iteration of the plan only 🙇 Thank you!

And even if that were the case, the warning coming with the pointer already being replaced is just strange; not sure why that was the chosen behaviour. I totally agree with what you recommend.

kathyavini:
@SayakaOno I remember wrapping up on Friday a bit uneasy about the way that the non `--refresh` behaviour would let the archive and the pointer differ silently like that, so I'm glad you brought it up! The script now produces a hard stop when that warning is generated, with no pointer replacement, which I much prefer, and `--refresh` updates both at the same time.

The report issues should be okay -- I think I also got a flag from Claude about the `--from-bucket` report, but the code change required seemed more than it was worth, and now Claude (after update to 5.5? 😂) is saying not to do anything about this, since the skipped files are only meaningful in local. Ditto for an empty `survey_version` which is rejected before anything is written.

SayakaOno (approved):
Looks good!

